Issue #33, opened 2019-01-14
Mixed-case checksum address encoding


dahlia:
The current `Address` object provides hexadecimal formatting, but it always returns only lowercase letters.  Since we are following the same address scheme to Ethereum by design, we should follow [EIP 55][1] as well.

This task consists of two subtasks:

- [x] Making the formatting method to return a mixed-case EIP 55 string. → #43
- [x] Adding a new constructor to take a hexadecimal address string.  If a string is mixed-case and the checksum is invalid it should throw an exception. → #53

[1]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md

dahlia:
Closed by #43 and #53.

PR #43, opened 2019-01-19
Implement the formatting method to return a mixed-case EIP 55 string

earlbread:
This Implements the formatting method to return a mixed-case EIP 55 string(#33).


Libplanet/Address.cs
```diff
@@ -166,13 +167,45 @@ public override string ToString()
...
+        /// </summary>
+        /// <example>A returned string looks like
+        /// <c>0x87Ae4774E20963fd6caC967CF47aDCF880C3e89B</c>.</example>
+        /// <returns>A <c>0x</c>-prefixed Mixed-case checksum address of 42
+        /// letters that represent this <see cref="Address"/>.</returns>
+        [Pure]
+        public string ToChecksumAddress()
+        {
+            var hex = ToHex();
+            var bytes = Encoding.ASCII.GetBytes(hex);
```
longfin: Instead of re-parsing, we can call `ToByteArray()` here.

Libplanet/Address.cs
```diff
@@ -166,13 +167,45 @@ public override string ToString()
...
+        /// <summary>
+        /// Gets a <c>0x</c>-prefixed Mixed-case checksum address of 42 letters
+        /// that represent this <see cref="Address"/>.
+        /// </summary>
+        /// <example>A returned string looks like
+        /// <c>0x87Ae4774E20963fd6caC967CF47aDCF880C3e89B</c>.</example>
+        /// <returns>A <c>0x</c>-prefixed Mixed-case checksum address of 42
+        /// letters that represent this <see cref="Address"/>.</returns>
+        [Pure]
+        public string ToChecksumAddress()
```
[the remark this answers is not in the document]
longfin: Good point, I agree.

Libplanet/Address.cs
```diff
@@ -166,13 +167,45 @@ public override string ToString()
...
+        {
+            var hex = ToHex();
```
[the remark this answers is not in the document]
earlbread: Thank you, I'll fix it.

Libplanet/Address.cs
earlbread: I think the results are different. the bytes in the code is 40 bytes and the result of `ToByteArray()` is 20 bytes.

Libplanet/Address.cs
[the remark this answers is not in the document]
earlbread: I agree. I'll fix `ToHex()` method instead of creating `ToChecksumAddress()`

Libplanet/Address.cs
longfin: My bad, I've mistaken.

Libplanet/Address.cs
```diff
@@ -145,15 +146,27 @@ public ImmutableArray<byte> ByteArray
...
         public string ToHex()
         {
-            return ByteUtil.Hex(ToByteArray());
+            string hex = ByteUtil.Hex(ToByteArray());
+            byte[] bytes = Encoding.ASCII.GetBytes(hex);
+            byte[] hash = CalculateHash(bytes);
+            string hashHex = ByteUtil.Hex(hash);
+            string address = string.Empty;
+
+            for (var i = 0; i < hex.Length; i++)
```
dahlia: As `string` implements `IEnumerable<char>`, this can be adjusted to use `foreach` instead.

Libplanet/Address.cs
```diff
@@ -130,11 +131,11 @@ public ImmutableArray<byte> ByteArray
         public byte[] ToByteArray() => ByteArray.ToArray();
 
         /// <summary>
-        /// Gets a hexadecimal string of 40 letters that represent this
-        /// <see cref="Address"/>.
+        /// Gets a mixed-case hexadecimal string of 40 letters that represent
+        /// this <see cref="Address"/>.
         /// </summary>
```
dahlia: Could you append a mention that a returned hexadecimal string follows [EIP 55](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md)?

dahlia:
Note that a build error during pushing docs was fixed by PR #45.  You should rebase this PR on the current master again (20386c5f95383b8ae6b4fc378d6f4c2a6e496148 at least).

Libplanet/Address.cs
longfin: It seems to use `i` as the index of `hexHash`.

dahlia: @longfin Fair enough.  This should be as is.  @earlbread Nevermind!

Libplanet/Address.cs
earlbread: I rebased on master and added a mention about EIP 55!

dahlia:
Thanks for your contribution!

PR #47, opened 2019-01-22
Separate message related features from Swarm.

longfin:
This PR moves message related features (e.g. parsing, signing) from `Swarm` to `Libplanet.Net.Messages`. 

- In this process, I've made `Message` abstract class for more naturally abstracting of `NetMQMessage`, and added a concrete class for each message.
- It also applied the missing message sign/verify to all messages using common code in `Message`.
- Added a missing `stop` parameter to `Swarm.GetBlocksAsync()` in the previous PR(#34).

### DIFF
--- a/Libplanet/Net/Messages/GetBlocks.cs
+++ b/Libplanet/Net/Messages/GetBlocks.cs
@@ -1,0 +1,58 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using NetMQ;
+
+namespace Libplanet.Net.Messages
+{
+    internal class GetBlocks : Message
+    {
+        public GetBlocks(
+            BlockLocator locator, HashDigest<SHA256>? stop)
+        {
+            Locator = locator;
+            Stop = stop;
+        }
+
+        public BlockLocator Locator { get; }
+
+        public HashDigest<SHA256>? Stop { get; }
+
+        protected override MessageType Type => MessageType.GetBlocks;
+
+        protected override IEnumerable<NetMQFrame> DataFrames
+        {
+            get
+            {
+                yield return new NetMQFrame(
+                    NetworkOrderBitsConverter.GetBytes(Locator.Count()));
+
+                foreach (HashDigest<SHA256> hash in Locator)
+                {
+                    yield return new NetMQFrame(hash.ToByteArray());
+                }
+
+                if (Stop is HashDigest<SHA256> stop)
+                {
+                    yield return new NetMQFrame(stop.ToByteArray());
+                }
+                else
+                {
+                    yield return NetMQFrame.Empty;
+                }
+            }
+        }
+
+        public static Message ParseBody(NetMQFrame[] frames)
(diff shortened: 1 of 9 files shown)


dahlia:
Note that a build error during pushing docs was fixed by PR #45.  You should rebase this PR on the current master again (20386c5f95383b8ae6b4fc378d6f4c2a6e496148 at least).

longfin:
Rebased.

Libplanet/Net/Messages/Message.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            /// <summary>
+            /// Peer's delta set to sync peer list.
+            /// </summary>
+            PeerSetDelta = 0x03,
+
+            /// <summary>
+            /// Request to query block hashes.
+            /// </summary>
+            GetBlocks = 0x4,
```
dahlia: Though it's really trivial, values for `Ping`/`Pong`/`PeerSetDelta` follow `0x0?` style (note that zero), while `GetBlocks`/`Inventory` follow `0x?` style.  It could follow the consistent style.

Libplanet/Net/Messages/Message.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        protected enum MessageType : byte
+        {
+            /// <summary>
+            /// Check message to determine peer is alive.
+            /// </summary>
+            Ping = 0x01,
+
+            /// <summary>
+            /// Reply of `Ping`.
```
dahlia: FYI unfortunately C# doc comments use an ancient XML-based syntax instead of Markdown.  Don't mind this, I will write docs for them later.

longfin:
Rebased. (https://github.com/planetarium/libplanet.net/pull/47#discussion_r249784507 fixed)

longfin:
Rebased because #43 was merged.

Libplanet/Net/Messages/GetBlocks.cs
```diff
@@ -0,0 +1,58 @@
...
+        public HashDigest<SHA256>? Stop { get; }
+
+        protected override MessageType Type => MessageType.GetBlocks;
+
+        protected override IEnumerable<NetMQFrame> DataFrames
+        {
+            get
+            {
+                yield return new NetMQFrame(
+                    NetworkOrderBitsConverter.GetBytes(Locator.Count()), 4);
```
ipdae: What is the reason for limiting the length of the buffer?

longfin: There isn't serious reason, `Locator.Count()` is `int` and the size of `int` is 4 (bytes).

longfin: Now that I think about it, it's fine to remove buffer size. I'll remove it.

longfin:
Rebased. (due to https://github.com/planetarium/libplanet.net/pull/47#discussion_r250057877)

PTAL @ipdae @kijun @dahlia

PR #48, opened 2019-01-23
Aesgcm → SymmetricKey

dahlia:
- Renamed `Libplanet.Crypto.Aesgcm` to `SymmetricKey` because it's easier to understand even to users who are not aware what AES and GCM mean.
- Changed the return type of `PrivateKey.ExchangeKey()` method from `byte[]` to `SymmetricKey`.
- To make `SymmetricKey` more end-usable, implemented `IEquatable<SymmetricKey>` interface for it, and added members like `ByteArray`/`ToByteArray()`.
- Wrote the missing API docs for the type.


Libplanet/Libplanet.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
longfin: is there BOM?

Libplanet.Tests/Libplanet.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
longfin: is there BOM? (or removed?)

Libplanet/Crypto/SymmetricKey.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// <returns>A new mutable <see cref="byte"/> array which encodes this
+        /// key.  Since it is created every time the method is called, any
+        /// mutation on that does not affect this <see cref="SymmetricKey"/>
+        /// object.</returns>
+        /// <seealso cref="ByteArray"/>
+        /// <seealso cref="SymmetricKey(byte[])"/>
+        [Pure]
+        public byte[] ToByteArray()
+        {
+            byte[] copied = new byte[_key.Length];
```
longfin: Can we re-use `ByteArray` here? (or, is it performance related decision?)

Libplanet.Tests/Crypto/SymmetricKeyTest.cs
```diff
@@ -0,0 +1,44 @@
...
+            var actual = aes.Decrypt(ciphertext);
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
+        public void ByteArray()
+        {
+            var aes = new SymmetricKey(KeyBytes);
+            Assert.Equal(KeyBytes, aes.ByteArray.ToArray());
```
longfin: To verify the immutability, it seems more accurate to compare with `ImmutableArray<byte>` if we can.

Libplanet.Tests/Libplanet.Tests.csproj
dahlia: It's like accidentally inserted.  I'll remove it.

Libplanet/Crypto/SymmetricKey.cs
dahlia: It was a concern about performance thing, but I will rewrite it to use `.ToArray()` which is more readable.

Libplanet.Tests/Crypto/SymmetricKeyTest.cs
dahlia: I'm going to fix that.

dahlia:
@longfin Fixed things you reviewed and amended the existing commit.

PR #50, opened 2019-01-24
Implement GetData / Block message 

longfin:
This PR implements `GetData` / `Block` message to request and transfer blocks. (it relies [AsyncEnumerator](https://github.com/Dasync/AsyncEnumerable) to create async-generator)

Also it has some other minor changes.

- Renamed bencodex related methods. (see also. https://github.com/planetarium/libplanet.net/pull/46#pullrequestreview-195368085)
- Separated message processing logic from `ReceiveMesseageAsync()` to (new) `ProcessMessageAsync()`
- Replaced `CancellationToken` overload in `Swarm` and `NetMQSocketExtension` with default parameters

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -414,47 +414,16 @@ namespace Libplanet.Net
                 try
                 {
                     Message message = Message.Parse(rawMessage, reply: false);
-                    switch (message)
+
+                    // Queue a task per message to avoid blocking.
+                    #pragma warning disable CS4014
+                    Task.Run(async () =>
                     {
-                        case Ping ping:
-                            _logger.Debug($"Ping received.");
-                            var reply = new Pong
-                            {
-                                Identity = ping.Identity,
-                            };
-                            await SendAsync(reply, cancellationToken);
-                            break;
-
-                        case Messages.PeerSetDelta peerSetDelta:
-                            _deltas.Enqueue(peerSetDelta.Delta);
-                            break;
-
-                        case GetBlocks getBlocks:
-                            IEnumerable<HashDigest<SHA256>> inventories =
-                                blockchain.FindNextHashes(
-                                    getBlocks.Locator, getBlocks.Stop, 500);
-
-                            var inv = new Inventory(inventories)
-                            {
-                                Identity = getBlocks.Identity,
-                            };
-                            await SendAsync(inv, cancellationToken);
-                            break;
-
-                        case GetData getData:
-                            #pragma warning disable CS4014
-                            Task.Run(() =>
-                            {
-                                TransferBlocks(
-                                    blockchain, getData, cancellationToken);
-                            });
-                            #pragma warning restore CS4014
-                            break;
-
-                        default:
-                            Trace.Fail($"Can't handle message. [{message}]");
-                            break;
-                    }
+                        // it's still async because some method it relies are
+                        // async yet.
+                        await ProcessMessageAsync(blockchain, message, cancellationToken);
+                    });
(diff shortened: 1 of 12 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -436,6 +432,77 @@ IEnumerator IEnumerable.GetEnumerator()
...
+            Blockchain<T> blockchain,
+            GetData getData,
+            CancellationToken cancellationToken)
+            where T : IAction
+        {
+            foreach (HashDigest<SHA256> hash in getData.BlockHashes)
+            {
+                if (blockchain.Blocks.TryGetValue(hash, out Block<T> block))
+                {
+                    Message response = new Block(block.ToBencodex(true, true))
```
kijun: The `Block` that is being instantiated here is `Message.Block` not `Blocks.Block`, right?

longfin: Yes, its' `Libplanet.Net.Messgeas.Block` (without generic parameter `T`).

kijun: 👌

Libplanet/Net/Swarm.cs
```diff
@@ -353,6 +327,51 @@ IEnumerator IEnumerable.GetEnumerator()
                 $"but {parsedMessage}");
         }
 
+        // TODO: Add TxIds
+        public IAsyncEnumerable<Block<T>> GetDataAsync<T>(
```
dahlia: Would it be directly used by library users?  To a library user, `IAsyncEnumerable<T>` can be unfamiliar.

Libplanet/Net/Swarm.cs
```diff
@@ -436,6 +432,77 @@ IEnumerator IEnumerable.GetEnumerator()
             }
         }
 
+        private async Task ProcessMessageAsync<T>(
+            Blockchain<T> blockchain,
+            Message message,
+            CancellationToken cancellationToken)
+            where T : IAction
+        {
+            switch (message)
```
dahlia: Couldn't this switch be restructured so that the `Message` subtypes has a polymorphic (virtual) method to process the equivalent thing?

longfin: I've tried replace with a sort of `Prcoessor` or `Message.Process()`, but there are also tricky points.

- Some message (e.g. `PeerSetDelta`) requires `Swarm`'s internal properties. (`_deltas`, `_peers`)
- (`Message.Process()`) I'm not sure that it is correct to deal with the logic of how a message is processed, rather than simply representing the data.

Libplanet/Net/Swarm.cs
longfin: > Would it be directly used by library users? 

It was not decided at this stage.

> To a library user, IAsyncEnumerable<T> can be unfamiliar.

I agree. should we provide `IEnumerable<T>` version? (but, it may be synchronous.)

Libplanet/Net/Swarm.cs
dahlia: My concern is that such function, which consists of a `switch` branch, is prone to miss cases when a new case (e.g., a new `Message` subtype) is added, which leads it to be [partial][1] not to try.  Unfortunately C# does not warn even if a `switch` is not total.  On the other hand, making a method `virtual` and leaving a subtype to not implement it, of course, is disallowed by C#.  We can be safe from a new case without its case handling.  For that reason, I believe it's better to be polymorphic and should be eventually, even though we could leave it as is in this PR.

[1]: https://wiki.haskell.org/Partial_functions

PR #53, opened 2019-02-03
Add a constructor which takes a hexadecimal address string

earlbread:
It adds a constructor which takes a hexadecimal address string(#33).

### DIFF
--- a/Libplanet.Tests/AddressTest.cs
+++ b/Libplanet.Tests/AddressTest.cs
@@ -47,6 +47,37 @@ namespace Libplanet.Tests
         }
 
         [Fact]
+        public void HexadecimalAddressConstructor()
+        {
+            Assert.Equal(
+                new Address(
+                    new byte[20]
+                    {
+                        0x45, 0xa2, 0x21, 0x87, 0xe2, 0xd8, 0x85, 0x0b, 0xb3, 0x57,
+                        0x88, 0x69, 0x58, 0xbc, 0x3e, 0x85, 0x60, 0x92, 0x9c, 0xcc,
+                    }
+                ),
+                new Address(
+                    "45a22187e2d8850bb357886958bc3e8560929ccc"
+                )
+            );
+
+            var address = new Address("45a22187e2d8850bb357886958bc3e8560929ccc");
+            Assert.Equal(
+                "45a22187e2D8850bb357886958bC3E8560929ccc",
+                address.ToHex()
+            );
+        }
+
+        [Fact]
+        public void CanDetectInvalidMixedCaseChecksum()
+        {
+            Assert.Throws<ArgumentException>(() =>
+                new Address("45A22187E2D8850BB357886958BC3E8560929CCC")
+            );
+        }
+
+        [Fact]
         public void AddressMustBe20Bytes()
         {
             for (int size = 0; size < 25; size++)
--- a/Libplanet.Tests/AddressTest.cs
+++ b/Libplanet.Tests/AddressTest.cs
@@ -73,6 +73,14 @@ namespace Libplanet.Tests
         }
 
         [Fact]
+        public void HexAddressConstructorOnlyTakesASCIICharacters()
+        {
+            Assert.Throws<ArgumentException>(
+                () => new Address("45a22187e2d8850bb357886958BC3E8560929£한글")
+            );
+        }
+
+        [Fact]
         public void CanDetectInvalidMixedCaseChecksum()
         {
             Assert.Throws<ArgumentException>(() =>
--- a/Libplanet.Tests/AddressTest.cs
+++ b/Libplanet.Tests/AddressTest.cs
@@ -53,13 +53,13 @@ namespace Libplanet.Tests
                 new Address(
                     new byte[20]
                     {
-                        0x45, 0xa2, 0x21, 0x87, 0xe2, 0xd8, 0x85, 0x0b, 0xb3,
-                        0x57, 0x88, 0x69, 0x58, 0xbc, 0x3e, 0x85, 0x60, 0x92,
-                        0x9c, 0xcc,
+                        0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef, 0xab,
+                        0xcd, 0xef, 0xab, 0xcd, 0xef, 0xab, 0xcd, 0xef, 0xab,
+                        0xcd, 0xef,
                     }
                 ),
                 new Address(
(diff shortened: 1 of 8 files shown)


Libplanet/Address.cs
```diff
@@ -186,6 +187,22 @@ public override string ToString()
             return $"0x{ToHex()}";
         }
 
+        private static string ToChecksumAddress(string hex)
+        {
+            byte[] bytes = Encoding.ASCII.GetBytes(hex);
```
dahlia: What if the given `hex` has any characters out of ASCII character set?  This case should be addressed.  See also `ByteUtil.ParseHex()` method's code.

Libplanet/Address.cs
```diff
@@ -96,6 +96,18 @@ public Address(PublicKey publicKey)
         {
         }
 
+        /// <summary>
+        /// Derives the corresponding <see cref="Address"/> from a hexadecimal
+        /// address string. If a string is mixed-case and the checksum is
+        /// invalid it throws an exception.
+        /// </summary>
+        /// <param name="hex">A 40 bytes hexadecimal address string to derive
+        /// the corresponding <see cref="Address"/> from.</param>
```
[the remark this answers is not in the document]
earlbread: Sure, I'll fix it.

Libplanet/Address.cs
```diff
@@ -202,5 +219,27 @@ private static byte[] DeriveAddress(PublicKey key)
 
             return output.Skip(output.Length - 20).ToArray();
         }
+
+        private static byte[] DeriveAddress(string hex)
+        {
+            if (hex == null)
+            {
+                throw new NullReferenceException(
```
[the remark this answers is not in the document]
earlbread: Ok, I'll use ArgumentNullException instead of NullReferenceException.

Libplanet/Address.cs
```diff
@@ -202,5 +219,27 @@ private static byte[] DeriveAddress(PublicKey key)
...
+                    "address hex must not be null"
+                );
+            }
+
+            if (hex.Length != 40)
+            {
+                throw new ArgumentException("address hex must be 40 bytes");
```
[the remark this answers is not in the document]
earlbread: Ok, I'll add paramName to Argument exceptions.

Libplanet/Address.cs
earlbread: Thanks, I'll add code for the case.

Libplanet/Address.cs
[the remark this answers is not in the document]
dahlia: Could you mention (and mention to) [EIP 55] too as we did in the previous PR #43?

[EIP 55]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md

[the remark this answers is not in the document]
earlbread: Sure, I Added a mention about EIP 55.

Libplanet/Address.cs
```diff
@@ -96,6 +96,18 @@ public Address(PublicKey publicKey)
...
         }
 
+        /// <summary>
+        /// Derives the corresponding <see cref="Address"/> from a hexadecimal
+        /// address string. If a string is mixed-case and the checksum is
+        /// invalid it throws an exception.
+        /// </summary>
+        /// <param name="hex">A 40 bytes hexadecimal address string to derive
+        /// the corresponding <see cref="Address"/> from.</param>
+        public Address(string hex)
```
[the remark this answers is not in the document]
earlbread: I replaced the constructor patterns to new constructor.

Libplanet/Address.cs
```diff
@@ -96,6 +96,28 @@ public Address(PublicKey publicKey)
...
         }
 
+        /// <summary>
+        /// Derives the corresponding <see cref="Address"/> from a hexadecimal
+        /// address string.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">Thrown when <c>null</c> was
+        /// passed to <paramref name="hex"/>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the given <paramref
+        /// name="hex"/> did not lengthen 40 bytes.</exception>
```
dahlia: It's technically not bytes but characters.

Libplanet/Address.cs
```diff
@@ -96,6 +96,28 @@ public Address(PublicKey publicKey)
...
+        /// <exception cref="ArgumentException">Thrown when the given <paramref
+        /// name="hex"/> is mixed-case and the checksum is invalid.</exception>
+        /// <exception cref="ArgumentException">Thrown when the given <paramref
+        /// name="hex"/> does not consist of ASCII characters.</exception>
+        /// <param name="hex">A 40 bytes hexadecimal address string to derive
```
dahlia: We should adjust the word “bytes” here too.

Libplanet.Tests/AddressTest.cs
```diff
@@ -46,6 +46,48 @@ public void DerivingConstructor()
...
+                "45a22187e2D8850bb357886958bC3E8560929ccc",
+                address.ToHex()
+            );
+        }
+
+        [Fact]
+        public void HexAddressConstructorOnlyTakesASCIICharacters()
+        {
+            Assert.Throws<ArgumentException>(
+                () => new Address("45a22187e2d8850bb357886958BC3E8560929£한글")
```
dahlia: Even if a string consists of only ASCII characters, it should throw an exception unless it only consists of `[A-Fa-f0-9]`.  It also must not be longer nor shorter than 40 characters.  Could we add tests for those three cases?

- 40 ASCII characters which contain any `[^A-Fa-f0-9]` character
- 41 (or longer) of valid hexadecimal characters
- 39 (or shorter) of valid hexadecimal characters

Libplanet/Address.cs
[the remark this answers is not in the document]
dahlia: Thanks!

Libplanet/Address.cs
earlbread: When given `hex` has some non-ASCII characters, `Encoding.ASCII.GetBytes` doesn't raise an exception, but `ByteUtil.ParseHex` raise `FormatException` instead. So I caught it and re-raised `ArgumentException`.

Libplanet/Address.cs
earlbread: You're right. I'll fix it.

Libplanet.Tests/AddressTest.cs
earlbread: That would be good, I'll add the test cases.

earlbread:
I fixed docs and added test cases.

Issue #49, opened 2019-01-23
Pass a context (environment?) object to IAction.Execute() method


dahlia:
Some operations such as [random generation][1] can depend on information determined by a mined block, that a transaction belongs to.  However, since simply passing a `Block` object to `IAction.Execute()` may accidentally cause unpredictable dependencies in the future, the information should be limited to only what we actually need at present.

A type for a such information unit can be named `Context` or `Environment`, I think? @kijun @longfin Any suggestions are welcome.

`IAction.Execute(Address, AddressStateMap)`'s signature also should be changed like:

```csharp
AddressStateMap Execute(Address from, Address to, AddressStateMap states, IContext ctx);
```

The `IContext` interface would look like at first:

```csharp
public interface IContext {
    System.Random Random { get; }
}
```

[1]: https://docs.nekoyu.me/white_paper.html#hash-random

dahlia:
#55 implemented this.  Thanks for @hanc1208.

PR #55, opened 2019-02-07
Pass a context including Random object to IAction.Execute() method

hanc1208:
It implements #49.

Context includes Random object.
Seed of Random object is generated using block hash and extra seed.
Extra seed is incremented for each transaction starting at 0.
Thus, each action uses same random seed, but each transaction will use a different random seed.

### DIFF
--- a/Libplanet.Tests/Action/ActionContextTest.cs
+++ b/Libplanet.Tests/Action/ActionContextTest.cs
@@ -1,0 +1,24 @@
+using Libplanet.Action;
+using Xunit;
+
+namespace Libplanet.Tests.Action
+{
+    public class ActionContextTest
+    {
+        [Fact]
+        public void RandomShouldBeDeterministic()
+        {
+            (int seed, int expected)[] testCases =
+            {
+                (0, 1559595546),
+                (1, 534011718),
+            };
+            foreach (var (seed, expected) in testCases)
+            {
+                var context = new ActionContext(seed);
+                IRandom random = context.Random;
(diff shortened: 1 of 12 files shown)


Libplanet/Action/IContext.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Libplanet.Action
+{
+    public class IContext
+    {
+        public Random Random { get; protected set; }
```
longfin: Do we need to provide a setter?

Libplanet/Action/Context.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+using System.Runtime.CompilerServices;
+using Libplanet.Blocks;
+
+[assembly: InternalsVisibleTo("Libplanet.Tests")]
```
longfin: Minor, but InternalsVisibleTo is a global setting so you do not have to declare it again.

Libplanet/Action/IContext.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Libplanet.Action
+{
+    public class IContext
```
longfin: `IContext` seems a little confusing name. Is `IExecutionContext` too messy?

dahlia: I suggest `IActionContext` for an alternative name.

dahlia: It should be `interface` instead of `class`.

PR #57, opened 2019-02-07
Move all parameters of IAction.Execute to IActionContext

hanc1208:
Side effect
- Increment random seed for each actions (was for each transactions)


longfin:
@hanc1208 Thanks for patch.

PR #58, opened 2019-02-07
Introduce Menees.Analyzers.2017 and line length limit

dahlia:
This patch introduces [Menees.Analyzers.2017][1] to limit maximum characters per line: 80 columns.  This analyzer also suggests several best practices such as preferring `DateTime.UtcNow` over` DateTime.Now` and [preferring `NotSupportedException` over `NotImplementedException`][2].

Except of rule settings, all changes in this patch are all about styles.  No semantic changes.

[1]: https://www.nuget.org/packages/Menees.Analyzers.2017/
[2]: https://blogs.msdn.microsoft.com/brada/2004/07/29/the-difference-between-notsupportedexception-and-notimplementedexception/


longfin:
I'm not sure if 80 characters are appropriate. 

- In Mono Project, [line limit is 180 columns.](https://www.mono-project.com/community/contributing/coding-guidelines/#line-length-and-alignment)
- bmenees(author of menees analyzer) seems to [use 160 / 180 characters in line length limit.](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1808#issuecomment-165850634)

On the other hand, there are [other opinions](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1808#issuecomment-158788915) about 'Is line length restriction itself necessary?'

dahlia:
> On the other hand, there are [other opinions](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1808#issuecomment-158788915) about 'Is line length restriction itself necessary?'

Indeed, Microsoft's open source projects written in C# (e.g., Roslyn, CoreFX, ASP.NET Core, PowerShell) do not have any restriction on the maximum characters per a line.

(Before leaving this comment, note that it's not teaching someone, but for a recording our discussion and how we made the decision for the future.  I believe I and @longfin already have consensus on limiting line length for the most part.  We've actually worked together long time.)

One of the most mentioned reasons why we should limit line length is because it can display multiple source files horizontally at a time, or be printed to a US Letter paper.  This particular rationale seems no more valid now, because no one uses VT100 anymore, displays become wider and wider, and the most of code editors have soft wrapping.  Should we put code into print to review that?  I even have never done that.

However, a reason why I still believe we should limit line length is, because it affects how we write code.  [It leads us to avoid deep indentations (which is not very readable), complicated expressions in a single statement (which does not work well with debugger).][1]  There's also a [another reason that we still have to use tools having only narrow columns, like GitHub's two-columned diff view or three-way merge][2].

So I believe it is still better to limit line length.  The problem is how long line is acceptable.

> I'm not sure if 80 characters are appropriate.

Agree.  80 is the number chosen by me because I've coded in Python, which is limiting 80 columns is still a norm (according to [PEP 8]).  C# is different from Python.  It tends to have more verbose names and longer lines than Python.

[Even Linus Torvalds recently changed his mind on this. He's suggesting “132 characters, which is another perfectly fine historical limit.”][3]

I am okay with 160, but it's like somewhat long to me.  It's my personal preference, but it would be good if it's longer than 120 and shorter than 160.

[1]: https://lkml.org/lkml/2016/12/15/749
[2]: https://news.ycombinator.com/item?id=5436993
[PEP 8]: https://www.python.org/dev/peps/pep-0008/
[3]: https://lkml.org/lkml/2009/12/17/229

longfin:
> However, a reason why I still believe we should limit line length is, because it affects how we write code. 

I thought so and basically still now. just an opinion in [mentioned comments](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1808#issuecomment-158788915) that 'line limit infer code reviews.' was so interesting. (I'm sorry if I did not have enough explanation.)

But on second thought, an automated linting is still better than review like 'It's too complicated code.' even if we limit creativity.

> I am okay with 160, but it's like somewhat long to me. It's my personal preference, but it would be good if it's longer than 120 and shorter than 160.

I think we can adjust in future. but in this time, I don't have idea about ideal length. (that's why I approved it. 😃 )

dahlia:
I rebased on the latest master as #57 was merged.

dahlia:
As Travis CI didn't run a build the previous push for an uncertain reason, I amended the commit and force-pushed it again.

PR #59, opened 2019-02-08
Add Swarm.BroadcastBlocksAsync()

longfin:
This PR implements `Swarm.BroadcastBlocksAsync()` for miner can broadcast mined block to known peers. the following points have also been changed to accomplish it.

- Added `DeleteIndex()` to `IStore` / `FileStore` to able to give up own block indexes when other longest chain appears. 
- Added `Blockchain.DeleteAfter()` to same reason.
- Added `Blockchain.GetBlockLocator()` to query block hashes to broadcasting peer.
- `Blockchain.FindNextHash()` returns result containing branch point.

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Async;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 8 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -611,12 +628,98 @@ CancellationToken cancellationToken
...
+                    blockchain.Append(block);
+                });
+
+                await AppendBlocksAsync(
+                    socket, blockchain, blocks, cancellationToken);
+            }
+            else
+            {
+                _logger.Information(
+                    "Received index is lower than current chain's tip." +
```
dahlia: The word “lower” here is assuming the highest is the tip and the lower is the genesis.  I doesn't find it necessarily intuitive, so the term should be more concrete IMHO, like “older.”

Libplanet/Store/FileStore.cs
```diff
@@ -134,6 +134,30 @@ public override long AppendIndex(HashDigest<SHA256> hash)
...
+                        found = true;
+                    }
+                    else
+                    {
+                        writer.Write(
+                            idx.ToByteArray(), 0, HashDigest<SHA256>.Size);
+                    }
+                }
+
+                File.WriteAllBytes(_indexPath, writer.ToArray());
```
dahlia: It doesn't seem very scalable to buffer the whole index on the memory.  It should be fixed in the future.  It also would be good if we write a `FIXME` comment here.

Libplanet/Net/Swarm.cs
```diff
@@ -611,12 +628,98 @@ CancellationToken cancellationToken
...
+        private async Task AppendBlocksAsync<T>(
+            DealerSocket socket,
+            Blockchain<T> blockchain,
+            List<Block<T>> blocks,
+            CancellationToken cancellationToken)
+            where T : IAction
+        {
+            // We assume that the blocks are sorted in order.
+            Block<T> oldest = blocks.First();
+            Block<T> newest = blocks.Last();
```
dahlia: It's more natural to be the `latest` than the `newest` here to me?

Libplanet/Net/Swarm.cs
```diff
@@ -611,12 +628,98 @@ CancellationToken cancellationToken
...
+            HashDigest<SHA256>? tip = blockchain.Store.IndexBlockHash(-1);
+
+            if (tip == null || oldest.PreviousHash == tip)
```
dahlia: Does it mean the `newest` is expected to be the most closest to the genesis, unlike the intuition its name give us?

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -391,6 +398,83 @@ public async Task CanBroadcastTx()
...
+            {
+                await Task.WhenAll(
+                    swarmA.InitContextAsync(),
+                    swarmB.InitContextAsync(),
+                    swarmC.InitContextAsync());
+
+                #pragma warning disable CS4014
+                Task.Run(async () => await swarmA.RunAsync(chainA, 250));
+                Task.Run(async () => await swarmB.RunAsync(chainB, 250));
+                Task.Run(async () => await swarmC.RunAsync(chainC, 250));
```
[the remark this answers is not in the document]
longfin: We can't await these tasks because they run in infinite loop.
but we can cancel these task in `finally` block. I'll add it.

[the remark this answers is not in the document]
dahlia: Ah, make sense. 👌🏼

Libplanet.Tests/Store/FileStoreTest.cs
```diff
@@ -9,13 +10,13 @@
...
-    public class FileStoreTest : IClassFixture<FileStoreFixture>
+    public class FileStoreTest : IDisposable
     {
         private readonly FileStoreFixture _fx;
 
-        public FileStoreTest(FileStoreFixture fixture)
+        public FileStoreTest()
         {
-            _fx = fixture;
+            _fx = new FileStoreFixture();
```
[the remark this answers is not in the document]
longfin: xUnit's fixture is shared within the running cases, which causes unnecessary side effects.

Libplanet/Blockchain.cs
```diff
@@ -42,6 +42,8 @@ public Blockchain(IStore store)
 
         public IStore Store { get; }
 
+        public Block<T> Tip => this[-1];
```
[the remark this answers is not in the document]
longfin: Yes.

Libplanet/Net/Swarm.cs
longfin: No, `newest` is most far from genesis. 

This condition means that if `oldest` in received blocks is next to our tip, we can connect it directly.

longfin:
In fixing https://github.com/planetarium/libplanet.net/pull/59#discussion_r255567497, I realize current `Swarm`'s interface is little weird. so I'll do the related work in #69 .

longfin:
I've rebased due to merging #63. PTAL @dahlia @kijun

PR #70, opened 2019-02-12
Loosen Block.Timestamp constraints (#65)

### LINKED ISSUE #65: Loosen constraint about Block.Timestamp
Currently we require that the generated block has a timestamp higher than the timestamp of the previous block. there are a few things to consider.

- `Block.Timestamp` isn't always accurate because they are based on the local time of the system.
- Additional things (like #64 ) are required for automated testing, since the latter block requires a later timestamp.

I think we'd better loosen this constraint like Ethereum or Bitcoin style.

- In Bitcoin, a timestamp is accepted as valid if it is greater than the median timestamp of previous 11 blocks, and less than the network-adjusted time + 2 hours. ([see also](https://en.bitcoin.it/wiki/Block_timestamp))
- In Ethereum, a timestamp is accepted as valid if `block.timestamp <= now + 900` and is `block.timestamp >= parent.timestamp`. ([see also](https://github.com/ethereum/wiki/blob/c02254611f218f43cbb07517ca8e5d00fd6d6d75/Block-Protocol-2.0.md#block-validation-algorithm))

There are pros and cons, but I think the Ethereum style is more intuitive and easier to implement.

@dahlia @kijun @ipdae any ideas?

longfin:
It resolves #65. as mentioned in #65, it loosens the `Block.Timestamp`'s constraint [like Ethereum](https://github.com/ethereum/wiki/blob/c02254611f218f43cbb07517ca8e5d00fd6d6d75/Block-Protocol-2.0.md#block-validation-algorithm).

Also, the validation related to the current time has been moved from `Blockchain.Validate()` to `Block.Validate()`.

### DIFF
--- a/Libplanet.Tests/Blocks/BlockTest.cs
+++ b/Libplanet.Tests/Blocks/BlockTest.cs
@@ -2,6 +2,7 @@ using System;
 using System.Security.Cryptography;
 using Libplanet.Blocks;
 using Libplanet.Tests.Common.Action;
+using Libplanet.Tx;
 using Xunit;
 using static Libplanet.Tests.TestUtils;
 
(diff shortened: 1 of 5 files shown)


longfin:
PTAL @dahlia @kijun 

image

Libplanet.Tests/Blocks/BlockTest.cs
```diff
@@ -160,6 +160,23 @@ public void CanValidate()
...
+        [Fact]
+        public void CanDetectInvalidTimestamp()
+        {
+            var invalidBlock = new Block<BaseAction>(
+                index: _fx.Next.Index,
+                difficulty: _fx.Next.Difficulty,
+                nonce: new Nonce(new byte[] { 0x00 }),
+                rewardBeneficiary: _fx.Next.RewardBeneficiary,
+                previousHash: _fx.Next.PreviousHash,
+                timestamp: DateTime.UtcNow + TimeSpan.FromDays(1),
```
dahlia: It would be better to try to falsify the closest case to the boundary; how about testing against 901 seconds later rather than a day later?  It's also a good idea to check if timestamp 899 seconds later is valid.

PR #78, opened 2019-02-15
Make intervals between blocks configurable

dahlia:
This patch abstracts out the standard between what kind of blocks are valid and what are invalid from blockchain objects.  It separates `IBlockPolicy<T>` (`where T : IAction`) and its default implementation `BlockPolicy<T>` from `BlockChain<T>` (renamed from `Blockchain<T>`).  Hard-coded interval between blocks (it was 5 seconds) now can be configured through `BlockPolicy<T>.BlockInterval`.  A new namespace `Libplanet.Blockchain` is also introduced.

### DIFF
--- a/Libplanet/Blockchain/Policies/BlockPolicy.cs
+++ b/Libplanet/Blockchain/Policies/BlockPolicy.cs
@@ -23,7 +23,7 @@ namespace Libplanet.Blockchain.Policies
         /// <param name="blockIntervalSeconds">Configures
         /// <see cref="BlockInterval"/> in seconds.  5 seconds by default.
         /// </param>
-        public BlockPolicy(uint blockIntervalSeconds = 5)
+        public BlockPolicy(int blockIntervalSeconds = 5)
             : this(
                 TimeSpan.FromSeconds(blockIntervalSeconds)
             )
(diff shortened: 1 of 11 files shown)


Libplanet.Tests/Libplanet.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
longfin: Is there any change needed?

Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -0,0 +1,190 @@
...
+        where T : IAction
+    {
+        /// <summary>
+        /// Creates a <see cref="BlockPolicy{T}"/> with configuring
+        /// <see cref="BlockInterval"/> in seconds.
+        /// </summary>
+        /// <param name="blockIntervalSeconds">Configures
+        /// <see cref="BlockInterval"/> in seconds.  5 seconds by default.
+        /// </param>
+        public BlockPolicy(uint blockIntervalSeconds = 5)
```
longfin: Could we expose `int` instead of `uint`? (for CLS compliant)

Libplanet.Tests/Libplanet.Tests.csproj
dahlia: Rider repeatedly tries to insert BOM to *.csproj* files.  I'm going to trim this.

PR #80, opened 2019-02-17
API docs for Libplanet.Tx namespace and some adjustments

dahlia:
Wrote API docs for `Libplanet.Tx` namespace, and fixed some error-prone behaviors:

- The public constructor of `Transaction` had thrown `NullReferenceException` when some parameters of reference types.  It's changed to throw `ArgumentNullException` instead.  See also added tests.

- `Transaction.Signature` property had shared its internal state with others.  That means, if a `byte` array returned by the property or will be passed to the constructor mutates it affects the `Transaction` object too.  It's now changed to be copied when it is taken through the public constructor or returned through the property.  See also added tests.

    It maybe better to be `ImmutableArray<byte>` or to have its own type named `Signature`.

### DIFF
--- a/Libplanet.Tests/Tx/TransactionTest.cs
+++ b/Libplanet.Tests/Tx/TransactionTest.cs
@@ -629,6 +629,37 @@ namespace Libplanet.Tests.Tx
             tx.Validate();
         }
 
+        [Fact]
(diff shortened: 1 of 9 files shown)


dahlia:
@longfin I rebased it upon the current master.  Could you review this again?

PR #87, opened 2019-02-18
Contributor guide and BOM checker

dahlia:
I wrote a small shell script to check if any text files have BOM (and trim with `--apply` option), and some guide for contributors.


dahlia:
@kijun @longfin I rebased this on the current master.  Please approve this again.

PR #88, opened 2019-02-18
Network polishing

longfin:
This PR contains some enhancements / bug fixes that occured while testing on Unity3d as below.

- Replaced `Swarm._listenUrl` with `_ipAddress` (optional) and `_listenPort`(also optional) for easy configuration.
- Added some log messages for debugging.
- Exposed broadcast related methods (`Swarm.BroadcastBlocksAsync()` / `Swarm.BroadcastTxsAsync()`)
- Made to create a new DealerSocket for each request to avoid response overlapping.
- Fixed an issue where `Swarm` does not correctly respond to peer that are registered but never actually connected.
- Fixed `ProcessBlockHash()`

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -1213,6 +1213,9 @@ namespace Libplanet.Net
             Peer original = peer;
             var connected = false;
 
+            // Peer can have more than 2 Url due to NAT and so on.
+            // Therefore, DialPeerAsync () checks the accessible URLs
+            // and puts them at the front.
             foreach (var (url, i) in peer.Urls.Select((url, i) => (url, i)))
             {
                 var dealer = new DealerSocket();
@@ -1362,6 +1365,7 @@ namespace Libplanet.Net
                 throw new ArgumentNullException(nameof(peer));
             }
 
+            // See DialPeerAsync().
             Uri url = peer.Urls.FirstOrDefault();
             return url?.ToString().TrimEnd('/');
         }
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -6,6 +6,8 @@ using System.Collections.Immutable;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Net;
+using System.Net.Sockets;
 using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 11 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -81,9 +89,14 @@ public partial class Swarm : ICollection<Peer>, IDisposable
...
             _receiveMutex = new AsyncLock();
+            _blockSyncMutex = new AsyncLock();
 
+            _ipAddress = ipAddress ?? GetLocalIPAddress();
+            _listenPort = listenPort;
+
+            string loggerId = _privateKey.PublicKey.ToAddress().ToHex();
             _logger = Log.ForContext<Swarm>()
-                .ForContext("Swarm_listenUrl", _listenUrl.ToString());
+                .ForContext("Swarm_id", loggerId);
```
dahlia: What's the convention?

longfin: There is no convention at this time. 😢  I'll rename to `SwarmId`.

Libplanet/Net/Swarm.cs
```diff
@@ -426,47 +509,41 @@ CancellationToken cancellationToken
...
+                        {
+                            Block<T> block = Block<T>.FromBencodex(
+                                blockMessage.Payload);
+                            await yield.ReturnAsync(block);
+                            hashCount--;
+                        }
+                        else
+                        {
+                            throw new InvalidMessageException(
+                                $"The response of getdata isn't block. " +
```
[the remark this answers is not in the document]
longfin: No, it's just mistake. I'll fix it.

longfin:
I've rebased to apply reviews and add a test case.

longfin:
I've found some problem in `SwarmTest.CanBroadcastWhileMining()` and fixed it.

Libplanet/Net/Swarm.cs
```diff
@@ -33,37 +35,43 @@ public partial class Swarm : ICollection<Peer>, IDisposable
...
-            TimeSpan? dialTimeout = null,
+            TimeSpan dialTimeout,
+            IPAddress ipAddress = null,
+            int? listenPort = null,
             DateTimeOffset? createdAt = null)
         {
-            _privateKey = privateKey;
-            _listenUrl = listenUrl;
-            _dialTimeout = dialTimeout ?? TimeSpan.FromMilliseconds(15000);
+            _runningEvent = new TaskCompletionSource<object>();
```
dahlia: This seems redundant; this is initialized at the bottom of the constructor:

```c#
Running = false;
```

This property invokes its setter, which creating a new instance of `TaskCompletionSource<object>`:

```c#
public bool Running
{
    get => _runningEvent.Task.Status == TaskStatus.RanToCompletion;

    private set
    {
        if (value)
        {
            _runningEvent.TrySetResult(null);
        }
        else
        {
            _runningEvent = new TaskCompletionSource<object>();
        }
    }
}
```

longfin: The exact reason is unknown, but it seems that the destructor is called before calling the setter that you talked about. (#97) so I've put this line here as workaround.

But, moving set timing and eliminating redundant is an orthogonal problem, so I'll put `Running = false` here.

PR #89, opened 2019-02-19
Replace DateTime with DateTimeOffset

### LINKED ISSUE #77: Replace DateTime with DateTimeOffset
[`System.DateTime`][1], which has been there since very first of .NET Framework, has a concept of [`Kind`][2].  This property could be one of [three states][3]:

- `Local`
- `Unspecified` → This is default.
- `Utc`

It even does not hold any information about time zone even if its `Kind` is `Local`.  According to the official docs:

> In a time zone-aware application, you must rely on some external mechanism to determine the time zone in which a [`DateTime`][1] object was created. You could use a structure that wraps both the [`DateTime`][1] value and the [`TimeZoneInfo`][4] object that represents the [`DateTime`][1] value's time zone.

This leads us bug-prone.  Instead of this archaic type, we'd better use [`System.DateTimeOffset`][5] instead.   This type actually was born to be an alternative to the existing [`DateTime`][1] and indeed fixed the problem.  According to the docs:

> Because it exactly defines a date and time relative to UTC, the [`DateTimeOffset`][5] structure does not include a `Kind` member, as the [`DateTime`][1] structure does.

Note that it is not a pair of date time and time zone, but merely a UTC assuming date time:

> Although a [`DateTimeOffset`][5] value includes an offset, it is not a fully time zone-aware data structure. While an offset from UTC is one characteristic of a time zone, it does not unambiguously identify a time zone. Not only do multiple time zones share the same offset from UTC, but the offset of a single time zone changes if it observes daylight saving time. This means that, as soon as a [`DateTimeOffset`][5] value is disassociated from its time zone, it can no longer be unambiguously linked back to its original time zone.

This might be bad for end-user-facing applications, for us this makes things rather simpler. 

Although it is relatively newer than [`DateTime`][1], it still has been in .NET Framework since the version 2.0 and is part of .NET Standard 1.0 as well.

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.datetime
[2]: https://docs.microsoft.com/en-us/dotnet/api/system.datetime.kind#System_DateTime_Kind
[3]: https://docs.microsoft.com/en-us/dotnet/api/system.datetimekind
[4]: https://docs.microsoft.com/en-us/dotnet/api/system.timezoneinfo
[5]: https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset

earlbread:
It resolves #77.
I replaced all `DateTime` with `DateTimeOffset`, except `ToDateTime` in `Libplanet/Serialization/BencodexFormatterConverter.cs`.

### DIFF
--- a/Libplanet.Explorer/ViewModels/TransactionViewModel.cs
+++ b/Libplanet.Explorer/ViewModels/TransactionViewModel.cs
@@ -9,7 +9,7 @@ namespace Libplanet.Explorer.ViewModels
     {
         public string Id { get; set; }
         public byte[] Signature { get; set; }
-        public DateTime Timestamp { get; set; }
+        public DateTimeOffset Timestamp { get; set; }
         public string Sender { get; set; }
         public string Recipient { get; set; }
         public IEnumerable<Dictionary<string, object>> Actions { get; set; }
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -106,7 +106,7 @@ namespace Libplanet.Tests.Blockchain
                 new PrivateKey(),
                 _fx.Address1,
                 actions1,
-                DateTime.UtcNow
+                DateTimeOffset.UtcNow
             );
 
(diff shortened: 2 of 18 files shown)


dahlia:
Thanks for your contribution!

PR #90, opened 2019-02-19
Awaiting method to ensure a Swarm instance is running

dahlia:
Many tests in the `SwarmTest` class assume `Swarm` objects are running after spawn them through threads.  The assumption is usually true for the most part, but on some slow machines like Travis CI workers, spawning a new thread and a NetMQ socket binding take longer time, so that tests get run in the main thread before things are all prepared.

In order to make tests to satisfy the assumption, I added a new method named `Swarm.WaitForRunning()` which returns a `Task` completed when `Swarm.Running` becomes `true`.  I also a façade method named `SwarmTest.StartAsync()` to eliminate duplicates.

Hope this patch to stabilize CI builds.

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -99,7 +99,7 @@ namespace Libplanet.Tests.Net
             BlockChain<BaseAction> chain = _blockchains[0];
 
             await swarm.StopAsync();
-            Task task = await StartAsync(swarm, chain, 250);
+            Task task = await StartAsync(swarm, chain);
 
             Assert.True(swarm.Running);
             await swarm.StopAsync();
@@ -126,7 +126,7 @@ namespace Libplanet.Tests.Net
 
             Task producerTask = Task.Run(async () =>
             {
-                await swarm.StartAsync(chain, 250);
+                await swarm.StartAsync(chain);
             });
 
             await consumerTask;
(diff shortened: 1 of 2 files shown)


Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -128,9 +155,9 @@ public async Task WorksAsExpected()
             {
                 try
                 {
-                    var at = Task.Run(async () => await a.StartAsync(chain, 250));
-                    var bt = Task.Run(async () => await b.StartAsync(chain, 250));
-                    var ct = Task.Run(async () => await c.StartAsync(chain, 250));
+                    await StartAsync(a, chain, 250);
```
longfin: I think we can remove `250`. (it didn't have serious meaning 😃 )

Libplanet/Net/Swarm.cs
```diff
@@ -256,7 +270,7 @@ public void CopyTo(Peer[] array, int arrayIndex)
 
                 _dealers.Clear();
 
-                Running = false;
+                _runningEvent = new TaskCompletionSource<object>();
```
longfin: It seems to be able to use `setter` of `Running` to eliminate duplicates. (it's trivial.)

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -15,6 +15,7 @@
 using Libplanet.Tests.Common.Action;
 using Libplanet.Tests.Store;
 using Libplanet.Tx;
+using Nito.AsyncEx;
```
longfin: Do we need this using statement?

dahlia: No, this was a mistake. I'll remove this.

Libplanet.Tests/Net/SwarmTest.cs
dahlia: Okay, it will leave them omitted.  👌

Libplanet/Net/Swarm.cs
dahlia: 👌

PR #92, opened 2019-02-20
Prevent AppDomain to workaround Xunit's fatal error on Mono

dahlia:
As Mono has no proper AppDomain, we prevent it. This works around Xunit's fatal error on Mono.

See also <https://github.com/xunit/xunit/issues/1357>.


dahlia:
On the other hand, Xunit seems to crash without `AppDomain` for uncertain reason on .NET Framework, so I changed the MSBuild script to deny `AppDomain` only on Mono.

@longfin Please review this one more time.

PR #93, opened 2019-02-21
Added BlockChain.Fork()

longfin:
It introduces `BlockChain.Fork()` to handle branch on block receiving safely. to accomplish that, it also added namespace concept to `IStore` / `FileStore` (using subdirectory).

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -17,13 +17,23 @@ namespace Libplanet.Blockchain
     public class BlockChain<T> : IEnumerable<Block<T>>
         where T : IAction
     {
+        private readonly Guid _id;
+
         public BlockChain(IBlockPolicy<T> policy, IStore store)
+            : this(policy, store, Guid.NewGuid())
         {
+        }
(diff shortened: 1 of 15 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -17,13 +17,23 @@ namespace Libplanet.Blockchain
     public class BlockChain<T> : IEnumerable<Block<T>>
         where T : IAction
     {
+        private readonly Guid _id;
+
         public BlockChain(IBlockPolicy<T> policy, IStore store)
+            : this(Guid.NewGuid(), policy, store)
```
dahlia: As the `id` parameter is optional in users' perspective, shouldn't it be the last parameter?

longfin: User can see only `BlockChain(policy, store)` because `BlockChain(id, policy, store)` is private. 
but I also think that moving `id` to last as you say is better. I'll change.

longfin:
I've missed catching master before pushed this, so I rebased. cc @dahlia

PR #103, opened 2019-02-23
Add BlockIndex to ActionContext. Closes #102

### LINKED ISSUE #102: Add block height to `ActionContext`.
Add block height to `ActionContext` so that game logic can reference it.

My usecase for this it to make a game where a player is only permitted one action per block.

qria:
Haven't tested it yet but it probably works.


Libplanet.Tests/Action/ActionContextTest.cs
```diff
@@ -18,6 +18,7 @@ public void RandomShouldBeDeterministic()
                 var context = new ActionContext(
                     from: new Address("21744f4f08db23e044178dafb8273aeb5ebe6644"),
                     to: new Address("21744f4f08db23e044178dafb8273aeb5ebe6644"),
+                    blockIndex: 1,
```
longfin: It would be better that adding blockIndex related assertion in here (or, adding other test case seems also good.)

qria: @longfin Done and done

Issue #105, opened 2019-02-25
Renaming the repository name to libplanet (no trailing “.net”)?


dahlia:
We do not have any plan to support or rewrite Libplanet in other languages than C# in the near future.  To be consistent with the names for branding, NuGet package, `namespace`, and so on, I suggest rename the repository also to planetarium/libplanet, no “.net” postfix.

kijun:
agreed

longfin:
Basically, I think that even if there are implementations in other languages / platforms, names that are not awkward are better.

But in reality

- As you said, we're not preparing implementations in other languages / platforms in near term.
- Even if other language / platform implementations are released, they will use names that are distinguishable from existing `Libplanet`. (e.g. `libplanet++`,  `libplanet-python`)

so I don't think that is a big problem.

dahlia:
I'm going to rename the repository to planetarium/libplanet in 24 hours.

dahlia:
Renamed the repository.  Redirection from existing links seems to work well, except of GitHub Pages: <https://planetarium.github.io/libplanet.net/>.  We should manually replace the links to the old name with <https://planetarium.github.io/libplanet/>.

dahlia:
Give the docs new url: <https://docs.libplanet.io/>.

PR #107, opened 2019-02-25
Replace links to the old repository name with the new name

### LINKED ISSUE #95: Bare http:// link in the NuGet package description
https://github.com/planetarium/libplanet.net/blob/3b425b2fd067a01af6d35f217a2d61309dd1462d/Libplanet/Libplanet.csproj#L9

it would be better `https://` rather than `http://`. (it may be typo.)

dahlia:
As we discussed in #105, I renamed the repository from planetarium/libplanet.net to planetarium/libplanet.  Although GitHub maintains the old links and redirects them to the new links, the redirection does not work for GitHub Pages.  With this situation, I gave the new domain docs.libplanet.net to the docs, and replaced all old links to the docs with the new domain.

This fixes #95 and #105.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,7 @@ builds the whole Libplanet solution:
     msbuild -r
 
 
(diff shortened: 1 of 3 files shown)


dahlia:
Rebased this on the current master.

PR #109, opened 2019-02-26
Release 0.1.0

dahlia:
If this is merged I'm going to tag the merge commit with 0.1.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,12 @@ For a merge commit build, a version name is like `0.1.0-dev.123+20181231` where
 `123` is a build number in our CI system and `20181231` is a date of the build.
 For a daily build, a version name is like `0.1.0-nightly.20181231`.
 
+Unfortuneatly, Unity currently does not support NuGet.  There are some Unity
+plug-ins to deal with NuGet package system, and these seem immature at present.
+To use Libplanet on Unity, you need to manually extract *Libplanet.dll*
+from *Libplanet.\*.nupkg* file and place it inside of your Unity project.
+We are acknowledging the fact Libplanet is currently not very usable together
+with Unity, and promise to make it better in next few minor releases.
 In the near future, we are going to submit it to [Unity Asset Store] too.
 
 [releases]: https://github.com/planetarium/libplanet/releases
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,17 +25,36 @@ server.  Under the hood, it incorporates many features (e.g.,
...
 -----
 
-Although this project has never released any version, for everyday and every
-merge commit, it is packed into a *.nupkg* and uploaded to [NuGet] with
-a hyphen-suffixed version name.
+For every stable release, we pack Libplanet into a *.nupkg* and
+upload it to [NuGet] and GitHub [releases] page.
+(You can find changelog for versions from [releases] page.)
+To use Libplanet in your game, your project needs to add a dependency
+to *Libplanet* package.  On Visual Studio IDE, run the following command
```
longfin: This guide can be quite tricky because...

- I don't know how familiar Unity developers are with nuget.
- Actually, this command will not work on a Unity + VS, because VS dependencies aren't resolved as Unity asset properly.

dahlia: How should this guide be rewritten instead then?  At this moment, we do not pack it as a Unity asset…

longfin: It would be nice to mention that it's currently on nuget so that you can use it in common .NET applications, and that we are preparing for the build for Unity.

dahlia: FYI although it's not added in this patch, the following statement has been already there:

> In the near future, we are going to submit it to [Unity Asset Store] too.

I'm going to explain the current status that it's not very usable with Unity yet.

[Unity Asset Store]: https://assetstore.unity.com/

dahlia: I added the following paragraph (1d1a4dc4d6f338500215c266e957be039335f0c7):

> Unfortuneatly, Unity currently does not support NuGet.  There is some Unity plug-ins to deal with NuGet package system, these seem immature at present. To use Libplanet on Unity, you need to manually extract *Libplanet.dll* from *Libplanet.\*.nupkg* file and place it inside of your Unity project. We are acknowledging the fact Libplanet is currently not very usable together with Unity, and promise to make it better in next few minor releases. In the near future, we are going to submit it to [Unity Asset Store] too.

[Unity Asset Store]: https://assetstore.unity.com/

kijun:
Wooooooooooot 🎉🎉 🎉 🎉 🎉 🎉 🎉

Issue #73, opened 2019-02-13
Connecting P2P over NAT


longfin:
Currently, the implementation of P2P through `Swarm` assumes that the peer can connect via TCP with public IP. however, clients targeted by libplanet are not directly connected to the Internet with such public IP, but are often connected through NAT. In such a connection method, it is often difficult for one peer to directly access another peer.

Using techniques such as [UDP hole punching](https://en.wikipedia.org/wiki/UDP_hole_punching), it is necessary to support natural connections in these environments.

longfin:
I've researched about it and leave note as below.

- Option 1: we can modify `Swarm` to act as a repeater.
  - it changes `Swarm`'s basic communications to broadcast and peers with public IP help communication with peer that don't
  - there is no need for a external server or entity, but the implementation is complicated and the load on peer with public IP is increased.
  - to be honest, I'm not sure if the implementation is possible. 😕 
- Option 2: we can modify `Swarm` to rely [STUN] / [TURN] protocols.
  - `Swarm` uses STUN to determine its public IP and NAT type to traversal.
  - `Swarm` uses TURN to send actual packet with relay.
  - although it requires two type of external servers (STUN / TURN), it's an open protocol as rfc and can be operated by anyone who wants to operate (including us).
    - but there is not a lot of public TURN servers (compared to STUN).

[STUN]: https://en.wikipedia.org/wiki/STUN
[TURN]: https://en.wikipedia.org/wiki/Traversal_Using_Relays_around_NAT

longfin:
I've made #117 for sharing progress to other teammates.

longfin:
Related PRs have been merged. (#117 , #130) so let's close this issue and try it for real game implementation. 😄

PR #117, opened 2019-03-07
TURN relay implementation

longfin:
This PR introduces [TURN](https://tools.ietf.org/html/rfc5766) client as relay mechanism to accomplish [NAT traversal](https://en.wikipedia.org/wiki/NAT_traversal). (see also #73) it consists of three parts as below.

- Added some helpers
  - CRC32
  - bytes ↔ uint (for masking) converter
  - Proxy for NetworkStream
- Added STUN / TURN message protocols
  - STUN / TURN attributes
  - STUN / TURN messages
- Added TURN protocol client

Currently, it doesn't have any open interface. so I'll add `[skip changelog]`. (perhaps it is necessary to write in the next PR.)

# TODO
- [x] Make the test case optional
- [x] ~~Fill STUN / TURN enum description.~~, I'll handle this at next PR.


Libplanet/Net/Stun/TurnClient.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+            while (response is AllocateErrorResponse && retry < AllocateRetry);
+
+            if (response is AllocateSuccessResponse allocOk)
+            {
+                return allocOk.RelayedEndPoint;
+            }
+            else
+            {
+                throw new TurnClientException("Allocate Failed.", response);
```
dahlia: Although it's trivial, F in the second word should be small, I think?

Libplanet/Net/Stun/TurnClient.cs
```diff
@@ -0,0 +1,154 @@
...
+using System.Net;
+using System.Net.Sockets;
+using System.Threading.Tasks;
+using Libplanet.Net.Stun.Messages;
+
+namespace Libplanet.Net.Stun
+{
+    internal class TurnClient : IStunContext, IDisposable
+    {
+        private const int AllocateRetry = 5;
```
dahlia: Is this an arbitrary number we decided?

Libplanet/Net/Stun/Messages/ConnectRequest.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
+using System.Net;
+using Libplanet.Net.Stun.Attributes;
+
+namespace Libplanet.Net.Stun.Messages
+{
+    internal class ConnectRequest : StunMessage
+    {
+        public ConnectRequest(IPEndPoint peerAddr)
```
dahlia: We'd better to have a consistent term for `IPEndPoint`.  Sometimes we name `IPEndPoint` values “address” and sometimes “end point.”

Libplanet/Net/Stun/Messages/CreatePermissionSucessResponse.cs
```diff
@@ -0,0 +1,9 @@
+namespace Libplanet.Net.Stun.Messages
+{
+    internal class CreatePermissionSucessResponse : StunMessage
```
dahlia: A typo here.  S should be twice in “sucess” → “success.”

Libplanet/Net/Stun/TurnClient.cs
longfin: Yes, it's an arbitrary.

ipdae:
😢 Window build is broken.

longfin:
> * How long would it take if we separate `Libplanet.Net.Strun`'s own project from Libplanet project?

Can't make sure, but I'll try on next PR.

> * Could we omit those `Can` prefixes from test methods when it's appropriate?

Actually, I'm confused about the naming convention for test cases 😕 .

longfin:
I've rebased to apply reviews.

And, I removed `Can` prefix in test cases. (I'm not sure about the convention yet, but I'm sure it's more natural than having a `Can`)

Issue #98, opened 2019-02-22
Replace AddressStateMap with IImmutableDictionary<Address, object>


dahlia:
The current implementation of `AddressStateMap` is just a thin wrapper around `ImmutableDictionary<Address, object>`, and implements `IImmutableDictionary<Address, object>` as well at a time.  This was intended to mimic `typedef IImmutableDictionary<Address, object> AddresStateMap;` in C#. (Although C# allows `using AddresStateMap = IImmutable<Address, object>;`, the alias is scoped to only a file.)  However, as `AddressStateMap` is not completely equivalent to `IImmutableDictionary<Address, object>` but only a subtype of it, it tends to cast things up and down, which leads runtime type unsafety.  For instance, when we implement a `IAction.Execute()` method, it looks like:

```c#
public AddressStateMap Execute(IActionContext context)
{
    AddressStateMap states = context.PreviousStates;
    IImmutableDictionary<Address, object> newStates =
        states.SetItem(context.To, "new value");
    return (AddressStateMap) newStates;  // downcast
}
```

If we completely replace `AddressStateMap` with `IImmutableDictionary<Address, object>` it would look like instead:

```c#
public AddressStateMap Execute(IActionContext context)
{
    IImmutableDictionary<Address, object> states = context.PreviousStates;
    IImmutableDictionary<Address, object> newStates =
        states.SetItem(context.To, "new value");
    return newStates;  // no cast
}
```

longfin:
@dahlia Is this still a valid issue?

dahlia:
Yes, it is still valid in some degree, but the ideas I suggested above are now unnecessary anymore for the most part.  I am going to make a patch to completely close this issue soon!

PR #121, opened 2019-03-08
Refactor the interface for actions updating states

dahlia:
This patch retouched `IAction` and `Transaction<T>` APIs to make library users easier to implement their own `IAction` classes.  Two key changes are:

 -  `Transaction<T>`'s `Sender`–`Recipient` model was replaced by `Signer`–`UpdatedAddresses` model.   Unlike cryptocurrencies, transactions in games are not necessarily a transfer of assets, so it is difficult to determine what type of assert is transferred or who will receives the asset.  A more useful perspective is, like what kind of transformation is performed, or what states are changed.

    To be close to this perspective, we decided to get rid of `Transaction<T>.Recipient` (which is singular) and have `Transaction<T>.UpdatedAddresses` (which is plural) instead. As there is no more asset to transfer, the term `Sender` was also renamed to `Signer`, which fits more to the new perspective.

 -  The `Address`es `IAction` tries to update no more need to be manually coded using `IAction.RequestStates()` method.  That method was removed at all, and updated `Address`es became automatically scanned using dirty-tracking for the most cases.

More detailed changes are written in *CHANGES.md* file.

----

### Tasks

- [x] `BlockChain` needs to validate if all `Transaction<T>.UpdatedAddresses` match to the actual addresses affected by `Transaction<T>.Actions`.   Note that as it depends on precedent states, cannot be done by `Transaction<T>` in itself.
- [x] `Block<T>.Validate()` had been not validating its `Transactions`.  `Transaction<T>.Validate()` also should be called inside that method.
- [x] Currently `UpdatedAddresses` must be coded by human when a `Transaction<T>` is instantiated.  There should be a façade factory to play the given actions and automatically determines the updated addresses.
- [ ] Get rid of the whole `AddressStateMap` thing?  It no differs from `IImmutableDictionary<Address, object>`.  See also #98.
- [x] Changelogs should be more organized and the whole change need to be explained at a higher level.


dahlia:
@longfin Could you review this?

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -3,6 +3,7 @@
 using System.Collections.Immutable;
 using System.Drawing.Design;
 using System.Linq;
+using System.Runtime.CompilerServices;
```
longfin: Is this using statement necessary?

Libplanet/Tx/Transaction.cs
```diff
@@ -225,62 +236,147 @@ public static Transaction<T> FromBencodex(byte[] bytes)
...
+
+            if (!actionsArray.IsEmpty)
+            {
+                IAccountStateDelta delta = new Transaction<T>(
+                    signer,
+                    publicKey,
+                    updatedAddresses,
+                    ts,
+                    actionsArray
+                ).EvaluateActions(
```
longfin: It's important to point out that `IAction.Execute()` is being executed in `Transaction<T>.Create()`.

Libplanet.Tests/Blockchain/BlockChainTest.cs
dahlia: No, it's mistaken.  I will remove them!

Libplanet/Tx/Transaction.cs
dahlia: It's for filling out `UpdatedAddresses` automatically, and documented in `<remarks>` section.

dahlia:
@longfin I amended a commit to fix the build failure on Windows. Can you approve this again?

PR #123, opened 2019-03-09
Change Peer Uri to IPEndPoint

### LINKED ISSUE #120: Change type of Peer.Urls to IPEndPoint
We've chosen [`Uri`](https://docs.microsoft.com/en-US/dotnet/api/system.uri?view=netstandard-2.0) as element type of `Peer.Urls`  to represent peer's address. but `Uri` isn't appropriate type because it also contains information that we do not need. (e.g. path, protocol)

So, I suggest [`IPEndPoint`](https://docs.microsoft.com/en-US/dotnet/api/system.net.ipendpoint?view=netstandard-2.0) to peer's endpoint type.  it has only IP address and port.

ipdae:
resolve #120

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ Version 0.2.0
 
 To be released.
 
- -  Change type of Peer.`Urls` to `IPEndPoint`.
+ -  Change type of `Peer.Urls` to `Peer.EndPoints`.
 
 
 Version 0.1.1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ Version 0.2.0
 
 To be released.
 
- -  Change type of `Peer.Urls` to `Peer.EndPoints`.
+ -  The type of `Peer.Urls` property is changed from `Uri` to `IPEndPoint`.
 
 
 Version 0.1.1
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -6,6 +6,8 @@ Version 0.2.0
 
 To be released.
 
+ -  Change type of Peer.`Urls` to `IPEndPoint`.
```
dahlia: Please write the sentence in past tense and I think `Peer.Urls` should be quoted altogether.

> The type of `Peer.Urls` property is changed from `Uri` to `IPEndPoint`.

Libplanet/Net/Peer.cs
```diff
@@ -1,77 +1,89 @@
...
+            new Peer(PublicKey, EndPoints.Add(endPoint));
 
         public void GetObjectData(
             SerializationInfo info,
             StreamingContext context
         )
         {
             info.AddValue("public_key", PublicKey.Format(true));
-            info.AddValue("urls", Urls.Select(u => u.ToString()).ToList());
+            info.AddValue(
```
longfin: I think we can accomplish it just like this. (please let me know if it doesn't works.)

```
info.AddValue("end_points", EndPoints.ToList());
```

ipdae: it raise `InvalidCastException` when deserialize `Peer`. because `IPEndPoint` is not implement `IConvirtible`.

Issue #126, opened 2019-03-11
Turn Peer.Endpoints to Peer.Endpoint


longfin:
We've assumed that `Peer` can have multiple endpoints to publish its own network address. but in #73 and #117 , we planed to using TURN to relay server, and using its relayed address to peer's connectable address. With this plan, it is more natural to assume that `Peer` has the only one `EndPoint` to which it can accept other peer.

PR #127, opened 2019-03-11
Peer.EndPoints to Peer.EndPoint

longfin:
This PR replaces `Peer.EndPoints` with `Peer.EndPoint`. (as suggested in #126)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,8 @@ Version 0.2.0
 
 To be released.
 
- -  The type of `Peer.Urls` property is changed from `Uri` to `IPEndPoint`.
+ - Since we decided to depend on TURN ([RFC 5766]) and STUN ([RFC 5389]) to work around NAT so that `Peer`'s endpoints don't have to be multiple,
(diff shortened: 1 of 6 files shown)


longfin:
Some tests seem to fail. I'll fix it.

CHANGES.md
```diff
@@ -6,7 +6,7 @@ Version 0.2.0
 
 To be released.
 
- -  The type of `Peer.Urls` property is changed from `Uri` to `IPEndPoint`.
+ - `Peer.Urls` is changed to `Peer.EndPoint` represented as single `IPEndPoint`.
```
dahlia: There were a renaming and a retyping, so we'd better to sum up the explanation at higher view, with a link to the issue #120 and #126.

>  -  Since we decided to depend on TURN ([RFC 5766]) and STUN ([RFC 5389]) to work around NAT so that `Peer`'s endpoints don't have to be multiple, `Peer.Urls` was renamed to `Peer.EndPoint` and its type also was changed from `IImmutableList<Uri>` to `IPEndPoint`.  [#120, #126]

[RFC 5766]: https://tools.ietf.org/html/rfc5766
[RFC 5389]: https://tools.ietf.org/html/rfc5389

Issue #131, opened 2019-03-19
Actions should be able to know if it's “rehearsal mode” during execution


dahlia:
The previously merged patch #121 introduced the concept of “rehearsal mode” (although I named it so right now) in order to automatically determine the `Transaction<T>.UpdatedAddresses` property.  However, as this mode does not provide any `PreviousStates` and mocks `Random` and `BlockIndex` (which is set to 0 during that mode), it tends to lead two quite different consequences for both modes.  Although the logical structure of an action should be consistent whether it's rehearsal or not, in some cases we need to make conditions depending on the mode.

In my opinion, the easiest way to implement would be to add a Boolean property like `IActionContext.Rehearsal`.

PR #135, opened 2019-03-20
IActionContext.Rehearsal property

dahlia:
This implements #131.


dahlia:
> When naming this type of action, I heard a log more of dry-run. Is there a particular reason for choosing Rehearsal?

No particular reason.  Just the word “rehearsal” came into my mind first.  The word “dry-run” is also difficult to conjugate to me because it is basically a two-word verb and we need to insert a hyphen in the middle in order to make it noun.

Issue #132, opened 2019-03-19
Exceptions thrown by actions during rehearsal mode should be wrapped


dahlia:
As I explained in the issue #131, rehearsal mode tends to lead an action to reach an unexpected case.  To a library user, it is likely to seem like it's their fault (I mean, it looks like the action won't work without rehearsal mode too, which is not necessarily so).

So that users can distinguishes errors due to rehearsal mode from other general errors, exceptions thrown by actions during that mode should be wrapped, like `UnexpectedlyTerminatedTxRehearsalException` with an inner exception.

PR #136, opened 2019-03-20
UnexpectedlyTerminatedTxRehearsalException

dahlia:
This implements #132.

### DIFF
--- a/Libplanet/Tx/UnexpectedlyTerminatedTxRehearsalException.cs
+++ b/Libplanet/Tx/UnexpectedlyTerminatedTxRehearsalException.cs
@@ -1,0 +1,41 @@
+using System;
+using Libplanet.Action;
+
+namespace Libplanet.Tx
+{
+    /// <summary>
+    /// The exception that is thrown during a <see cref="Transaction{T}"/>
+    /// is being created when one of <see cref="Transaction{T}.Actions"/>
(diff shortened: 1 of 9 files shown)


Libplanet/Tx/UnexpectedlyTerminatedTxRehearsalException.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Libplanet.Tx
+{
+    /// <summary>
+    /// The exception that is thrown during a <see cref="Transaction{T}"/>
+    /// is being created when one of <see cref="Transaction{T}.Actions"/>
+    /// throws some exception during rehearsal mode.
+    /// <para>The actual exception that the <see cref="Action"/> threw
+    /// is stored in the <see cref="Exception.InnerException"/> property.</para>
+    /// </summary>
+    public class UnexpectedlyTerminatedTxRehearsalException : Exception
```
longfin: - Could we seal this class?
- It would be better if `[Serializable]` was attached to make it serializable.

dahlia: Both ideas are good.  I'll do that!

dahlia: @longfin I've just amended the last commit.

Issue #96, opened 2019-02-22
Provide an option to remove invalid actions (i.e., repair)


ipdae:
If a block is created with an invalid tx, a miner never works since then and shuts down with an error.  The only solution to recover the situation is to delete the whole blockchain directory on filesystem.  I think It would be nice if we provide an option to remove invalid actions like [`nekoyume repair`](https://github.com/nekoyume/nekoyume/blob/0.1.4/nekoyume/cli.py#L141).

PR #141, opened 2019-03-21
IStore.ListNamespaces() method

dahlia:
Although I frankly am not sure if it's really necessary, I added a new method `ListNamespaces()` to `IStore` and implemented `FileSystem.ListNamespaces()` too.  I am going to use this to make an ad-hoc solution to #96.


dahlia:
@longfin @earlbread I rebased this on the current master and added a changelog too. Could you review this again?

PR #143, opened 2019-03-22
Fix NuGet packaging on Travis CI

dahlia:
The arguments syntax of `nuget pack` seems quite picky.  According to the [official docs][1], `-Properties` has to go last and its items have to be separated by semicolons.

[1]: https://docs.microsoft.com/en-us/nuget/tools/cli-ref-pack#options


dahlia:
@earlbread @longfin I added more commits.  Could you review this again?

Issue #28, opened 2019-01-07
Should check and warn if an action type is not annotated with ActionTypeAttribute


dahlia:
It currently crashes without any warning or meaningful error message if an action type is not annotated with `ActionTypeAttribute`.  This behavior can confuse library users whether they are aware that an action should be annotated with the attribute and just miss it or not aware of it at all.

- We should make`Blockchain<T>` to throw a runtime exception (e.g., `InvalidActionTypeException`) with a well-guided error message if `T` is not annotated with the attribute.
- We could write a small [Roslyn Analyzer][1] so that this situation is warned at compile-time.

(Reported by @ipdae.)

[1]: https://github.com/dotnet/roslyn-analyzers

dahlia:
Since the patch #169, `ActionTypeAttribute` has been no more used by all `IAction` implementations, but only actions used with `PolymorphicAction<T>`.  Currently, a `MissingActionTypeException` is thrown when an action type which lacks `ActionTypeAttribute` is passed to `PolymorphicAction<T>` class.  This issue is not closed because we don't have a Roslyn Analyzer to detect this at compile-time.

PR #144, opened 2019-03-22
Throw exception when action type is not annotated

earlbread:
- Made threw an `InvalidActionTypeException` if an action type is not annotated with `actionTypeAttribute`. #28

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -99,6 +99,8 @@ To be released.
     from `IImmutableList<Uri>` to `IPEndPoint`.
     [[#120], [#123] by Yang Chun Ung, [#126], [#127]]
  -  Added `IStore.ListNamespaces()` method.
+ -  `Transaction<T>` now throws an `InvalidActionTypeException` if an action type
+    is not annotated with `ActionTypeAttribute`.  [#144]
 
 [#98]: https://github.com/planetarium/libplanet/issues/98
 [#99]: https://github.com/planetarium/libplanet/issues/99
@@ -113,6 +115,7 @@ To be released.
 [#132]: https://github.com/planetarium/libplanet/issues/132
 [#135]: https://github.com/planetarium/libplanet/pull/135
 [#136]: https://github.com/planetarium/libplanet/pull/136
+[#144]: https://github.com/planetarium/libplanet/pull/144
 [RFC 5389]: https://tools.ietf.org/html/rfc5389
 [RFC 5766]: https://tools.ietf.org/html/rfc5766
 
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -279,6 +279,17 @@ namespace Libplanet.Tests.Blockchain
             Assert.Equal(states[TestEvaluateAction.BlockIndexKey], blockIndex);
         }
 
+        [Fact]
+        public void DetectInvalidActionType()
+        {
+            var privateKey = new PrivateKey();
+            var action = new ActionNotAttributeAnnotated();
+
+            Assert.Throws<InvalidActionTypeException>(
+                () => Transaction<BaseAction>.Create(
+                privateKey, new[] { action }));
+        }
+
         [ActionType("test")]
         private class TestEvaluateAction : BaseAction
         {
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -99,6 +99,8 @@ To be released.
     from `IImmutableList<Uri>` to `IPEndPoint`.
     [[#120], [#123] by Yang Chun Ung, [#126], [#127]]
  -  Added `IStore.ListNamespaces()` method.
+ -  `Transaction` now throws an `InvalidActionTypeException` if an action type
```
dahlia: Although it's trivial, can you refer to the type more specifically, like `Transaction<T>`?  In .NET, two different types of the same name can exist at a time in a namespace if one is generic and other one is non-generic (e.g., [`ImmutableDictionary<TKey,TValue>`][1] and [` ImmutableDictionary`][2]).

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2
[2]: https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -279,6 +280,21 @@ public void EvaluateActions()
             Assert.Equal(states[TestEvaluateAction.BlockIndexKey], blockIndex);
         }
 
+        [Fact]
+        public void CanDetectInvalidActionType()
```
dahlia: Could you omit the prefix `Can`?  We've recently started to omit that prefix.  See also [this](https://github.com/planetarium/libplanet/pull/117#pullrequestreview-213330287) and [this](https://github.com/planetarium/libplanet/pull/117#issuecomment-472338131).

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -303,5 +319,21 @@ public override IAccountStateDelta Execute(IActionContext context)
                     .SetState(BlockIndexKey, context.BlockIndex);
             }
         }
+
+        private class InvalidAction : BaseAction
```
dahlia: It would be nice if the class name is more self-descriptive.  There are many ways for `IAction` implementation to be invalid.  I suggest the name `ActionNotAttributeAnnotated`.

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Reflection;
```
dahlia: Is this actually used?

Libplanet/Tx/Transaction.cs
```diff
@@ -602,6 +602,12 @@ internal RawTransaction ToRawTransaction(bool includeSign)
         private static T ToAction(IDictionary<string, object> arg)
         {
             var typeStr = (string)arg["type_id"];
+            if (typeStr is null)
+            {
+                throw new InvalidActionTypeException(
+                    "Action type should be annotated with ActionTypeAttribute");
```
dahlia: It's more likely to be a serialization problem rather than the lack of attribute annotation.  It would be better to throw this exception when an action is serialized.

And coul you insert a period to the end of the statement?

Libplanet/Tx/InvalidActionTypeException.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace Libplanet.Tx
+{
+    [Serializable]
+    public class InvalidActionTypeException : Exception
```
longfin: Could we seal this exception? (or, should we open to inherit?)

Libplanet.Tests/Blockchain/BlockChainTest.cs
longfin: it seems to be needed to using `TargetInvocationException`.

earlbread: Yes, it is used for `TargetInvocationException`.

Libplanet/Tx/InvalidActionTypeException.cs
earlbread: Sure, it doesn't need to be inherited.

earlbread:
I fixed the code. Please take a look. @dahlia @longfin

Libplanet/Action/ActionTypeAttribute.cs
```diff
@@ -40,11 +41,20 @@ public ActionTypeAttribute(string typeIdentifier)
         /// <c>null</c>.</returns>
         public static string ValueOf(Type actionType)
         {
-            return actionType
+            var typeIdentifier = actionType
```
longfin: it would be better if the type of `typeIdentifier` is specified. we specify variable type unless the type of the variable is self-evident.

earlbread: I added type and amended the commit. Please take a look @longfin.

PR #147, opened 2019-03-22
Make .nupkg to include Libplanet.Stun.dll assembly

### LINKED ISSUE #146: .nupkg does not include Libplanet.Stun.dll assembly
```
error: Unable to find package Libplanet.Stun. No packages exist with this id in source(s): nuget.org
```

dahlia:
This fixes #146.

Note that I replaced all `/o` & `/Option` style options with `-o` & `-Option` style options.  It purposes to work around Cygwin's automatic conversion between Windows/DOS-style path and Unix-style path (or it might be done by MinGW64; I'm not sure).  Since a Windows path like *C:\foo\bar* is represented as */c/foo/bar* in Cygwin, when a command like `cmd /o` is executed, it tries to reinterpret this like `cmd o:\`.  Fortunately `msbuild` supports both styles of options, I decided to use only `-o` & `-Option` style options.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ install:
     # MSBuild
     {
       echo "#!/bin/bash"
-      echo "'/c/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/MSBuild/15.0/Bin/MSBuild.exe' $@"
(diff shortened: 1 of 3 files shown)


longfin:
It seems that `msbuild` in Windows [doesn't know `DebugSymbols:true`](https://travis-ci.com/planetarium/libplanet/jobs/187131051#L2952). (it's weird 😕)

dahlia:
@earlbread @longfin Could you review this?  The build passes now.

PR #154, opened 2019-03-26
Simplify file structure and improve forking performance

longfin:
This PR shares blocks and Tx files between forked chains to prevent unnecessary file creation and improve forking performance.

- Removed `@namespace` argument from block/tx related methods in `IStore` and `FileStore`
- Removed `@namespace` property in `BaseIndex<T>` and inherited classes.
- Removed unused `BlockChain<T>.Addresses`.

Alos, It isn't  directly related to this, but the following changes have been made.

- Changed return type of `IStore.CountIndex()` to `int`.
- Removed chain cloning code in `BlockChain<T>.Validate()`

# TODO

- [x] Modify `AddressTransactionSet<T>`
- [x] Documentation


Libplanet/Store/IStore.cs
```diff
@@ -14,7 +14,7 @@ public interface IStore
         /// <returns>Existing namespaces.</returns>
         IEnumerable<string> ListNamespaces();
 
-        long CountIndex(string @namespace);
+        int CountIndex(string @namespace);
```
dahlia: If this change is merely for being along with `CountTransactions()`/`CountBlocks()`, we'd better make these three methods to all return `long` instead.

Libplanet/Store/IStore.cs
```diff
@@ -37,49 +37,42 @@ Address address
...
             AddressStateMap states
         );
 
-        int CountTransactions(string @namespace);
+        int CountTransactions();
 
-        int CountBlocks(string @namespace);
+        int CountBlocks();
 
         int CountAddresses(string @namespace);
```
dahlia: Is this method still needed?

CHANGES.md
```diff
@@ -103,6 +103,11 @@ To be released.
  -  Fixed a bug that occured when `Swarm` was handling multiple responses at the
     same time.
  -  Added `IStore.ListNamespaces()` method.
+ -  `IStore.CountIndex()` became to return `int` as same as
+    `IStore.CountBlockHashes()`.
+ -  Block and Tx related methods in `IStore` and `BaseIndex<T>` no longer
+    accept namespace as arguments. it means that even if a forking occurs, the
+    block and tx files are shared.
```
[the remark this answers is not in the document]
longfin: I think `BlockChain<T>.Addresses` is okay because it was an `internal`.

Libplanet/Store/IStore.cs
longfin: Nope, I'll remove.

CHANGES.md
```diff
@@ -102,7 +102,18 @@ To be released.
     already has and deal with only longer (higher) blocks.
  -  Fixed a bug that occured when `Swarm` was handling multiple responses at the
     same time.
+ -  Removed following methods in `IStore` and `AddressTransactionSet` that
```
dahlia: The `AddressTransactionSet` class in itself was also removed (not only those methods in it).

longfin: it means 'methods in `IStore`' and '`AddressTransactionSet`'. should I re-write like "methods in `IStore` and `AddressTransactionSet` class"?

dahlia: Ah, make sense.  As the statement is ambiguous, we would disambiguate it by changing the word order like:

> Removed `AddressTransactionSet` which handles handle `Address` to `IEnumerable<TxId>` indices, and the following methods in `IStore`.

longfin:
Rebased to catch up master. PTAL @dahlia @earlbread

PR #156, opened 2019-03-28
Fix solution syntax

earlbread:
- Fixed `Solution 'Libplanet' load failed:` due to a syntax error.


earlbread:
I rebased this on the master branch. Could you review this again? @dahlia @longfin

PR #159, opened 2019-03-28
Add BigInteger to ToBencodexValue()

earlbread:
- Added `BigInteger` to `ToBencodexValue`.


Libplanet.Tests/Serialization/Bencodex/BencodexFormatterTest.cs
```diff
@@ -115,6 +122,7 @@ protected Person(SerializationInfo info, StreamingContext context)
                 .ToList();
             Secret = info.GetString("Secret");
             IsStudent = info.GetBoolean("IsStudent");
+            Assets = (BigInteger)info.GetValue("Assets", typeof(BigInteger));
```
longfin: We already have [`GetValue<T>()`](https://github.com/planetarium/libplanet/blob/bd4345abde3b31d1f5cc4504da63a7601614d6fe/Libplanet/Serialization/SerializationInfoExtension.cs#L8), so you can use it.

earlbread:
I added a commit. Please take a look. @dahlia @longfin

Libplanet.Tests/Serialization/Bencodex/BencodexFormatterTest.cs
```diff
@@ -109,12 +116,12 @@ protected Person(SerializationInfo info, StreamingContext context)
         {
             Name = info.GetString("Name");
             Age = (int)info.GetInt64("Age");
-            Names =
-                ((List<object>)info.GetValue("Names", typeof(List<object>)))
+            Names = info.GetValue<List<object>>("Names")
```
dahlia: Is this change intended?  The previous code looks rather proper.

earlbread: I thought `info.GetValue<List<object>>("Names")` is better than `((List<object>)info.GetValue("Names", typeof(List<object>)))`. Do you think previous one is better?

dahlia: I thought it is not like better or worse, but just it's unlikely to be compiled.  If the type checker passes I don't mind then.

dahlia: To elaborate, the changed version looks like trying to putting a `List<object>` value into `List<string>` property, which C#'s type checker does not allow AFAIK.

earlbread: There is more code below:
```csharp
.Select(s => (string)s)
.ToList();
```
So I think it will pass the type checker.

PR #160, opened 2019-03-28
Modify Block Explorer to have GUID argument

earlbread:
- Added GUID parameter to API to get a specific blockchain.

### DIFF
--- a/Libplanet.Explorer/Controllers/ExplorerController.cs
+++ b/Libplanet.Explorer/Controllers/ExplorerController.cs
@@ -25,18 +25,19 @@ namespace Libplanet.Explorer.Controllers
             Store = store;
         }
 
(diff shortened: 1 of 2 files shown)


longfin:
> Putting a chain ID (i.e., namespace) into URLs seems inconvenient to me. Whereas we could see and compare all branches (i.e., forks) for debugging purpose in this way, we also need to type a chain ID at front of every URL. I thought we would be like adding a `ChainId` property into the `IBlockchainStore` interface so that we type the chain ID only once when we invoke the web server on the command line.
> 
> @longfin What's your opinion? Would be this way more useful?

I also think it's a little uncomfortable to put ID in the URL every time. Using CLI option seems a bit easier at this time.

earlbread:
Okay, I'll change it to get GUID as a CLI option.

Issue #152, opened 2019-03-25
Block explorer


dahlia:
Although there has been *Libplanet.Explorer* project, it is virtually not usable.  We need to make it actually usable by Libplanet developers and users.

- [x] Check if the existing *Libplanet.Explorer* builds.
- [x] Try to make it standalone. (Currently as it requires a concrete `IAction` implementation it cannot be a standalone executable but only a library.)

dahlia:
We made a separated repository for Libplanet.Explorer: <https://github.com/planetarium/libplanet-explorer>.  Now executable binaries for macOS and Windows are built for every push.

PR #161, opened 2019-03-29
Store actions in serialized format

earlbread:
- Changed `Transaction` to store serialized actions to workaround `ToAction` when exploring blockchain(#152).


dahlia:
It seems to take some time (maybe longer than a day) to review in-depth.  I will review this until next Monday EOD!

dahlia:
Sorry, I was late.  If I understand this properly, the idea is `Transaction<T>` having only raw representations of actions instead of proper `IAction` instance to prevent determining types of actions.  And the reason we need to prevent determining types of actions is because we want *Libplanet.Explorer* to load `Transaction<T>` objects made by multiple games while we are still agnostic about each game and do not load each game's assembly DLL.

I believe it can be achieved now since PR #169 was merged.  We could merely define a dumb action store its raw plain value as its only field like:

```c#
class RawAction : IAction
{
    public ImmutableDictionary<string, object> _rawData;
    public void LoadPlainValue(ImmutableDictionary<string, object> plainValue) =>
        _rawData = plainValue;
    public ImmutableDictionary<string, object> PlainValue => _rawData;
    public IAccountStateDelta Execute(IActionContext context) => ...;
}
```

PR #163, opened 2019-03-29
Modify Block Explorer to get chain ID

earlbread:
- Added `ChainId` to IBlockchainStore to get a specific blockchain.

### DIFF
--- a/Libplanet.Explorer/Controllers/ExplorerController.cs
+++ b/Libplanet.Explorer/Controllers/ExplorerController.cs
@@ -17,18 +17,21 @@ namespace Libplanet.Explorer.Controllers
     public class ExplorerController<T> : Controller where T : IAction
     {
         private readonly IBlockchainStore Store;
+        private readonly Guid _chainId;
 
         public string TimestampFormat = "yyyy-MM-ddTHH:mm:ss.ffffffZ";
 
         public ExplorerController(IBlockchainStore store)
         {
             Store = store;
+            _chainId = store.ChainId;
         }
 
         public BlockChain<T> GetBlockChain()
         {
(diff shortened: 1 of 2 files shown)


Libplanet.Explorer/Controllers/ExplorerController.cs
```diff
@@ -17,18 +17,21 @@ namespace Libplanet.Explorer.Controllers
     public class ExplorerController<T> : Controller where T : IAction
     {
         private readonly IBlockchainStore Store;
+        private Guid _guid;
```
longfin: It would be better if `_guid` is `readonly`.

Libplanet.Explorer/Interfaces/IBlockchainStore.cs
```diff
@@ -5,5 +5,6 @@ namespace Libplanet.Explorer.Interfaces
     public interface IBlockchainStore
     {
         IStore Store { get; }
+        string ChainId { get; }
```
longfin: Could we use `Guid` directly here?

Libplanet.Explorer/Controllers/ExplorerController.cs
```diff
@@ -17,18 +17,21 @@ namespace Libplanet.Explorer.Controllers
     public class ExplorerController<T> : Controller where T : IAction
     {
         private readonly IBlockchainStore Store;
+        private readonly Guid _guid;
```
dahlia: Not sure whether it is a good name; how about `_chainId`?

earlbread:
Applied the reviews and rebased this on the master. Could you please review again? @dahlia @longfin

PR #166, opened 2019-04-01
Change type of Peer.EndPoint to DnsEndPoint

### LINKED ISSUE #165: Change type of Peer.EndPoint to enable DNS lookup
We've chosen `IPEndPoint` to represent `Peer.EndPoint` in #120 . but `IPEndPoint` can't represent address using DNS. (e.g. `seed.nekoyume.game`) so we should replace it to with [`DnsEndPoint`](https://docs.microsoft.com/en-us/dotnet/api/system.net.dnsendpoint?view=netstandard-2.0) or [`EndPoint`](https://docs.microsoft.com/en-us/dotnet/api/system.net.endpoint?view=netstandard-2.0)

longfin:
This PR changes `Peer.EndPoint`'s type to `DnsEndPoint` to resolve #165.

In this process, I've found a problem that the node behind the NAT can't connect properly, but I will treat it as a separate issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -149,8 +149,8 @@ To be released.
  -  Since we decided to depend on TURN ([RFC 5766]) and STUN ([RFC 5389]) to
     work around NAT so that `Peer`'s endpoints don't have to be multiple,
     `Peer.Urls` was renamed to `Peer.EndPoint` and its type also was changed
-    from `IImmutableList<Uri>` to `IPEndPoint`.
-    [[#120], [#123] by Yang Chun Ung, [#126], [#127]]
+    from `IImmutableList<Uri>` to `DnsEndPoint`.
+    [[#120], [#123] by Yang Chun Ung, [#126], [#127], [#165]]
  -  `Swarm` became to ignore tip blocks of the same height (`Index`) that it
     already has and deal with only longer (higher) blocks.
  -  Fixed a bug that occured when `Swarm` was handling multiple responses at the
@@ -181,7 +181,6 @@ To be released.
 [#98]: https://github.com/planetarium/libplanet/issues/98
 [#99]: https://github.com/planetarium/libplanet/issues/99
 [#120]: https://github.com/planetarium/libplanet/issues/120
-[#167]: https://github.com/planetarium/libplanet/issues/167
 [#121]: https://github.com/planetarium/libplanet/pull/121
 [#123]: https://github.com/planetarium/libplanet/pull/123
 [#124]: https://github.com/planetarium/libplanet/pull/124
@@ -196,6 +195,8 @@ To be released.
 [#144]: https://github.com/planetarium/libplanet/pull/144
 [#151]: https://github.com/planetarium/libplanet/pull/151
 [#159]: https://github.com/planetarium/libplanet/pull/159
+[#165]: https://github.com/planetarium/libplanet/issues/165
+[#167]: https://github.com/planetarium/libplanet/issues/167
 [#169]: https://github.com/planetarium/libplanet/pull/169
 [#170]: https://github.com/planetarium/libplanet/pull/170
 [RFC 5389]: https://tools.ietf.org/html/rfc5389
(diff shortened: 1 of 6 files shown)


longfin:
@dahlia @earlbread I've amend the commit to cover `Dns.GetAddressesAsync()` in unittest. PTAL.

longfin:
I've append [another bug fix](https://github.com/planetarium/libplanet/pull/166/files#diff-8b1c3fd0d4a6765c16dfd18509182f9dR107).

CHANGES.md
```diff
@@ -98,12 +98,14 @@ To be released.
  -  Since we decided to depend on TURN ([RFC 5766]) and STUN ([RFC 5389]) to
     work around NAT so that `Peer`'s endpoints don't have to be multiple,
     `Peer.Urls` was renamed to `Peer.EndPoint` and its type also was changed
-    from `IImmutableList<Uri>` to `IPEndPoint`.
-    [[#120], [#123] by Yang Chun Ung, [#126], [#127]]
+    from `IImmutableList<Uri>` to `DnsEndPoint`.
+    [[#120], [#123] by Yang Chun Ung, [#126], [#127], [#165]]
```
dahlia: It would be good if it mentions this PR #166 as well.

CHANGES.md
```diff
@@ -98,12 +98,14 @@ To be released.
...
  -  `Swarm` became to ignore tip blocks of the same height (`Index`) that it
     already has and deal with only longer (higher) blocks.
  -  Fixed a bug that occured when `Swarm` was handling multiple responses at the
     same time.
+ -  Fixed a bug that hangs when calling the `Swarm` constructor in certain
```
```suggestion
 -  Fixed a bug that the `Swarm` constructor had hanged in certain runtimes
```

CHANGES.md
```diff
@@ -98,12 +98,14 @@ To be released.
...
+    runtimes. (e.g. Unity3d)
```
```suggestion
    like Unity engine.
```

Unity seems to recently have changed their name from <q>Unity3D</q> to simply <q>Unity</q>.

Libplanet/Net/Swarm.cs
```diff
@@ -75,7 +76,7 @@ public partial class Swarm : ICollection<Peer>, IDisposable
         public Swarm(
             PrivateKey privateKey,
             TimeSpan dialTimeout,
-            IPAddress ipAddress = null,
+            string host = null,
```
dahlia: What if an invalid address is passed?  Does the constructor throw an exception, or merely `Swarm` crashes at unspecified moments?

Libplanet/Net/Swarm.cs
```diff
@@ -107,23 +108,19 @@ public partial class Swarm : ICollection<Peer>, IDisposable
...
-                _turnClient = IceServer.CreateTurnClient(iceServers).Result;
-            }
-
-            if (ipAddress == null && _turnClient == null)
+            _iceServers = iceServers?.ToList();
+            if (_host == null && (_iceServers == null || !_iceServers.Any()))
             {
                 throw new ArgumentException(
-                    $"Swarm needs {nameof(ipAddress)} or {iceServers}.");
+                    $"Swarm needs {nameof(host)} or {nameof(iceServers)}.");
```
```suggestion
                    $"Swarm requires either {nameof(host)} or {nameof(iceServers)}.");
```

Libplanet/Net/Swarm.cs
```diff
@@ -75,7 +76,7 @@ public partial class Swarm : ICollection<Peer>, IDisposable
         public Swarm(
             PrivateKey privateKey,
             TimeSpan dialTimeout,
-            IPAddress ipAddress = null,
+            string host = null,
             int? listenPort = null,
```
dahlia: Is it okay to pass only `host` and omit `listenPort`?

longfin: Yes, in that case, it binds arbitrary ports through `_router.BindRandomPort()`.

Libplanet/Net/Swarm.cs
longfin: Neither. `Swarm` will advertise the address as its endpoint, and as a result will be isolated from the network.

Do we need to check connectability? (It's probably hard on the constructor, and it's possible in `StartAsync()`).

dahlia: Although it would good to have basic validations on the given `ipAddress`, at this point we necessarily need that IMHO.

longfin: `host` can be an IP address, but it can also be any host name. so I think it is a bit tricky to check here statically.

PR #169, opened 2019-04-02
Extract PolymorphicAction<T> from Transaction<T> & IAction

dahlia:
This patch extracts the subtype dispatcher which had been coupled to `Transaction<T>` and `IAction` into a new class `PolymorphicAction<T>`.  This is both a concrete `IAction` implementation and a decorator of an abstract action class having subclasses.  This also is an alternative approach to make *Libplanet.Explorer* standalone (see also #152) besides @earlbread's PR #161.

Please read added changelogs and XML comments on `IAction` and `PolymorphicAction<T>`for details.

Additionally, I renamed `InvalidActionTypeExtension` (introduced by #144) to `MissingActionTypeException`.


Libplanet/Tx/Transaction.cs
```diff
@@ -584,11 +580,12 @@ internal RawTransaction ToRawTransaction(bool includeSign)
                     a.ToByteArray()).ToArray(),
                 publicKey: PublicKey.Format(false),
                 timestamp: Timestamp.ToString(TimestampFormat),
-                actions: Actions.Select(a => new Dictionary<string, object>
-                {
-                    { "type_id", ActionTypeAttribute.ValueOf(a.GetType()) },
-                    { "values", a.PlainValue },
-                })
+                actions: Actions.Select(a =>
+                    a.PlainValue.ToDictionary(
```
longfin: Maybe it's better to change the type of `RawTransaction.Actions`... 😕

Libplanet.Tests/Common/Action/DumbAction.cs
```diff
@@ -0,0 +1,76 @@
...
+            string item,
+            bool recordRehearsal = false
+        )
+        {
+            TargetAddress = targetAddress;
+            Item = item;
+            RecordRehearsal = recordRehearsal;
+        }
+
+        public static ThreadLocal<ImmutableList<(Address, string)>>
```
longfin: (trivial) In the test, asynchronous processing is often performed on `Task` base without threading directly, so it is good to use [`AsyncLocal<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=netframework-4.7.2).

CHANGES.md
```diff
@@ -73,6 +73,52 @@ To be released.
      -  Added `Transaction<T>.EvaluateActions()` method.
      -  Added `Block<T>.EvaluateActions()` generator method.
 
+ -  The built-in subtype polymorphism of `IAction` and `Transaction<T>` was
+    moved to a separated new `PolymorphicAction<T>` abstract class.
+    Polymorphic actions now should be wrapped by `PolymorphicAction<T>`.
+    For example, the following code:
+
+    ~~~~ csharp
+    public abstract class BaseAction : IAction { ... }
```
longfin: Can we omit `Base` prefix in docs?

dahlia: It was an intention to emphasize the presence of an `abstract class` as an elaborate example.  Do you prefer simply `Action` over `BaseAction` here?

Libplanet/Tx/Transaction.cs
```diff
@@ -601,10 +598,8 @@ internal RawTransaction ToRawTransaction(bool includeSign)
 
         private static T ToAction(IDictionary<string, object> arg)
         {
-            var typeStr = (string)arg["type_id"];
-            var action = (T)Activator.CreateInstance(Types[typeStr]);
-            var values = (IDictionary<string, object>)arg["values"];
-            action.LoadPlainValue(values.ToImmutableDictionary());
+            var action = new T();
+            action.LoadPlainValue(arg.ToImmutableDictionary());
```
earlbread: Can we deserialize the actions properly without concrete action type?

dahlia: Did you mean, what if `T` is not a concrete type?

CHANGES.md
longfin: I thought it would be appropriate to show as we were following the [Framework guidelines](https://stackoverflow.com/questions/429470/naming-conventions-for-abstract-classes/429494#429494). but it would be fine if it was intended.

If you intend to make it more clear, it may seem okay to attach an `Abstract` prefix.

Libplanet/Tx/Transaction.cs
earlbread: Right, what if we use like `DumbAction`?

dahlia: That's fine because `DumbAction` is a concrete class.  According to the [official docs][1], `new()` constraint is satisfied by only concrete class, neither `abstract class` nor `interface`:

> To use the new constraint, the type cannot be abstract.

That means, from now on we only accept a concrete class as an action type — even `PolymorphicAction<T>` in itself is concrete.  This is a point of getting rid of built-in subtype dispatcher.  Note that `T` of `PolymorphicAction<T>` does not have `new()` constraint on the other hand.  It can take an `abstract class` or an `interface`.

[1]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-constraint

Libplanet/Tx/Transaction.cs
dahlia: Currently `RawTransaction` consists of many mutable types like `byte[]` and `Dictionary<K, V>`.  If we want `RawTransaction` to be freer from state I believe we need to adjust other fields too.  Seems like it should be a separated patch.

CHANGES.md
```diff
@@ -73,6 +73,52 @@ To be released.
...
+        actions: new[] { new Attack(...) }.Select(
+            a => new PolymorphicAction<BaseAction>(a)
+        )
+    );
+    ~~~~
+
+    [[#169]]
+
+     -  The type parameter `T` of `Transaction<T>`, `Block<T>`, and
+        `BlockChain<T>` became to require having a parameterless constructor
```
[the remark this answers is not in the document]
dahlia: Applied: 3c2315068acc3edf6f60958853873830a72cdf75.

CHANGES.md
dahlia: Applied: 7a6ebc034e102fd2f91ec9e83d4e6f0041a744a7.

Libplanet.Tests/Common/Action/DumbAction.cs
dahlia: Applied: 7a6ebc034e102fd2f91ec9e83d4e6f0041a744a7.

dahlia:
I adjusted the code according @longfin's review comments, wrote more tests, and added `implicit operator PolymorphicAction<T>(T)` as well for convenience.

@longfin @earlbread Please review this again.

Issue #167, opened 2019-04-02
Add version information at the network message level


earlbread:
We need to add version information to determine how different versions of the node work.

PR #170, opened 2019-04-02
Add protocol version and ignore peer if protocol version is different

earlbread:
- Added network protocol version. #167 
- Made the peers ignore each other if the protocol version is different.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -174,9 +174,8 @@ To be released.
     `MissingActionTypeException` at runtime.  [[#28], [#144], [#169]]
  -  Turn into parameter in `BlockPolicy`'s constructor to milliseconds. [[#151]]
  -  `BencodexFormatter` became able to serialize `BigInteger`.  [[#159]]
- -  Made `Swarm` possible to configure its (application-specific)
-    network `protocolVersion` and, to ignore peers if their
-    version is different.  [[#167]], [[#170]]
+ -  Made `Swarm` possible to configure its network `appProtocolVersion` and,
+    to ignore peers if their version is different.  [[#167]], [[#170]]
 
 [#28]: https://github.com/planetarium/libplanet/issues/28
 [#98]: https://github.com/planetarium/libplanet/issues/98
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -174,7 +174,8 @@ To be released.
     `MissingActionTypeException` at runtime.  [[#28], [#144], [#169]]
  -  Turn into parameter in `BlockPolicy`'s constructor to milliseconds. [[#151]]
  -  `BencodexFormatter` became able to serialize `BigInteger`.  [[#159]]
- -  Added the network protocol version and change to ignore peers if the
+ -  Made `Swarm` possible to configure its (application-specific)
+    network `protocolVersion` and, to ignore peers if their
     version is different.  [[#167]], [[#170]]
 
 [#28]: https://github.com/planetarium/libplanet/issues/28
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -174,11 +174,14 @@ To be released.
     `MissingActionTypeException` at runtime.  [[#28], [#144], [#169]]
  -  Turn into parameter in `BlockPolicy`'s constructor to milliseconds. [[#151]]
  -  `BencodexFormatter` became able to serialize `BigInteger`.  [[#159]]
+ -  Added the network protocol version and change to ignore peers if the
+    version is different.  [[#167]], [[#170]]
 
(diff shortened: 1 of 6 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -1330,12 +1374,20 @@ CancellationToken cancellationToken
...
                     ToNetMQAddress(peer),
                     dealer,
                     cancellationToken);
                 _logger.Debug($"DialAsync({peer.EndPoint}) is complete.");
 
+                if (pong.ProtocolVersion != _protocolVersion)
+                {
+                    dealer.Dispose();
+                    throw new DifferentProtocolVersionException(
+                        $"Peer protocol version({pong.ProtocolVersion}) is " +
```
longfin: it would be better if current version(`_protocolVersion`) exists in the error message.

Libplanet/Net/Swarm.cs
```diff
@@ -61,14 +62,16 @@ public partial class Swarm : ICollection<Peer>, IDisposable
             IPAddress ipAddress = null,
             int? listenPort = null,
             DateTimeOffset? createdAt = null,
-            IEnumerable<IceServer> iceServers = null)
+            IEnumerable<IceServer> iceServers = null,
+            int protocolVersion = 1)
```
longfin: I'm not sure if it would be okay to omit `protocolVersion`. but since we need to improve the protocol version system,  I also think it can be addressed at that time.

earlbread: If omitting `protocolVersion` is not proper, I'll change it in a different pull request.

Libplanet/Net/Swarm.cs
earlbread: I added the current version in the error message. @longfin

earlbread:
I rebased on the current master. Could you please review again? @longfin @dahlia

CHANGES.md
```diff
@@ -174,11 +174,14 @@ To be released.
     `MissingActionTypeException` at runtime.  [[#28], [#144], [#169]]
  -  Turn into parameter in `BlockPolicy`'s constructor to milliseconds. [[#151]]
  -  `BencodexFormatter` became able to serialize `BigInteger`.  [[#159]]
+ -  Added the network protocol version and change to ignore peers if the
```
```suggestion
 -  Made `Swarm` possible to configure its (application-specific)
    network `protocolVersion` and, to ignore peers if their
```

Libplanet/Net/Swarm.cs
dahlia: Since it's not about Libplanet's internals but application's own speak, `appProtocolVersion` or such name might be better.

And I still it's better not to have the default version, as it's prone to be unaware of this.

Libplanet/Net/DifferentProtocolVersionException.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+
+namespace Libplanet.Net
+{
+    [Serializable]
+    public sealed class DifferentProtocolVersionException : Exception
```
dahlia: Could we have XML comments for docs here?

Libplanet/Net/Swarm.cs
```diff
@@ -1330,12 +1374,21 @@ CancellationToken cancellationToken
...
                     ToNetMQAddress(peer),
                     dealer,
                     cancellationToken);
                 _logger.Debug($"DialAsync({peer.EndPoint}) is complete.");
 
+                if (pong.ProtocolVersion != _protocolVersion)
+                {
+                    dealer.Dispose();
+                    throw new DifferentProtocolVersionException(
+                        $"Peer protocol version({pong.ProtocolVersion}) is " +
```
dahlia: We could have rather a well-structured exception type.  I mean, how about having a separated properties like `DifferentProtocolVersionException.ExpectedVersion` and `DifferentProtocolVersionException.ActualVersion`?  This approach is a more programmable way.  See also `DifferentProtocolVersionException` as an example of this approach.

CHANGES.md
```diff
@@ -174,11 +174,15 @@ To be released.
     `MissingActionTypeException` at runtime.  [[#28], [#144], [#169]]
  -  Turn into parameter in `BlockPolicy`'s constructor to milliseconds. [[#151]]
  -  `BencodexFormatter` became able to serialize `BigInteger`.  [[#159]]
+ -  Made `Swarm` possible to configure its (application-specific)
+    network `protocolVersion` and, to ignore peers if their
```
```suggestion
    network `appProtocolVersion` and, to ignore peers if their
```

earlbread:
I applied the review comments. Could you please review this again? @dahlia @longfin

PR #185, opened 2019-04-09
Different protocol version of peer handling

earlbread:
- Added `appProtocolVersion` to `Peer`.
- Made ignore or remove `Peer` if the `appProtocolVersion` is different.
- Added EventHandler to handle different version of Peer.

Please review the name of classes as well. I'm not sure they are appropriate.

### DIFF
--- a/Libplanet/Net/Peer.cs
+++ b/Libplanet/Net/Peer.cs
@@ -8,10 +8,23 @@ using Uno;
 
 namespace Libplanet.Net
 {
+    /// <summary>
+    /// A representation of peer node.
+    /// </summary>
+    /// <seealso cref="Swarm"/>
     [Serializable]
     [GeneratedEquality]
     public partial class Peer : ISerializable
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Peer"/> class.
+        /// </summary>
+        /// <param name="publicKey">A <see cref="PublicKey"/> of the
+        /// <see cref="Peer"/>.</param>
+        /// <param name="endPoint">A <see cref="DnsEndPoint"/> consisting of the
+        /// host and port of the <see cref="Peer"/>.</param>
+        /// <param name="appProtocolVersion">An application protocol version
+        /// that the <see cref="Peer"/> is using.</param>
         public Peer(
             PublicKey publicKey, DnsEndPoint endPoint, int appProtocolVersion)
         {
@@ -38,20 +51,32 @@ namespace Libplanet.Net
             AppProtocolVersion = info.GetInt32("app_protocol_version");
         }
 
+        /// <summary>
+        /// The corresponding <see cref="Libplanet.Crypto.PublicKey"/> of
+        /// this peer.
+        /// </summary>
         [EqualityKey]
         [Pure]
         public PublicKey PublicKey { get; }
 
+        /// <summary>
+        /// The corresponding <see cref="DnsEndPoint"/> of this peer.
+        /// </summary>
         [EqualityKey]
         [Pure]
         public DnsEndPoint EndPoint { get; }
 
+        /// <summary>
+        /// The corresponding application protocol version of this peer.
+        /// </summary>
+        /// <seealso cref="Swarm.DifferentVersionPeerEncountered"/>
         [Pure]
         public int AppProtocolVersion { get; }
 
         [Pure]
         public Address Address => new Address(PublicKey);
 
+        /// <inheritdoc/>
         public void GetObjectData(
             SerializationInfo info,
             StreamingContext context
@@ -63,6 +88,7 @@ namespace Libplanet.Net
             info.AddValue("app_protocol_version", AppProtocolVersion);
         }
 
+        /// <inheritdoc/>
         public override string ToString()
         {
             return $"{Address}.{EndPoint}.{AppProtocolVersion}";
--- a/Libplanet/Net/Peer.cs
+++ b/Libplanet/Net/Peer.cs
@@ -12,7 +12,8 @@ namespace Libplanet.Net
     [GeneratedEquality]
     public partial class Peer : ISerializable
     {
-        public Peer(PublicKey publicKey, DnsEndPoint endPoint)
+        public Peer(
+            PublicKey publicKey, DnsEndPoint endPoint, int appProtocolVersion)
         {
             if (publicKey == null)
             {
@@ -25,6 +26,7 @@ namespace Libplanet.Net
 
             PublicKey = publicKey;
             EndPoint = endPoint;
+            AppProtocolVersion = appProtocolVersion;
         }
 
         protected Peer(SerializationInfo info, StreamingContext context)
@@ -33,6 +35,7 @@ namespace Libplanet.Net
             EndPoint = new DnsEndPoint(
                 info.GetString("end_point_host"),
                 info.GetInt32("end_point_port"));
+            AppProtocolVersion = info.GetInt32("app_protocol_version");
         }
 
         [EqualityKey]
@@ -42,6 +45,9 @@ namespace Libplanet.Net
         [EqualityKey]
         [Pure]
         public DnsEndPoint EndPoint { get; }
+
+        [Pure]
+        public int AppProtocolVersion { get; }
 
         [Pure]
         public Address Address => new Address(PublicKey);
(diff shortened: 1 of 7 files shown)


Libplanet/Net/Peer.cs
```diff
@@ -33,19 +42,35 @@ protected Peer(SerializationInfo info, StreamingContext context)
...
         [EqualityKey]
         [Pure]
         public DnsEndPoint EndPoint { get; }
 
+        /// <summary>
+        /// The corresponding app protocol version of this peer.
+        /// </summary>
+        /// <seealso cref="Swarm"/>
+        [Pure]
+        public int AppProtocolVersion { get; }
```
dahlia: This changed also should be logged.

Libplanet/Net/Peer.cs
```diff
@@ -8,11 +8,19 @@
...
+    /// </summary>
+    /// <seealso cref="Swarm"/>
     [Serializable]
     [GeneratedEquality]
     public partial class Peer : ISerializable
     {
-        public Peer(PublicKey publicKey, DnsEndPoint endPoint)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Peer"/> class.
+        /// </summary>
```
dahlia: :+1: Could you write `<param>` docs as well?

Libplanet/Net/Swarm.cs
```diff
@@ -145,6 +145,13 @@ public partial class Swarm : ICollection<Peer>, IDisposable
             }
         }
 
+        /// <summary>
+        /// The <see cref="EventHandler" /> called when the different version of
+        /// <see cref="Peer" /> is discovered.
+        /// </summary>
+        public event EventHandler<DifferentProtocolVersionEventArgs>
+            DifferentVersionPeerDiscovered;
```
dahlia: It would be better if the `Swam` constructor can take this as an optional parameter.

Libplanet/Net/Peer.cs
```diff
@@ -33,19 +42,35 @@ protected Peer(SerializationInfo info, StreamingContext context)
...
+        /// The corresponding <see cref="DnsEndPoint"/> of this peer.
+        /// </summary>
         [EqualityKey]
         [Pure]
         public DnsEndPoint EndPoint { get; }
 
+        /// <summary>
+        /// The corresponding app protocol version of this peer.
+        /// </summary>
+        /// <seealso cref="Swarm"/>
```
dahlia:
```suggestion
        /// <seealso cref="Swarm.DifferentVersionPeerDiscovered"/>
```

Libplanet/Net/Swarm.cs
dahlia: @kijun @longfin Which word is better, discovered vs. encountered?  Or any other word?

Libplanet/Net/Swarm.cs
```diff
@@ -1381,6 +1408,19 @@ CancellationToken cancellationToken
...
+                    DifferentProtocolVersionEventArgs args =
+                        new DifferentProtocolVersionEventArgs
+                        {
+                            ExpectedVersion = _appProtocolVersion,
+                            ActualVersion = pong.AppProtocolVersion,
+                        };
+
+                    if (DifferentVersionPeerDiscovered != null)
+                    {
+                        DifferentVersionPeerDiscovered(this, args);
```
dahlia: Currently game programmers (library users) seem to have no much control on this situation, like ignoring this peer or treating this peer as compatible.

To me, if this event callback cannot return nor decide anything, it doesn't look useful.  Nodes can be aware that there are some peers having different versions of protocol, but they don't have option to ignore/reject them.

Libplanet/Net/Swarm.cs
kijun: Hmm maybe this is not an option, but ProtocolVersionConflict sounds more concise... thoughts?

dahlia: Even if a peer have a different protocol we still treat them as compatible, right?  It's one of options game programmers have.  So naming it “conflict” feels like oversimplified to me.

kijun: Ah, I wasn't aware of this. I'd say then `Encountered` is slightly better since we're not actively trying to find peers with different protocol version.

Libplanet/Net/Swarm.cs
earlbread: For now, I only thought of options to ignore or deny. I think we can make version compatible options later. How do you think about it?

dahlia: > For now, I only thought of options to ignore or deny.

What does “to deny” mean here?

earlbread: "deny" was a wrong word. Just "to ignore". A peer is not added to the peer list if the protocol version is different. And also, if a different version of peer sends peer delta, it is ignored.

dahlia: If we can only ignore them and there's no other option, when would `DifferentVersionPeerDiscovered` be used mostly?  Why would a game need to receive this event and what would it do for this event?

earlbread: For example, using `DifferentVersionPeerDiscovered` the game can notify users of things like version upgrades.

dahlia: Ah, make sense. 👌

Libplanet/Net/Swarm.cs
earlbread: Ok, So I'll change `Discovered` to `Encountered`.

earlbread:
I applied review comments and amended a commit. Could you please review this again? @dahlia @kijun @longfin

Libplanet/Net/Peer.cs
```diff
@@ -8,11 +8,25 @@
...
     {
-        public Peer(PublicKey publicKey, DnsEndPoint endPoint)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Peer"/> class.
+        /// </summary>
+        /// <param name="publicKey">A <see cref="PublicKey"/> of the
+        /// <see cref="Peer"/>.</param>
+        /// <param name="endPoint">A <see cref="DnsEndPoint"/> consisting of the
+        /// host and port of the <see cref="Peer"/>.</param>
+        /// <param name="appProtocolVersion">A application protocol version that
```
dahlia:
```suggestion
        /// <param name="appProtocolVersion">An application protocol version that
```

Libplanet/Net/Swarm.cs
```diff
@@ -145,6 +145,13 @@ public partial class Swarm : ICollection<Peer>, IDisposable
             }
         }
 
+        /// <summary>
+        /// The <see cref="EventHandler" /> called when the different version of
+        /// <see cref="Peer" /> is discovered.
+        /// </summary>
+        public event EventHandler<DifferentProtocolVersionEventArgs>
+            DifferentVersionPeerEncountered;
```
dahlia: > It would be better if the Swam constructor can take this as an optional parameter.

This previous review doesn't seem addressed yet.

earlbread:
I applied suggestions. Please take a look.

Libplanet/Net/Swarm.cs
```diff
@@ -1381,6 +1414,19 @@ CancellationToken cancellationToken
...
                     dealer.Dispose();
+
+                    DifferentProtocolVersionEventArgs args =
+                        new DifferentProtocolVersionEventArgs
+                        {
+                            ExpectedVersion = _appProtocolVersion,
+                            ActualVersion = pong.AppProtocolVersion,
+                        };
+
+                    if (DifferentVersionPeerEncountered != null)
```
longfin: We can omit this null check like below

```csharp
DifferentVersionPeerEncountered?.Invoke(this, args)
```

Libplanet/Net/Swarm.cs
```diff
@@ -1161,6 +1174,11 @@ CancellationToken cancellationToken
 
             if (IsUnknownPeer(sender))
             {
+                if (IsDifferentProtocolVersion(sender))
+                {
+                    return;
```
longfin: Is it ok to return without calling `DifferentVersionPeerEncountered`?

Libplanet/Net/Swarm.cs
```diff
@@ -1381,6 +1414,19 @@ CancellationToken cancellationToken
                 if (pong.AppProtocolVersion != _appProtocolVersion)
```
longfin: Do we still need the `AppProtocolVersion` in the `Pong`?

Libplanet/Net/Swarm.cs
earlbread: I thought only when the swarm try to add another peer, but I think it would be good to call the handler in this case too.

Libplanet/Net/Swarm.cs
earlbread: I think the `Pong` should have `AppProtocolVersion` so that the sender can know the version of the `Peer`.

earlbread:
I applied review comments and amended commits. Please take a look.

Issue #187, opened 2019-04-10
Add Initial Block Download (a.k.a. IBD) phase


longfin:
Current `Swarm` implementations don't do anything unless another peer introduces the blocks they have mined through `BroadcastBlocks<T>()`. this complicates the process of downloading a block when the node first joins the network that has the formed chain.

Therefore, in `Swarm.StartAsync()`, it is necessary to specify the process of requesting a block to the registered peer and delay the other processing of `Swarm` until this phase is completed.

PR #190, opened 2019-04-10
Implement simple IBD

longfin:
This PR addresses #187. the following changes were made for this purpose.

- Added `TipIndex` as current chain tip Index to `Pong` to determine peer that has the longest chain.
- Made `Swarm.StartAsync()` to compare other chain height using `Pong.TipIndex` and download blocks if necessary before starts other tasks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,8 +9,13 @@ To be released.
  - Added `Swarm.DifferentVersionPeerEncountered` event handler that can handle
    events when a different version of a peer is discovered.  [[#167]], [[#185]]
  - Added `Peer.AppProtocolVersion` property.  [[#185]]
+ - `Swarm.StartAsync()` now receives the height of blocks (tip `Index`) from
+   other known peers and synchronizes the blocks if necessary before starts
+   before propagating/receiving pinpointed recent blocks to prevent inefficient
+   round-trips [[#187]]
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
+[#187]: https://github.com/planetarium/libplanet/issues/187
 
 
 Version 0.2.1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,12 +10,13 @@ To be released.
    events when a different version of a peer is discovered.  [[#167]], [[#185]]
  - Added `Peer.AppProtocolVersion` property.  [[#185]]
  - `Swarm.StartAsync()` now receives the height of blocks (tip `Index`) from
-   other known peers and synchronizes the blocks if necessary before starts
+   other known peers and synchronizes the blocks if necessary
    before propagating/receiving pinpointed recent blocks to prevent inefficient
-   round-trips [[#187]]
+   round-trips.  [[#187], [#190]]
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
+[#190]: https://github.com/planetarium/libplanet/pull/190
 
 
 Version 0.2.1
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -478,15 +478,7 @@ namespace Libplanet.Net
                     {
                         try
                         {
-                            Peer replacedPeer = await DialPeerAsync(
-                                peer,
-                                cancellationToken
-                            );
-                            if (replacedPeer != peer)
-                            {
-                                _peers[replacedPeer] = _peers[peer];
-                                _peers.Remove(peer);
-                            }
+                            await DialPeerAsync(peer, cancellationToken);
                         }
                         catch (TimeoutException e)
                         {
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -931,6 +931,85 @@ namespace Libplanet.Net
             }
         }
 
+        private async Task<BlockChain<T>> SyncPreviousBlocksAsync<T>(
+            BlockChain<T> blockChain,
+            Peer peer,
+            HashDigest<SHA256>? stop,
+            CancellationToken cancellationToken)
+            where T : IAction, new()
+        {
+            // Fix the tip here because it may change while receiving the block
+            // hashes.
+            Block<T> tip = blockChain.Tip;
+
+            _logger.Debug("Trying to find branchpoint...");
+            BlockLocator locator = blockChain.GetBlockLocator();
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -9,8 +9,12 @@ To be released.
  - Added `Swarm.DifferentVersionPeerEncountered` event handler that can handle
    events when a different version of a peer is discovered.  [[#167]], [[#185]]
  - Added `Peer.AppProtocolVersion` property.  [[#185]]
+ - `Swarm.StartAsync()` now receives the height(`Index`) of the block from
```
```suggestion
 - `Swarm.StartAsync()` now receives the height of blocks (tip `Index`) from
```

CHANGES.md
```diff
@@ -9,8 +9,12 @@ To be released.
...
+   other known peers and synchronizes the blocks if necessary before starts
+   another process. [[#187]]
```
dahlia: What does “another process” here mean?

Libplanet/Net/Swarm.cs
```diff
@@ -944,104 +962,118 @@ private async Task RefreshAllocate()
...
+            CancellationToken cancellationToken
+        )
+            where T : IAction, new()
+        {
+            // We assume that the blocks are sorted in order.
+            Block<T> oldest = blocks.First();
+            Block<T> latest = blocks.Last();
+            Block<T> tip = blockChain.Tip;
+
+            if (tip == null || latest.Index > tip.Index)
```
```suggestion
            if (tip is null || latest.Index > tip.Index)
```

Libplanet/Net/Swarm.cs
```diff
@@ -944,104 +962,118 @@ private async Task RefreshAllocate()
...
+            {
+                _logger.Debug("Trying to fill up previous blocks...");
+                BlockChain<T> toAppend = await SyncExistingBlocksAsync(
```
dahlia: How about naming this `previousBlocks` to be consistent with the logged message?

Libplanet/Net/Swarm.cs
```diff
@@ -944,104 +962,118 @@ private async Task RefreshAllocate()
...
-                    $"{ByteUtil.Hex(branchPoint.ToByteArray())}"
-                );
+            _logger.Debug(
+                $"Branchpoint is " +
+                $"{ByteUtil.Hex(branchPoint.ToByteArray())}"
+            );
 
-                BlockChain<T> toSync;
+            BlockChain<T> synced;
+            if (tip == null || branchPoint == tip.Hash)
```
```suggestion
            if (tip is null || branchPoint == tip.Hash)
```

Libplanet/Net/Swarm.cs
```diff
@@ -759,6 +717,64 @@ private async Task BindingProxies()
...
+            foreach (Peer peer in _peers.Keys)
+            {
+                try
+                {
+                    Pong pong = await DialPeerAsync(peer, cancellationToken);
+                    if (pong.TipIndex > longestTipIndex)
+                    {
+                        longestPeer = peer;
+                        longestTipIndex = pong.TipIndex.Value;
+                    }
```
dahlia: Instead of manually comparing the longest peer in the loop, how about making a simple array of pairs like `(Peer, long)[] peerLengths` and determining the longest peer after the loop ends like `(Peer longestPeer, long longestTipIndex) = peerLengths.Max(pair => pair.Item2);`?

CHANGES.md
longfin: Receiving and processing messages(`Swarm.ReceiveMessageAsync()`) Repeating delta distribution(`Swarm.RepeatDeltaDistributionAsync()`), and so on.

Libplanet/Net/Swarm.cs
longfin: Interesting code. but I'm not sure that there is a big advantage enough to managing a separate array and looping it one more time.

dahlia: At first glance, it was hard to guess the intention, because code is not that short, and finding the peer who has the longest blockchain takes more lines than actual “dialing to existing peers” (as the method name says).  If we stick with the current version I suggest to write an explanation as a short comment.

longfin: Okay, I'll leave comments and/or consider renaming `DialToExistingPeers()`.

longfin:
I've amend to reflect reviews.

> Okay, I'll leave comments and/or consider renaming `DialToExistingPeers()`.

At first I tried to have a descriptive name, but it seemed more intuitive to split the method that dials into the known peer and the method that receives the block from the longest chain.

Libplanet/Net/Swarm.cs
```diff
@@ -759,6 +723,75 @@ private async Task BindingProxies()
...
+                    {
+                        _logger.Error(
+                            e,
+                            $"Protocol Version is different ({peer}).");
+                    }
+                }
+            });
+        }
+
+        private async Task SyncPreviousBlocksFromPeers<T>(
```
dahlia: Another comment about naming thing: how about calling this “behind blocks” instead of “previous blocks” here?  For example, GitHub uses the term “behind commits” when comparing two branches.

CHANGES.md
dahlia: It would be better if that's elaborated, e.g.:

> … before propagating/receiving pinpointed recent blocks to prevent inefficient round-trips.

longfin:
Amended. PTAL @dahlia @earlbread

CHANGES.md
```diff
@@ -9,8 +9,13 @@ To be released.
  - Added `Swarm.DifferentVersionPeerEncountered` event handler that can handle
    events when a different version of a peer is discovered.  [[#167]], [[#185]]
  - Added `Peer.AppProtocolVersion` property.  [[#185]]
+ - `Swarm.StartAsync()` now receives the height of blocks (tip `Index`) from
+   other known peers and synchronizes the blocks if necessary before starts
```
```suggestion
   other known peers and synchronizes the blocks if necessary
```

CHANGES.md
```diff
@@ -9,8 +9,13 @@ To be released.
...
+   before propagating/receiving pinpointed recent blocks to prevent inefficient
+   round-trips [[#187]]
```
```suggestion
   round-trips.  [[#187], [#190]]
```

CHANGES.md
```diff
@@ -9,8 +9,13 @@ To be released.
...
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
+[#187]: https://github.com/planetarium/libplanet/issues/187
```
```suggestion
[#187]: https://github.com/planetarium/libplanet/issues/187
[#190]: https://github.com/planetarium/libplanet/pull/190
```

Issue #189, opened 2019-04-10
BlockChain<T>.GetStates() returns significantly slow if the height is too high


dahlia:
[`BlockChain<T>.GetStates(A)`][1] where `A` is nonexistent (i.e., never used) returns significantly slow if the height is enough high.  Currently [`IStore.GetBlockStates()`][2] returns only a set of updated states, not a complete set of existent accounts.  This means that `BlockChain<T>.GetStates(A)` needs to descend to the bottom (i.e., the genesis block) in order to determine if `A` is nonexistent.

[1]: https://docs.libplanet.io/master/api/Libplanet.Blockchain.BlockChain-1.html#Libplanet_Blockchain_BlockChain_1_GetStates_System_Collections_Generic_IEnumerable_Libplanet_Address__System_Nullable_Libplanet_HashDigest_System_Security_Cryptography_SHA256___
[2]: https://docs.libplanet.io/master/api/Libplanet.Store.IStore.html#Libplanet_Store_IStore_GetBlockStates_Libplanet_HashDigest_System_Security_Cryptography_SHA256__

dahlia:
It seems to happen even if a request address is existent.  The current implementation compares two sets (more exactly, intended to be a set but represented as `IEnumerable<Address>`) by arbitrarily ordered sequences:

https://github.com/planetarium/libplanet/blob/65bc8cdbc729991bc4aeb940c983fe99405ddc9c/Libplanet/Blockchain/BlockChain.cs#L178

PR #192, opened 2019-04-12
Make BlockChain<T>.GetStates() to only drill down until all requested addresses are found 

dahlia:
This patch fixed a part of the issue #189.  (For the case of nonexistent addresses, as it needs to change `IStore` interface to fix that, I am going to submit a separated patch to the master.)

The problem is that the condition to test if all requested addresses are satisfied had depended on the order of two sets of addresses, which is arbitrary.  I wrote a regression test to reproduce this problem, which involves testing utilities like `NullPolicy<T>` and `StoreTracker`, and fixed it by replacing `SequenceEquals()` with `SetEquals()`.


Libplanet.Tests/Store/StoreTrackerTest.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _tracker.ListNamespaces();
+            _tracker.CountIndex(Guid.NewGuid().ToString());
+            Assert.Equal(2, _tracker.Logs.Count);
+
+            _tracker.ClearLogs();
+            Assert.Empty(_tracker.Logs);
+        }
+    }
+}
```
longfin:
```suggestion
}

```

Libplanet.Tests/Store/StoreTracker.cs
```diff
@@ -0,0 +1,161 @@
...
+            _store.StageTransactionIds(txids);
+        }
+
+        public void UnstageTransactionIds(ISet<TxId> txids)
+        {
+            _logs.Add((nameof(UnstageTransactionIds), txids, null));
+            _store.UnstageTransactionIds(txids);
+        }
+    }
+}
```
longfin:
```suggestion
}

```

PR #197, opened 2019-04-12
Make BlockChain<T>.GetStates() returns early for nonexistent addresses

dahlia:
This patch completely fixes the bug #189.  Besides the patch #192, this deals with nonexistent addresses.

This involves a mask-based index to approximates a set of `Addresses`.  The key idea is that if a certain address had been ever used and included into the mask, the mask ensures to have only non-zero bits (i.e., trues) for the same positions that the address has non-zero bits.  You can also think of it as a kind of bloom filter but no hash functions.  The point is that although a mask cannot ensure if an address is in it but can ensure an address is definitely not in it.

To store a mask for each block, I added a new parameter `Address addressesMask` to `IStore.PutBlock<T>()` method, and a new method named `IStore.GetAddressesMask()` to query this.  I wrote XML comments on these methods.  Please read comments for API details.

Although I added a mask as an ad-hoc field of `IStore` at this stage, I suggest put this information into `Block<T>` type for the future so that mask is also validated within block's cryptographic hash.  (More detailed proposal is in the commit message of e01da78a4d13b9a9c3c3d05a54412e31cb9385ec.)

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -165,19 +165,56 @@ namespace Libplanet.Blockchain
                 _rwlock.ExitReadLock();
             }
 
-            ImmutableHashSet<Address> requestedAddresses =
-                addresses.ToImmutableHashSet();
+            ImmutableDictionary<Address, BitArray> requestedAddresses =
+                addresses.Select(addr =>
+                    new KeyValuePair<Address, BitArray>(
+                        addr,
+                        new BitArray(addr.ToByteArray())
+                    )
+                ).ToImmutableDictionary();
+            ImmutableHashSet<Address> requestedAddressSet =
+                requestedAddresses.Keys.ToImmutableHashSet();
+
+            bool IsPossiblyIn(BitArray addr, BitArray maskPattern)
+            {
+                var match = (BitArray)maskPattern.Clone();
+                match.And(addr);
+                return match.OfType<bool>().Any(bit => bit);
+            }
+
+            var coveredAddresses = new HashSet<Address>();
             var states = new AddressStateMap();
+
             while (offset != null)
             {
                 states = (AddressStateMap)states.SetItems(
                     Store.GetBlockStates(offset.Value)
                     .Where(
-                        kv => requestedAddresses.Contains(kv.Key) &&
+                        kv => requestedAddresses.ContainsKey(kv.Key) &&
                         !states.ContainsKey(kv.Key))
                     );
 
-                if (requestedAddresses.SetEquals(states.Keys))
+                // If there is no mask for the offset (the store made in
+                // the older versions of Libplanet may lack mask files),
+                // make the mask to match to all possible addresses.
+                var mask = new BitArray(
+                    offset is HashDigest<SHA256> hash &&
+                        Store.GetAddressesMask(hash) is Address a
+                        ? a.ToByteArray()
+                        : Enumerable.Repeat((byte)0xff, Address.Size).ToArray()
+                );
+
+                // Addresses that are definitely not existent.
+                coveredAddresses.UnionWith(
+                    requestedAddresses.Where(
+                        pair => !IsPossiblyIn(pair.Value, mask)
+                    ).Select(pair => pair.Key)
+                );
+
+                // Already gethered addresses.
+                coveredAddresses.UnionWith(states.Keys);
+
+                if (requestedAddressSet.SetEquals(coveredAddresses))
                 {
                     break;
                 }
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -201,7 +201,7 @@ namespace Libplanet.Blockchain
                     offset is HashDigest<SHA256> hash &&
                         Store.GetAddressesMask(hash) is Address a
                         ? a.ToByteArray()
-                        : Enumerable.Repeat((byte)0xff, Address.Size).ToArray()
+                        : BlockSet<T>.WildcardMask
                 );
 
                 // Addresses that are definitely not existent.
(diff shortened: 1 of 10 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -165,19 +165,51 @@ IEnumerator IEnumerable.GetEnumerator()
...
+            // versions of Libplanet may lack mask files),
+            // make the mask to match to all possible addresses.
+            var mask = new BitArray(
+                offset is HashDigest<SHA256> hash &&
+                    Store.GetAddressesMask(hash) is Address a
+                    ? a.ToByteArray()
+                    : Enumerable.Repeat((byte)0xff, Address.Size).ToArray()
+            );
+
+            Func<BitArray, BitArray, bool> isPossibilyIn = (addr, maskPat) =>
```
earlbread:
```suggestion
            Func<BitArray, BitArray, bool> isPossiblyIn = (addr, maskPat) =>
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -165,19 +165,51 @@ IEnumerator IEnumerable.GetEnumerator()
...
                     .Where(
-                        kv => requestedAddresses.Contains(kv.Key) &&
+                        kv => requestedAddresses.ContainsKey(kv.Key) &&
                         !states.ContainsKey(kv.Key))
                     );
 
-                if (requestedAddresses.SetEquals(states.Keys))
+                // A set of addresses that are definitely not existent.
+                ImmutableHashSet<Address> nonexistents = requestedAddresses
+                    .Where(pair => !isPossibilyIn(pair.Value, mask))
```
earlbread:
```suggestion
                    .Where(pair => !isPossiblyIn(pair.Value, mask))
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -165,19 +165,51 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                requestedAddresses.Keys.ToImmutableHashSet();
+
+            // If there is no mask for the offset (the store made in the older
+            // versions of Libplanet may lack mask files),
+            // make the mask to match to all possible addresses.
+            var mask = new BitArray(
+                offset is HashDigest<SHA256> hash &&
+                    Store.GetAddressesMask(hash) is Address a
+                    ? a.ToByteArray()
+                    : Enumerable.Repeat((byte)0xff, Address.Size).ToArray()
```
longfin: I think this code can be precomputed, right?

Libplanet/Blockchain/BlockChain.cs
longfin: I think we can use local function.

```suggestion
            bool isPossiblyIn(BitArray addr, BitArray maskPat)
            Func<BitArray, BitArray, bool> isPossibilyIn = (addr, maskPat) =>
```

Libplanet/Store/BlockSet.cs
```diff
@@ -49,21 +51,57 @@ public override ICollection<Block<T>> Values
                     throw new KeyNotFoundException();
                 }
 
-                Trace.Assert(block.Hash == key);
+                Trace.Assert(block.Hash.Equals(key));
```
longfin: Should we prefer `Equals()` instead of `==`?

dahlia: Rider warned me that `HashDigest<SHA256>` hadn't implemented this operator, so I made this change.  AFAIK this operator is implemented through *Uno.CodeGen* though.

Libplanet/Blockchain/BlockChain.cs
dahlia: Oh no, it definitely is a bug.  I'm going to address this.

dahlia: Fixed this and amended the commit.

Libplanet/Blockchain/BlockChain.cs
dahlia: Addressed.

Libplanet/Blockchain/BlockChain.cs
dahlia: Addressed.

Libplanet/Blockchain/BlockChain.cs
dahlia: Addressed.

dahlia:
@longfin I turned that `byte[]` entirely filled with `0xff` into a constant (kinda, it's `static readonly`). f928bfd960499dfb0612327e1de143094990ac07

Libplanet/Store/BlockSet.cs
```diff
@@ -49,21 +51,57 @@ public override ICollection<Block<T>> Values
...
+                    {
+                        prevMask = maskAddr.ToByteArray();
+                    }
+                    else
+                    {
+                        // If there is no mask for the previous hash
+                        // (the store made in the older versions of Libplanet
+                        // may lack mask files), make the mask to match to
+                        // all possible addresses.
+                        prevMask = Enumerable.Repeat((byte)0xff, Address.Size)
```
longfin: We can also use [WildcardMask](https://github.com/planetarium/libplanet/pull/197/files#diff-f9c16aaa8d3a312223ca86876f791202R21) here, right?

dahlia: `WildcardMask` is defined in the `BlockChain<T>` class and it's `private`.  Making `BlockSet<T>` depends on `BlockChain<T>` also seems unnatural.  So I'm going to define the same constant here too.

It's like we should define a distinct type like `AddressMask` in the future.

dahlia: Hmm, as `BlockChain<T>` already has depended on `BlockSet<T>` (though the inverse is not true), I'm going to move the `WildcardMask` constant to `BlockSet<T>` and make it `internal` instead of `private`.

dahlia: Addressed and amended the commit: 23170446098fc02655e7127b4fc45e8b52e6b7cd.

dahlia:
@earlbread @longfin Could you review this again?

PR #198, opened 2019-04-14
Fix disconnecting on TURN relay session

longfin:
This PR aids problem the problem of TURN relay session being disconnected. if the TURN server doesn't receive `CreatePermission` on time, it deletes the registered peer and rejects any request from that. to prevent this, I added a task that iterates `CreatePermission`.

It also fixes the following issues:

- Fixed a bug that `Swarm `attempted to use TURN relay even though `_host` was given.
- Adjusted a lock in `Append()` to increase parallelism.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,8 @@ To be released.
     `PutBlock<T>(Block<T>, Address)`.  [[#189]]
  -  Improved the read throughput of `BlockChain<T>` when calling
     `BlockChain<T>.Append()`.
+ -  Fixed a bug that `Swarm` attempted to use TURN relay even though `_host` was
+    given.
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,8 @@ To be released.
     and its all implementations.  [[#189]]
  -  The signature of `IStore.PutBlock<T>(Block<T>)` method was changed to
     `PutBlock<T>(Block<T>, Address)`.  [[#189]]
+ -  Improved the read throughput of `BlockChain<T>` when calling
+    `BlockChain<T>.Append()`.
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,8 @@ To be released.
     `BlockChain<T>.Append()`.
  -  Fixed a bug that `Swarm` attempted to use TURN relay even though `_host` was
     given.
+ -  Fixed a bug that TURN relay was disconnected when connecting for more than 5
+    minutes.
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,11 +24,10 @@ To be released.
     and its all implementations.  [[#189]]
  -  The signature of `IStore.PutBlock<T>(Block<T>)` method was changed to
     `PutBlock<T>(Block<T>, Address)`.  [[#189]]
- -  Improved the read throughput of `BlockChain<T>` when calling
-    `BlockChain<T>.Append()`.
- -  Fixed a bug that `Swarm` attempted to use TURN relay even though `_host` was
+ -  Improved the read throughput of `BlockChain<T>.Append()`.
+ -  Fixed a bug that `Swarm` had attempted to use TURN relay even if the `host` argument was
     given.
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -15,6 +15,12 @@ To be released.
    round-trips.  [[#187], [#190]]
  - Improved the read throughput of `BlockChain<T>` while mining through
    `BlockChain<T>.MineBlock()`
+ - Improved the read throughput of `BlockChain<T>` when calling
+   `BlockChain<T>.Append()`.
+ - Fixed a bug that `Swarm `attempted to use TURN relay even though `_host` was
```
```suggestion
 - Fixed a bug that `Swarm` attempted to use TURN relay even though `_host` was
```

longfin:
I've rebased.

Libplanet/Net/Swarm.cs
```diff
@@ -33,6 +33,9 @@ public partial class Swarm : ICollection<Peer>, IDisposable
         private static readonly TimeSpan TurnAllocationLifetime =
             TimeSpan.FromSeconds(777);
 
+        private static readonly TimeSpan TurnPermissionLifetime =
+            TimeSpan.FromMinutes(5);
```
dahlia: Is this time interval defined by the specification, or just arbitrary?  If it's arbitrary, it would be good to parametrize this.

CHANGES.md
```diff
@@ -24,6 +24,12 @@ To be released.
     and its all implementations.  [[#189]]
  -  The signature of `IStore.PutBlock<T>(Block<T>)` method was changed to
     `PutBlock<T>(Block<T>, Address)`.  [[#189]]
+ -  Improved the read throughput of `BlockChain<T>` when calling
+    `BlockChain<T>.Append()`.
+ -  Fixed a bug that `Swarm` attempted to use TURN relay even though `_host` was
```
dahlia: As the `_host` field is just internal details that library users are not aware of, we'd better to refer to the `Swarm` contstructor's `host` parameter instead:

```suggestion
 -  Fixed a bug that `Swarm` had attempted to use TURN relay even if the `host` argument was
```

CHANGES.md
```diff
@@ -24,6 +24,12 @@ To be released.
     and its all implementations.  [[#189]]
  -  The signature of `IStore.PutBlock<T>(Block<T>)` method was changed to
     `PutBlock<T>(Block<T>, Address)`.  [[#189]]
+ -  Improved the read throughput of `BlockChain<T>` when calling
```
```suggestion
 -  Improved the read throughput of
```

CHANGES.md
```diff
@@ -24,6 +24,12 @@ To be released.
...
+    `BlockChain<T>.Append()`.
+ -  Fixed a bug that `Swarm` attempted to use TURN relay even though `_host` was
+    given.
+ -  Fixed a bug that TURN relay was disconnected when connecting for more than 5
```
```suggestion
 -  Fixed a bug that TURN relay had been disconnected when being connected for longer than 5
```

Libplanet/Net/Swarm.cs
```diff
@@ -436,7 +439,7 @@ IEnumerator IEnumerable.GetEnumerator()
                 throw new SwarmException("Swarm is already running.");
             }
 
-            if (_iceServers != null)
+            if (_host is null && _iceServers != null)
```
dahlia: For consistency of style:

```suggestion
            if (_host is null && _iceServers is null)
```

Libplanet/Net/Swarm.cs
```diff
@@ -801,6 +805,17 @@ private async Task RefreshAllocate()
...
         }
 
+        private async Task RefreshPermissions()
+        {
+            TimeSpan lifetime = TurnPermissionLifetime;
+            while (Running)
+            {
+                await Task.Delay(lifetime - TimeSpan.FromMinutes(1));
+                await Task.WhenAll(
+                    _peers.Keys.Select(p => CreatePermission(p)));
```
```suggestion
                    _peers.Keys.Select(CreatePermission));
```

It's trivial though.

Libplanet/Net/Swarm.cs
longfin: `_iceServers != null` and `_iceServers is null` seem to be opposite. did you mean `!(_iceServers is null)` ?

Libplanet/Net/Swarm.cs
longfin: It was defined in RFC. 

>    The Permission Lifetime MUST be 300 seconds (= 5 minutes).

https://tools.ietf.org/html/rfc5766#section-8

dahlia: Good to link to that in a comment.

Libplanet/Net/Swarm.cs
dahlia: Ah, it's my fault.  You're right.

```suggestion
            if (_host is null && !(_iceServers is null))
```

Libplanet/Net/Swarm.cs
longfin: I've mentioned.

PR #199, opened 2019-04-15
Use previous two blocks to calculate difficulty

earlbread:
Currently, we need to traverse from the first block to get the next block difficulty.
This fixes `GetNextBlockDifficulty()` to use only the last two blocks of the chain to calculate the next block difficulty.

### DIFF
--- a/Libplanet/Blockchain/Policies/BlockPolicy.cs
+++ b/Libplanet/Blockchain/Policies/BlockPolicy.cs
@@ -131,9 +131,20 @@ namespace Libplanet.Blockchain.Policies
         /// <inheritdoc />
         public int GetNextBlockDifficulty(IEnumerable<Block<T>> blocks)
         {
-            return ExpectDifficulties(blocks, yieldNext: true)
-                .First(t => t.Block == null)
-                .Difficulty;
+            int blockCount = blocks.Count();
+
+            if (blockCount <= 1)
+            {
+                return blockCount;
+            }
+
+            Block<T>[] prevBlocks = blocks.Skip(Math.Max(0, blockCount - 2))
+                .ToArray();
+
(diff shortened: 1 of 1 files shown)


Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -131,9 +131,20 @@ DateTimeOffset currentTime
...
-                .First(t => t.Block == null)
-                .Difficulty;
+            int blockCount = blocks.Count();
+
+            if (blockCount <= 1)
+            {
+                return blockCount;
+            }
+
+            Block<T>[] prevBlocks = blocks.Skip(Math.Max(0, blockCount - 2))
```
longfin: It seems to be better than before but still a loop with as much complexity as O(n). How about using the index in `BlockChain <T> .Store`?

Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -131,9 +131,20 @@ DateTimeOffset currentTime
...
+                .ToArray();
+
+            return GetNextDifficultyFromPrevTimestamp(
+                prevBlocks[0].Timestamp,
+                prevBlocks[1].Timestamp,
+                prevBlocks[1].Difficulty);
         }
 
         private IEnumerable<DifficultyExpectation> ExpectDifficulties(
```
longfin: Does this function still used?

earlbread: Right, this is still used in `ValidateBlocks()`. I'll find a way to get rid of `ExpectDifficulties()`.

earlbread:
I opened another PR to address other issues.

PR #205, opened 2019-04-16
Change BlockChain to implement IReadOnlyList

earlbread:
This changes `BlockChain<T>` to implement `IReadOnlyList<Block<T>>`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,11 +29,18 @@ To be released.
     given.
  -  Fixed a bug that TURN relay had been disconnected when being connected for longer than 5
     minutes.
+ -  `BlockChain<T>` became to implement `IReadOnlyList<Block<T>>`.  [[#205]]
+ -  `BlockChain<T>.Validate()` became to receive `IReadOnlyList<Block<<T>>`
+    instead of `IEnumerable<Block<T>>`.  [[#205]]
+ -  `IBlockPolicy<T>.ValidateBlocks()` and
+    `IBlockPolicy<T>.GetNextBlockDifficulty()` became to receive
+    `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.  [[#205]]
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
 [#190]: https://github.com/planetarium/libplanet/pull/190
 [#193]: https://github.com/planetarium/libplanet/pull/193
+[#205]: https://github.com/planetarium/libplanet/pull/205
 
 
 Version 0.2.2
--- a/Libplanet.Tests/Blockchain/Policies/BlockPolicyTest.cs
+++ b/Libplanet.Tests/Blockchain/Policies/BlockPolicyTest.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using System.Security.Cryptography;
 using Libplanet.Blockchain.Policies;
@@ -62,27 +61,27 @@ namespace Libplanet.Tests.Blockchain.Policies
 
             Assert.Equal(
                 0,
-                policy.GetNextBlockDifficulty(blocks.Take(0))
+                policy.GetNextBlockDifficulty(blocks.Take(0).ToList())
             );
             Assert.Equal(
                 1,
-                policy.GetNextBlockDifficulty(blocks.Take(1))
+                policy.GetNextBlockDifficulty(blocks.Take(1).ToList())
             );
             Assert.Equal(
                 2,
-                policy.GetNextBlockDifficulty(blocks.Take(2))
+                policy.GetNextBlockDifficulty(blocks.Take(2).ToList())
             );
             Assert.Equal(
                 3,
-                policy.GetNextBlockDifficulty(blocks.Take(3))
+                policy.GetNextBlockDifficulty(blocks.Take(3).ToList())
             );
             Assert.Equal(
                 2,
-                policy.GetNextBlockDifficulty(blocks.Take(4))
+                policy.GetNextBlockDifficulty(blocks.Take(4).ToList())
             );
             Assert.Equal(
                 3,
-                policy.GetNextBlockDifficulty(blocks.Take(5))
+                policy.GetNextBlockDifficulty(blocks.Take(5).ToList())
             );
             Assert.Equal(
                 2,
@@ -104,8 +103,7 @@ namespace Libplanet.Tests.Blockchain.Policies
                         {
                             fields.Index = 1;
                             return fields;
-                        }
-                    ),
+                        }).ToList(),
                     FixtureEpoch + new TimeSpan(0, 1, 0)
                 )
             );
@@ -119,8 +117,7 @@ namespace Libplanet.Tests.Blockchain.Policies
                         {
                             fields.PreviousHash = default(HashDigest<SHA256>);
                             return fields;
-                        }
-                    ),
+                        }).ToList(),
                     FixtureEpoch + new TimeSpan(0, 1, 0)
                 )
             );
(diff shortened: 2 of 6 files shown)


CHANGES.md
```diff
@@ -29,11 +29,13 @@ To be released.
     given.
  -  Fixed a bug that TURN relay had been disconnected when being connected for longer than 5
     minutes.
+ -  `BlockChain<T>` now implements `IReadOnly<Block<T>>`.  [[#205]]
```
```suggestion
 -  `BlockChain<T>` now implements `IReadOnlyList<Block<T>>`.  [[#205]]
```

Libplanet.Tests/Blockchain/Policies/BlockPolicyTest.cs
```diff
@@ -148,32 +144,30 @@ public void ValidateBlocks()
...
                     FixtureEpoch + new TimeSpan(3, 1, 0)
                 )
             );
             Assert.Null(
                 policy.ValidateBlocks(
                     MineBlocks(
-                        new[] { (0, 0), (1, 1), (5, 2), (9, 1) }
-                    ),
+                        new[] { (0, 0), (1, 1), (5, 2), (9, 1) })
+                        .ToList(),
```
dahlia: Is this indentation intended?

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -236,8 +243,9 @@ public void Append(Block<T> block, DateTimeOffset currentTime)
                 block.Validate(
                     currentTime,
                     a => GetStates(new[] { a }, tip).GetValueOrDefault(a));
-                Validate(Enumerable.Append(this, block), currentTime);
+                Enumerable.Append(this, block);
 
+                // Validate(this, currentTime);
```
dahlia: Isn't this a miss?

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -111,7 +118,7 @@ public Block<T> Tip
         }
 
         public void Validate(
-            IEnumerable<Block<T>> blocks,
+            IReadOnlyList<Block<T>> blocks,
```
dahlia: As this is a `public` method, this change also should be written in the changelog.

Libplanet/Blockchain/Policies/IBlockPolicy.cs
```diff
@@ -43,6 +43,6 @@ DateTimeOffset currentTime
         /// followed by a new <see cref="Block{T}"/> to be mined.</param>
         /// <returns>A right <see cref="Block{T}.Difficulty"/>
         /// for a new <see cref="Block{T}"/> to be mined.</returns>
-        int GetNextBlockDifficulty(IEnumerable<Block<T>> blocks);
+        int GetNextBlockDifficulty(IReadOnlyList<Block<T>> blocks);
```
dahlia: Changes in this file also break API backward compatibility, we need to write the corresponding changelogs.

Libplanet/Blockchain/BlockChain.cs
earlbread: Oh, it's my mistake. I'll fix it.

Libplanet.Tests/Blockchain/Policies/BlockPolicyTest.cs
earlbread: I used Rider's fix spacing, but it can be written in one line. I'll fix it.

earlbread:
I applied review comments and amended the commit. Please take a look. @dahlia @longfin

CHANGES.md
```diff
@@ -29,11 +29,19 @@ To be released.
     given.
  -  Fixed a bug that TURN relay had been disconnected when being connected for longer than 5
     minutes.
+ -  `BlockChain<T>` now implements `IReadOnlyList<Block<T>>`.  [[#205]]
+    `BlockChain<T>.Validate()` now receives `IReadOnlyList<Block<<T>>`
```
dahlia: A bullet seems missed.

```suggestion
 -  `BlockChain<T>.Validate()` now receives `IReadOnlyList<Block<<T>>`
```

CHANGES.md
```diff
@@ -29,11 +29,19 @@ To be released.
...
+    instead of `IEnumerable<Block<T>>`.  [[#205]]
+ -  `IBlockPolicy<T>.ValidateBlocks()` and
+    `IBlockPolicy<T>.GetNextBlockDifficulty()`
+    now receives `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.
```
dahlia: Plural present tense.

```suggestion
    now receive `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.
```

As a non-native speaker, I prefer to use “became to do” instead of “now does/do”, because in the latter way a verb should inflect depending on its subject (singular or plural).  On the other hand, a verb following “became to do” is always singular present, so less prone to make a grammatical mistake.  🙄

earlbread:
I fixed the changelog and amended the commit. Could you please review this again? @dahlia @longfin

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -83,8 +83,15 @@ public Block<T> Tip
...
         }
 
+        /// <inheritdoc/>
+        int IReadOnlyCollection<Block<T>>.Count =>
+            checked((int)Store.CountIndex(Id.ToString()));
+
         internal IStore Store { get; }
 
+        /// <inheritdoc/>
+        public Block<T> this[int index] => this[(long)index];
```
dahlia: @earlbread Although it's late, we should make it throw [`ArgumentOutOfRangeException`][1] instead of [`IndexOutOfRangeException`][2], because collection types in [`System.Collections.Generic`][3] throw `ArgumentOutOfRangeException` and only arrays throw `IndexOutOfRangeException`.  See also [this SO answer][4]:

> Note that `List<T>` throws `ArgumentOutOfRangeException` for the same cases where arrays use `IndexOutOfRangeException`.

Could you change this in the next patch?

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception
[2]: https://docs.microsoft.com/en-us/dotnet/api/system.indexoutofrangeexception
[3]: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic
[4]: https://stackoverflow.com/a/20940980/383405

earlbread: Sure, I'll change it in the next patch!

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -236,8 +243,9 @@ public void Append(Block<T> block, DateTimeOffset currentTime)
                 block.Validate(
                     currentTime,
                     a => GetStates(new[] { a }, tip).GetValueOrDefault(a));
-                Validate(Enumerable.Append(this, block), currentTime);
+                Enumerable.Append(this, block);
```
dahlia: Although it's late, this seems a miss; [`Enumerable.Append()`][1] method returns a new `IEnumerable<T>` that has one more item, but this code doesn't use its return value.  According to the [docs][2]:

> This method does not modify the elements of the collection. Instead, it creates a copy of the collection with the new element.

To me, the code before changed seems correct.  @earlbread Could you fix this in the pull request #210?

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.append
[2]: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.append#remarks

dahlia: It seems this code was changed because the type of `Validate()` method's first parameter was changed from `IEnumerable<Bock<T>>` to `IReadOnlyList<Block<T>>` in this pull request.  To fix this quickly, it could be like `Validate(Enumerable.Append(this, block).ToList(), currentTime);`, I'm skeptic if it's good though.

earlbread: This seems not needed anymore. I'll remove it in #210.

PR #206, opened 2019-04-17
Add Swarm.PreloadAsync()

longfin:
This PR adds `Swarm.PreloadAsync()` to explicitly downloads the blocks before starting. and, it also adds `BlockDownloadState` to provide current download state to caller(game).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@ To be released.
     round-trips.  [[#187], [#190]]
  -  Added `Swarm.PreloadAsync()` method to explicitly and preemptively download
     initial blocks before `Swarm.StartAsync<T>()` being called.
-    [[#204]] [[#206]]
+    [[#204]], [[#206]]
  -  Added `BlockDownloadState` class to represents a block downloading state.
     [[#204]], [[#206]]
  -  Improved overall read throughput of `BlockChain<T>` while blocks are being
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@ To be released.
     round-trips.  [[#187], [#190]]
  -  Added `Swarm.PreloadAsync()` that explicitly synchronizes the blocks before
     starting and `BlockDownloadState` that represents downloading states.
-    [[#204]] [[#206]]
+    [[#204]], [[#206]]
  -  Improved overall read throughput of `BlockChain<T>` while blocks are being
     mined by `BlockChain<T>.MineBlock()`.
  -  Fixed a bug that `TurnClientException` had been thrown by Swarm when a STUN
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,8 +13,10 @@ To be released.
     other known peers and synchronizes the blocks if necessary
     before propagating/receiving pinpointed recent blocks to prevent inefficient
     round-trips.  [[#187], [#190]]
- -  Added `Swarm.PreloadAsync()` that explicitly synchronizes the blocks before
-    starting and `BlockDownloadState` that represents downloading states.
+ -  Added `Swarm.PreloadAsync()` method to explicitly and preemptively download
+    initial blocks before `Swarm.StartAsync<T>()` being called.
+    [[#204]] [[#206]]
+ -  Added `BlockDownloadState` class to represents a block downloading state.
     [[#204]], [[#206]]
  -  Improved overall read throughput of `BlockChain<T>` while blocks are being
     mined by `BlockChain<T>.MineBlock()`.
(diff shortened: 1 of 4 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -545,6 +541,40 @@ IEnumerator IEnumerable.GetEnumerator()
...
+        /// A cancellation token used to propagate notification that this
+        /// operation should be canceled.
+        /// </param>
+        /// <typeparam name="T">An <see cref="IAction"/> type. It should match
+        /// to <see cref="BlockChain{T}"/>'s type parameter.</typeparam>
+        /// <returns>
+        /// No object or value is returned by this method when it completes.
+        /// </returns>
+        public async Task PreloadAsync<T>(
+            BlockChain<T> blockChain,
```
dahlia: This parameter seems common in methods in `Swarm`.  Is it inappropriate to have a field for this instead?

Libplanet/Net/Swarm.cs
```diff
@@ -545,6 +541,40 @@ IEnumerator IEnumerable.GetEnumerator()
...
+        /// </param>
+        /// <param name="cancellationToken">
+        /// A cancellation token used to propagate notification that this
+        /// operation should be canceled.
+        /// </param>
+        /// <typeparam name="T">An <see cref="IAction"/> type. It should match
+        /// to <see cref="BlockChain{T}"/>'s type parameter.</typeparam>
+        /// <returns>
+        /// No object or value is returned by this method when it completes.
+        /// </returns>
```
dahlia: If we omit `<returns>` tag at all does the compiler warns of it?

CHANGES.md
```diff
@@ -13,6 +13,9 @@ To be released.
     other known peers and synchronizes the blocks if necessary
     before propagating/receiving pinpointed recent blocks to prevent inefficient
     round-trips.  [[#187], [#190]]
+ -  Added `Swarm.PreloadAsync()` that explicitly synchronizes the blocks before
```
```suggestion
 -  Added `Swarm.PreloadAsync()` method to explicitly and preemptively download
```

CHANGES.md
```diff
@@ -13,6 +13,9 @@ To be released.
...
+    starting and `BlockDownloadState` that represents downloading states.
```
```suggestion
    initial blocks before `Swarm.StartAsync<T>()` being called.
    [[#204]] [[#206]]
 -  Added `BlockDownloadState` class to represents a block downloading state.
```

Libplanet/Net/Swarm.cs
longfin: Yes. linter says '[SA1615](http://www.stylecop.soyuz5.com/SA1615.html): Element return value shuld be documented.'

Libplanet/Net/Swarm.cs
longfin: It's also appropriate, but I want to keep `Swarm`  as `T` agnostic, so I only get arguments for methods that use block chains.

From another point of view, it would be better to refactor by dividing into the appropriate classes because `Swarm` is currently too large. I'll create an issue about this.

dahlia: Yes, it's anyway a refactoring.  We'd better have a separated patch for that.

CHANGES.md
```diff
@@ -13,6 +13,11 @@ To be released.
     other known peers and synchronizes the blocks if necessary
     before propagating/receiving pinpointed recent blocks to prevent inefficient
     round-trips.  [[#187], [#190]]
+ -  Added `Swarm.PreloadAsync()` method to explicitly and preemptively download
+    initial blocks before `Swarm.StartAsync<T>()` being called.
+    [[#204]], [[#206]]
+ -  Added `BlockDownloadState` class to represents a block downloading state.
```
dahlia: I made a typo in the previous suggestion. 🙄

```suggestion
 -  Added `BlockDownloadState` class to represent a block downloading state.
```

Libplanet/Net/Swarm.cs
```diff
@@ -545,6 +541,40 @@ IEnumerator IEnumerable.GetEnumerator()
...
+        /// An instance that receives progress updates for block downloads.
+        /// </param>
+        /// <param name="cancellationToken">
+        /// A cancellation token used to propagate notification that this
+        /// operation should be canceled.
+        /// </param>
+        /// <typeparam name="T">An <see cref="IAction"/> type. It should match
+        /// to <see cref="BlockChain{T}"/>'s type parameter.</typeparam>
+        /// <returns>
+        /// No object or value is returned by this method when it completes.
```
```suggestion
        /// A task without value.
        /// You only can <c>await</c> until the method is completed.
```

PR #210, opened 2019-04-18
Validate only the next block to be appended when append a block

earlbread:
- Added `ValidateNextBlock(IReadOnlyList<Block<T>>, Block<T>)` to `IBlockPolicy`.
- Changed `BlockChain<T>.Append()` to validate only the next block to be appended.
- Removed `ExpectDifficulties()`.
- Changed `IndexOutOfRangeException` in `BlockChain<T>` to `ArgumentOutOfRangeException`.
- Moved `IBlockPolicy<T>.ValidateBlocks()` to `IBlockPolicyExtension.ValidateBlocks<T>()`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,10 @@ To be released.
  -  `IBlockPolicy<T>.ValidateBlocks()` and
     `IBlockPolicy<T>.GetNextBlockDifficulty()` became to receive
     `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.  [[#205]]
+ -  Added `ValidateBlockToAppend(IReadOnlyList<Block<T>>, Block<T>)` to
+    `IBlockPolicy<T>`.  [[#210]]
+ -  Instead of validating the entire blocks, `BlockChain<T>.Append()` became
+    to validate only the next block to be appended.  [[#210]]
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
@@ -48,6 +52,7 @@ To be released.
 [#204]: https://github.com/planetarium/libplanet/issues/204
 [#205]: https://github.com/planetarium/libplanet/pull/205
 [#206]: https://github.com/planetarium/libplanet/pull/206
+[#210]: https://github.com/planetarium/libplanet/pull/210
 
 
 Version 0.2.2
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,15 +37,18 @@ To be released.
  -  `BlockChain<T>` became to implement `IReadOnlyList<Block<T>>`.  [[#205]]
  -  `BlockChain<T>.Validate()` became to receive `IReadOnlyList<Block<<T>>`
     instead of `IEnumerable<Block<T>>`.  [[#205]]
- -  `IBlockPolicy<T>.GetNextBlockDifficulty()` became to receive
+ -  `IBlockPolicy<T>.GetNextBlockDifficulty()` method became to receive
     `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.  [[#205]]
- -  Added `ValidateNextBlock(IReadOnlyList<Block<T>>, Block<T>)` to
-    `IBlockPolicy<T>`.  [[#210]]
- -  `IBlockPolicy<T>.ValidateBlocks()` is removed.  [[#210]]
+ -  Added
+    `IBlockPolicy<T>.ValidateNextBlock(IReadOnlyList<Block<T>>, Block<T>)`
+    method.  [[#210]]
+ -  Removed `IBlockPolicy<T>.ValidateBlocks()` method.  [[#210]]
  -  Added `IBlockPolicyExtension.ValidateBlocks<T>(IBlockPolicy<T>,
-    IReadOnlyList<Block<T>>, DateTimeOffset)`.  [[#210]]
- -  Instead of validating the entire blocks, `BlockChain<T>.Append()` became
-    to validate only the next block to be appended.  [[#210]]
+    IReadOnlyList<Block<T>>, DateTimeOffset)` method.  [[#210]]
+ -  `BlockChain<T>[int]` became to throw `ArgumentOutOfRangeException` instead
+    of `IndexOutOfRangeException`.  [[#210]]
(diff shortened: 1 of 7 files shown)


Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -66,134 +64,141 @@ public BlockPolicy(TimeSpan blockInterval)
...
-                prevTimestamp = block.Timestamp;
+                secondLastBlock = lastBlock;
+                lastBlock = block;
             }
 
             return null;
         }
 
+        /// <inheritdoc/>
+        public InvalidBlockException ValidateBlockToAppend(
```
longfin: Do we need to expose the validation interface used when adding blocks? if `BlockChain<T>` knows detail that uses the last block instead of a `BlockPolicy`, it also seems to be a problem that can be solved using the existing `IBlockPolicy.ValidateBlocks()`.

Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -66,134 +64,141 @@ public BlockPolicy(TimeSpan blockInterval)
...
+            Block<T> lastBlock,
+            Block<T> blockToAppend)
         {
-            DateTimeOffset? prevTimestamp = null;
-            DateTimeOffset? prevPrevTimestamp = null;
+            int index = blockCount;
+            int difficulty = 0;
 
-            if (yieldNext)
+            if (lastBlock is Block<T> block)
```
longfin: This test may seem a bit confusing.

```suggestion
            if (!(lastBlock is null))
```

Libplanet/Blockchain/Policies/BlockPolicy.cs
longfin: But again, depending on the implementation, how many previous blocks we see may change. 🤔

Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -66,134 +64,141 @@ public BlockPolicy(TimeSpan blockInterval)
...
+                return new InvalidBlockDifficultyException(
+                    $"the expected difficulty of the block #{index} " +
+                    $"is {difficulty}, but its difficulty is " +
+                    $"{blockToAppend.Difficulty}'");
+            }
 
-                if (block != null)
+            if (!blockToAppend.PreviousHash.Equals(prevHash))
+            {
+                if (prevHash == null)
```
longfin:
```suggestion
                if (prevHash is null)
```

Libplanet/Blockchain/Policies/BlockPolicy.cs
earlbread: Do you mean to put `ValidateBlockToAppend()` as a private method in `BlockChain<T>` instead of `IBlockPolicy`?

longfin: yes, exactly, my first thought was removing `ValidateBlockToAppend()` and moving these logics to `BlockChain<T>.Append()`.

longfin:
@earlbread it seems to have conflicts. could you rebase it?

CHANGES.md
```diff
@@ -35,12 +35,17 @@ To be released.
  -  `IBlockPolicy<T>.ValidateBlocks()` and
     `IBlockPolicy<T>.GetNextBlockDifficulty()` became to receive
     `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.  [[#205]]
+ -  Added `ValidateBlockToAppend(IReadOnlyList<Block<T>>, Block<T>)` to
+    `IBlockPolicy`.  [[#210]]
```
```suggestion
    `IBlockPolicy<T>`.  [[#210]]
```

Libplanet/Blockchain/Policies/BlockPolicy.cs
earlbread: To avoid problems when implementing similar interfaces, I removed `ValidateBlocks()` from` IBlockPolicy<T>` and created it as an extension method.

Libplanet.Tests/Blockchain/Policies/BlockPolicyTest.cs
```diff
@@ -89,6 +112,123 @@ public void GetNextBlockDifficulty()
             );
         }
 
+        [Fact]
+        public void ValidateNextBlock()
+        {
+            _blockChain.Append(_genesis);
```
dahlia: Does not `BlockChain<T>.Append()` method in itself call `IBlockPolicy<T>.ValidateNextBlock()` method under the hood?  It's like not minimum reproducable code.  Could we use a plain array/list instead of `BlockChain<T>` for fixture here?

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -52,7 +52,7 @@ public Block<T> Tip
                 {
                     return this[-1];
                 }
-                catch (IndexOutOfRangeException)
+                catch (ArgumentOutOfRangeException)
```
dahlia: This interfacial change also should be in the changelog.

CHANGES.md
```diff
@@ -37,9 +37,15 @@ To be released.
...
     instead of `IEnumerable<Block<T>>`.  [[#205]]
- -  `IBlockPolicy<T>.ValidateBlocks()` and
-    `IBlockPolicy<T>.GetNextBlockDifficulty()` became to receive
+ -  `IBlockPolicy<T>.GetNextBlockDifficulty()` became to receive
     `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.  [[#205]]
+ -  Added `ValidateNextBlock(IReadOnlyList<Block<T>>, Block<T>)` to
+    `IBlockPolicy<T>`.  [[#210]]
+ -  `IBlockPolicy<T>.ValidateBlocks()` is removed.  [[#210]]
+ -  Added `IBlockPolicyExtension.ValidateBlocks<T>(IBlockPolicy<T>,
+    IReadOnlyList<Block<T>>, DateTimeOffset)`.  [[#210]]
```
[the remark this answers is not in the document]
earlbread: The first way seems more concise to me too.

earlbread:
I applied suggestions and added commits. Please take a look. @dahlia @longfin

earlbread:
I added a test to make sure the `Append` function does a block validation. @dahlia @longfin

Issue #31, opened 2019-01-09
Provide API to do post-process right after a certain action executed


ipdae:
In games that currently use libplanet, there is no way to know created action has been processed.
so, in the game, we waiting for process has finished like [this](https://github.com/planetarium/nekoyume-unity/blob/master/nekoyume/Assets/_Scripts/Game/Stage.cs#L103).
However, this can cause the game to hang if the block containing the Tx is not included in the block chain.
In my opinion, if libplanet provides an event based on the result of the process, the game will be handling the internal state change as event-driven and solved the problem.

dahlia:
I guess @ipdae wanted to link to <https://github.com/planetarium/nekoyume-unity/blob/4e2e942d1916bd76f43233fb4dea3e7111512696/nekoyume/Assets/_Scripts/Game/Stage.cs#L103>.

dahlia:
Ah, that repository seems not public yet.  ☹️

dahlia:
Quote the code he referred to:

```c#
private IEnumerator SleepAsync()
{
    var tables = this.GetRootComponent<Data.Tables>();
    Data.Table.Stats statsData;
    if (tables.Stats.TryGetValue(ActionManager.Instance.Avatar.Level, out statsData))
    {
        ActionManager.Instance.Sleep(statsData);
        while (ActionManager.Instance.Avatar.CurrentHP == ActionManager.Instance.Avatar.HPMax)
        {
            yield return new WaitForSeconds(1.0f);
        }
    }
    OnRoomEnter();
}
```

dahlia:
The quoted code looks helpful to explain his intention though.

To sum up, the current API enforces game apps to poll every state change of the blockchain in order to execute game logic depending on that that state change, which is inefficient.

moreal:
I agree about that opinion, the pattern based on event  
It'll be so helpful for reducing useless waiting.

dahlia:
I started to work on this yesterday, and have sketched my rough idea on it.  Let me explain my current idea.

The reason why we need to be aware when an action is *actually* processed (whatever it technically means) is not to create a new consequent action that depends on previous states causes by the previous action.  Making an action depending on another action is already possible and easy; you only need to put both actions into two different blocks.

[...]

PR #212, opened 2019-04-20
IAction.Render() & IAction.Unrender()

dahlia:
This patch implements a minimum viable subset of the idea I had suggested in https://github.com/planetarium/libplanet/issues/31#issuecomment-481699147.

There are missing behaviors yet; `Render()` does not guarantee it's called after the action is confirmed in the network (we need to define what “confirmation” means here), etc, for instance.  The currently guaranteed attributes are:

 -  `IAction.Render()` and `IAction.Unrender()` take an equivalent `IActionContext` object with a clean state (which means its `Random` state is never consumed) to `IAction.Execute()` took.
 -  If some blocks become “stale,” `Unrender()` methods of actions in these blocks are invoked in reverse order.
 -  For each action, `Render()` is only once called unless it's stale so that its `Unrender()` is called.
 -  For each action, the number of `Unrender()` calls is the same to or one less than the number of `Render()` calls.


Libplanet/Blocks/Block.cs
```diff
@@ -150,6 +150,52 @@ public byte[] ToBencodex(bool hash, bool transactionData)
...
+        /// the <see cref="Transactions"/> and their
+        /// <see cref="Transaction{T}.Actions"/> (e.g., tx&#xb9;-act&#xb9;,
+        /// tx&#xb9;-act&#xb2;, tx&#xb2;-act&#xb9;, tx&#xb2;-act&#xb2;,
+        /// &#x2026;).
+        /// Note that each <see cref="IActionContext.Random"/> object has
+        /// a unconsumed state.
+        /// </returns>
+        [Pure]
+        public
+        IEnumerable<(Transaction<T>, T, IActionContext, IAccountStateDelta)>
```
longfin: Quadruples seems to be a bit complicated as a return type for public methods. How about creating a type?

Libplanet/Tx/Transaction.cs
```diff
@@ -460,31 +466,40 @@ public byte[] ToBencodex(bool sign)
         /// <c>false</c>.
         /// </exception>
         [Pure]
-        public IAccountStateDelta EvaluateActions(
+        public IEnumerable<(T, IActionContext, IAccountStateDelta)>
```
longfin: This triples also seems to be a bit complicated.

Libplanet/Tx/Transaction.cs
```diff
@@ -460,31 +466,40 @@ public byte[] ToBencodex(bool sign)
...
+            ActionContext CreateActionContext(
+                IAccountStateDelta prevStates,
+                int randomSeed
+            ) =>
+                new ActionContext(
+                    signer: Signer,
+                    miner: minerAddress,
+                    blockIndex: blockIndex,
+                    previousStates: prevStates,
+                    randomSeed: randomSeed,
```
longfin: Is it okay not to have `unchecked`?

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -468,6 +511,30 @@ internal BlockLocator GetBlockLocator(int threshold = 10)
         // we need to add a synchronization mechanism to handle this correctly.
         internal void Swap(BlockChain<T> other)
         {
+            // Finds the branch point.
+            Block<T> topmostCommon = null;
+            for (long i = Math.Min(other.Tip.Index, Tip.Index); i >= 0; i--)
```
longfin: We can use `Block<T>.PreviousHash` to avoid index lookup.

Libplanet.Tests/Blocks/BlockTest.cs
```diff
@@ -208,35 +246,70 @@ public void EvaluateActions()
...
+                .ToImmutableArray();
+            i = 0;
+            expectations = new[]
+            {
+                (0, 0, new[] { "A,D", "B", "C", null, null }),
+                (1, 0, new[] { "A,D", "B", "C", "E", null }),
+                (2, 0, new[] { "A,D", "B", "C", "E", "RecordRehearsal:False" }),
+            };
+            Assert.Equal(expectations.Length, tuples.Length);
+            foreach (var tuple in tuples)
```
earlbread: How about using `for` loop if you use `i`?

Libplanet.Tests/Blocks/BlockTest.cs
```diff
@@ -208,35 +246,70 @@ public void EvaluateActions()
...
+            {
+                var expect = expectations[i++];
```
earlbread: Rider says we can use deconstruction like:
```csharp
var (transaction, dumbAction, actionContext, accountStateDelta) = tuples[i];
var (item1, item2, strings) = expectations[i];
```
How about using deconstruction?
It looks more readable.

Libplanet.Tests/Common/Action/DumbAction.cs
```diff
@@ -62,16 +87,113 @@ public IAccountStateDelta Execute(IActionContext context)
...
             RecordRehearsal = (bool)plainValue["record_rehearsal"];
+            RecordRandom =
+                plainValue.ContainsKey("record_random") &&
+                plainValue["record_random"] is bool r &&
+                r;
+        }
+
+        public bool Equals(DumbAction other)
+        {
+            return !ReferenceEquals(null, other) && (
```
earlbread: Is this different from `!(other is null)`?

dahlia: No, it means the same.  It's generated by Rider.  I will changed it to use pattern matching syntax.

Libplanet/Blocks/Block.cs
dahlia: Fixed: 19089e75db6287a8e9de6eb47b92f5d72eef0d1d.

Libplanet/Tx/Transaction.cs
dahlia: Fixed: 19089e75db6287a8e9de6eb47b92f5d72eef0d1d.

Libplanet.Tests/Blocks/BlockTest.cs
dahlia: Fixed by adding a distinct type: 19089e75db6287a8e9de6eb47b92f5d72eef0d1d.

Libplanet.Tests/Common/Action/DumbAction.cs
dahlia: Fixed: 10f136b87996339da9678896db76bdfa6ead6807.

Libplanet.Tests/Blocks/BlockTest.cs
dahlia: Fixed by using `Enumerable.Zip()`: 78faaa43e34ec37cebf0793e13e7d30db92ca11d.

Libplanet/Blockchain/BlockChain.cs
dahlia: Fixed: 9db8f48c4dd539b17559cf3723a9703f28cbeedd.

Libplanet/Tx/Transaction.cs
dahlia: Fixed: 604f9b57060498eecd3c03783be742c39f0260af.

dahlia:
@earlbread @longfin I rebased this on the current master and addressed review feedback.  Could you review this again?

CHANGES.md
```diff
@@ -70,18 +71,30 @@ To be released.
...
 [#205]: https://github.com/planetarium/libplanet/pull/205
 [#206]: https://github.com/planetarium/libplanet/pull/206
 [#210]: https://github.com/planetarium/libplanet/pull/210
 [#213]: https://github.com/planetarium/libplanet/pull/213
 [#215]: https://github.com/planetarium/libplanet/pull/215
+<<<<<<< HEAD
 [Ethereum Homestead algorithm]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md
+=======
+[#216]: https://github.com/planetarium/libplanet/pull/216
+>>>>>>> Format/writing adjustments on CHANGES.md
```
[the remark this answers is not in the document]
dahlia: Fixed!

dahlia:
@longfin Please review this again.

PR #214, opened 2019-04-21
Do not read block hash of out of index

### LINKED ISSUE #208: BlockChain raises KeyNotFoundException when index is not in it
If there is no value at index when `BlockChain[index]` is called, [ArgumentOutofRangeException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) should occur. 
But it raises ` KeyNotFoundException`, because `Store.IndexBlockHash` returns `default(T)` instead of `null`.

image (4)

gurrpi:
This PR fixes #208. An `ArgumentOutOfRangeException` will be thrown instead of `KeyNotFoundException` when `Blockchain<T>[int]` calls the index of a block that does not exist locally.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -85,6 +85,9 @@ To be released.
  -  Fixed `BlockChain<T>.GetStates()` had descended to the bottom
     (i.e., the genesis block) where a given `Address` refers to
     a nonexistent account (i.e., never used before).  [[#189], [#192]]
+ -  Fixed a bug that `KeyNotFoundException` had been thrown instead of
+    `ArgumentOutOfRangeException` when `Blockchain<T>[int]` called while the
+    index of a block that does not exist locally.  [[#208]], [[#214]]
  -  Fixed a bug that a TURN connection had turned unavailable after
     it once failed to parse a message (due to a corrupted packet).
     [[#215]]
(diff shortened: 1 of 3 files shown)


longfin:
@gurrpi Thanks for patch! could you add a mention about this fix on `CHANGES.md`?

gurrpi:
> @gurrpi Thanks for patch! could you add a mention about this fix on `CHANGES.md`?

Sure, I submitted another commits that updates `changelog` and also added test for  `RecentIndex(string @namespace)` — Newly added method by this PR — to suppress Codecov warning.

Libplanet/Store/IStore.cs
```diff
@@ -16,6 +16,8 @@ public interface IStore
 
         long CountIndex(string @namespace);
 
+        long RecentIndex(string @namespace);
```
dahlia: This method doesn't seem used anywhere.  Is this necessary?

gurrpi: Well, of course not. It was a mistake.
I will going to delete the code associated with the method.

gurrpi:
I applied the review comment. Please take a look. @dahlia

longfin:
@gurrpi Thank you for taking the time to work on it. unfortunately, #214 was resolved by another PR (#317) and this PR was closed automatically. 

We still appreciate your work and next contributions.

PR #215, opened 2019-04-24
Network polishing

longfin:
This PR fixes two problems.

1. Current TURN client and relay implementations are vulnerable to packet corruption because they do not retry if parsing fails. To fix this, I modified to ignore exceptions and retry.
2. When forking to apply a chain from another node, use `BlockChain<T>.Fork()` to rewind the current chain to the branch point. I improved performance by eliminating unnecessary validation at this time.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,7 @@ To be released.
     given.
  -  Fixed a bug that TURN relay had been disconnected when being connected for longer than 5
     minutes.
+ -  Improved connection stability of TURN relay.
  -  Instead of validating the entire blocks, `BlockChain<T>.Append()` method
     became to validate only the next block to be appended.  [[#210]]
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,7 @@ To be released.
  -  Improved connection stability of TURN relay.
  -  Instead of validating the entire blocks, `BlockChain<T>.Append()` method
     became to validate only the next block to be appended.  [[#210]]
+ -  Improved `BlockChain<T>.Fork()` performance by omitting unnecessary validation.
 
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,10 +47,13 @@ To be released.
     given.
  -  Fixed a bug that TURN relay had been disconnected when being connected for longer than 5
     minutes.
- -  Improved connection stability of TURN relay.
+ -  Fixed a bug that a TURN connection had turned unavailable after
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -47,8 +47,10 @@ To be released.
     given.
  -  Fixed a bug that TURN relay had been disconnected when being connected for longer than 5
     minutes.
+ -  Improved connection stability of TURN relay.
```
```suggestion
-  Fixed a bug that a TURN connection had turned unavailable after
   it once failed to parse a message (due to a corrupted packet).
   [[#215]]
```

CHANGES.md
```diff
@@ -47,8 +47,10 @@ To be released.
...
  -  Instead of validating the entire blocks, `BlockChain<T>.Append()` method
     became to validate only the next block to be appended.  [[#210]]
+ -  Improved `BlockChain<T>.Fork()` performance by omitting unnecessary validation.
```
```suggestion
 -  Improved `BlockChain<T>.Fork()` performance by avoiding double validation
    of already validated blocks.  [[#215]]
```

Libplanet/Net/Swarm.cs
```diff
@@ -737,20 +737,30 @@ private async Task BindingProxies()
...
-                #pragma warning disable CS4014
-                Task.Run(async () =>
+                try
                 {
-                    using (var proxy = new NetworkStreamProxy(stream))
+                    NetworkStream stream =
+                        await _turnClient.AcceptRelayedStreamAsync();
+
+#pragma warning disable CS4014
+                    Task.Run(async () =>
```
dahlia: Even if we don't need the result, wouldn't it be good to `await` the task at the end of this `while` loop?

CHANGES.md
dahlia: Also need to add `[#215]: https://github.com/planetarium/libplanet/pull/185` below.

Libplanet/Net/Swarm.cs
longfin: If we `await` this task at here, we can't accept other relay stream same time because `await`ing will blocks this loop.

dahlia: @longfin Yes, of course we shouldn't `await` them *in the loop*.  So I suggested `await` them *at the end of this loop* so that these threads are joined to the current thread after the swarm stops.

dahlia: Indeed, apparently my suggestion cannot be applied without more structural change — as we can't maintain the unlimited numbers of `Task` objects (it depends on the number of requests, right?).  In the future, we should make an explicit thread pool and also take the number of workers to be pooled as a constructor parameter, I think.

dahlia:
Need a link reference destination too.

```markdown
[#215]: https://github.com/planetarium/libplanet/pull/185
```

Issue #82, opened 2019-02-18
Links to GitHub source code in the docs seem broken


longfin:
Source links in documentation page seem to broken. (e.g. [link](https://github.com/planetarium/libplanet.net/blob/merge/Libplanet/Address.cs/#L94) in [`Libplanet.Address`](https://planetarium.github.io/libplanet.net/pulls/54/api/Libplanet.Address.html))

dahlia:
This bug is even more complicated in recent builds: <https://planetarium.github.io/libplanet.net/master/api/Libplanet.Address.html>.  The generated `href` is like: <https://github.com/planetarium/libplanet.net/blob/master/Libplanet/obj/Release/netstandard2.0/g/EqualityGenerator/Address_17837EF5.g.cs/#L11>.  It seems because Uno.CodeGen generates partial files under the hood.

PR #216, opened 2019-04-24
.NET Framework 4.6 & Uno.CodeGen → .NET Core 2.2 & Equals.Fody

dahlia:
This patch replaces Uno.CodeGen, a barrier to adopting .NET Core, with [Equals.Fody], and makes the build process depends on `dotnet` CLI instead of `msbuild`.  (Of course, `msbuild` also still works.)

Pros of .NET Core are:

 -  Behaviors across platforms are more consistent than .NET Framework v. Mono.
 -  Both tests (`Libplanet.Tests` & `Libplanet.Stun.Tests`) can be run at a time on CLI (`dotnet test`).
 -  Tests can be `--filter`ed on CLI as well (whereas xunit.runner.msbuild is only able to run the whole tests.)

Pros of Equals.Fody:

 -  As it rewrites code at IL level, types don't have to be `partial`.
     -  So it's entirely free from the bug #82 without any tricks.
     -  We don't need to add lines of preprocessor directives to suppress warnings.
 -  Works with .NET Framework, Mono, and .NET Core.

Also note that there are some limitations on Equals.Fody when it's compared to Uno.CodeGen:

 -  It doesn't work with generic `struct`s (and generic `class`es?  Not sure).  See `HashDigest<T>` as well; its equality methods are written by hand.
 -  Fields to compare are opt-in by default and can be opt-out by annotating `[IgnoreDuringEquals]` attribute.  This leads types with one identity property and other many properties to look dirty.  See `Block<T>`for example; its properties all have an `[IgnoreDuringEquals]` attribute except for the `Hash` property.
 -  Apparently `operator ==` and `operator !=` are not generated.

[Equals.Fody]: https://github.com/Fody/Equals

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@ Version 0.3.0
 To be released.
 
  -  `BlockChain<T>` became to implement `IReadOnlyList<Block<T>>`.  [[#205]]
- -  `BlockChain<T>.Validate()` method became to receive 
+ -  `BlockChain<T>.Validate()` method became to receive
     `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.  [[#205]]
  -  `IBlockPolicy<T>.GetNextBlockDifficulty()` method became to receive
     `IReadOnlyList<Block<<T>>` instead of `IEnumerable<Block<T>>`.  [[#205]]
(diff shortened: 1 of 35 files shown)


CHANGES.md
```diff
@@ -31,6 +31,10 @@ To be released.
     [[#204]], [[#206]]
  -  Added `BlockDownloadState` class to represent a block downloading state.
     [[#204]], [[#206]]
+ -  Removed `KeyEquals()` methods from all classes and structs.
+ -  `Swarm` class now does not `IEquatable<Swarm>` anymore and
```
```suggestion
 -  `Swarm` class now does not implement `IEquatable<Swarm>` anymore and
```

carldebilly:
@dahlia `Uno.Code` now works well with dotnet core 3.0.

PR #217, opened 2019-04-25
Tune lock in BlockChain<T>

longfin:
This PR adjusts the locks on `BlockChain<T>`.

- Removed unnecessary writer lock in `BlockChain<T>.StageTransactions()`
- Adjust `BlockChain<T>.GetState()` locks to reduce locking period.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -75,6 +75,8 @@ To be released.
     longer than 5 minutes.  [[#198]]
  -  Fixed a bug that `Swarm` had attempted to use TURN relay even if the `host`
     argument was given.  [[#198]]
+ -  Removed unnecessary writer lock on `BlockChain<T>.StageTransactions()`. [[#217]]
+ -  Improve concurrency of `BlockChain<T>.Append()`. [[#217]]
 
 [Ethereum Homestead algorithm]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md
 [#185]: https://github.com/planetarium/libplanet/pull/185
@@ -92,6 +94,7 @@ To be released.
 [#213]: https://github.com/planetarium/libplanet/pull/213
 [#215]: https://github.com/planetarium/libplanet/pull/215
 [#216]: https://github.com/planetarium/libplanet/pull/216
+[#217]: https://github.com/planetarium/libplanet/pull/217
 
 
 Version 0.2.2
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -70,6 +70,8 @@ To be released.
  -  The difficulty was changed from representing the number of leading zeros of
     target number to representing a divisor to obtain the target number.
     [[#213]]
+ -  Removed unnecessary reader lock on `BlockChain<T>.StageTrasactions()`. [[#217]]
+ -  Improve concurrency of `BlockChain<T>.GetState()`. [[#217]]
```
dahlia: It should be `GetStates()`.

CHANGES.md
```diff
@@ -70,6 +70,8 @@ To be released.
  -  The difficulty was changed from representing the number of leading zeros of
     target number to representing a divisor to obtain the target number.
     [[#213]]
+ -  Removed unnecessary reader lock on `BlockChain<T>.StageTrasactions()`. [[#217]]
```
dahlia: Was the removed thing a reader lock?  Wasn't it a writer lock?

longfin: You're right. it was a writer lock.

CHANGES.md
dahlia: Hmm… the change was actually made in `Append()`.  Would this explanation be just fine?

longfin: It's true that it has been competing mostly with `GetStates`(), but it's probably better to write that the `Append()` because the change was made in `Append()` like you say.

CHANGES.md
```diff
@@ -70,6 +70,8 @@ To be released.
  -  The difficulty was changed from representing the number of leading zeros of
     target number to representing a divisor to obtain the target number.
     [[#213]]
+ -  Removed unnecessary writer lock on `BlockChain<T>.StageTrasactions()`. [[#217]]
```
```suggestion
 -  Removed unnecessary writer lock on `BlockChain<T>.StageTransactions()`. [[#217]]
```

longfin:
I've rebased (and made some changes related #212 ). cc @dahlia

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -600,32 +591,34 @@ internal void Swap(BlockChain<T> other)
...
                     return null;
                 }
-            }).ToArray();
-            IAccountStateDelta lastStates = tuples.Any()
-                ? tuples[tuples.Length - 1].Item2.OutputStates
-                : null;
+            }).Select(t => t.Item2).ToArray();
+        }
 
+        private void SetStateFromEvaluations(
```
dahlia: We'd better use the plural term “states” here for consistency (with `SetStates()`/`SetBlockStates()` etc).

PR #223, opened 2019-04-29
Add BlockChain<T>.UnstageTransactions()

earlbread:
This patch adds `BlockChain<T>.UnstageTransactions()` method.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -265,16 +265,35 @@ namespace Libplanet.Blockchain
         public void Append(Block<T> block, DateTimeOffset currentTime) =>
             Append(block, currentTime, render: true);
 
-        public void StageTransactions(ISet<Transaction<T>> txs)
+        /// <summary>
+        /// Adds <paramref name="transactions"/> to the pending list so that
+        /// a next <see cref="Block{T}"/> to be mined contains these
+        /// <paramref name="transactions"/>.
+        /// </summary>
+        /// <param name="transactions">
+        /// <see cref="Transaction{T}"/>s to add to the pending list.</param>
+        public void StageTransactions(ISet<Transaction<T>> transactions)
         {
-            foreach (Transaction<T> tx in txs)
+            foreach (Transaction<T> tx in transactions)
             {
                 Transactions[tx.Id] = tx;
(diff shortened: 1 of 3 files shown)


dahlia:
What would this method be used for?

earlbread:
> What would this method be used for?

We need this method to avoid duplicate rewards in games.

dahlia:
Could you rebase this on the current master?

earlbread:
I rebased this on the master branch. @dahlia

earlbread:
I rebased again. Could you please review again? @dahlia @longfin

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -265,18 +265,42 @@ bool IsPossiblyIn(BitArray addr, BitArray maskPattern)
...
+        /// </summary>
+        /// <param name="transactions"><see cref="Transaction{T}"/>s
+        /// to remove from the pending list.</param>
+        /// <seealso cref="StageTransactions"/>
+        public void UnstageTransactions(ISet<Transaction<T>> transactions)
+        {
+            foreach (Transaction<T> tx in transactions)
+            {
+                Transactions.Remove(tx.Id);
+            }
```
dahlia: @longfin Should we remove these transactions from pool as well?

This could break things, for example:

```csharp
b = Block<Action>.Mine(..., transactions: txs);
blockChain.Append(b);
blockChain.UnstageTransactions(txs);
Block b2 = blockChain[b.blockHash];  // This will crash.
```

longfin: Yes, it can be problem. and, I think that `UnstageTransactions()` should remove only staged transactions, not mined transaction.

earlbread: You're right. So should we make like `BlockChain<T>.UnstageTransactionIds()` instead of `BlockChain<T>.UnstageTransactions()`?

dahlia: The current name looks better to me, because it exposes less implementation details.

earlbread:
I made `BlockChain<T>.UnstageTransactionIds()` method remove only staged transactions, not mined transaction. Please take a look. @dahlia @longfin

PR #227, opened 2019-04-30
Fix UnstageTransactions() test

earlbread:
Currently, `UnstageTransactions()` sometimes fails because of the order of `TxId`. To successfully take the test in any order, I changed `List<TxId>` to `HashSet<TxId>`.


longfin:
It seems to need changelog or `[changelog skip]`.

earlbread:
I added `[changelog skip]`. Could you please review this again? @dahlia @longfin

PR #229, opened 2019-05-03
Build docs using GitHub Actions

dahlia:
As currently Mono has not been installed in Travis CI builds, DocFX (which is based on .NET Framework/Mono) has not resolved types in the standard library.  So I removed a process to build docs from Travis CI, and added a GitHub Actions workflow to build docs using DocFX with Mono.

### DIFF
--- a/Docs/publish.sh
+++ b/Docs/publish.sh
@@ -1,8 +1,12 @@
-#!/bin/bash
+#!/bin/ash
 # Publish docs to GitHub Pages.
-# Note that this script is intended to be run by Travis CI.
-if ! (env | grep '^TRAVIS_'); then
-  echo "This script is intended to be run by Travis CI." > /dev/stderr
+# Note that this script is intended to be run by GitHub Actions.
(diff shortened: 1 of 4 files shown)


Docs/publish.sh
```diff
@@ -1,8 +1,12 @@
-#!/bin/bash
+#!/bin/ash
```
earlbread:
```suggestion
#!/bin/sh
```

dahlia: It's not typo, but means [Almquist Shell][1] (hence `ash`), which an extended variant of POSIX Bourne Shell (`sh`).  I switched from Bourne-Again Shell (`bash`) to `ash` here, for better portability. Some Linux distros like Alpine Linux, which is a de facto standard for Docker images, lacks `bash` by default.

[1]: https://en.wikipedia.org/wiki/Almquist_shell

earlbread: That's good. I didn't know about `ash`.

PR #230, opened 2019-05-03
Remove storing addresses mask into blocks

earlbread:
- This patch reverts #197 to make indexing for states.


earlbread:
I will open another PR with indexing for states.

PR #231, opened 2019-05-07
Remove validation when getting block

earlbread:
- Removed `block.Validate()` when getting a block using `Blocks[index]`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -70,6 +70,7 @@ To be released.
  -  The difficulty was changed from representing the number of leading zeros of
     target number to representing a divisor to obtain the target number.
     [[#213]]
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -70,6 +70,7 @@ To be released.
  -  The difficulty was changed from representing the number of leading zeros of
     target number to representing a divisor to obtain the target number.
     [[#213]]
+ -  `BlockSet<T>[int]` changed so as not to validate a block.
```
longfin: I also think that previous behavior (validate when getting) wasn't expectable, so we can skip this phrase. (or organize at later.)

earlbread: It is already in behavioral changes. Did you mean bug fixes?

PR #232, opened 2019-05-07
Add block hash index which has address states

earlbread:
- Added `IStore.LookupStateReference<T>()` method.
- Added `IStore.StoreStateReference<T>()` method.
- Added `IStore.ForkStateReferences<T>()` method.
- Changed `BlockChain<T>.GetStates()` to use a state reference, which is a `Block<T>.Hash` that has the state of the `Address`.
- Added documentation of `BlockChain<T>.GetStates()`.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -354,7 +354,14 @@ namespace Libplanet.Blockchain
                         .ToImmutableHashSet();
 
                     Store.UnstageTransactionIds(txIds);
-                    Store.SetAddressStateBlockHash(Id.ToString(), block);
+
+                    ImmutableHashSet<Address> updatedAddresses =
+                        block.Transactions
+                        .SelectMany(tx => tx.UpdatedAddresses)
+                        .ToImmutableHashSet();
+
+                    Store.SetAddressStateBlockHash(
+                        Id.ToString(), block, updatedAddresses);
                 }
                 finally
                 {
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -172,56 +172,19 @@ namespace Libplanet.Blockchain
                 _rwlock.ExitReadLock();
             }
 
-            ImmutableDictionary<Address, BitArray> requestedAddresses =
-                addresses.Select(addr =>
-                    new KeyValuePair<Address, BitArray>(
-                        addr,
-                        new BitArray(addr.ToByteArray())
-                    )
-                ).ToImmutableDictionary();
-            ImmutableHashSet<Address> requestedAddressSet =
-                requestedAddresses.Keys.ToImmutableHashSet();
-
-            bool IsPossiblyIn(BitArray addr, BitArray maskPattern)
-            {
-                var match = (BitArray)maskPattern.Clone();
-                match.And(addr);
-                return match.OfType<bool>().Any(bit => bit);
-            }
-
-            var coveredAddresses = new HashSet<Address>();
+            ImmutableHashSet<Address> requestedAddresses =
+                addresses.ToImmutableHashSet();
             var states = new AddressStateMap();
-
             while (offset != null)
             {
                 states = (AddressStateMap)states.SetItems(
                     Store.GetBlockStates(offset.Value)
                     .Where(
-                        kv => requestedAddresses.ContainsKey(kv.Key) &&
+                        kv => requestedAddresses.Contains(kv.Key) &&
                         !states.ContainsKey(kv.Key))
                     );
 
-                // If there is no mask for the offset (the store made in
-                // the older versions of Libplanet may lack mask files),
-                // make the mask to match to all possible addresses.
-                var mask = new BitArray(
-                    offset is HashDigest<SHA256> hash &&
-                        Store.GetAddressesMask(hash) is Address a
-                        ? a.ToByteArray()
-                        : BlockSet<T>.WildcardMask
-                );
-
-                // Addresses that are definitely not existent.
-                coveredAddresses.UnionWith(
-                    requestedAddresses.Where(
-                        pair => !IsPossiblyIn(pair.Value, mask)
-                    ).Select(pair => pair.Key)
-                );
-
-                // Already gathered addresses.
-                coveredAddresses.UnionWith(states.Keys);
-
-                if (requestedAddressSet.SetEquals(coveredAddresses))
+                if (requestedAddresses.SetEquals(states.Keys))
                 {
                     break;
                 }
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -172,24 +172,37 @@ namespace Libplanet.Blockchain
                 _rwlock.ExitReadLock();
             }
 
+            var states = new AddressStateMap();
+
+            if (offset == null)
+            {
+                return states;
+            }
+
+            Block<T> block = Blocks[offset.Value];
+            long blockIndex = block.Index;
+
             ImmutableHashSet<Address> requestedAddresses =
                 addresses.ToImmutableHashSet();
-            var states = new AddressStateMap();
-            while (offset != null)
+            var hashValues = new HashSet<HashDigest<SHA256>>();
+
+            foreach (var address in requestedAddresses)
+            {
+                var hashDigest =
+                    Store.GetAddressStateBlockHash(address, blockIndex);
+                if (!(hashDigest is null))
+                {
+                    hashValues.Add(hashDigest.Value);
+                }
+            }
+
+            foreach (var hashValue in hashValues)
             {
                 states = (AddressStateMap)states.SetItems(
-                    Store.GetBlockStates(offset.Value)
-                    .Where(
-                        kv => requestedAddresses.Contains(kv.Key) &&
-                        !states.ContainsKey(kv.Key))
-                    );
-
-                if (requestedAddresses.SetEquals(states.Keys))
-                {
-                    break;
-                }
-
-                offset = Blocks[offset.Value].PreviousHash;
+                        Store.GetBlockStates(hashValue)
+                        .Where(
+                            kv => requestedAddresses.Contains(kv.Key) &&
+                            !states.ContainsKey(kv.Key)));
             }
 
             return states;
@@ -330,6 +343,7 @@ namespace Libplanet.Blockchain
                         .ToImmutableHashSet();
 
                     Store.UnstageTransactionIds(txIds);
+                    Store.SetAddressStateBlockHash(block);
                 }
                 finally
                 {
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -471,9 +471,12 @@ namespace Libplanet.Blockchain
 
                 var toUpdateAddresses = new HashSet<Address>();
 
-                for (var i = pointBlock.Index + 1; i <= Tip.Index; i++)
+                for (
+                    Block<T> block = Tip;
+                    block.PreviousHash is HashDigest<SHA256> hash
+                    && !block.Hash.Equals(point);
+                    block = Blocks[hash])
                 {
-                    Block<T> block = this[i];
                     toUpdateAddresses.UnionWith(
                         block.Transactions
                             .SelectMany(tx => tx.UpdatedAddresses)
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -188,8 +188,8 @@ namespace Libplanet.Blockchain
 
             foreach (var address in requestedAddresses)
             {
-                var hashDigest =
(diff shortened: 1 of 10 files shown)


Libplanet/Store/IStore.cs
```diff
@@ -85,6 +65,59 @@ void PutBlock<T>(Block<T> block, Address addressesMask)
...
+        /// <param name="block">The <see cref="Block{T}"/> which has the state
+        /// of the <see cref="Address"/>.</param>
+        /// <typeparam name="T">An <see cref="IAction"/> class used with
+        /// <paramref name="block"/>.</typeparam>
+        /// <seealso cref="GetAddressStateBlockHash"/>
+        void SetAddressStateBlockHash<T>(string @namespace, Block<T> block)
+            where T : IAction, new();
+
+        /// <summary>
+        /// Fork <see cref="Block{T}.Hash"/>es which has the state of the
```
dahlia: If a sentence in XML comment starts with a verb with a subject omitted it should be third-person singular tense.  See also other existing XML comments.

Libplanet/Store/IStore.cs
```diff
@@ -85,6 +65,59 @@ void PutBlock<T>(Block<T> block, Address addressesMask)
...
+        /// <paramref name="targetNamespace"/>.
+        /// </summary>
+        /// <param name="sourceNamespace">The namespace to fork
+        /// <see cref="Block{T}.Hash"/>es.</param>
+        /// <param name="targetNamespace">The namespace where the forked
+        /// <see cref="Block{T}.Hash"/> will be stored.</param>
+        /// <param name="branchPointIndex">The index of branch point to fork.
+        /// </param>
+        /// <param name="toUpdateAddresses">The set of <see cref="Address"/>es
+        /// to be updated by fork in the <paramref name="targetNamespace"/>.
```
dahlia: The parameter name and its description seem to neither its intention nor behavior.  It takes a set of addresses that *had been updated* from `branchPointIndex` to the *source*, so its description <q>to be updated by fork in the `targetNamespace`</q> seems wrong and misleading, right?  To me, it looks like addresses to be stripped or invalidated.

Libplanet/Store/IStore.cs
```diff
@@ -85,6 +65,59 @@ void PutBlock<T>(Block<T> block, Address addressesMask)
...
+        /// <summary>
+        /// Gets a <see cref="Block{T}.Hash"/> which has the state of the
+        /// <paramref name="address"/>.
+        /// </summary>
+        /// <param name="namespace">The namespace to get
+        /// <see cref="Block{T}.Hash"/>.</param>
+        /// <param name="address">The <see cref="Address"/> to find the
+        /// <see cref="Block{T}.Hash"/>.</param>
+        /// <param name="offsetIndex">The offset at which to begin looking for
+        /// a block hash.</param>
```
dahlia: This parameter and `ForkAddressStateBlockHash()` method's `branchPointIndex` takes a block index instead of a block hash, but is there any intention on this?  To me, these parameters can take a block hash and it's less error-prone as block hashes are hard to overlap by mistake (whereas block indexes are easily overlapped and the same index can refer to different blocks depending on branch).

dahlia: Oh, I got it.  These parameters cannot take a block hash because these operations depend on `<=`/`>=` comparisons.

Libplanet/Store/FileStore.cs
```diff
@@ -545,6 +532,145 @@ AddressStateMap states
...
+        {
+            HashDigest<SHA256> blockHash = block.Hash;
+            long blockIndex = block.Index;
+            int hashSize = HashDigest<SHA256>.Size;
+            int blockInfoSize = hashSize + sizeof(long);
+            var buffer = new byte[blockInfoSize];
+
+            ImmutableHashSet<Address> updatedAddresses = block.Transactions
+                .SelectMany(tx => tx.UpdatedAddresses)
+                .ToImmutableHashSet();
```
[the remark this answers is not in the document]
earlbread: Right, this behavior has to same regardless of implementation. I'll move this to the caller and pass the result as an argument.

Libplanet/Store/FileStore.cs
```diff
@@ -545,6 +532,145 @@ AddressStateMap states
...
+            long branchPointIndex,
+            IImmutableSet<Address> toUpdateAddresses)
+        {
+            string sourceDir = GetAddressStateBlockHashPath(sourceNamespace);
+            string targetDir = GetAddressStateBlockHashPath(targetNamespace);
+            var copied = CopyDirectory(sourceDir, targetDir);
+
+            if (!copied)
+            {
+                return;
```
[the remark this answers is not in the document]
earlbread: `copied` is `false` when `sourceDirectory` does not exist. I thought if there are no states in the chain that are updated by actions, the `sourceDirectory` may not exist.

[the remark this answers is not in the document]
dahlia: IMO that case shouldn't be silent no-op but should throw something.

Libplanet/Store/IStore.cs
earlbread: You're right, I meant <q>addresses to be stripped</q>. I'll fix the parameter name and its description.

earlbread:
I applied review comments and rebased on the master. Also, I renamed methods using a word <q>StateReference</q>. Please take a look. @dahlia @longfin

Libplanet/Store/FileStore.cs
```diff
@@ -545,6 +532,141 @@ AddressStateMap states
...
+            Block<T> branchPoint,
+            IImmutableSet<Address> addressesToStrip)
+        {
+            string sourceDir = GetStateReferencePath(sourceNamespace);
+            string targetDir = GetStateReferencePath(destNamespace);
+            bool copied = CopyDirectory(sourceDir, targetDir);
+
+            if (!copied && addressesToStrip.Any())
+            {
+                throw new DirectoryNotFoundException(
```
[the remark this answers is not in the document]
earlbread: Agree, I'll add a new exception for this case.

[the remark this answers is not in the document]
earlbread: I made `NamespaceNotFoundException`.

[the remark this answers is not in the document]
dahlia: That's also should be added to the changelog.

Libplanet/Store/NamespaceNotFoundException.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// </summary>
+    [Serializable]
+    public sealed class NamespaceNotFoundException : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the
+        /// <see cref="NamespaceNotFoundException"/> class.
+        /// </summary>
+        /// <param name="message">The message that describes the error.</param>
+        public NamespaceNotFoundException(string message)
```
dahlia: It would be good if this exception has a property to refer to the namespace tried to find.

earlbread:
I amended a commit. Please take a look.

Libplanet/Store/IStore.cs
```diff
@@ -85,6 +65,80 @@ void PutBlock<T>(Block<T> block, Address addressesMask)
...
+        /// <summary>
+        /// Lookup a state reference, which is a <see cref="Block{T}.Hash"/>
+        /// that has the state of the <paramref name="address"/>.
+        /// </summary>
+        /// <param name="namespace">The namespace to lookup a state reference.
+        /// </param>
+        /// <param name="address">The <see cref="Address"/> to find lookup.
+        /// </param>
+        /// <param name="lookupFrom">The starting point to lookup a state
+        /// reference. Blocks after <paramref name="lookupFrom"/> are ignored.
```
dahlia: How about changing the description like: <q>The upper bound (i.e., the latest block) of the search range.  Blocks after `lookupUntil` are ignored.</q>

Libplanet/Store/IStore.cs
```diff
@@ -85,6 +65,80 @@ void PutBlock<T>(Block<T> block, Address addressesMask)
...
 
+        /// <summary>
+        /// Lookup a state reference, which is a <see cref="Block{T}.Hash"/>
+        /// that has the state of the <paramref name="address"/>.
+        /// </summary>
+        /// <param name="namespace">The namespace to lookup a state reference.
+        /// </param>
+        /// <param name="address">The <see cref="Address"/> to find lookup.
+        /// </param>
+        /// <param name="lookupFrom">The starting point to lookup a state
```
dahlia: I suggest another parameter name: `lookupUntil`.

Libplanet/Store/IStore.cs
```diff
@@ -85,6 +65,80 @@ void PutBlock<T>(Block<T> block, Address addressesMask)
...
+            string @namespace,
+            IImmutableSet<Address> addresses,
+            Block<T> block)
+            where T : IAction, new();
+
+        /// <summary>
+        /// Forks state references, which are <see cref="Block{T}.Hash"/>es that
+        /// have the state of the <see cref="Address"/>es, from
+        /// <paramref name="sourceNamespace"/> to
+        /// <paramref name="destNamespace"/>.
```
dahlia: To me `destinationNamespace` looks better.

Libplanet/Store/IStore.cs
earlbread: Changed description and parameter name.

Libplanet/Store/IStore.cs
earlbread: Changed.

Libplanet/Store/BaseStore.cs
```diff
@@ -52,11 +53,7 @@ ISet<TxId> txids
         public abstract Block<T> GetBlock<T>(HashDigest<SHA256> blockHash)
             where T : IAction, new();
 
-        /// <inheritdoc />
-        public abstract Address? GetAddressesMask(HashDigest<SHA256> blockHash);
-
-        /// <inheritdoc />
-        public abstract void PutBlock<T>(Block<T> block, Address addressesMask)
+        public abstract void PutBlock<T>(Block<T> block)
```
longfin:
```suggestion
        /// <inheritdoc />
        public abstract void PutBlock<T>(Block<T> block)
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -472,6 +466,25 @@ internal BlockChain<T> Fork(HashDigest<SHA256> point)
                         break;
                     }
                 }
+
+                Block<T> pointBlock = Blocks[point];
+
+                var toUpdateAddresses = new HashSet<Address>();
+
+                for (var i = pointBlock.Index + 1; i <= Tip.Index; i++)
```
longfin: (minor) We can omit block index lookup by looping reverse order using `Tip.PreviousHash` and `Blocks[Hash]`.

earlbread:
I rebased on the master and applied review comments. Please take a look.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -472,6 +466,28 @@ internal BlockChain<T> Fork(HashDigest<SHA256> point)
...
                     }
                 }
+
+                Block<T> pointBlock = Blocks[point];
+
+                var toUpdateAddresses = new HashSet<Address>();
+
+                for (
+                    Block<T> block = Tip;
+                    block.PreviousHash is HashDigest<SHA256> hash
```
dahlia: This condition won't work if `block` refers to a block right after genesis, which means addresses updated by the second block (where `blockIndex == 1`) are always excluded from the index.

earlbread: I think this condition is break when the block refers to the genesis block, not the second block. But I thought it is ok because every forking must have the genesis block.
I added tests for this case and found another bug and fixed it.

earlbread:
I have fixed the part that uses `Transaction.UpdatedAddress`. Because it is for validation, it may be different from the result of `BlockChain<T>.SetStates`.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -359,14 +345,19 @@ bool render
...
+                    SetStates(block, evaluations);
 
                     Store.AppendIndex(Id.ToString(), block.Hash);
                     ISet<TxId> txIds = block.Transactions
                         .Select(t => t.Id)
                         .ToImmutableHashSet();
 
                     Store.UnstageTransactionIds(txIds);
+
+                    ImmutableHashSet<Address> updatedAddresses =
```
longfin: Is this variable needed yet?

earlbread: No, it doesn't. I removed it.

longfin:
> I have fixed the part that uses `Transaction.UpdatedAddress`. Because it is for validation, it may be different from the result of `BlockChain<T>.SetStates`.

It would be also better if there is a regression test about it.

longfin:
> > I have fixed the part that uses `Transaction.UpdatedAddress`. Because it is for validation, it may be different from the result of `BlockChain<T>.SetStates`.
> 
> It would be also better if there is a regression test about it.

I've missed. it's added 👍

PR #235, opened 2019-05-10
Nupkg distribution using GitHub Actions

dahlia:
The process that builds a *Libplanet.\*.nupkg* and uploads it to NuGet Gallery/GitHub Releases has moved from Travis CI to GitHub Actions.

As GitHub Actions doesn't have build number (unlike Travis CI), the number after the version suffix `dev.` becomes a timestamp like `0.3.0-dev.20190511051659`.


.github/bin/dist-github-release.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/ash
```
longfin:
```suggestion
#!/bin/sh
```

or, did you intend [amquist shell](https://en.wikipedia.org/wiki/Almquist_shell)? (If so, is there any special reason?)

dahlia: Yes, it's Amquist shell by intention, because it's closest to Bash and there's no bash in this Docker image. 😭

longfin: So are these scripts okay to run with `/bin/sh`?

- https://github.com/planetarium/libplanet/pull/235/files#diff-4a84523e74f96c4e221c2e75647af2ceR1
- https://github.com/planetarium/libplanet/pull/235/files#diff-906da86a1ded732ffc59f7d5e075a846R1

dahlia: These are not intention, maybe. I will make these to use ash too.

dahlia: Ah, these were intention! These two scripts are run on `mcr.microsoft.com/dotnet/core/sdk:2.2` image, which have neither bash nor ash… it has only POSIX sh. 😇

PR #236, opened 2019-05-11
Fix errors during unittest on Mono

longfin:
This PR aids some unittest errors on Mono runtime.

- Made `Swarm.StopAsync()` closes internal resources(e.g. `_broadcastQueue`) to avoid too many open files error.
  - Also, `SwarmTest` now cleans NetMQ resources at the end of each test case.
- Bump .netframework version of `Libplanet.Tests` used by Mono to avoid missing method exception.

I omitted changelog because there is no difference from previous version.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -85,6 +85,7 @@ To be released.
  -  `BlockSet<T>[int]` changed so as not to validate a block.  [[#231]]
  -  Improved read performance of `Block<T>.Hash` and `Transaction<T>.Id`. [[#228],
     [#234]]
+ -  `Swarm.StartAsync()` doesn't call `Swarm.StopAsync()` anymore.
 
 # Bug fixes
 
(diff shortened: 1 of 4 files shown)


longfin:
I've add some commit.

- In macOS + .NET Core, `NetMQConfig.Cleanup()` seems to hanging. so add condition as workaround.
- Remove `StopAsync()` in `Swarm.StartAsync()`.

CHANGES.md
```diff
@@ -85,6 +85,7 @@ To be released.
  -  `BlockSet<T>[int]` changed so as not to validate a block.  [[#231]]
  -  Improved read performance of `Block<T>.Hash` and `Transaction<T>.Id`. [[#228],
     [#234]]
+ -  `Swarm.StartAsync()` doesn't call `Swarm.StopAsync()` anymore.
```
dahlia: So users need some action due to this change?

longfin: If there is a code that relied on this process (without explicitly calling `StopAsync()`), they should now explicitly call `StopAsync()`.

PR #238, opened 2019-05-14
Use Azure Pipelines as our main CI instead of Travis CI

dahlia:
Although we'd switched from Azure Pipelines to Travis CI, the reason behind this decision was that Azure Pipelines does not support tag-push-triggered builds.  Tag-push triggers was needed to automate releases, but now we automate it using GitHub Actions and we currently do not use Travis CI's tag-push triggers.

On the other hand, Azure Pipelines is way faster than Travis CI, because Travis CI's .NET/C# environment is community-supported, whereas .NET/C# in Azure Pipelines is the first-class citizen and the primary target (as both are made by Microsoft).

For smoother migration path, I didn't entirely remove *.travis.yml* yet.  We'd better remove it entirely after we wait and see if it works well.  So we would work under the dual system for the mean time.

Plus: I made that docs for pull requests are also generated.


Docs/publish.sh
```diff
@@ -32,6 +32,21 @@ if [ "$GHPAGES_SSH_KEY" = "" ]; then
...
+    chmod +x /usr/local/bin/jq
+  fi
+  pr_number="$(jq '.pull_request.number' "$GITHUB_EVENT_PATH")"
+  slug="pulls/$pr_number"
+else
+  slug="$(echo -n "$GITHUB_REF" | sed -e 's/^refs\/\(heads\|tags\)\///g')"
+fi
+[ "$slug" != "" ]
+
+echo "Slug: $slug" > /dev/stderr
```
earlbread: Is there a reason to send this to stderr?

dahlia:
Because it is just a log. This should be excluded if there is a process that takes its stdout as their stdin through pipe.

PR #239, opened 2019-05-15
Bump netmq version to 4.0.0.207-pre

longfin:
This PR bumps the netmq version to [4.0.0.207-pre](https://www.nuget.org/packages/NetMQ/4.0.0.207-pre) to avoid unexpected errors on macOS. 

I omitted change log because it doesn't change behavior.

### DIFF
--- a/Libplanet/Libplanet.csproj
+++ b/Libplanet/Libplanet.csproj
@@ -38,7 +38,7 @@ https://docs.libplanet.io/</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
     <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <NoWarn>$(NoWarn);CS1591;NU5104</NoWarn>
     <!-- FIXME: CS1591 should be turned on eventually. -->
     <IsTestProject>false</IsTestProject>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


dahlia:
Could you amend the last commit hash and force-push it again, so that a new build gets started?  The failed build was started before I configured the codecov.io token in Azure Pipelines settings, and it's why it failed.

Libplanet/Libplanet.csproj
```diff
@@ -38,7 +38,7 @@ https://docs.libplanet.io/</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
     <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
dahlia: Why was this removed?

longfin: It wasn't intended. I'll restore them.

Libplanet/Libplanet.csproj
```diff
@@ -38,7 +38,7 @@ https://docs.libplanet.io/</Description>
...
+    <NoWarn>1701;1702;CS1591;NU5104</NoWarn>
```
dahlia: Could you disambiguate `1701` and `1702` by prepending `CS` prefix, like `CS1701` and `CS1702`?

dahlia: A pinpoint control is probably possible instead of ignoring a warning in the entire project, like:

```
<PackageReference Include="NetMQ" Version="4.0.0.207-pre" />
  <NoWarn>NU5104</NoWarn>
</PackageReference>
```

longfin: In my env, `dotnet build` looks like it isn't affected by `<NoWarn>` in `<PackageReference>`...

dahlia: Okay, it seems due to `NU5104` is for a project that depends.

PR #241, opened 2019-05-15
Remove state comparison in EvaluateActionsGradually

longfin:
This PR removes state comparison in `AccountStateDeltaImpl.SetState()` (used by `IAction.Execute()`) to improve performance of action evaluation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -91,6 +91,9 @@ To be released.
     [[#228], [#234]]
  -  `Swarm.StartAsync()` now does not call `Swarm.StopAsync()` anymore,
     therefore `Swarm.StopAsync()` should be explicitly called.  [[#236]]
+ -  `Transaction<T>.EvaluateActionsGradually()` became to record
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -91,6 +91,8 @@ To be released.
     [[#228], [#234]]
  -  `Swarm.StartAsync()` now does not call `Swarm.StopAsync()` anymore,
     therefore `Swarm.StopAsync()` should be explicitly called.  [[#236]]
+ -  `Transaction<T>.EvaluateActionsGradually()` became to produce next states
+    without comparison even if states are not changed. [[#241]]
```
dahlia: This seems quite ambiguous, because it sounds like it will produce some delta even if I code like `IAction.Execute() => actionContext.PreviousStates`.  How about rewarding this like:

> `Transaction<T>.EvaluateActionsGradually()` became to record `IAccountStateDelta.SetState()` calls even if its argument is the same value to the previous state.

Issue #125, opened 2019-03-09
Miners can fabricate transactions (i.e., replay attack) at some degree


dahlia:
Currently miners can fabricate transactions signed by others by replicating them.

Suppose we have a transaction *t* from *a¹* to *a²* having an action of “*a¹* transferring $1 to *a²*” like:

```c#
AddressStateMap Execute(IActionContext context)
{
    return (AddressStateMap)context.PreviousStates.SetItem(
        context.To,
        ((int)context.PreviousStates[context.To]) + 1
    ).SetItem(
        context.From,
        ((int)context.PreviousStates[context.From]) - 1
    );
}
```

In this scenario, if a miner shares incentive with *a²* to get more dollars they can replicate *t* and mine multiple blocks including the transaction, and *a²* can more than $1.  The current implementation does not find this kind of abusing invalid, because contents of a transaction and the signature derived from it is not affected by which block it belongs to.

In order to prevent this, we may need to add `Nonce` property referring the previous `Block<T>.Hash` to `Transaction<T>`.

dahlia:
As this requires a transaction to be signed again if it is failed to belong to the latest block, the concept of “staged transactions” (i.e., txpool) would be less useful.  From the application developer's perspective, the concept of “unsigned transactions to stage actions” might be more useful then.

PR #242, opened 2019-05-16
Prevent multiple transactions of the same signer

earlbread:
This patch prevents multiple transactions of the same signer from being stored in one block to prevent transaction replay.  #125

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -73,6 +73,7 @@ To be released.
     parameter.  This purposes to determine how long to wait for pending
     messages when a `Swarm` instance is requested to terminate.
  -  Added `NamespaceNotFoundException` class.  [[#232]]
+ -  Added `SimultaneousTxsException` class.  [[#242]]
 
 ### Behavioral changes
 
@@ -94,6 +95,8 @@ To be released.
  -  `Transaction<T>.EvaluateActionsGradually()` became to record
     `IAccountStateDelta.SetState()` calls even if its argument is the same
     to the previous state. [[#241]]
+ -  A signer became to allowed to have only one transaction at most in 
+    `Block<T>.Transactions`. [[#125]], [[#242]]
 
 ### Bug fixes
 
@@ -125,6 +128,7 @@ To be released.
 
 [Ethereum Homestead algorithm]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md
 [#31]: https://github.com/planetarium/libplanet/issues/31
+[#125]: https://github.com/planetarium/libplanet/issues/125
 [#185]: https://github.com/planetarium/libplanet/pull/185
 [#187]: https://github.com/planetarium/libplanet/issues/187
 [#190]: https://github.com/planetarium/libplanet/pull/190
@@ -151,6 +155,7 @@ To be released.
 [#236]: https://github.com/planetarium/libplanet/pull/236
 [#240]: https://github.com/planetarium/libplanet/pull/240
 [#241]: https://github.com/planetarium/libplanet/pull/241
+[#242]: https://github.com/planetarium/libplanet/pull/242
 
 
 Version 0.2.2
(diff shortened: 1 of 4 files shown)


Libplanet/Blocks/Block.cs
```diff
@@ -390,6 +390,17 @@ internal void Validate(DateTimeOffset currentTime)
                 );
             }
 
+            var hasTxsOfSameSigner = Transactions
+                .GroupBy(tx => tx.Signer)
+                .Any(s => s.Count() > 1);
+
+            if (hasTxsOfSameSigner)
+            {
+                throw new MultipleTxsOfSameSignerException(
```
longfin: It would be better if there is signer info who violate policy.

CHANGES.md
```diff
@@ -91,6 +91,8 @@ To be released.
     [[#228], [#234]]
  -  `Swarm.StartAsync()` now does not call `Swarm.StopAsync()` anymore,
     therefore `Swarm.StopAsync()` should be explicitly called.  [[#236]]
+ -  `Block<T>` became to store only one transaction per signer.
```
dahlia: Since `Block<T>` does nothing with persist storage, we'd better use an other verb instead of “store” here, e.g.:

> A signer became to allowed to have only one transaction at most in a `Block<T>`.

Libplanet/Blocks/MultipleTxsOfSameSignerException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+
+namespace Libplanet.Blocks
+{
+    /// <summary>
+    /// The exception that is thrown when the namespace of
+    /// <see cref="Block{T}"/> has multiple transactions of the same signer.
+    /// </summary>
+    [Serializable]
+    public sealed class MultipleTxsOfSameSignerException : Exception
```
dahlia: This added class also should be in the changelog.

Libplanet/Blocks/MultipleTxsOfSameSignerException.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+
+namespace Libplanet.Blocks
+{
+    /// <summary>
+    /// The exception that is thrown when the namespace of
```
dahlia: What the term “namespace” does mean here?

Libplanet/Blocks/Block.cs
```diff
@@ -390,6 +390,17 @@ internal void Validate(DateTimeOffset currentTime)
...
+                    "Only one transaction per signer " +
+                    "can be stored in the block.");
```
dahlia: How about this: “A signer is allowed to have only one transaction at most in a block.”

Libplanet/Blocks/MultipleTxsOfSameSignerException.cs
dahlia: Would there be a better name for this? A type name having a preposition like “of” seems quite awkward to me.

Libplanet/Blocks/MultipleTxsOfSameSignerException.cs
earlbread: My mistake. Fixed.

Libplanet/Blocks/MultipleTxsOfSameSignerException.cs
earlbread: I renamed it to `SingersHaveMoreThanOneTxsException`.

Libplanet/Blocks/Block.cs
```diff
@@ -261,6 +261,10 @@ public byte[] ToBencodex(bool hash, bool transactionData)
         /// any <see cref="IAction"/> of <see cref="Transactions"/> tries
         /// to update the states of <see cref="Address"/>es not included
         /// in <see cref="Transaction{T}.UpdatedAddresses"/>.</exception>
+        /// <exception cref="SimultaneousTxsException">Thrown when a
+        /// <see cref="Transaction{T}.Signer"/> has more than one
+        /// <see cref="Transaction{T}"/>s in <see cref="Transactions"/>.
```
dahlia: The last word seems misleading (it's not incorrect though).   How about rewording a simple term “block” here?

```suggestion
        /// <see cref="Transaction{T}"/>s in a block.
```

earlbread: Fixed.

PR #245, opened 2019-05-19
Run tests on Mono and .NET Framework as well

dahlia:
To be more portable and cover more runtime/platforms, this patch extends CI configuration so that Libplanet is tested with Mono on Linux/macOS/Windows and .NET Framework on Windows.

However, I encountered some problems for certain combinations (especially on Windows):

- With .NET Core on Windows, tests depending on `TURN_SERVER_URL` tend to never end.  `IceServerTest` in particular has always reached timeout.  So I temporarily made these tests skipped.
- With Mono on Windows, `SwarmTest` throws an error like `System.Net.Sockets.SocketException: The operation completed successfully` and never ends in the same manner.  I haven't found what a certain test case causes the problem, so just made `SwarmTest` skipped.  This also should be fixed in the future.

This patch also fixes a bug that docs haven't been properly generated for pull requests from forked repositories.


.github/main.workflow
```diff
@@ -8,11 +8,6 @@ workflow "pull request" {
   resolves = ["docs:status"]
 }
 
-workflow "everyday" {
-  on = "schedule(59 14 * * *)"
-  resolves = ["dist:nuget"]
-}
-
```
earlbread: Don't need to build nightly builds anymore?

.azure-pipelines/windows-net461.yml
```diff
@@ -0,0 +1,47 @@
...
+    targetType: inline
+    script: |
+      ${{ parameters.testPrefix }} \
+        xunit.runner.console.*/tools/net472/xunit.console.exe \
+          *.Tests/bin/$(configuration)/net*/*.Tests.dll \
+          ${{ parameters.testArguments }}
+    # FIXME: For unknown reason, on Windows tests depending on TURN_SERVER_URL
+    #        seems to never end or to take too long time.  We should diagnose
+    #        this and make Windows builds to run these tests too.
+    # FIXME: For unknown reason, on Windows + Mono SwarmTests seem never end.
```
longfin: It may be [`AsyncIO.ForceDotnet.Force()`](https://github.com/somdoron/AsyncIO/blob/986d7424dff95aa0bc87d10cb56b85451c2d8981/Source/AsyncIO/ForceDotNet.cs#L11) related issue. I'll fix in next PR.

.github/main.workflow
dahlia: No, it's a mistake.  I'm going to fix it.

dahlia: Fixed and amended: abd4ac6068546dfc2d58ea5ddf50e0ee61925e7f.

PR #247, opened 2019-05-20
Force dotnet if running on mono

longfin:
- This PR mainly aids hanging unit tests running on Mono in Windows by calling `AsyncIO.ForceDotNet.Force()` to enforce using sockets in .NET, not Windows native.
- It's also prevents some other possible hanging unit tests by eliminating `Swarm.Add()` usages.
- It's also fixes `ArgumentOutOfRangeException` that occurred when requesting a hash when `BlockChain <T>` was empty.

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -275,51 +275,6 @@ namespace Libplanet.Tests.Net
         }
 
         [Fact(Timeout = Timeout)]
-        public async Task WorksAsCollection()
-        {
-            Swarm a = _swarms[0];
-            Swarm b = _swarms[1];
-            Swarm c = _swarms[2];
-
-            // Obtaining swarm's endpoint...
-            await Task.WhenAll(
-                StartAsync(a, _blockchains[0]),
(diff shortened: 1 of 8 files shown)


Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -269,51 +274,6 @@ public async Task CanExchangePeer()
...
-                a.CopyTo(peers, 2);
-            });
-
-            a.CopyTo(peers, 1);
-
-            Assert.Equal(
-                new HashSet<Peer> { null, b.AsPeer, c.AsPeer },
-                peers.ToHashSet());
-        }
-
```
earlbread: `Swarm` is not going to be used as an `ICollection` anymore?

longfin: I will try to do so. Would it be better to handle this issue?

earlbread: No, I think it would be better to handle this in another PR.

longfin:
It still appears intermittently failing, but seems to be more stabilized than before. PTAL @earlbread @dahlia

Issue #262, opened 2019-05-28
Nightly build seems to fail


longfin:
```
### STARTED dist:pack 15:00:06Z

Already have image (with digest): gcr.io/github-actions-images/action-runner:latest
Unable to find image 'mcr.microsoft.com/dotnet/core/sdk:2.2' locally
2.2: Pulling from dotnet/core/sdk
Digest: sha256:3bbfbc6a7ea235322f3af249141e88b20ba06f06d07b5bda92a903bbc2dd7023
Status: Downloaded newer image for mcr.microsoft.com/dotnet/core/sdk:2.2
GITHUB_EVENT_PATH=/github/workflow/event.json
GITHUB_WORKFLOW=everyday
GITHUB_ACTION=dist:pack
GITHUB_REPOSITORY=planetarium/libplanet
GITHUB_WORKSPACE=/github/workspace
GITHUB_SHA=40ba2c2a6b49b07df81635c84ae4e06cc8cc9345
GITHUB_ACTOR=dahlia
GITHUB_EVENT_NAME=schedule

Welcome to .NET Core!
---------------------
Learn more about .NET Core: https://aka.ms/dotnet-docs
Use 'dotnet --help' to see available commands or visit: https://aka.ms/dotnet-cli-docs

Telemetry
---------
The .NET Core tools collect usage data in order to help us improve your experience. The data is anonymous and doesn't include command-line arguments. The data is collected by Microsoft and shared with the community. You can opt-out of telemetry by setting the DOTNET_CLI_TELEMETRY_OPTOUT environment variable to '1' or 'true' using your favorite shell.

Read more about .NET Core CLI Tools telemetry: https://aka.ms/dotnet-cli-telemetry

ASP.NET Core
------------
Successfully installed the ASP.NET Core HTTPS Development Certificate.
To trust the certificate run 'dotnet dev-certs https --trust' (Windows and macOS only). For establishing trust on other platforms refer to the platform specific documentation.
For more information on configuring HTTPS see https://go.microsoft.com/fwlink/?linkid=848054.
Microsoft (R) Build Engine version 16.1.76+g14b0a930a7 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

/usr/share/dotnet/sdk/2.2.300/NuGet.targets(121,5): error : '1.0.0-nightly.20190527+' is not a valid version string. [/github/workspace/Libplanet.sln]
[...]

PR #263, opened 2019-05-29
Fix broken nightly builds

dahlia:
This patch fixes the bug #262.  It caused because GitHub Actions' `schedule` events refer to the commit hash directly (not through the branch name indirectly) to check out the source code, while the *.github/bin/dist-version.ps1* script had assumed *.git/HEAD* has always an indirect ref (e.g., `ref: refs/heads/master`).  This fixes the script able to see both direct and indirect refs.

### DIFF
--- a/.github/bin/dist-version.ps1
+++ b/.github/bin/dist-version.ps1
@@ -21,22 +21,24 @@ Write-Output $VersionPrefix > obj/version_prefix.txt
 
 $Event = Get-Content $env:GITHUB_EVENT_PATH | ConvertFrom-Json
 if ($Event.head_commit.id -ne $null) {
-  $HeadCommit = $Event.head_commit.id
+  $HeadCommitHash = $Event.head_commit.id
 }
-if ($HeadCommit -eq $null) {
-  $HeadCommit = Get-Content .git/HEAD
-  $refPair = $HeadCommit.Split()
+if ($HeadCommitHash -eq $null) {
+  $headCommit = Get-Content .git/HEAD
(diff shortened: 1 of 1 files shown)


.github/bin/dist-version.ps1
```diff
@@ -19,8 +19,18 @@ New-Item -ItemType directory -Path obj -ErrorAction SilentlyContinue
...
 Write-Output $VersionPrefix > obj/version_prefix.txt
 
-$HeadCommit = `
-  (Get-Content ".git/$((Get-Content .git/HEAD).Split()[1])").Substring(0, 7)
+$Event = Get-Content $env:GITHUB_EVENT_PATH | ConvertFrom-Json
+if ($Event.head_commit.id -ne $null) {
+  $HeadCommit = $Event.head_commit.id
+}
+if ($HeadCommit -eq $null) {
+  $HeadCommit = Get-Content .git/HEAD
```
longfin: it's a bit confusing because `$HeadCommit` isn't a commit hash yet.

dahlia: It's actually a *commit*, it could be not a *commit hash* yet though…

longfin: Of course it is. but I think representing *commit* and *commit hash* in same variable (`$HeadCommit`) is a bit confusing.

dahlia: Fixed! 5b4b5125b9776b4c3bad4817f32fd278147f1f67

PR #269, opened 2019-06-03
Add LiteDBStore

longfin:
This PR adds `LiteDBStore` as a new storage class that using [LiteDB](https://www.litedb.org/) as its backend.

Compared to `FileStore`, the following improvements have been made.

- Because LiteDB has its own cache layer, write performance has increased slightly. (20~30% faster on my laptop)
- Easy to manage because the entire DB is managed with a single file.


longfin:
I also added commit that introduces [Microsoft.DotNet.Analyzers.Compatibility](https://www.nuget.org/packages/Microsoft.DotNet.Analyzers.Compatibility) to check API compatibility.

CHANGES.md
```diff
@@ -10,10 +10,14 @@ To be released.
 
 ### Added interfaces
 
+ -  Added `LiteDBStore` that using [LiteDB] as backend. [[#269]]
```
```suggestion
 -  Added `LiteDBStore` backend that uses [LiteDB] under the hood.  [[#269]]
```

Libplanet/Store/LiteDBStore.cs
```diff
@@ -0,0 +1,631 @@
...
+
+        private LiteCollection<TxIdDoc> StagedTxIds =>
+            _db.GetCollection<TxIdDoc>("staged_txids");
+
+        /// <inheritdoc/>
+        public IEnumerable<string> ListNamespaces()
+        {
+            return _db.GetCollectionNames()
+                .Where(name => name.StartsWith("indices_"))
+                .Select(name => name.Split('_')[1]);
```
dahlia: How about this?

```csharp
const string prefix = "indices_";
readonly int prefixLength = prefix.Length;
return _db.GetCollectionNames()
    .Where(name => name.StartsWith(prefix))
    .Select(name => name.Substring(prefixLength));
```

Libplanet.Tests/Store/LiteDBStoreTest.cs
```diff
@@ -0,0 +1,336 @@
...
+using System.Security.Cryptography;
+using Libplanet.Action;
+using Libplanet.Blocks;
+using Libplanet.Tests.Common.Action;
+using Libplanet.Tx;
+using Xunit;
+
+namespace Libplanet.Tests.Store
+{
+    public class LiteDBStoreTest : IDisposable
```
dahlia: I believe we can write some black box tests (a kind of property-based tests?) that all `IStore` implementations should commonly pass, besides this kind of white block tests.  E.g.:

```csharp
store.PutBlock(block);
Assert.Equal(block, store.GetBlock(block.Hash));
```

Such tests also could help us to eliminate duplicated assertions between `FileStoreTest` and `LiteDBStoreTest`.

Libplanet/Store/LiteDBStore.cs
longfin: I'll refactor with some class level constant. (e.g. `const string _indexPrefix`).

Libplanet/Store/LiteDBStore.cs
```diff
@@ -0,0 +1,631 @@
...
+                if (stream.Length % nonceEntrySize != 0)
+                {
+                    throw new FileLoadException(
+                        $"Nonce file size {stream.Length} should be " +
+                        $"a multiple of nonce entry size {nonceEntrySize}");
+                }
+
+                var buffer = new byte[nonceEntrySize];
+                var offset = (int)(stream.Length - buffer.Length);
+                stream.Read(buffer, offset, buffer.Length);
```
earlbread: The offset should be 0 because the buffer length and the size to read are the same.
I think we should move the file pointer before we read the file.

Libplanet.Tests/Store/FileStoreFixture.cs
```diff
@@ -1,144 +1,30 @@
...
-using Libplanet.Tests.Common.Action;
-using Libplanet.Tx;
 
 namespace Libplanet.Tests.Store
 {
-    public class FileStoreFixture : IDisposable
+    public class FileStoreFixture : StoreFixture, IDisposable
     {
         public FileStoreFixture()
+        : base()
```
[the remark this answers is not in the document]
longfin: I've removed.

Libplanet.Tests/Store/LiteDBStoreFixture.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.IO;
+using Libplanet.Store;
+using LiteDB;
```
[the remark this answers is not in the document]
longfin: I've removed.

Libplanet/Store/LiteDBStore.cs
longfin: I think moving and reading the file pointer and specifying the `offset` to read(=the last entry's location) are the same behavior. Is it different?

earlbread: According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=netframework-4.8#System_IO_Stream_Read_System_Byte___System_Int32_System_Int32_), the offset parameter is the starting point of the buffer to be copied, not the file stream position.

Libplanet/Store/LiteDBStore.cs
```diff
@@ -0,0 +1,642 @@
...
+                        srcStream.Read(nonceBytes, 0, nonceBytes.Length);
+
+                        destStream.Write(hashBytes, 0, hashBytes.Length);
+                        destStream.Write(indexBytes, 0, indexBytes.Length);
+                        destStream.Write(nonceBytes, 0, nonceBytes.Length);
+
+                        long currentIndex =
+                            BitConverter.ToInt64(indexBytes, 0);
+
+                        if (branchPointIndex <= currentIndex)
```
earlbread: When we fork, the first state is copied regardless of the current index.

Libplanet/Store/LiteDBStore.cs
```diff
@@ -0,0 +1,642 @@
...
+                        srcStream.Read(hashBytes, 0, hashBytes.Length);
+                        srcStream.Read(indexBytes, 0, indexBytes.Length);
+
+                        destStream.Write(hashBytes, 0, hashBytes.Length);
+                        destStream.Write(indexBytes, 0, indexBytes.Length);
+
+                        long currentIndex =
+                            BitConverter.ToInt64(indexBytes, 0);
+
+                        if (branchPointIndex <= currentIndex)
```
earlbread: When we fork, the first state is copied regardless of the current index.

Issue #273, opened 2019-06-04
Sending multiple blocks at once


earlbread:
Currently, when node A requested blocks to node B, B sends one block per message. So if A requested 10 blocks, B sends 10 messages.

However, it takes too much time to verify messages if there are many blocks requested because the recipient of the message verifies the signature of the message when it receives the message.

To shorten message verification time, we need to change the `Block` message so that the block sender can send multiple blocks at once.

PR #276, opened 2019-06-05
Send multiple blocks in one message

earlbread:
This makes `Swarm` send and receive multiple blocks with one message instead of one block per message. #273


Libplanet/Net/Swarm.cs
```diff
@@ -682,12 +682,15 @@ public void BroadcastTxs<T>(IEnumerable<Transaction<T>> txs)
...
                             cancellationToken: token);
                         Message parsedMessage = Message.Parse(response, true);
-                        if (parsedMessage is Block blockMessage)
+                        if (parsedMessage is Messages.Blocks blockMessage)
                         {
-                            Block<T> block = Block<T>.FromBencodex(
-                                blockMessage.Payload);
-                            await yield.ReturnAsync(block);
-                            hashCount--;
+                            var payloads = blockMessage.Payloads.ToArray();
```
longfin: Could you specify type instead of `var`?

Libplanet/Net/Swarm.cs
```diff
@@ -1337,18 +1340,37 @@ private void TransferTxs<T>(BlockChain<T> blockChain, GetTxs getTxs)
...
                         Identity = getData.Identity,
                     };
                     _replyQueue.Enqueue(response);
+                    blocks.Clear();
                 }
             }
 
+            if (blocks.Any())
+            {
+                Message response = new Messages.Blocks(blocks.ToArray())
```
longfin: `response`'s type seems to be obvious.
```suggestion
                var response = new Messages.Blocks(blocks.ToArray())
```

Libplanet/Net/Swarm.cs
```diff
@@ -1337,18 +1340,37 @@ private void TransferTxs<T>(BlockChain<T> blockChain, GetTxs getTxs)
...
                 if (blockChain.Blocks.TryGetValue(hash, out Block<T> block))
                 {
-                    Message response = new Block(block.ToBencodex(true, true))
+                    var payload = block.ToBencodex(true, true);
+                    blocks.Add(payload);
+                }
+
+                if (blocks.Count == getData.ChunkSize)
+                {
+                    Message response = new Messages.Blocks(blocks.ToArray())
```
longfin: Should we re-materialize `blocks` with `.ToArray()`? `blocks` seems to be already non-lazy enumerable.

Libplanet/Net/Swarm.cs
```diff
@@ -1337,18 +1340,37 @@ private void TransferTxs<T>(BlockChain<T> blockChain, GetTxs getTxs)
...
             _logger.Debug("Trying to transfer blocks...");
+
+            var blocks = new List<byte[]>();
+
             foreach (HashDigest<SHA256> hash in getData.BlockHashes)
             {
                 if (blockChain.Blocks.TryGetValue(hash, out Block<T> block))
                 {
-                    Message response = new Block(block.ToBencodex(true, true))
+                    var payload = block.ToBencodex(true, true);
```
```suggestion
                    byte[] payload = block.ToBencodex(true, true);
```

CHANGES.md
```diff
@@ -12,12 +12,17 @@ To be released.
 
 ### Behavioral changes
 
--  `BlockChain<T>.GetNonce()` became to count staged transactions too during nonce
-   computation. [[#270]]
+ -  `BlockChain<T>.GetNonce()` became to count staged transactions too during
+    nonce computation.  [[#270]]
+ -  `Swarm` became to send and receive multiple blocks with one message instead 
+    of one block per message.  [[#273], [#276]]
```

> A message `Swarm` makes became to have multiple blocks within it, which means round trips on the network are now much reduced.

Libplanet/Net/Swarm.cs
longfin: Also, should we re-materialize blocks with `.ToArray()`?

Libplanet/Net/Messages/Blocks.cs
```diff
@@ -0,0 +1,40 @@
+using System.Collections.Generic;
+using System.Linq;
+using NetMQ;
+
+namespace Libplanet.Net.Messages
+{
+    internal class Blocks : Message
+    {
+        public Blocks(IEnumerable<byte[]> payloads)
```
dahlia: As a `Blocks` message cannot have more than 2<sup>32</sup> blocks, it would be better to check the number of `payloads`.

Libplanet/Net/Messages/Message.cs
```diff
@@ -51,9 +51,9 @@ internal enum MessageType : byte
             GetTxs = 0x08,
 
             /// <summary>
-            /// Message containing serialized block.
+            /// Message containing serialized blocks.
             /// </summary>
-            Block = 0x09,
+            Blocks = 0x09,
```
dahlia: At least in theory, it's a distinct message type which is incompatible with the previous `Block` message type, and we'd released multiple Libplanet versions used `Block` message type, IMHO we'd better bump the magic number here to `0x0a`.  @longfin Opinion?

longfin: Basically, We should comply backward compatibility as you say. however, I think that there is a burden to leave the code that will not be used.

Considering libplanet's current user base, I guess this burden will be greater than the importance of backward compatibility.

Libplanet/Net/Swarm.cs
earlbread: You're right. However, if I call Clean after putting the `blocks` in a queue, the messages in the queue are affected. So I called ToArray to copy it.

Libplanet/Net/Messages/Message.cs
dahlia: It don't think we should leave any code for removed message types, but what I suggest is only to use `0x0a` here instead of `0x09`.

CHANGES.md
dahlia: Also we should mention that an existing message type (`Block`) was gone and a new message type (`Blocks`) was introduced instead.

Libplanet/Net/Swarm.cs
longfin: Ok, I see. on the other hand, it seems that copying from the `Blocks` constructor is safer than the caller's side if copy is needed.

Libplanet/Net/Messages/Message.cs
longfin: Oh, I see. it seems to be fine.

CHANGES.md
earlbread: Should this be a change of behavior?

> Also we should mention that an existing message type (`Block`) was gone and a new message type (`Blocks`) was introduced instead.

Should this be `Behavioral changes`? This is Backward-incompatible but it's not about the interface.

dahlia: It might be good to have a new section for protocol-level compatibility, but at the moment <q>behavior changes</q> seems the most appropriate section.

earlbread:
I applied the review suggestions. Please take a look.

PR #277, opened 2019-06-08
Improve branch point finding

longfin:
This PR improves performance of `BlockChain<T>.Blocks.Contains()`, `BlockChain<T>.FindBranchPoint()` and `BlockChain<T>.FindNextHashes()` to reduce response time for block hashes query.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,8 @@ To be released.
     round trips on the network are now much reduced.  [[#273], [#276]]
  -  `Message.Block` has been replaced by `Message.Blocks` and the magic number
     has been changed to `0x0a`.  [[#276]]
+ -  Improved performance of `Swarm`'s response time to `GetBlockHashes`
+    request messages.  [[#277]]
 
 ### Bug fixes
 
@@ -30,6 +32,7 @@ To be released.
 [#273]: https://github.com/planetarium/libplanet/issues/273
 [#275]: https://github.com/planetarium/libplanet/pull/275
 [#276]: https://github.com/planetarium/libplanet/pull/276
+[#277]: https://github.com/planetarium/libplanet/pull/277
 
 
 Version 0.3.0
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -18,12 +18,14 @@ To be released.
     round trips on the network are now much reduced.  [[#273], [#276]]
  -  `Message.Block` has been replaced by `Message.Blocks` and the magic number
     has been changed to `0x0a`.  [[#276]]
+ -  Improved reponse performance for block hash queries.  [[#277]]
```
```suggestion
 -  Improved response performance for block hash queries.  [[#277]]
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -438,12 +438,16 @@ internal HashDigest<SHA256> FindBranchPoint(BlockLocator locator)
             {
                 _rwlock.EnterReadLock();
 
+                ImmutableHashSet<HashDigest<SHA256>> indices = Store
+                    .IterateIndex(Id.ToString())
+                    .ToImmutableHashSet();
```
dahlia: It would be more defensive if we make it to chunk these indices to load instead of loading them entirely at a time.  If it's difficult to improve this in that way we'd better leave a `FIXME` comment at least.

CHANGES.md
```diff
@@ -18,6 +18,7 @@ To be released.
     round trips on the network are now much reduced.  [[#273], [#276]]
  -  `Message.Block` has been replaced by `Message.Blocks` and the magic number
     has been changed to `0x0a`.  [[#276]]
+ -  Improved response performance for block hash queries.  [[#277]]
```
```suggestion
 -  Improved performance of `Swarm`'s response time to `GetBlockHashes`
    request messages.  [[#277]]
```

Libplanet/Blockchain/BlockChain.cs
longfin: I've added `FIXME`. (https://github.com/planetarium/libplanet/pull/277/commits/dd197fbca7e766654ebca1210a26205b8b85a325)

Issue #222, opened 2019-04-28
Let's widen the line length limit!


dahlia:
Although we've adopted 80 columns for the line limit, but [it was like a procrastination rather than a proper decision][1]. It's because we didn't arrived at a consensus on the maximum number of columns to allow, but only on that we need to limit columns per line anyway.

However, we become more and more aware that 80 columns are too short to ordinary C# code.  C# is neither C nor Python, and have way longer names, and it already starts with second-level depth due to namespaces.

So I suggest to widen the line limit.

What number would be suitable for this project?  I suggest 132 characters as [Linus Torvalds did in a decade ago][2]:

> I'd be ok with changing the warning to 132 characters, which is another perfectly fine historical limit.

Except that for me the historical reason is not primary; to me the 132 characters limit looks good because it is not that wider than other candidates, e.g., 160 or 180.  If we adopt a wide line limit once, it will hardly become narrow again.  So we'd better have a small step.

[1]: https://github.com/planetarium/libplanet/pull/58#issuecomment-461724764
[2]: https://lkml.org/lkml/2012/2/3/394

earlbread:
How about 120 characters? It is the maximum line length without horizontal scrolling of the Rider when splitting the screen into two at 1920x1080 resolution.

dahlia:
> How about 120 characters? It is the maximum line length without horizontal scrolling of the Rider when splitting the screen into two at 1920x1080 resolution.

Doesn't it depend on font size? Anyway, 132 characters seem indeed too many for narrow displays.  In fact, 120 characters also are not fit in two columns on my main monitor…

dahlia:
100 columns seem fit in two columns on my main monitor.

longfin:
💯 columns seems to good.

PR #280, opened 2019-06-11
Allow null to Peer.AppProtocolVersion for easy configuration

longfin:
This patch allows null value to `Peer.AppProtocolVersion`. so this adds version-less `Peer` constructor and `Peer.WithAppProtocolVersion()`.

Also, this PR widens the line length limit to 100 (as #222).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@ To be released.
 
 ### Backward-incompatible interface changes
 
+ -  `Peer.AppProtocolVersion` became nullable to represent `Peer` whose version
+    is unknown.
+
 ### Added interfaces
 
  -  Added `LiteDBStore` backend that uses [LiteDB] under the hood.  [[#269]]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,8 +13,8 @@ To be released.
  -  Added `LiteDBStore` backend that uses [LiteDB] under the hood.  [[#269]]
  -  All `*Async()` methods belonging to `TurnClient` class became to have
     `cancellationToken` option.  [[#287]]
- -  Added a version-less `Peer` constructor to create a `Peer` whose version
-    is unknown.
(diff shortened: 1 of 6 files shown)


Libplanet/Net/Peer.cs
```diff
@@ -15,6 +15,20 @@ namespace Libplanet.Net
     [Equals]
     public class Peer : ISerializable
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Peer"/> class.
```
dahlia:
```suggestion
        /// Initializes a new instance of the <see cref="Peer"/> class
        /// with omitting <see cref="AppProtocolVersion"/>.
```

Libplanet/Net/Peer.cs
```diff
@@ -79,7 +94,7 @@ protected Peer(SerializationInfo info, StreamingContext context)
         /// <seealso cref="Swarm.DifferentVersionPeerEncountered"/>
         [IgnoreDuringEquals]
         [Pure]
-        public int AppProtocolVersion { get; }
+        public int? AppProtocolVersion { get; }
```
dahlia: As this breaks the existing interface (IL-wise at least) we should add this to *CHANGES.md*'s *Backward-incompatible interface changes* section.

CHANGES.md
```diff
@@ -13,6 +13,8 @@ To be released.
  -  Added `LiteDBStore` backend that uses [LiteDB] under the hood.  [[#269]]
  -  All `*Async()` methods belonging to `TurnClient` class became to have
     `cancellationToken` option.  [[#287]]
+ -  Added a version-less `Peer` constructor to create a `Peer` whose version
```
```suggestion
 -  Added a `Peer` constructor omitting `appProtocolVersion` parameter
```

CHANGES.md
```diff
@@ -13,6 +13,8 @@ To be released.
...
+    is unknown.
```
```suggestion
    to create a `Peer` whose version is unknown.
```

PR #281, opened 2019-06-12
Fix bugs when forking TxNonce

earlbread:
- Fixed a bug where the first `TxNonce` of an address is not stripped when forking using `FileStore.ForkTxNonce()` method.
- Fixed a bug where `LiteDBStore.GetTxNonce()` method throws a `System.IO.IOException` when reading a zero-length file.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,8 @@ To be released.
     [[#275]]
  -  Fixed a bug where the oldest `TxNonce` of an address is not invalidated
     when forking using `FileStore.ForkTxNonce()` method.  [[#281]]
+ -  Fixed a bug where `LiteDBStore.GetTxNonce()` method throws a
+    `System.IO.IOException` after forking.  [[#281]]
 
 [LiteDB]: https://www.litedb.org/
 [#269]: https://github.com/planetarium/libplanet/pull/269
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,8 +16,8 @@ To be released.
 
  -  `BlockChain<T>.GetNonce()` became to count staged transactions too during
     nonce computation.  [[#270]]
- -  A message `Swarm` became to have multiple blocks within it, which means 
-    round trips on the network are now much reduced.  [[#273], [#276]]
+ -  A message `Swarm` makes became to have multiple blocks within it, which
+    means round trips on the network are now much reduced.  [[#273], [#276]]
  -  `Message.Block` has been replaced by `Message.Blocks` and the magic number
     has been changed to `0x0a`.  [[#276]]
  -  Improved performance of `Swarm`'s response time to `GetBlockHashes`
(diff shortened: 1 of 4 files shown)


Libplanet/Store/LiteDBStore.cs
```diff
@@ -459,15 +472,20 @@ public long GetTxNonce(string @namespace, Address address)
 
             using (var stream = new MemoryStream())
             {
+                file.CopyTo(stream);
+
+                if (stream.Length == 0)
```
dahlia: Isn't this condition can be detected without copying `file` contents to `stream`?  IMO we could use `LiteFileInfo.Length` property.

dahlia: By the way, how could this condition happen?  Is there any code path that creates a nonce file and immediately closes it without writing anything?  Even if such a code path exists, isn't *that* code a bug, not this code?

CHANGES.md
```diff
@@ -26,9 +26,13 @@ To be released.
 ### Bug fixes
 
  -  Fixed a bug that `Swarm` reported `TaskCanceledException` as an unknown
-   exception while stopping. [[#275]]
+    exception while stopping. [[#275]]
  -  Fixed a bug that `Swarm` didn't stop properly during `Swarm.Preload()`.
-   [[#275]]
+    [[#275]]
+ -  Fixed a bug where the first `TxNonce` of an address is not stripped when
```
dahlia: What does <q>stripped</q> mean here?

CHANGES.md
```diff
@@ -26,9 +26,13 @@ To be released.
...
+    forking using `FileStore.ForkTxNonce()` method.  [[#281]]
+ -  Fixed a bug where `LiteDBStore.GetTxNonce()` method throws a
+    `System.IO.IOException` when reading a zero-length file.  [[#281]]
```
dahlia: The expression <q>a zero-length file</q> here seems too much about implementation details.  Actually even I didn't get it (I believe it's relevant to a question I left below, in *LiteDBStore.cs* file).

CHANGES.md
```diff
@@ -16,7 +16,7 @@ To be released.
 
  -  `BlockChain<T>.GetNonce()` became to count staged transactions too during
     nonce computation.  [[#270]]
- -  A message `Swarm` became to have multiple blocks within it, which means 
+ -  A message `Swarm` became to have multiple blocks within it, which means
```
```suggestion
 -  A message `Swarm` makes became to have multiple blocks within it, which means
```

Libplanet/Store/LiteDBStore.cs
earlbread: This condition can happen [here](https://github.com/planetarium/libplanet/commit/03caf772deef950236387d4502bd2eed9ab23e17#diff-3fe3c469228f579007de2d2703d077d9R554). After open `destFile`, you don't write anything if `currentIndex` is less than `branchPointIndex`. Do you think it is better to fix it?

CHANGES.md
earlbread: <q>strip</q> means [this](https://github.com/planetarium/libplanet/blob/03caf772deef950236387d4502bd2eed9ab23e17/Libplanet/Store/FileStore.cs#L759). To fork `TxNonce`s,  we copy and strip them.

dahlia: Term suggestions:

- the first `TxNonce` → the oldest `TxNonce`
- stripped → invalidated

Libplanet/Store/LiteDBStore.cs
dahlia: To make the code more defensive, we'd better change both side.

earlbread: Fixed to delete the file if nothing is written.

PR #284, opened 2019-06-14
Fix transaction sort order in ValidateNonce

earlbread:
This fixes transaction sort order in `ValidateNonce`.
This bug was introduced in 0.4.0-dev, so I omitted the changelog.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -410,7 +410,9 @@ namespace Libplanet.Blockchain
         internal void ValidateNonce(Block<T> block)
         {
             var nonces = new Dictionary<Address, long>();
-            foreach (Transaction<T> tx in block.Transactions)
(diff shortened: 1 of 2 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -410,7 +410,9 @@ bool render
         internal void ValidateNonce(Block<T> block)
         {
             var nonces = new Dictionary<Address, long>();
-            foreach (Transaction<T> tx in block.Transactions)
+            IEnumerable<Transaction<T>> transactions =
```
longfin: it seems to be already sorted by nonce in [`.MineBlock()`](https://github.com/planetarium/libplanet/commit/fbca5c679061a74af55db1994037b10a8da581d7#diff-f9c16aaa8d3a312223ca86876f791202)

earlbread: You're right. This was not the cause of the bug.

Issue #272, opened 2019-06-04
Trusted state synchronization on a restricted basis for faster IBD


dahlia:
In principle, every single node should calculate their own states from the given transactions and actions from scratch, even though every node will eventually have the same deterministic states. Unlike transactions signatures, the only way to verify states in itself is to simply calculate transactions from scratch and compare them with the calculated result.  There's no way to digest the calculation in itself; calculating and verifying states cost the same.

The problem is that evaluating actions is hard to predict how long they will take.  As a rule of thumb, according to @longfin, validating actions (which requires evaluating them) take more than 10% of the time during receiving new blocks.

If a node wants to leverage states already calculated by peers it must trust them first.  However, blindly trusting peers makes the network fragile, because a malicious (or malfunctioning) node could send invalid states to peers.  Therefore if we're willing to trust some peers, it has to be allowed on only a restricted basis.

So I suggest an option to trust peers to use states calculated by them in a restricted-enough manner.  API-wise this would basically look like: the `Swarm` constructor has a new optional parameter named `trustedStateValidators`.  The type of this is a set of peers, i.e., `IImmutableSet<PublicKey>`.  Such a whitelist approach implements the <q>decentralized trust model</q> in the concept of [web of trust].  The basic idea is that a game player should decide what peers to trust, through a configuration file or, ideally, GUI.  Although each node could make some wrong decision on whom to trust, these mistakes do not impact on the entire network on a bigger scale.

There should be one more restriction on it IMHO: precalculated states should be leveraged only for [IBD].  If a node once catches up the topmost block of the major blockchain, a node should calculate states from the received transactions by itself.

Opinions?

[web of trust]: https://en.wikipedia.org/wiki/Web_of_trust
[IBD]: https://bitcoin.org/en/glossary/initial-block-download

earlbread:
I agree with your suggestion.
It sounds good to have `trustedStateValidators` and use them only for IBD.

longfin:
![image](https://user-images.githubusercontent.com/128436/58863446-d4f97480-86ed-11e9-8643-02c9a44aee53.png)

PR #285, opened 2019-06-14
Allow BlockChain<T> to have incomplete states

dahlia:
This patch makes `BlockChain<T>` to have incomplete states so that we can implement #272 on this.  Where a `BlockChain<T>` does not have complete states and a query to states that it lacks is required, there are two options:

- Throwing `IncompleteBlockStatesException` (which is default).
- Evaluating actions preceding the requested states and filling the states on the fly (`completeStates: true`).

See also *CHANGES.md* as well.

### DIFF
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -743,7 +743,7 @@ namespace Libplanet.Tests.Blockchain
 
             // As the store lacks the states for blocks other than the tip,
             // the following GetStates() calls should throw an exception.
-            foreach (Address addr in addresses.SkipLast(1))
+            foreach (Address addr in addresses.Take(addresses.Length - 1))
             {
                 Assert.Throws<IncompleteBlockStatesException>(() =>
                     chain.GetStates(new[] { addr })
@@ -759,7 +759,7 @@ namespace Libplanet.Tests.Blockchain
             IStore store = chain.Store;
 
             chain.GetStates(new[] { addresses.Last() }, completeStates: true);
-            foreach (Block<DumbAction> block in chain.SkipLast(1))
+            foreach (Block<DumbAction> block in chain.Take(chain.Count() - 1))
             {
                 Assert.Null(store.GetBlockStates(block.Hash));
             }
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -8,6 +8,7 @@ using Libplanet.Blockchain;
 using Libplanet.Blockchain.Policies;
 using Libplanet.Blocks;
 using Libplanet.Crypto;
+using Libplanet.Store;
 using Libplanet.Tests.Common.Action;
 using Libplanet.Tests.Store;
 using Libplanet.Tx;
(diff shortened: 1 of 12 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -208,11 +228,49 @@ IEnumerator IEnumerable.GetEnumerator()
...
+
+                            ActionEvaluation<T>[] evaluations =
+                                b.Evaluate(
+                                    DateTimeOffset.UtcNow,
+                                    a => GetStates(
+                                        new[] { a },
+                                        b.PreviousHash
+                                    ).GetValueOrDefault(a)
+                                ).ToArray();
+                            SetStates(b, evaluations);
```
earlbread: I think `StateReference` and `TxNonce` should be cleared before call `SetStates`. Because they were made on the assumption that it is stored in order.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -208,11 +228,49 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                    if (completeStates)
+                    {
+                        // Calculates and fills the incomplete states
+                        // on the fly.
+                        foreach (Block<T> b in this)
+                        {
+                            if (!(Store.GetBlockStates(b.Hash) is null))
+                            {
+                                continue;
+                            }
```
earlbread: If we skip `SetStates` when the state is already calculated, the order of `StateReference` and `TxNonce` may be mixed.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -165,10 +165,30 @@ IEnumerator IEnumerable.GetEnumerator()
...
+        /// <see cref="IncompleteBlockStatesException"/> instead
+        /// for the same situation.
+        /// Just-in-time calculation of states could take long time so that
+        /// overall latency of an application may rise.</param>
         /// <returns>The <see cref="AddressStateMap"/> of given
         /// <paramref name="addresses"/>.</returns>
+        /// <exception cref="IncompleteBlockStatesException">Thrown when
+        /// the <see cref="BlockChain{T}"/> instance does not contain
+        /// states dirty of the block which lastly updated states of a requested
+        /// address, because actions in the block has never been executed.
```
earlbread:
```suggestion
        /// address, because actions in the block have never been executed.
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -165,10 +165,30 @@ IEnumerator IEnumerable.GetEnumerator()
...
         /// start finding the state. It will be The tip of the
         /// <see cref="BlockChain{T}"/> if it is <c>null</c>.</param>
+        /// <param name="completeStates">When the <see cref="BlockChain{T}"/>
+        /// instance does not contain states dirty of the block which lastly
+        /// updated states of a requested address, this option makes
+        /// the incomplete states calculated and filled on the fly.
+        /// If this option is turned off (which is default) this method throws
+        /// <see cref="IncompleteBlockStatesException"/> instead
+        /// for the same situation.
+        /// Just-in-time calculation of states could take long time so that
```
earlbread:
```suggestion
        /// Just-in-time calculation of states could take a long time so that
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -165,10 +165,30 @@ IEnumerator IEnumerable.GetEnumerator()
...
+        /// overall latency of an application may rise.</param>
```
earlbread:
```suggestion
        /// the overall latency of an application may rise.</param>
```

Libplanet/Blockchain/BlockChain.cs
dahlia: It's assuming that there are already complete `StateReference`s but only block states are absence.

Libplanet/Blockchain/BlockChain.cs
dahlia: I wrote a test code for the case you mentioned, and it indeed failed.  I fixed `GetStates()` not to build indices (`TxNonce` and `StateReference`): 8708fd1de8127a2753f77175ca02b5f83662dbb9.

Libplanet/Blockchain/BlockChain.cs
dahlia: See also 8708fd1de8127a2753f77175ca02b5f83662dbb9.

dahlia:
@longfin @earlbread Rebased this on the current master and addressed reviewed points.  Please review this again.

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -814,6 +910,105 @@ public void GetNonce()
...
+        /// Builds a fixture that has incomplete states for blocks other
+        /// than the tip, to test <c>GetStates()</c> method's
+        /// <c>completeStates: true</c> option and
+        /// <see cref="IncompleteBlockStatesException"/>.
+        ///
+        /// <para>The fixture this makes has total 5 addresses (i.e., accounts;
+        /// these go to the second item of the returned triple) and 11 blocks
+        /// (these go to the third item of the returned triple). Every block
+        /// contains a transaction within an action that mutates one account
+        /// state except of the genesis block.  All transactions in the fixture
```
earlbread:
```suggestion
        /// state except for the genesis block.  All transactions in the fixture
```

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -814,6 +910,105 @@ public void GetNonce()
...
+        /// are signed by one private key (its address goes to the first item
+        /// of the returned triple).  The most important thing is that
+        /// these blocks all lack its states except of the last block (tip).
```
earlbread:
```suggestion
        /// these blocks all lack its states except for the last block (tip).
```

dahlia:
@earlbread @longfin Rebased again.  Please review this again.

Issue #244, opened 2019-05-17
Evaluation order of transactions in a block


dahlia:
Currently the evaluation order of transactions signed by multiple signers in a block are arbitrary and can be reoredered (fabricated) by a miner.  This indeterministic order makes games unfair.

For example, suppose there are two actions, *Attack* which decreases target's health and *Defend* which invalidates one succeeding *Attack* against on the signer.  *A* signs a transaction *t<sup>A</sup>* to *Attack* *B*, and *B* signs a transaction  *t<sup>B</sup>* to *Defend* itself.  If two transactions try to belong to a block, a block miner can determine whether to decrease *B*'s health or not.  If *t<sup>A</sup>* is followed by  *t<sup>B</sup>* *B* gets damage, whereas if *t<sup>B</sup>* is followed by  *t<sup>A</sup>* *A* loses its chance.

Therefore, the evaluation order of transactions in a block should not be determined by any certain party.

One idea I have is to make the evaluation order determined after a block is mined so that a miner can't deal with it:

 1. First of all, make `Block<T>.Mine()` to take a set of transactions instead of a sequence of transactions.  These transactions are unordered.
 2. When transactions (actions) in a block is being evaluated, sort them in the order of each one's *nonce*, which is derived from both a block hash and a transaction's signature.

This makes the order of evaluation impossible to predict until a block is mined.  The order is still arbitrary, but it makes game play in general more fair.

Any opinions are welcome.

moreal:
I started to work on this issue :grin:

moreal:
It's closed by #355.

moreal:
In #355, I decided to order transactions by its nonce for hash. But it can cause undetermined hash, which also affects the order in which transactions are executed.

PR #294, opened 2019-06-18
Combine creating and staging transaction interface

earlbread:
- Combines creating and staging of transaction interfaces with the `BlockChain<T>.MakeTransaction()` method.

### DIFF
--- a/Libplanet/Blocks/Block.cs
+++ b/Libplanet/Blocks/Block.cs
@@ -38,6 +38,9 @@ namespace Libplanet.Blocks
             Miner = miner;
             PreviousHash = previousHash;
             Timestamp = timestamp;
+
+            // FIXME: We need to fix this to solve
+            // https://github.com/planetarium/libplanet/issues/244.
             Transactions = transactions.OrderBy(tx => tx.Nonce).ToList();
             Hash = Hashcash.Hash(ToBencodex(false, false));
         }
--- a/Libplanet/Blocks/Block.cs
+++ b/Libplanet/Blocks/Block.cs
@@ -38,7 +38,7 @@ namespace Libplanet.Blocks
             Miner = miner;
             PreviousHash = previousHash;
             Timestamp = timestamp;
-            Transactions = transactions;
+            Transactions = transactions.OrderBy(tx => tx.Nonce).ToList();
             Hash = Hashcash.Hash(ToBencodex(false, false));
         }
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@ To be released.
  -  `Peer.AppProtocolVersion` became nullable to represent `Peer` whose version
     is unknown.
  -  Added `IStore.ListAddresses()` method.  [[#272], [#285]]
+ -  Renamed `BlockChain<T>.GetNonce()` method to `BlockChain<T>.GetNextTxNonce()`
+    method.  [[#294]]
 
 ### Added interfaces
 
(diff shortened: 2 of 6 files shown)


Libplanet/Blocks/Block.cs
```diff
@@ -38,7 +38,7 @@ public class Block<T> : ISerializable
             Miner = miner;
             PreviousHash = previousHash;
             Timestamp = timestamp;
-            Transactions = transactions;
+            Transactions = transactions.OrderBy(tx => tx.Nonce);
```
longfin: It would be safer if we also materialize `transactions` at here.

longfin: also, I guess that we can remove ordering in `BlockChain<T>.MineBlock()` if we'll sort here.

earlbread: > also, I guess that we can remove ordering in `BlockChain<T>.MineBlock()` if we'll sort here.

I removed the ordering [here](https://github.com/planetarium/libplanet/pull/294/files#diff-f9c16aaa8d3a312223ca86876f791202L328).

earlbread: > It would be safer if we also materialize `transactions` at here.

I moved `ToList()` to `Block` constructor as well.

Libplanet/Tx/Transaction.cs
```diff
@@ -268,180 +268,6 @@ public static Transaction<T> FromBencodex(byte[] bytes)
...
-            return new Transaction<T>(
-                nonce,
-                signer,
-                publicKey,
-                updatedAddresses,
-                ts,
-                actionsArray,
-                sig
-            );
-        }
```
dahlia: Does this method have to be removed at all?  I thought this is fine to remain, and `BlockChain<T>.MakeTransaction()` method could be a thin façade on this.

Libplanet/Blocks/Block.cs
```diff
@@ -38,7 +38,7 @@ public class Block<T> : ISerializable
             Miner = miner;
             PreviousHash = previousHash;
             Timestamp = timestamp;
-            Transactions = transactions;
+            Transactions = transactions.OrderBy(tx => tx.Nonce).ToList();
```
dahlia: Could you elaborate the intention behind this change?  I guess it is to address the issue #244, but the suggested change seems not completely solve that, because…

1. If a miner does not sort transactions in that way but the way for their advantage other nodes don't reject such blocks.
2. A transaction signer could adjust their nonce to fabricate evaluation order of actions so that their actions execute earlier or later.

Libplanet/Tx/Transaction.cs
earlbread: I think it's okay to leave this, but is there a use case?

Libplanet/Blocks/Block.cs
earlbread: This is not intended to solve #244, but just to move of [this code](https://github.com/planetarium/libplanet/pull/294/files#diff-f9c16aaa8d3a312223ca86876f791202L328) from `MineBlock` to `Block` constructor. The intention of the code is sorting transactions to prevent `InvalidTxNonceException` when mining. Later, we need to fix the code to solve #244. I'll add FIXME to prevent confusion.

earlbread:
I made `Transaction<T>.Create()`, `BlockChain<T>.GetNonce()` and `BlockChain<T>.StageTransactions()` to public again to have options to create `Transaction`.

PR #297, opened 2019-06-21
Make Swarm broadcast transactions periodically

earlbread:
This makes `Swarm` broadcast transactions periodically. To do this, I make staged `Transaction`s store whether to broadcast.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,13 @@ To be released.
     is unknown.
  -  Added `IStore.ListAddresses()` method.  [[#272], [#285]]
  -  Removed `BlockChain<T>.GetNonce()` method.  [[#294]]
+ -  `BlockChain<T>.StageTransactions` became to receive
+    `IDictionary<Transaction<T>, bool>` instead of `ISet<Transaction<T>>`.
+ -  `IStore.StageTransactionIds()` method became to receive
+    `IDictionary<TxId, bool>` instead of `ISet<TxId>`.
+ -  `IStore.IterateStagedTransactionIds()` method became to receive
+    `bool toBroadcast` which is whether to iterate only the TxId set to
+    broadcast.
 
 ### Added interfaces
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ To be released.
 ### Backward-incompatible interface changes
 
  -  `Peer.AppProtocolVersion` became nullable to represent `Peer` whose version
-    is unknown.
+    is unknown.  [[#280]]
  -  Added `IStore.ListAddresses()` method.  [[#272], [#285]]
  -  Removed `BlockChain<T>.GetNonce()` method.  [[#294]]
  -  `BlockChain<T>.StageTransactions` became to receive
@@ -29,7 +29,7 @@ To be released.
  -  All `*Async()` methods belonging to `TurnClient` class became to have
     `cancellationToken` option.  [[#287]]
  -  Added a `Peer` constructor omitting `appProtocolVersion` parameter
-    to create a `Peer` whose version is unknown.
+    to create a `Peer` whose version is unknown.  [[#280]]
  -  Added `IncompleteBlockStatesException` class.  [[#272], [#285]]
  -  Added `completeStates` option to `BlockChain<T>.GetStates()` method.
     [[#272], [#285]]
@@ -83,6 +83,7 @@ To be released.
 [#275]: https://github.com/planetarium/libplanet/pull/275
 [#276]: https://github.com/planetarium/libplanet/pull/276
 [#277]: https://github.com/planetarium/libplanet/pull/277
+[#281]: https://github.com/planetarium/libplanet/pull/280
 [#281]: https://github.com/planetarium/libplanet/pull/281
 [#285]: https://github.com/planetarium/libplanet/pull/285
 [#287]: https://github.com/planetarium/libplanet/pull/287
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@ To be released.
  -  `IStore.IterateStagedTransactionIds()` method became to receive
     `bool toBroadcast` which is whether to iterate only the TxId set to
     broadcast.
+ -  `Swarm.StartAsync()` method became to receive `broadcastTxInterval`
+    (or `millisecondsBroadcastTxInterval`) parameter.
 
 ### Added interfaces
 
@@ -49,6 +51,7 @@ To be released.
  -  Added IPv6 support to `Libplanet.Stun.StunAddress`. [[#267], [#271]]
  -  `IStore.GetBlockStates()` became able to return `null` to represent an absence
     of states (i.e., incomplete states).  [[#272], [#285]]
+ -  `Swarm` became to broadcast staged `Transaction`s periodically.
 
 ### Bug fixes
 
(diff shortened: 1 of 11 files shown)


Libplanet/Store/LiteDBStore.cs
```diff
@@ -163,6 +167,12 @@ public void UnstageTransactionIds(ISet<TxId> txids)
             StagedTxIds.Delete(tx => txids.Contains(tx.TxId));
         }
 
+        /// <inheritdoc />
+        public IEnumerable<TxId> IterateBroadcastTransactionIds()
+        {
+            return StagedTxIds.FindAll().Where(t => t.Broadcast == true).Select(t => t.TxId);
```
longfin:
```suggestion
            return StagedTxIds.FindAll().Where(t => t.Broadcast).Select(t => t.TxId);
```

Libplanet/Net/Swarm.cs
```diff
@@ -976,6 +982,34 @@ private async Task RefreshAllocate(CancellationToken cancellationToken)
             }
         }
 
+        private async Task BroadcastTxAsync<T>(
+            BlockChain<T> blockChain,
+            TimeSpan broadcastTxInterval,
+            CancellationToken cancellationToken)
+            where T : IAction, new()
+        {
+            TimeSpan lifetime = TurnPermissionLifetime;
```
longfin: I guess it's unnecessary.

Libplanet/Net/Swarm.cs
```diff
@@ -1356,8 +1390,8 @@ private void TransferTxs<T>(BlockChain<T> blockChain, GetTxs getTxs)
 
             IAsyncEnumerable<Transaction<T>> fetched = GetTxsAsync<T>(
                 peer, unknownTxIds, cancellationToken);
-            var toStage = new HashSet<Transaction<T>>(
-                await fetched.ToListAsync(cancellationToken));
+            List<Transaction<T>> txs = await fetched.ToListAsync(cancellationToken);
+            var toStage = txs.ToDictionary(tx => tx, _ => true);
```
longfin: It seems to re-broadcast all transactions from other peer. is it intended?

Libplanet/Store/BaseStore.cs
```diff
@@ -32,14 +32,18 @@ HashDigest<SHA256> hash
         /// <inheritdoc />
         public abstract IEnumerable<Address> ListAddresses(string @namespace);
 
+        /// <inheritdoc />
         public abstract void StageTransactionIds(
-            ISet<TxId> txids
+            IDictionary<TxId, bool> txids
```
longfin: I prefer `IEnumerable<(TxId, bool)>` to `IDictionary<TxId, bool>` because we have no plans to lookup.

Libplanet/Net/Swarm.cs
earlbread: This is to stage transactions from other peers. Did you talk about [this](https://github.com/planetarium/libplanet/pull/297/files#diff-9501da6c2e844a81d5e3ad3d3d5456ccR996)? I intend to broadcast all transactions again from others, do you think it is inappropriate?

longfin: > Did you talk about this?

Yes. if we stage in this method, we also resend them through mentioned code.

> I intend to broadcast all transactions again from others, do you think it is inappropriate?

No, it seems to be appropriate. but it would be better if related test cases exist.

Libplanet/Store/BaseStore.cs
earlbread: I tried to change it to `IEnumerable<(TxId, bool)>`, but when `txids` is `IEnumerable` the same `TxId` can have different values.

Libplanet/Net/Swarm.cs
```diff
@@ -976,6 +983,45 @@ private async Task RefreshAllocate(CancellationToken cancellationToken)
...
+                try
+                {
+                    await Task.Delay(broadcastTxInterval, cancellationToken);
+
+                    await Task.Run(
+                        () =>
+                        {
+                            List<TxId> txIds = blockChain.Store
+                                .IterateBroadcastTransactionIds()
+                                .ToList();
```
dahlia: We need to test if non-broadcasting staged transactions are actually never broadcasted to peers.  If I've read this patch correctly the current test suite would pass even if we replace `.IterateBroadcastTransactionIds()` with `.IterateStagedTransactionIds()` here.

earlbread:
I applied review suggestions and fixed a problem with test timeout. Please take a look.

CHANGES.md
```diff
@@ -42,6 +49,9 @@ To be released.
  -  Added IPv6 support to `Libplanet.Stun.StunAddress`. [[#267], [#271]]
  -  `IStore.GetBlockStates()` became able to return `null` to represent an absence
     of states (i.e., incomplete states).  [[#272], [#285]]
+ -  `Swarm` became to broadcast staged `Transaction`s periodically.  [[#274], [#297]]
```
```suggestion
 -  `Swarm` became to broadcast staged `Transaction`s periodically
     so that game apps no more need to maintain their own thread to
     broadcast staged transactions.  [[#274], [#297]]
```

CHANGES.md
```diff
@@ -42,6 +49,9 @@ To be released.
...
+ -  When `Swarm` receives a TxIds message, it has been changed to request transactions
+    only if there are unknown `TxId`.
```
dahlia: Could you elaborate on this?

Libplanet/Net/Swarm.cs
dahlia: I believe the commit 6b8d2e080ffac22350d4def7b56b32b2e9fe49c3 does not fix what I addressed above…  I meant our test code *should not* pass if we replace `.IterateBroadcastTransactionIds()` with `.IterateStagedTransactionIds()` here.  In other words, it's about test coverage.

earlbread: That commit you mentioned is not for this case but for using lock. I added test for this case [here](https://github.com/planetarium/libplanet/pull/297/files#diff-29be8fcde7b01328ce4fda0a59dffa1cR604). Could you please look at it?

CHANGES.md
```diff
@@ -42,6 +52,13 @@ To be released.
  -  Added IPv6 support to `Libplanet.Stun.StunAddress`. [[#267], [#271]]
  -  `IStore.GetBlockStates()` became able to return `null` to represent an absence
     of states (i.e., incomplete states).  [[#272], [#285]]
+ -  `Swarm` became to broadcast staged `Transaction`s periodically
+     so that game apps no more need to maintain their own thread to
+     broadcast staged transactions.  [[#274], [#297]]
+ -  Currently, Swarm sent `GetTxs` with an empty `txIds` when it receives a
+    `TxIds` message and if it does not have unknown `TxId`s.  Changed so that
+    the `GetTxs` message is not sent when there are no unknown `TxId`s.
+    [[#297]]
```
dahlia: Thanks for elaboration.  In changelog, don't a present tense or words like “current” denote the version after a change made, because people read this after we release a new version?  To me it'd better be:

```markdown
 -  Previously, `Swarm` had sent an empty `GetTxs` message when it receives
    an empty `TxIds` from peers, and it had made the network waste bandwidth for
    unnecessary messages.  `Swam` became to no more send such empty `GetTxs`.
    [[#297]]
```

Libplanet/Net/Swarm.cs
dahlia: Ah, okay.  I'm going to take a look!

dahlia: `TxStagedNotToBroadcast()` indeed addresses the case.  :100:

CHANGES.md
earlbread: I updated the changelog.

PR #298, opened 2019-06-24
Made LiteDBStore to be configurable.

longfin:
This PR adds `journal` and `cacheSize` as optional parameters to `LiteDBStore`'s constructor.

I omitted change log because the mention about `LiteDBStore` has been added `0.0.4` section.

### DIFF
--- a/Libplanet/Store/LiteDBStore.cs
+++ b/Libplanet/Store/LiteDBStore.cs
@@ -28,7 +28,7 @@ namespace Libplanet.Store
 
         private readonly LiteDatabase _db;
 
-        public LiteDBStore(string path)
+        public LiteDBStore(string path, bool journal = true, int cacheSize = 50000)
         {
             if (path is null)
             {
@@ -38,8 +38,8 @@ namespace Libplanet.Store
             var connectionString = new ConnectionString
             {
                 Filename = path,
-                Journal = false,
-                CacheSize = 50000,
+                Journal = journal,
+                CacheSize = cacheSize,
             };
 
             if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) &&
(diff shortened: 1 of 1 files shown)


Libplanet/Store/LiteDBStore.cs
```diff
@@ -16,6 +16,10 @@
 
 namespace Libplanet.Store
 {
+    /// <summary>
+    /// <seealso cref="IStore"/> implementation using <a href="https://www.litedb.org/">LiteDB</a>.
```
earlbread:
```suggestion
    /// <see cref="IStore"/> implementation using <a href="https://www.litedb.org/">LiteDB</a>.
```

Libplanet/Store/LiteDBStore.cs
```diff
@@ -28,7 +32,15 @@ public class LiteDBStore : IStore, IDisposable
 
         private readonly LiteDatabase _db;
 
-        public LiteDBStore(string path)
+        /// <summary>
+        /// Creates a new <seealso cref="LiteDBStore"/>.
+        /// </summary>
+        /// <param name="path">The path where the storage file will be saved.</param>
+        /// <param name="journal">
+        /// Enabled or disable double write check to ensure durability.
```
earlbread:
```suggestion
        /// Enables or disables double write check to ensure durability.
```

Libplanet/Store/LiteDBStore.cs
```diff
@@ -28,7 +32,15 @@ public class LiteDBStore : IStore, IDisposable
...
+        /// </param>
+        /// <param name="cacheSize">Max number of pages in cache.</param>
```
earlbread:
```suggestion
        /// <param name="cacheSize">Max number of pages in the cache.</param>
```

PR #308, opened 2019-06-27
Improve TurnClient connection

longfin:
This PR adds improves connectivity of `TurnClient._control` by adding `TurnClient.EnsureConnection` to ensure `_control` connection before processing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -71,6 +71,9 @@ To be released.
  -  Fixed a bug where `LiteDBStore.GetTxNonce()` method throws a
     `System.IO.IOException` after forking.  [[#281]]
  -  Fixed a bug that `TurnClient` had not stopped properly.  [[#287]]
+ -  Fixed a bug that `TurnClient` had been trying to use an already closed
+    connection. [[#303], [#308]]
+
 
 [LiteDB]: https://www.litedb.org/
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -59,6 +59,7 @@ To be released.
     an empty `TxIds` from peers, and it had made the network waste bandwidth for
     unnecessary messages.  `Swam` became to no more send such empty `GetTxs`.
     [[#297]]
+ -  Improved connection stability of `TurnClient`. [[#303], [#308]]
```
dahlia: Isn't this a bug fix?

longfin:
I've added https://github.com/planetarium/libplanet/pull/308/commits/1c914ed6e8d97680f4b352fc49c590f5f78069ae to avoid `ArgumentException` on `IceServer.CreateTurnClient()`.

Libplanet/Net/IceServer.cs
```diff
@@ -61,7 +61,7 @@ public class IceServer
 
                         return turnClient;
                     }
-                    catch (SocketException)
+                    catch (Exception)
```
dahlia: How about catching both `SocketException` and `ArgumentException` instead of using wildcard here?

Issue #309, opened 2019-06-28
When forking chain, Tx nonce should be set


earlbread:
After #307, Tx nonce is not set except `signersToStrip` addresses.
Reproduce scenario is below:

```
var privateKey1 = new PrivateKey();
var privateKey2 = new PrivateKey();

var address1 = privateKey1.PublicKey.ToAddress();
var address2 = privateKey2.PublicKey.ToAddress();

_blockChain.MineBlock(address1);

_blockChain.MakeTransaction(privateKey1, new[] { new DumbAction() });
Block<DumbAction> branchPoint = _blockChain.MineBlock(address1);

_blockChain.MakeTransaction(privateKey2, new[] { new DumbAction() });
_blockChain.MineBlock(address2);

BlockChain<DumbAction> forked = _blockChain.Fork(branchPoint.Hash);
Assert.Equal(1, forked.GetNextTxNonce(address1));
```

dahlia:
I'm going to fix this.

PR #310, opened 2019-06-28
Fix tx nonces being missed during fork

dahlia:
This fixes the bug #309 that tx nonces being missed during fork, because addresses that aren't *stale* are missing.  To correctly fix this, I've needed `IStore.ListTxNonces()` as well (note that `IStore.ListAddresses()` could miss addresses that have never had any states but ever signed some transactions).

### DIFF
--- a/Libplanet/Store/IStore.cs
+++ b/Libplanet/Store/IStore.cs
@@ -181,7 +181,7 @@ namespace Libplanet.Store
         /// <param name="namespace">The namespace to list <see cref="Address"/>es and their
         /// <see cref="Transaction{T}"/> nonces.</param>
         /// <returns>Pairs of an <see cref="Address"/> and its tx nonce.  All nonces are greater
-        /// than 0.  (If there are underlying entries having zero nonce these must be hidden.)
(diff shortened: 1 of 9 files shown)


Libplanet/Store/IStore.cs
```diff
@@ -174,6 +174,18 @@ AddressStateMap states
...
             where T : IAction, new();
 
+        /// <summary>
+        /// Lists all <see cref="Address"/>es that have ever signed <see cref="Transaction{T}"/>,
+        /// and their corresponding <see cref="Transaction{T}"/> nonces.
+        /// </summary>
+        /// <param name="namespace">The namespace to list <see cref="Address"/>es and their
+        /// <see cref="Transaction{T}"/> nonces.</param>
+        /// <returns>Pairs of an <see cref="Address"/> and its tx nonce.  All nonces are greater
+        /// than 0.  (If there are underlying entries having zero nonce these must be hidden.)
```
earlbread:
```suggestion
        /// than 0.  (If there are underlying entries having zero nonces these must be hidden.)
```

PR #317, opened 2019-06-29
FileStore.IndexBlockhash return null when index out of range

ipdae:
resolve #214

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -88,9 +88,13 @@ To be released.
  -  Fixed a bug that `TurnClient` had not stopped properly.  [[#287]]
  -  Fixed a bug that `TurnClient` had been trying to use an already closed
     connection. [[#303], [#308]]
+ -  Fixed a bug that `KeyNotFoundException` had been thrown instead of
+    `ArgumentOutOfRangeException` when `Blockchain<T>[int]` called while the
+    index of a block that does not exist locally.  [[#208]]
 
 
 [LiteDB]: https://www.litedb.org/
+[#208]: https://github.com/planetarium/libplanet/issues/208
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -88,9 +88,13 @@ To be released.
  -  Fixed a bug that `TurnClient` had not stopped properly.  [[#287]]
  -  Fixed a bug that `TurnClient` had been trying to use an already closed
     connection. [[#303], [#308]]
+ -  Fixed a bug that `KeyNotFoundException` had been thrown instead of
+    `ArgumentOutOfRangeException` when `Blockchain<T>[int]` called while the
+    index of a block that does not exist locally.  [[#208]]
```
```suggestion
    index of a block that does not exist locally.  [[#208], [#317]]
```

Libplanet.Tests/Store/StoreTest.cs
```diff
@@ -470,5 +470,15 @@ public void TxNonce()
...
             );
         }
+
+        [Fact]
+        public void IndexBlockHashReturnNull()
+        {
+            Fx.Store.PutBlock(Fx.Block1);
+            Fx.Store.AppendIndex(Fx.StoreNamespace, Fx.Block1.Hash);
+            var index = Fx.Store.CountIndex(Fx.StoreNamespace);
+            Assert.True(index == 1);
```
longfin:
```suggestion
            Assert.Equals(1, index);
```

Libplanet/Store/FileStore.cs
```diff
@@ -362,6 +362,13 @@ long index
                     return null;
                 }
             }
+            else
+            {
+                if (CountIndex(@namespace) < index)
```
longfin: it seems to can be replaced with `else if`.

PR #324, opened 2019-07-03
Make Swarm constructor have BlockChain<T>

earlbread:
This makes `Swarm` constructor receive `BlockChain<T>` and have a type parameter `T`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,10 @@ To be released.
         Instead, a newly added static class `StoreExtension` provides
         an extension method of the same name.
      -  Added `IStore.IterateStateReferences()` method.
+ -  `Swarm` constructor became to receive `BlockChain<T>`.  [[#324]]
+ -  `Swarm` became to have type parameter `T`.  [[#324]]
+ -  Methods that take `BlockChain<T>` as a parameter in `Swarm` no longer take
+    `BlockChain<T>`. And they are no longer the generic methods.  [[#324]]
 
 ### Added interfaces
 
@@ -120,6 +124,7 @@ To be released.
 [#310]: https://github.com/planetarium/libplanet/pull/310
 [#311]: https://github.com/planetarium/libplanet/pull/311
 [#317]: https://github.com/planetarium/libplanet/pull/317
+[#324]: https://github.com/planetarium/libplanet/pull/324
 
 
 Version 0.3.0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,10 +37,11 @@ To be released.
         an extension method of the same name.
      -  Added `IStore.IterateStateReferences()` method.
  -  `Swarm` constructor became to receive `BlockChain<T>`.  [[#324]]
- -  `Swarm` became to have type parameter `T` that reprsents action type as
+ -  `Swarm` became to have type parameter `T` to represent an action type as
     like as `BlockChain<T>`.  [[#324]]
- -  Methods that take `BlockChain<T>` as a parameter in `Swarm` no longer take
-    `BlockChain<T>`. And they are no longer the generic methods.  [[#324]]
+ -  Methods  in `Swarm` that had taken a parameter of `BlockChain<T>` type
+    became to neither longer take `BlockChain<T>` nor a generic method.
+    Because the `Swarm` constructor takes it instead.  [[#324]]
 
 ### Added interfaces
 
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -36,6 +36,8 @@ To be released.
         Instead, a newly added static class `StoreExtension` provides
         an extension method of the same name.
      -  Added `IStore.IterateStateReferences()` method.
+ -  `Swarm` constructor became to receive `BlockChain<T>` and have a type
+    parameter `T`.  [[#324]]
```
longfin: Actually, `Swarm`(not constructor, but class) became to have a type parameter. how about write them separately?

earlbread: I separated the changelog.

CHANGES.md
```diff
@@ -36,6 +36,8 @@ To be released.
         Instead, a newly added static class `StoreExtension` provides
         an extension method of the same name.
      -  Added `IStore.IterateStateReferences()` method.
+ -  `Swarm` constructor became to receive `BlockChain<T>`.  [[#324]]
+ -  `Swarm` became to have type parameter `T`.  [[#324]]
```
longfin: it would be better if there is mention about `T`.

```suggestion
 -  `Swarm` became to have type parameter `T` that reprsents action type as like as `BlockChain<T>`.  [[#324]]
```

Libplanet/Net/Swarm.cs
```diff
@@ -450,27 +455,22 @@ IEnumerator IEnumerable.GetEnumerator()
             return GetEnumerator();
         }
 
-        public async Task StartAsync<T>(
-            BlockChain<T> blockChain,
```
dahlia: This and other methods that had taken a `blockChain` argument should be addressed in the changelog.  Not only that it no more takes the argument, but also that it became no more a generic method.

earlbread: Added a description of the methods.

CHANGES.md
```diff
@@ -36,6 +36,11 @@ To be released.
         Instead, a newly added static class `StoreExtension` provides
         an extension method of the same name.
      -  Added `IStore.IterateStateReferences()` method.
+ -  `Swarm` constructor became to receive `BlockChain<T>`.  [[#324]]
+ -  `Swarm` became to have type parameter `T` that reprsents action type as
```
dahlia: A typo: “reprsents.”  Also it would be better:

```suggestion
 -  `Swarm` became to have type parameter `T` to represent an action type as
```

CHANGES.md
```diff
@@ -36,6 +36,11 @@ To be released.
...
+    like as `BlockChain<T>`.  [[#324]]
+ -  Methods that take `BlockChain<T>` as a parameter in `Swarm` no longer take
```
```suggestion
 -  Methods  in `Swarm` that had taken a parameter of `BlockChain<T>` type
```

CHANGES.md
```diff
@@ -36,6 +36,11 @@ To be released.
...
+    `BlockChain<T>`. And they are no longer the generic methods.  [[#324]]
```
```suggestion
    became to neither longer take `BlockChain<T>` nor a generic method.
    Because the `Swarm` constructor takes it instead.  [[#324]]
```

Libplanet/Net/Swarm.cs
```diff
@@ -1068,27 +1042,27 @@ private void BroadcastTxIds(IEnumerable<TxId> txIds)
...
-                        TransferTxs(blockChain, getTxs);
+                        TransferTxs(getTxs);
                         break;
                     }
 
                 case TxIds txIds:
                     {
                         await ProcessTxIds(
-                            txIds, blockChain, cancellationToken);
+                            txIds, cancellationToken);
```
longfin: it seems that we can write on the above line.

Libplanet/Net/Swarm.cs
```diff
@@ -1068,27 +1042,27 @@ private void BroadcastTxIds(IEnumerable<TxId> txIds)
...
                         break;
                     }
 
                 case BlockHashes blockHashes:
                     {
                         await ProcessBlockHashes(
-                            blockHashes, blockChain, cancellationToken);
+                            blockHashes, cancellationToken);
```
longfin: it seems that we can write on the above line.

earlbread:
Applied review comments. Please take a look.

PR #332, opened 2019-07-04
Remove `id` from BlockChain<T> constructor

### LINKED ISSUE #279: Hide the concept of “chain IDs”
Currently game apps that use Libplanet have to remember both of the path of `FileStore` and the major chain ID (which is a UUID type).  This current interface makes the implementation details (of how it distinguishes major/stale chains under the hood) hard to change in the future.  It's also cumbersome for library users because they need to store two interdependent data.

So I suggest to change `BlockChain` to hide the concept of “chain IDs” and automatically choose the major chain by itself.  This would improve both forward compatibility and usability.

longfin:
This PR resolves #279 by removing the `id` parameter from `BlockChain<T>` constructor. it also made `BlockChain<T>.Id` to be automatically determined by chain height.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,9 @@ To be released.
     Because the `Swarm<T>` constructor takes it instead.  [[#324]]
  -  `Swarm<T>` does not implement `ICollection<Peer>` anymore.  [[#326]]
  -  Added `IStore.DeleteNamespace()` method. [[#329]]
+ -  Removed the `id` parameter from the `BlockChain<T>` constructor, and it
+    became to automatically detect an appropriate `BlockChain<T>.Id`.
+    [[#279], [#332]]
 
 ### Added interfaces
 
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -44,6 +44,7 @@ To be released.
     Because the `Swarm<T>` constructor takes it instead.  [[#324]]
  -  `Swarm<T>` does not implement `ICollection<Peer>` anymore.  [[#326]]
  -  Added `IStore.DeleteNamespace()` method. [[#329]]
+ -  Removed `BlockChain<T>` constructor that receives `id`.  [[#TBD]]
```

> Removed the `id` parameter from the `BlockChain<T>` constructor, and it became to automatically detect an appropriate `BlockChain<T>.Id`.

longfin: I wrote mention about `BlockChain<T>.Id` in 'Behavioral changes' section. is better writing them here?

dahlia: As two changes are related each other, writing a single unified log seems better to me.

longfin:
I've renamed `BlockChain<T>.GetBestChain()` to `BlockChain<T>.GetCanonicalChain()`.

PR #344, opened 2019-07-09
Remove NetMQSocketExtension

### LINKED ISSUE #340: Introduce NetMQRuntime
Currently, we're using `async` extensions for NetMQSocket in `Swarm<T>`. but in 0MQ / NetMQ, a socket isn't thread-safe and can't be used in multiple threads. so these methods can be dangerous in some situation.

so, I suggest that remove them and replace with original methods in NetMQSocket. as they are synchronous, we may need to change some interface relying on them.

longfin:
This PR will resolve #340. (maybe?)


longfin:
I'll continue this in #485.

PR #346, opened 2019-07-11
Fix GetStates bug when there are multiple addresses

earlbread:
This fixes a bug where `GetStates` method does not return the latest state when there are multiple addresses. And also this changes the return value of `LookupStateReference()` method to fix the bug.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,8 @@ To be released.
 
 ### Bug fixes
 
- -  Fixed a bug where the GetStates method does not return the latest state
-    when there are multiple addresses.  [[#346]]
+ -  Fixed a bug where the `BlockChain<T>.GetStates()` method had not returned
+    the latest state when there are multiple addresses.  [[#346]]
 
 [#346]: https://github.com/planetarium/libplanet/pull/346
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,12 @@ Version 0.4.1
 
 To be released.
 
+### Bug fixes
+
+ -  Fixed a bug where the GetStates method does not return the latest state
+    when there are multiple addresses.  [[#346]]
+
+[#346]: https://github.com/planetarium/libplanet/pull/346
 
 Version 0.4.0
 -------------
--- a/Libplanet/Store/StoreExtension.cs
+++ b/Libplanet/Store/StoreExtension.cs
@@ -32,6 +32,19 @@ namespace Libplanet.Store
             Block<T> lookupUntil)
             where T : IAction, new()
         {
+            Tuple<HashDigest<SHA256>, long> sr = store.
+                LookupStateReferenceWithIndex(@namespace, address, lookupUntil);
+
+            return sr?.Item1;
+        }
+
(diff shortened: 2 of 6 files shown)


dahlia:
IMHO as this fixes a bug made in the previous minor release, it should be merged into 0.\*-maintenance instead of master.

earlbread:
I changed the target branch to 0.4-maintenance. @dahlia

Libplanet/Store/StoreExtension.cs
```diff
@@ -18,14 +18,15 @@ public static class StoreExtension
...
+        /// <returns>Returns a nullable tuple consisting of <see cref="Block{T}.Hash"/> and
+        /// <see cref="Block{T}.Index"/> of the <see cref="Block{T}"/> with the state of the
+        /// address.</returns>
         /// <typeparam name="T">An <see cref="IAction"/> class used with
         /// <paramref name="lookupUntil"/>.</typeparam>
         /// <seealso
         /// cref="IStore.StoreStateReference{T}(string, IImmutableSet{Address}, Block{T})"/>
         /// <seealso cref="IStore.IterateStateReferences(string, Address)"/>
-        public static HashDigest<SHA256>? LookupStateReference<T>(
+        public static Tuple<HashDigest<SHA256>, long> LookupStateReference<T>(
```
dahlia: How about making this method `internal` and renaming to like `LookupStateReferenceWithIndex<T>()`, and then adding a wrapper method named `LookupStateReference<T>()` around it in order to maintain backward compatibility?

dahlia: Elaboration: This makes a backward-incompatible change on the public interface, which should increase a minor version (not a patch version), according to the Semantic Versioning.  However, this kind of problem is definitely not about usability nor feature lacking, but a bug, so it should be a patch release as possible.  So we'd better to add a new internal method (`LookupStateReferenceWithIndex<T>()`) in this pull request, and then replace the old method (`LookupStateReference<T>()`) with the new one (`LookupStateReferenceWithIndex<T>()`) in a separated pull request which targets to the master branch.

earlbread: I added `LookupStateReferenceWithIndex<T>()` method and maintained `LookupStateReference<T>()` interface as you suggested.

CHANGES.md
```diff
@@ -6,6 +6,12 @@ Version 0.4.1
 
 To be released.
 
+### Bug fixes
+
+ -  Fixed a bug where the GetStates method does not return the latest state
```
```suggestion
 -  Fixed a bug where the `BlockChain<T>.GetStates()` method had not returned
```

CHANGES.md
```diff
@@ -6,6 +6,12 @@ Version 0.4.1
...
+    when there are multiple addresses.  [[#346]]
```
```suggestion
    the latest state when there are multiple addresses.  [[#346]]
```

earlbread:
> Shouldn't we add a regression test calling the `GetStates()` method?

The regression test is added [here](https://github.com/planetarium/libplanet/pull/346/files#diff-f350fcc8d10e89d73dd7a7ad7a6d48a5R845).

PR #353, opened 2019-07-15
Kademlia protocol implemented

limebell:
Swarm will not communicate with all the peers on the network. Choosing and connecting to the peers to communicate with is done by Kademlia protocol.
For this change, some message formats(`BlockHashes`, `TxIds`) and some public methods were changed.
Waiting for the reply from the dealer socket will not create block, it now uses `NetMQPoller` to handle messages.
`SwarmTest` modified due to swarm interface changes. `ProtocolTest` and some new tests in `SwarmTest` is added to check correctness of changed behaviour of swarm.


limebell:
`BootstrapAsync` needs summary since it's public method provided to users of this library.

limebell:
Currently Kademlia protocol here does not implement `refresh`, which should be called every 30 minutes. Refresh looks up nearby nodes from itself, and from some random addresses. This should be implemented to ensure network stability.

limebell:
The PR has some know issues which occurs occasionally:
- Exception at TurnClient.ProcessMessage()
- `System.ObjectDisposedException`: Safe handle has been closed

CHANGES.md
```diff
@@ -12,6 +12,10 @@ To be released.
  -  `StoreExtension.LookupStateReference<T>()` method became to return
     `Tuple<HashDigest<SHA256>, long>` which is a nullable tuple of `Block<T>.Hash`
     and `Block<T>.Index`.  [[#350]]
+
```
```suggestion
```

CHANGES.md
```diff
@@ -12,6 +12,10 @@ To be released.
  -  `StoreExtension.LookupStateReference<T>()` method became to return
     `Tuple<HashDigest<SHA256>, long>` which is a nullable tuple of `Block<T>.Hash`
     and `Block<T>.Index`.  [[#350]]
+    
+ - `Swarm<T>.AddPeersAsync()` method became internal method. To connect with
```
```suggestion
 -  `Swarm<T>.AddPeersAsync()` method became internal method. To connect with
```

CHANGES.md
```diff
@@ -12,6 +12,10 @@ To be released.
...
+    seed peers, use `Swarm<T>.BootstrapAsync()` method instead. [[#353]]
```
```suggestion
     seed peers, use `Swarm<T>.BootstrapAsync()` method instead. [[#353]]
```

CHANGES.md
```diff
@@ -12,6 +12,10 @@ To be released.
...
+
```
```suggestion
```

CHANGES.md
```diff
@@ -44,6 +48,12 @@ To be released.
  -  Added `StoreExtension.ListAllStateReferences(this IStore, string,
     HashDigest<SHA256>?, HashDigest<SHA256>?)` extension method.
     [[#363], [#384], [#385]]
+
```
```suggestion
```

CHANGES.md
```diff
@@ -44,6 +48,12 @@ To be released.
  -  Added `StoreExtension.ListAllStateReferences(this IStore, string,
     HashDigest<SHA256>?, HashDigest<SHA256>?)` extension method.
     [[#363], [#384], [#385]]
+    
+ - Added `Swarm<T>.BootstrapAsync()` method to connect with seed peers.  [[#353]]
```
```suggestion
 -  Added `Swarm<T>.BootstrapAsync()` method to connect with seed peers.  [[#353]]
```

CHANGES.md
```diff
@@ -44,6 +48,12 @@ To be released.
...
+
```
```suggestion
```

CHANGES.md
```diff
@@ -44,6 +48,12 @@ To be released.
  -  Added `StoreExtension.ListAllStateReferences(this IStore, string,
     HashDigest<SHA256>?, HashDigest<SHA256>?)` extension method.
     [[#363], [#384], [#385]]
+    
+ - Added `Swarm<T>.BootstrapAsync()` method to connect with seed peers.  [[#353]]
+    
+ -  Added `StoreExtension.ListAllStateReferences(this IStore, string)` extension
```
longfin: It seems duplicated.

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
 
  -  `BlockChain<T>.PreloadAsync()` method became to omit rendering of
     `IAction`s in the preloaded behind blocks.  [[#272], [#343]]
+
```
```suggestion
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
 
  -  `BlockChain<T>.PreloadAsync()` method became to omit rendering of
     `IAction`s in the preloaded behind blocks.  [[#272], [#343]]
+    
  -  `Swarm<T>` became to have two more message types: `GetRecentStates` (`0x0b`)
     and `RecentStates` (`0x0c`).  [[#272], [#343]]
+
```
```suggestion
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
 
  -  `BlockChain<T>.PreloadAsync()` method became to omit rendering of
     `IAction`s in the preloaded behind blocks.  [[#272], [#343]]
+    
  -  `Swarm<T>` became to have two more message types: `GetRecentStates` (`0x0b`)
     and `RecentStates` (`0x0c`).  [[#272], [#343]]
+    
+ - Peers now connected via Kademlia protocol. Peers are now selectively connected to
```
```suggestion
 -  Peers now connected via Kademlia protocol. Peers are now selectively connected to
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
...
+    each peer. [[#353]]
```
```suggestion
     each peer. [[#353]]
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
...
+
+- TxIds and Blocks are now broadcasted to selected peers from routing table of the host
+  peer. [[#353]]
+
+- `NetMQPoller` is not used anymore. [[#353]]
+
  -  `BlockChain<T>.MineBlock()` and `BlockChain<T>.GetNextTxNonce()` methods
     became to ignore transactions that didn't follow `Transaction<T>.Nonce`
     sequentially and treat them as pendings.  [[#365]]
+
```
```suggestion
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
...
  -  `Swarm<T>` became to have two more message types: `GetRecentStates` (`0x0b`)
     and `RecentStates` (`0x0c`).  [[#272], [#343]]
+    
+ - Peers now connected via Kademlia protocol. Peers are now selectively connected to
+    each peer. [[#353]]
+
+- TxIds and Blocks are now broadcasted to selected peers from routing table of the host
+  peer. [[#353]]
+
+- `NetMQPoller` is not used anymore. [[#353]]
```
longfin: Is this needed for libplanet user?

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
 
  -  `BlockChain<T>.PreloadAsync()` method became to omit rendering of
     `IAction`s in the preloaded behind blocks.  [[#272], [#343]]
+    
  -  `Swarm<T>` became to have two more message types: `GetRecentStates` (`0x0b`)
     and `RecentStates` (`0x0c`).  [[#272], [#343]]
+    
+ - Peers now connected via Kademlia protocol. Peers are now selectively connected to
+    each peer. [[#353]]
+
```
```suggestion
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
...
+- TxIds and Blocks are now broadcasted to selected peers from routing table of the host
```
```suggestion
 -  TxIds and Blocks are now broadcasted to selected peers from routing table of the host
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
...
+  peer. [[#353]]
```
```suggestion
    peer. [[#353]]
```

CHANGES.md
```diff
@@ -44,6 +48,12 @@ To be released.
  -  Added `StoreExtension.ListAllStateReferences(this IStore, string,
     HashDigest<SHA256>?, HashDigest<SHA256>?)` extension method.
     [[#363], [#384], [#385]]
+    
+ - Added `Swarm<T>.BootstrapAsync()` method to connect with seed peers.  [[#353]]
+    
+ -  Added `StoreExtension.ListAllStateReferences(this IStore, string)` extension
+    method.  [[#363]]
+
```
```suggestion
```

CHANGES.md
```diff
@@ -57,11 +67,22 @@ To be released.
...
+    
  -  `Swarm<T>` became to have two more message types: `GetRecentStates` (`0x0b`)
     and `RecentStates` (`0x0c`).  [[#272], [#343]]
+    
+ - Peers now connected via Kademlia protocol. Peers are now selectively connected to
+    each peer. [[#353]]
+
+- TxIds and Blocks are now broadcasted to selected peers from routing table of the host
+  peer. [[#353]]
+
```
```suggestion
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -109,10 +112,25 @@ public void BlockChain()
         }
 
         [Fact(Timeout = Timeout)]
-        public async Task CannotStartTwice()
+        public async Task CanNotStartBeforePrepare()
+        {
+            Swarm<DumbAction> swarm = _swarms[0];
+
+            Task t = swarm.StartAsync();
+
```
longfin: It seems that we need to wait `t`.

CHANGES.md
longfin: Also, it would be better if there is a link about [Kademlia](https://en.wikipedia.org/wiki/Kademlia).

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -127,14 +145,115 @@ public async Task CannotStartTwice()
             await swarm.StopAsync();
         }
 
+        [Fact(Timeout = Timeout)]
+        public async Task CanHandleReconnection()
```
longfin:
```suggestion
        public async Task HandleReconnection()
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -127,14 +145,115 @@ public async Task CannotStartTwice()
...
+            finally
+            {
+                await seed.StopAsync();
+                await swarmA.StopAsync();
+                await swarmB.StopAsync();
+            }
+        }
+
+        [Fact(Timeout = Timeout)]
+        public async Task CanBroadcastBlockToReconnectedPeer()
```
longfin:
```suggestion
        public async Task BroadcastBlockToReconnectedPeer()
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -175,6 +295,190 @@ public async Task CanWaitForRunning()
...
+            }
+            finally
+            {
+                await swarmA.StopAsync();
+                await swarmB.StopAsync();
+                await swarmC.StopAsync();
+            }
+        }
+
+        [Fact(Timeout = 2 * Timeout)]
```
longfin: is this intended?

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -769,42 +1042,37 @@ public async Task CanBroadcastBlock()
...
+                Log.Debug(swarmA.TraceTable());
+                Log.Debug(swarmB.TraceTable());
+                Log.Debug(swarmC.TraceTable());
 
                 swarmB.BroadcastBlocks(new[] { chainB.Last() });
 
-                await swarmC.BlockReceived.WaitAsync();
-                await swarmA.BlockReceived.WaitAsync();
-
-                Assert.Equal(chainB.AsEnumerable(), chainC);
```
longfin: is this intended?

Libplanet/Net/Messages/Message.cs
```diff
@@ -134,8 +145,13 @@ public static Message Parse(NetMQMessage raw, bool reply)
             return message;
         }
 
-        public NetMQMessage ToNetMQMessage(PrivateKey key)
+        public NetMQMessage ToNetMQMessage(PrivateKey key, Peer self)
```
longfin: How about just `peer` instead of `self`?

Libplanet/Net/Protocols/KBucket.cs
```diff
@@ -0,0 +1,102 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
+namespace Libplanet.Net.Protocols
+{
+    internal class KBucket
```
longfin: It would be great if we have unit tests for this class.

Libplanet/Net/Protocols/KBucket.cs
```diff
@@ -0,0 +1,102 @@
...
+            _peers = new List<Peer>();
+            ReplacementCache = new List<Peer>();
+
+            _lastUpdated = DateTimeOffset.UtcNow;
+        }
+
+        public int Count => _peers.Count;
+
+        // get most recently used peer.
+        public Peer Head => _peers.Count == 0 ? null : _peers[_peers.Count - 1];
```
longfin:
```suggestion
        public Peer Head => Empty() ? null : _peers[_peers.Count - 1];
```

Libplanet/Net/Protocols/KBucket.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        // get least recently used peer.
+        public Peer Tail => _peers.Count == 0 ? null : _peers[0];
```
longfin:
```suggestion
        public Peer Tail => Empty() ? null : _peers[0];
```

Libplanet/Net/Protocols/Kademlia.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Libplanet.Net.Protocols
+{
+    internal static class Kademlia
```
longfin: It would be great if we have unit tests for this class.

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+                _address, TableSize, BucketSize, _random);
+        }
+
+        public int Count => _routing.Count;
+
+        public ImmutableList<Peer> Peers
+        {
+            get
+            {
+                List<Peer> peers = new List<Peer>();
```
longfin:
```suggestion
                var peers = new List<Peer>();
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+            TimeSpan? pingSeedTimeout,
+            TimeSpan? findPeerTimeout,
+            CancellationToken cancellationToken)
+        {
+            if (bootstrapPeers is null)
+            {
+                throw new ArgumentNullException(nameof(bootstrapPeers));
+            }
+
+            List<Task> findPeerTasks = new List<Task>();
```
longfin:
```suggestion
            var findPeerTasks = new List<Task>();
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+                {
+                    await Task.WhenAll(tasks);
+                }
+                catch (TimeoutException)
+                {
+                }
+            }
+        }
+
+#pragma warning disable CS4014 // 이 호출을 대기하지 않으므로 호출이 완료되기 전에 현재 메서드가 계속 실행됩니다.
```
longfin:
```suggestion
#pragma warning disable CS4014
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+                case FindPeer findPeer:
+                    ReceiveFindPeerAsync(findPeer);
+                    break;
+
+                default:
+                    UpdateAsync(message.Remote, _cancellationToken);
+                    break;
+            }
+        }
+#pragma warning restore CS4014 // 이 호출을 대기하지 않으므로 호출이 완료되기 전에 현재 메서드가 계속 실행됩니다.
```
longfin:
```suggestion
#pragma warning restore CS4014
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+
+        public string Trace()
```
longfin: What purpose of this method?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+                await RemovePeerAsync(target);
+                throw;
+            }
+            catch (DifferentAppProtocolVersionException)
+            {
+                _logger.Debug("Different AppProtocolVersion encountered at PingAsync.");
+            }
+            catch (Exception)
+            {
+                throw;
```
longfin: This statement seems to can be omitted.

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+            {
+                cancellationToken = _cancellationToken;
+            }
+
+            if (cancellationToken.IsCancellationRequested)
+            {
+                throw new TaskCanceledException();
+            }
+
+            if (peer != null)
```
longfin:
```suggestion
            if (peer is Peer)
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+                    }
+                    catch (TimeoutException)
+                    {
+                        await RemovePeerAsync(peer);
+                        await _routing.AddPeerAsync(evictionCandidate);
+                    }
+                }
+            }
+
+            /*foreach (KBucket bucket in _routing.NonFullBuckets)
```
longfin: Is this block needed?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+            await ProcessFoundAsync(found, target, timeout, cancellationToken);
+        }
+
+        private async Task<ImmutableList<Peer>> QueryNeighboursAsync(
+            Address target,
+            TimeSpan? timeout,
+            CancellationToken cancellationToken)
+        {
+            List<Peer> neighbours = _routing.Neighbours(target, BucketSize).ToList();
+            List<Peer> found = new List<Peer>();
```
longfin:
```suggestion
            var found = new List<Peer>();
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+            return found.ToImmutableList();
+        }
+
+        private async Task<ImmutableList<Peer>> GetNeighbours(
+            Peer addressee,
+            Address target,
+            TimeSpan? timeout,
+            CancellationToken cancellationToken)
+        {
+            FindPeer findPeer = new FindPeer(target);
```
longfin:
```suggestion
            var findPeer = new FindPeer(target);
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+            _swarm.ReplyMessage(pong);
+        }
+
+        private async Task ProcessFoundAsync(
+            ImmutableList<Peer> found,
+            Address target,
+            TimeSpan? timeout,
+            CancellationToken cancellationToken)
+        {
+            List<Peer> peers = new List<Peer>();
```
longfin:
```suggestion
            var peers = new List<Peer>();
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+            }
+
+            if (foundPingTimeout)
+            {
+                _logger.Debug("All neighbours found are invalid.");
+                throw new TimeoutException();
+            }
+
+            List<Peer> closest_candidate = _routing.Neighbours(target, BucketSize).ToList();
+            List<Task> findPeerTasks = new List<Task>();
```
longfin:
```suggestion
            var findPeerTasks = new List<Task>();
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,525 @@
...
+                }
+            }
+
+            if (foundPingTimeout)
+            {
+                _logger.Debug("All neighbours found are invalid.");
+                throw new TimeoutException();
+            }
+
+            List<Peer> closest_candidate = _routing.Neighbours(target, BucketSize).ToList();
```
longfin:
```suggestion
            List<Peer> closestCandidates = _routing.Neighbours(target, BucketSize).ToList();
```

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -0,0 +1,209 @@
...
+
+                return count;
+            }
+        }
+
+        public List<KBucket> NonFullBuckets
+        {
+            get
+            {
+                List<KBucket> buckets = new List<KBucket>();
```
longfin:
```suggestion
                var buckets = new List<KBucket>();
```

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -0,0 +1,209 @@
...
+    internal class RoutingTable
+    {
+        private readonly Address _address;
+        private readonly int _tableSize;
+        private readonly int _bucketSize;
+        private readonly Random _random;
+        private readonly KBucket[] _buckets;
+        private readonly AsyncLock _bucketMutex;
+
+        public RoutingTable(
```
longfin: It would be great if we have unit tests for this class.

Libplanet/Net/Swarm.cs
```diff
@@ -441,46 +329,104 @@ private set
...
+        /// already <see cref="Running"/>.</exception>
+        /// <remarks>If the <see cref="BlockChain"/> has no blocks at all or there are long behind
+        /// blocks to caught in the network this method could lead to unexpected behaviors, because
+        /// this tries to <see cref="IAction.Render"/> <em>all</em> actions in the behind blocks
+        /// so that there are a lot of calls to <see cref="IAction.Render"/> method in a short
+        /// period of time.  This can lead a game startup slow.  If you want to omit rendering of
+        /// these actions in the behind blocks use <see cref=
+        /// "PreloadAsync(IProgress{PreloadState}, IImmutableSet{Address}, CancellationToken)"
+        /// /> method too.</remarks>
+        public async Task StartAsync(TimeSpan broadcastTxInterval)
```
longfin: `CancellationToken` is missed. is it intended?

Libplanet/Net/Swarm.cs
```diff
@@ -711,6 +707,179 @@ public void BroadcastTxs(IEnumerable<Transaction<T>> txs)
...
+                {
+                    tasks.Add(kp.PingAsync(
+                        peer,
+                        timeout: timeout,
+                        cancellationToken: cancellationToken));
+                }
+
+                await Task.WhenAll(tasks);
+            }
+            catch (TimeoutException)
```
longfin: I think that it's useless because `Take.WhenAll` wraps its exceptions into `AggregateException`.

Libplanet/Net/Swarm.cs
```diff
@@ -711,6 +707,179 @@ public void BroadcastTxs(IEnumerable<Transaction<T>> txs)
...
+                return;
+            }
+
+            DealerSocket dealer = await GetDealerSocket(peer);
+
+            try
+            {
+                _logger.Debug($"Trying to send [{message}] to [{peer.ToString()}]...");
+
+                // dealer.SendMultipartMessage(message.ToNetMQMessage(_privateKey, AsPeer));
```
longfin: Is it needed?

Libplanet/Net/Swarm.cs
```diff
@@ -1476,6 +1597,9 @@ CancellationToken cancellationToken
                     _blockChain.Swap(previousBlocks, render: true);
                     _logger.Debug("Swapping complete");
                 }
+
+                Message msg = new BlockHashes(peer.Address, blocks.Select(b => b.Hash));
```
```suggestion
                var msg = new BlockHashes(peer.Address, blocks.Select(b => b.Hash));
```

Libplanet/Net/Swarm.cs
```diff
@@ -2077,7 +1921,7 @@ private void ReceiveMessage(object sender, NetMQSocketEventArgs e)
                 Message message = Message.Parse(raw, reply: false);
                 _logger.Debug("A message has parsed: {0}", message);
 
-                // it's still async because some method it relies are async yet.
+                // it's still async because some method it relies are async yet.Task.Run(
```
```suggestion
                // it's still async because some method it relies are async yet.
```

Libplanet/Net/Swarm.cs
```diff
@@ -597,9 +593,9 @@ public void BroadcastTxs(IEnumerable<Transaction<T>> txs)
 
             try
             {
-                await SyncBehindsBlocksFromPeersAsync(
+                await SyncBehindsBlocksFromPeerAsync(
                     workspace,
-                    peersWithHeight,
+                    peersWithHeight.First(),
```
longfin: Is `peersWithHeight` sorted? (or, is `.First()` just arbitrary?)

Libplanet/Net/Messages/Neighbours.cs
```diff
@@ -0,0 +1,50 @@
...
+        protected override IEnumerable<NetMQFrame> DataFrames
+        {
+            get
+            {
+                yield return new NetMQFrame(
+                    NetworkOrderBitsConverter.GetBytes(Found.Count));
+
+                foreach (Peer peer in Found)
+                {
+                    yield return new NetMQFrame(SerializePeer(peer));
```
earlbread: If the maximum value of found peer is `int.MaxValue`, can we send it as a message?
Shouldn't we divide it into chunks?

Libplanet/Net/Protocols/KBucket.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            if (!(exists is null))
+            {
+                _peers.Remove(exists);
+                _peers.Add(peer);
+                return null;
+            }
+            else if (Full())
+            {
+                return Tail;
```
earlbread: The behavior is different from the comment above. Which is right?

Libplanet/Net/Swarm.cs
limebell: Cancellation is moved to PrepareAsync, which must be called before StartAsync.

Libplanet/Net/Swarm.cs
```diff
@@ -499,14 +445,59 @@ private set
...
+        /// not <see cref="Running"/>.</exception>
+        public async Task BootstrapAsync(
+            IEnumerable<Peer> seedPeers,
+            TimeSpan? pingSeedTimeout,
+            TimeSpan? findPeerTimeout,
+            CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (cancellationToken == default(CancellationToken))
+            {
+                cancellationToken = _cancellationToken;
```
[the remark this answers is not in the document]
limebell: If user calls this method without cancellation token, then uses the cancellation token given in PrepareAsync so that BootstrapAsync can be canceled when StopAsync is called.

Libplanet/Net/Swarm.cs
limebell: `peersWithHeight` is sorted with `Distance`, which is  determined by Kademlia protocol. So `.First()` will return closest peer from the caller.

Libplanet/Net/Protocols/Kademlia.cs
```diff
@@ -0,0 +1,59 @@
...
+            var length = 0;
+
+            foreach (var x in distance.ToByteArray())
+            {
+                var mask = 1 << 7;
+                while (length < 160 && mask != 0)
+                {
+                    if ((mask & x) != 0)
+                    {
+                        goto done;
```
earlbread: How about just return `length`?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -0,0 +1,524 @@
...
+
+            try
+            {
+                if (!(await _swarm.SendMessageWithReplyAsync(
+                    target,
+                    new Ping(),
+                    timeout,
+                    cancellationToken) is Pong pong))
+                {
+                    throw new Exception(
```
earlbread: Shouldn't we use the appropriate exception for this?

Libplanet/Net/Swarm.cs
limebell: Actually, `AggregateException` will be unwrapped in this case.
Please take a look at https://stackoverflow.com/questions/12007781/why-doesnt-await-on-task-whenall-throw-an-aggregateexception

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -0,0 +1,191 @@
...
+            _bucketMutex = new AsyncLock();
+        }
+
+        public int Count
+        {
+            get
+            {
+                return _buckets.Sum(bucket => bucket.Count);
+            }
+        }
```
earlbread: How about `public int Count => _buckets.Sum(bucket => bucket.Count);`?
We can also change the properties below.

Libplanet/Net/Swarm.cs
longfin: I'd confused with `.Wait()`.

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                if (bucket.Empty())
+                {
+                    continue;
+                }
+                else
+                {
+                    sorted.AddRange(bucket.Peers);
+                }
+            }
```
earlbread: How about using LINQ like this?
```
var sorted = _buckets
    .Where(b => !b.Empty())
    .SelectMany(b => b.Peers)
    .ToList();
```

Libplanet/Net/Protocols/RoutingTable.cs
limebell: I wanted to add tests using distance, but generating `PublicKey` with specified `Address` was quite picky. I will work on more precise tests later.

Libplanet/Net/Swarm.cs
```diff
@@ -363,62 +272,41 @@ private set
...
 
                     _dealers.Clear();
 
                     Running = false;
                 }
+                else if (!(_protocol is null))
+                {
+                    // FIXME: Should be move to destructor
+                    _router.Dispose();
+                }
```
earlbread: How about calling `_router.Dispose()` outside of this if statement?

Libplanet/Net/Messages/Neighbours.cs
limebell: `Found` is built in `RoutingTable.Neighbours()`, which returns maximum 32 peers.

Libplanet/Net/Protocols/KBucket.cs
limebell: Since this should return oldest one, `Tail` is correct.

CHANGES.md
```diff
@@ -1,6 +1,35 @@
...
+### Backward-incompatible interface changes
+
+ -  `Swarm<T>.AddPeersAsync()` method removed. To connect with seed peers, use
+    `Swarm<T>.BootstrapAsync()` method instead.  [[#353]]
+
+### Added interfaces
+
+ -  Added `Swarm<T>.BootstrapAsync()` method to connect with seed peers.  [[#353]]
+
```
longfin: I think we should introduce `Swarm<T>.PrepareAsync()`.

Libplanet/Net/Swarm.cs
longfin: `.PrepareAsync()` doesn't seem to use `cancellationToken` directly and it just assigns `_workerCacncellationToken` for later. 

so I think taking `cancellationToken` on `.StartAsync()` and `BootStrapAsync()`  is more natural than as-is.

Libplanet/Net/Messages/Neighbours.cs
earlbread: So shouldn't we check it's not over 32 [here](https://github.com/planetarium/libplanet/pull/353/files/6fbcae192e88d7092205041933d5df946e55c936#diff-bdc8d19aacbb2ffb302c9361c70c5b1fR14), too?

Libplanet/Net/Swarm.cs
limebell: It would not guarantee that this swarm is prepared.

Libplanet/Net/Messages/Neighbours.cs
limebell: Actually 32 is a value that might be changed due to protocol setting, so the line you mentioned is just used to prevent mistakes of users.

Libplanet.Tests/Net/SwarmTest.cs
limebell: Yes, it takes more than a minute.

Libplanet.Tests/Net/SwarmTest.cs
limebell: Yes, because swarmB may be only connected to swarmA, than swarmC's `BlockReceived` will never be invoked.

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: A test method to investigate the routing table of the peer.

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: This block should be moved where peer is actually removed.

limebell:
Sorry for late check, I totally missed those comments. Can you please check if I missed any other of your comments? @longfin

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                _buckets[i] = new KBucket(_bucketSize, _random);
+            }
+
+            _bucketMutex = new AsyncLock();
+        }
+
+        public int Count => _buckets.Sum(bucket => bucket.Count);
+
+        public List<KBucket> NonFullBuckets
```
dahlia: Couldn't this be `IEnumerable<KBucket>` instead?

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            get
+            {
+                return _buckets.Where(bucket => !bucket.Full()).ToList();
+            }
+        }
+
+        public List<KBucket> NonEmptyBuckets
```
dahlia: Couldn't this be `IEnumerable<KBucket>` instead?

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -0,0 +1,177 @@
...
+            if (peer is null)
+            {
+                Log.Debug($"Remove Peer null argument exception [{peer}]");
+                throw new ArgumentNullException(nameof(peer));
+            }
+
+            if (peer.Address.Equals(_address))
+            {
+                throw new ArgumentException(
+                    $"Cannot remove self from routing table");
```
dahlia:
```suggestion
                    "Cannot remove self from routing table.");
```

Libplanet/Net/Messages/FindPeer.cs
```diff
@@ -0,0 +1,30 @@
...
+    internal class FindPeer : Message
+    {
+        public FindPeer(Address target)
+        {
+            Target = target;
+        }
+
+        public FindPeer(NetMQFrame[] body)
+        {
+            Target = new Address(body[0].ToByteArray());
```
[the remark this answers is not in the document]
limebell: Do you mean by this?
```
Target = new Address(body[0].Buffer);
```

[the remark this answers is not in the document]
dahlia: Yeah, but it turns out `.ToByteArray()` has `copy` option and it's `false` by default.  Don't have to change.

Libplanet/Net/Swarm.cs
```diff
@@ -191,12 +191,16 @@ static Swarm()
 
         public Address Address => _privateKey.PublicKey.ToAddress();
 
-        public Peer AsPeer =>
-            new Peer(
+        public Peer AsPeer => EndPoint is null
+            ? new UnboundPeer(
+                _privateKey.PublicKey,
+                _appProtocolVersion,
+                _publicIPAddress) as Peer
```
longfin: Is this casting(`as Peer`) necessary?

Libplanet/Net/Swarm.cs
```diff
@@ -1298,17 +1387,11 @@ private void BroadcastTxIds(IEnumerable<TxId> txIds)
...
         {
-            if (!(message.Sender is Address from))
-            {
-                throw new NullReferenceException(
-                    "BlockHashes doesn't have sender address.");
-            }
-
-            Peer peer = _peers.Keys.FirstOrDefault(p => p.Address.Equals(from));
-            if (peer == null)
+            if (!(message.Remote is BoundPeer peer))
```
```suggestion
            if (message.Remote is UnBoundPeer)
```

longfin: Or, can `message.Remote` be `null`?

Libplanet/Net/Swarm.cs
```diff
@@ -1710,365 +1792,97 @@ private void TransferRecentStates(GetRecentStates getRecentStates)
...
-        {
-            DateTimeOffset timestamp = delta.Timestamp;
-
-            foreach (Peer peer in added)
-            {
-                _logger.Debug($"{timestamp} {delta.Sender} > +{peer}");
-            }
-
-            foreach (Peer peer in existing)
+            if (_turnClient != null)
```
```suggestion
            if (!(_turnClient is null))
```

Libplanet/Net/Swarm.cs
```diff
@@ -2095,15 +1913,16 @@ private void ReceiveMessage(object sender, NetMQSocketEventArgs e)
                         }
                         catch (Exception exc)
                         {
-                            _logger.Error(
-                                exc,
-                                $"Something went wrong during message parsing: {exc}"
-                            );
+                            _logger.Error("Something went wrong during message parsing: {0}", exc);
```
```suggestion
                            _logger.Error(exc, "Something went wrong during message parsing: {0}", exc);
```

Libplanet/Net/Swarm.cs
```diff
@@ -2095,15 +1913,16 @@ private void ReceiveMessage(object sender, NetMQSocketEventArgs e)
...
                             throw;
                         }
                     },
                     _cancellationToken);
             }
+            catch (DifferentAppProtocolVersionException)
+            {
+                _logger.Debug("Ignore message from peer with different version.");
```
longfin: How about logging peer's info?

Libplanet/Net/Swarm.cs
limebell: Without this, ternary operator causes error.

Libplanet/Net/Swarm.cs
limebell: Yes, it cannot be null in formal cases, but in case of malicious peer trying to send invalid messages, it can be null.

CHANGES.md
```diff
@@ -28,11 +28,25 @@ To be released.
...
+    `Swarm<T>.StartAsync()`.  [[#353]]
+ -  Added `Swarm<T>.BootstrapAsync()` method to connect with seed peers.  [[#353]]
+
 ### Behavioral changes
 
+ -  NetMQ instances are now initialized at `Swarm<T>.StartAsync()` instead of
+    `Swarm<T>()`.  [[#353]]
+ -  Peers now connected via [Kademlia protocol][Kademlia]. Peers are now selectively
+    connected to each peer.  [[#353]]
+ -  TxIds and Blocks are now broadcasted to selected peers from routing table of
```
```suggestion
 -  `TxId`s and `Block`s are now broadcasted to selected peers from routing table of
```

CHANGES.md
```diff
@@ -63,6 +77,10 @@ Released on August 28, 2019.
 [#468]: https://github.com/planetarium/libplanet/pull/468
 
 
+[#353]: https://github.com/planetarium/libplanet/pull/353
+[Kademlia]: https://en.wikipedia.org/wiki/Kademlia
+
+
```
dahlia: It seems a wrong place.

Libplanet/Net/BoundPeer.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Net;
+using System.Runtime.Serialization;
+using Libplanet.Crypto;
+using Libplanet.Serialization;
+
+namespace Libplanet.Net
+{
+    [Serializable]
+    [Equals]
+    public class BoundPeer : Peer
```
dahlia: It would be better if it's a `sealed class`.

Libplanet/Net/UnboundPeer.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Net;
+using System.Runtime.Serialization;
+using Libplanet.Crypto;
+using Libplanet.Serialization;
+
+namespace Libplanet.Net
+{
+    [Serializable]
+    [Equals]
+    public class UnboundPeer : Peer
```
dahlia: It would be better if it's a `sealed class`.

Libplanet/Net/BoundPeer.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Peer"/> class.
```
dahlia:
```suggestion
        /// Initializes a new instance of the <see cref="BoundPeer"/> class.
```

Libplanet/Net/Peer.cs
```diff
@@ -13,89 +12,40 @@ namespace Libplanet.Net
...
         [Pure]
-        public PublicKey PublicKey { get; }
+        public PublicKey PublicKey { get; protected set; }
 
         /// <summary>
         /// The corresponding <see cref="DnsEndPoint"/> of this peer.
         /// </summary>
         [Pure]
-        public DnsEndPoint EndPoint { get; }
+        public DnsEndPoint EndPoint { get; protected set; }
```
dahlia: Should this abstract class have this property?

Libplanet/Net/UnboundPeer.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        internal UnboundPeer(
+            PublicKey publicKey,
+            int appProtocolVersion,
+            IPAddress publicIPAddress)
+        : base(publicKey, null, appProtocolVersion, publicIPAddress)
+        {
+            PublicKey = publicKey ??
+                        throw new ArgumentNullException(nameof(publicKey));
+            EndPoint = null;
```
dahlia: IMHO we'd better get rid of this property from the abstract class `Peer`, and let the only `BoundPeer` have this property instead.

longfin:
It may be quite trivial, but I believe it would be better if `Peer` remains abstract for distinguishing `UnboundPeer` and `BoundPeer` as below.

```
UnboundPeer unboundPeer;
BoundPeer boundPeer;

Assert.True(unboundPeer is Peer);
Assert.True(boundPeer is Peer);

Assert.False(unboundPeer is BoundPeer);
Assert.False(boundPeer is UnboundPeer);
```

PR #359, opened 2019-07-22
Add rwlock for state reference

earlbread:
This adds rwlock for state reference.


earlbread:
I changed the lock of `GetEnumerator()` method to `UpgradeableReadLock` to get write lock.

PR #362, opened 2019-07-22
Re-stage transactions when receiving already unstaged

longfin:
This PR makes `Swarm<T>` to re-stage transactions when receiving already unstaged.

PR #364, opened 2019-07-22
Fixed infinte echo during ProcessTxIds()

longfin:
This PR fixes a bug that makes infinite broadcasting during `Swarm<T>.ProcessTxIds()`. (occurred by #362).


longfin:
It will be replaced by #365.

PR #369, opened 2019-07-24
Fix a bug when serializing default address

longfin:
This PR fixes a bug that `NullReferenceException` occurred when serializing default `Address`.


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -272,6 +272,7 @@ IEnumerator IEnumerable.GetEnumerator()
                                     ).GetValueOrDefault(a)
                                 ).ToArray();
 
+                            var l = new List<int>();
```
dahlia: This seems unused?

longfin: Removed.

PR #374, opened 2019-07-29
Don't execute actions in already appended blocks again during preloading

dahlia:


Libplanet/Net/Swarm.cs
```diff
@@ -619,8 +621,16 @@ public void BroadcastTxs(IEnumerable<Transaction<T>> txs)
 
             if (!received)
             {
-                foreach (Block<T> block in _blockChain)
+                long initHeight = initialTip is null || _blockChain[initialTip.Index] != initialTip
+                    ? 0
+                    : initialTip.Index;
+                foreach (Block<T> block in _blockChain.Skip((int)initHeight))
                 {
+                    if (block.Index < initHeight)
```
longfin: it looks like little redundant. can't we assume this with only `_blockChain.Skip()`?

dahlia: It's indeed redundant for most cases, but when `initHeight` is out of 32-bit integer this guard becomes necessary. Please `.Skip()` is a small optimization on this line.

PR #379, opened 2019-07-30
Skip sync recent states If it has already been evaluated

earlbread:
This makes `PreloadAsync()` skip state synchronization if the most recent state has already been evaluated.

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -957,6 +957,47 @@ namespace Libplanet.Tests.Net
         }
 
         [Fact(Timeout = Timeout)]
+        public async Task InitialBlockDownloadStates()
+        {
+            Swarm<DumbAction> minerSwarm = _swarms[0];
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -40,8 +40,10 @@ To be released.
  -  `BlockChain<T>.MineBlock()` and `BlockChain<T>.GetNextTxNonce()` methods
     became to ignore transactions that didn't follow `Transaction<T>.Nonce`
     sequentially and treat them as pendings.  [[#365]]
- - `BlockChain<T>` became to evaluate `IBlockPolicy<T>.BlockAction` and set the
-   state when a block is appended to the chain.  [[#319], [#367]]
+ -  `BlockChain<T>` became to evaluate `IBlockPolicy<T>.BlockAction` and set the
+    state when a block is appended to the chain.  [[#319], [#367]]
+ -  `PreloadAsync()` became to skip state synchronization from
+    `trustedStateValidators` if `render` is true.
```
dahlia: This change should not be logged because `trustedStateValidators` option will be introduced 0.5.

PR #382, opened 2019-07-31
Use our own fork of NetMQ

dahlia:
In order to fix <https://github.com/zeromq/netmq/issues/806> (see also @earlbread's PR: <https://github.com/zeromq/netmq/pull/808>), I made [our own fork of NetMQ][1] and [released it][2] (see also [NuGet][3]).  We can roll back this to the upstream when a new release is made in the upstream.

[1]: https://github.com/planetarium/netmq
[2]: https://github.com/planetarium/netmq/releases/tag/4.0.0.253-planetarium
[3]: https://www.nuget.org/packages/Planetarium.NetMQ/4.0.0.253-planetarium

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -91,10 +91,7 @@ namespace Libplanet.Tests.Net
                 s.StopAsync().Wait(DisposeTimeout);
             }
 
-            if (!(Type.GetType("Mono.Runtime") is null))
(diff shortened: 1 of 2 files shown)


dahlia:
The reason why the build on .NET Framework + Windows failed was the assembly was signed with a wrong key.  I [turned off assembly signing][1], and now it seems all builds pass.

[1]: 8ff8e33c211694bf57441bb8141caf28b79ab21b

Issue #395, opened 2019-08-05
LiteDBStore.StoreStateReference() appends entries in an inefficient way


dahlia:
The current implementation of `LiteDBStore.StoreStateReference()` method reads all the existing set and then appends a new entry to the set in memory, and then replaces the entire set with a new entire set. This should be replaced by an efficient append operation.

PR #398, opened 2019-08-05
More efficient way to store state references

dahlia:
This fixes <https://github.com/planetarium/libplanet/issues/395>.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,8 @@ To be released.
  -  `BlockSet<T>.ContainsKey()` and `TransactionSet<T>.ContainsKey()` methods
     became O(1) time complexity through omitting iteration and relying
     own retrieve implementations.  [[#390]]
+ -  The way `LiteDBStore` stores state references became efficient,
+    but the file-level backward compatibility was also broken.  [[#395], [#398]]
 
 ### Bug fixes
 
@@ -68,6 +70,8 @@ To be released.
 [#387]: https://github.com/planetarium/libplanet/pull/387
 [#389]: https://github.com/planetarium/libplanet/pull/389
 [#390]: https://github.com/planetarium/libplanet/pull/390
+[#395]: https://github.com/planetarium/libplanet/issues/395
+[#398]: https://github.com/planetarium/libplanet/pull/398
 [LiteDB #1268]: https://github.com/mbdavid/LiteDB/issues/1268
 
 
--- a/Libplanet.Tests/Store/LiteDBStoreTest.cs
+++ b/Libplanet.Tests/Store/LiteDBStoreTest.cs
@@ -1,4 +1,8 @@
 using System;
+using System.Security.Cryptography;
+using Libplanet.Crypto;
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -46,6 +46,8 @@ To be released.
  -  `BlockSet<T>.ContainsKey()` and `TransactionSet<T>.ContainsKey()` methods
     became O(1) time complexity through omitting iteration and relying
     own retrieve implementations.  [[#390]]
+ -  The way `LiteDBStore` stores state references became efficient,
+    but the file-level backward compatibilty was also broken.  [[#395], [#398]]
```
```suggestion
    but the file-level backward compatibility was also broken.  [[#395], [#398]]
```

Libplanet.Tests/Store/LiteDBStoreTest.cs
```diff
@@ -15,5 +19,28 @@ public void Dispose()
         {
             _fx?.Dispose();
         }
+
+        public void StateRefDocBlockHash()
```
earlbread:
```suggestion
        [Fact]
        public void StateRefDocBlockHash()
```

Libplanet/Store/LiteDBStore.cs
```diff
@@ -572,6 +461,13 @@ private string BlockStateFileId(HashDigest<SHA256> blockHash)
             return $"state/{blockHash}";
         }
 
+        private string StateRefId(string @namespace)
+        {
+            string ns =
+                ByteUtil.Hex(namespaceHasher.ComputeHash(Encoding.UTF8.GetBytes(@namespace)));
```
earlbread: Is there a reason to use the hash of namespace instead of the namespace?

Libplanet/Store/LiteDBStore.cs
```diff
@@ -424,59 +353,19 @@ public override AddressStateMap GetBlockStates(HashDigest<SHA256> blockHash)
             Block<T> branchPoint,
             IImmutableSet<Address> addressesToStrip)
         {
-            long branchPointIndex = branchPoint.Index;
-            List<LiteFileInfo> files =
-                _db.FileStorage
-                    .Find($"{StateRefIdPrefix}{srcNamespace}")
-                    .ToList();
+            string srcCollId = StateRefId(@srcNamespace);
+            string dstCollId = StateRefId(@destNamespace);
```
earlbread: I think we can omit the `@` prefix. Is this intended?

dahlia: No, it's not intended.  I'll remove them.

Libplanet/Store/LiteDBStore.cs
dahlia: Because [LiteDB's collections names allow only limited set of characters][1], and on `IStore`-level namespaces are just free text strings.  IMHO, namespaces should be rather `Guid` than `string`.

[1]: https://github.com/mbdavid/LiteDB/wiki/Collections

PR #406, opened 2019-08-07
Change total block count when received count is bigger

earlbread:
Currently, received count is displayed bigger than total block count if seed node mine another block while IBD. This patch changes the total block count when that situation happens.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -1364,6 +1364,8 @@ namespace Libplanet.Net
                     $"(tip: {blockChain.Tip?.Hash})"
                 );
 
+                totalBlockCount = Math.Max(totalBlockCount, receivedBlockCount + hashCount);
+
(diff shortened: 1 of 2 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -1364,6 +1364,11 @@ CancellationToken cancellationToken
                     $"(tip: {blockChain.Tip?.Hash})"
                 );
 
+                if (receivedBlockCount + hashCount > totalBlockCount)
```
longfin: I'd prefer `Math.max()` in these cases. (but it isn't a strong opinion.)

```
totalBlockCount = Math.Max(totalBlockCount, receivedBlockCount + hashCount);
```

earlbread: I changed it as your suggestion.

dahlia:
Could we have a unit test for this case?

PR #411, opened 2019-08-07
Read–write lock on TransactionSet

dahlia:
Since `IStore` implementations could have some problems with concurrent `PutTransaction<T>()` calls, I believe it's way safer to have read–write lock on `TransactionSet` interface.


longfin:
@dahlia it seems obsoleted by #413. right?

dahlia:
Yes, as we prefer the approach made in #413 over this, I will close this.

PR #414, opened 2019-08-07
Fix bug that NetworkStreamProxy.StartAsync() never stops

longfin:
This PR fixes a bug that `NetworkStreamProxy.StartAsync()` never stops even if the connection had disconnected by a remote peer.


longfin:
I rebased and added two commits.

- Replaced meaningless message content with frame count.
- Restore `_poller.Dispose()` in `Swarm<T>.StopAsync()`

PR #420, opened 2019-08-09
Boost PreloadAsync()

longfin:
This patch enhances the performance of `Swarm.PreloadAsync()` using two tricks as below.

- Instead of using `RecentStates.StateReferences` directly, `PreloadAsync` became to translate this message as `Dictionary<HashDigest<SHA256>, ISet<Address>>` to reduce file I/O.
  - To accomplish this, `IStore.StoreStateReference` became to take `blockHash` and `blockIndex`, not whole `Block`.
- `PreloadAsync()` are currently fork its chain to prevent corruption by unexpected termination. this patch reduces its file I/O by `IStore.ForkIndexes()`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,8 @@ To be released.
     method became replaced by
     `StoreStateReference(string, IImmutableSet<Address>, HashDigest<SHA256>, long)`
     method so that it takes hash and index of a block instead of an entire
-    block.
+    block.  [[#420]]
+ -  Added `IStore.ForkBlockIndexes()` method.  [[#420]]
 
 ### Added interfaces
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,12 @@ To be released.
 
 ### Backward-incompatible interface changes
 
+ -  `IStore.StoreStateReference<T>(string, IImmutableSet<Address>, Block<T>)`
+    method became replaced by
+    `StoreStateReference(string, IImmutableSet<Address>, HashDigest<SHA256>, long)`
+    method so that it takes hash and index of a block instead of an entire
+    block.
+
 ### Added interfaces
 
 ### Behavioral changes
@@ -17,6 +23,7 @@ To be released.
  -  Fixed a bug that `Swarm<T>` hadn't released its TURN releated resources on
     `Swarm<T>.StopAsync()`.  [[#450]]
 
+[#420]: https://github.com/planetarium/libplanet/pull/420
 [#450]: https://github.com/planetarium/libplanet/pull/450
 
 
--- a/Libplanet/Store/IStore.cs
+++ b/Libplanet/Store/IStore.cs
@@ -54,6 +54,26 @@ namespace Libplanet.Store
         bool DeleteIndex(string @namespace, HashDigest<SHA256> hash);
 
         /// <summary>
+        /// Forks block indexes from
+        /// <paramref name="sourceNamespace"/> to
+        /// <paramref name="destinationNamespace"/>.
+        /// </summary>
+        /// <param name="sourceNamespace">The namespace of block indexes to
+        /// fork.</param>
+        /// <param name="destinationNamespace">The namespace of destination
+        /// block indexes.</param>
+        /// <param name="branchPoint">The branch point <see cref="Block{T}"/>
+        /// to fork.</param>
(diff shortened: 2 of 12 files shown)


longfin:
I've rebased. PTAL.

longfin:
I've rebased to append it to 0.6.0.

CHANGES.md
```diff
@@ -17,9 +22,11 @@ To be released.
  -  Fixed a bug that `Swarm<T>` hadn't released its TURN releated resources on
     `Swarm<T>.StopAsync()`.  [[#450]]
 
+[#420]: https://github.com/planetarium/libplanet/pull/420
 [#450]: https://github.com/planetarium/libplanet/pull/450
 
 
+
```
dahlia: This blank line seems unintended?

CHANGES.md
```diff
@@ -8,6 +8,11 @@ To be released.
 
 ### Backward-incompatible interface changes
 
+ -  `IStore.StoreStateReference<T>()` became to take the hash and index of the
+    block instead of the block itself.  [[#420]]
+     -  Therefore, it also became to lose type parameter `T`.
```

> `IStore.StoreStateReference<T>(string, IImmutableSet<Address>, Block<T>)` method became replaced by `StoreStateReference(string, IImmutableSet<Address>, HashDigest<SHA256>, long)` method so that it takes hash and index of a block instead of an entire `Block` object.

Libplanet/Store/IStore.cs
```diff
@@ -193,25 +215,23 @@ AddressStateMap states
...
         /// <see cref="Address"/>es by the <see cref="Transaction{T}"/>s in the
-        /// <paramref name="block"/>.</summary>
+        /// <see cref="Block{T}" />.</summary>
         /// <param name="namespace">The namespace to store a state reference.
         /// </param>
         /// <param name="addresses">The <see cref="Address"/>es to store state
         /// reference.</param>
-        /// <param name="block">The <see cref="Block{T}"/> which has the state
+        /// <param name="blockHash">The <see cref="Block{T}.Hash"/> which has the state
+        /// of the <see cref="Address"/>.</param>
```
dahlia:
```suggestion
        /// of the <see cref="Address"/>.  This must refer to the same block
        /// that <paramref name="blockIndex"/> refers to.</param>
```

Libplanet/Store/IStore.cs
```diff
@@ -53,6 +53,26 @@ public interface IStore
 
         bool DeleteIndex(string @namespace, HashDigest<SHA256> hash);
 
+        /// <summary>
+        /// Forks block indexes from
```
dahlia: Does the term “block indexes” mean a sequence (chain) of block hashes and its index numbers?  Or does it include state references?  If it's latter the plural term seems natural for me as well, but if it's former I believe the singular term (i.e., “block index”) would listen more natural IMHO (as like the way other related methods are named, e.g., `AppendIndex`, `DeleteIndex`, `IterateIndex`).

longfin: It's former. but I've confused about that because….

- `AppendIndex()` and `DeleteIndex()` take only one argument. 
- Other iteration method use plural form. (e.g. `IterateBlockHashes()`, `IterateStateReferences()`)

so, How about keep `ForkIndexes()` in this PR and rename `IterateIndex()` to `IterateIndexes()` in other PR?

CHANGES.md
longfin: Nope, I'll remove it.

CHANGES.md
longfin: 👌

Libplanet/Store/IStore.cs
dahlia: IMO in the case of this a singular noun “index” is more natural than a plural noun “indexes”/“indices,” in the same manner we say “a set of blocks” instead of “sets of blocks” to mean `ISet[Block]` in verbal language.

dahlia: I thought about this terminology once more, and it turns out the plural term has its own advantage, and I agree your suggestion to rename these method names to use the plural term “indexes” instead of the singular “index.”  Some reasons:

- Although relational databases tend to say “an index” for values in a column, we also say “indices” in books.
- We already name a block's height by `Block.Index`, so it's natural to say a collection of `Block.Index`es as the plural “indexes.”

PR #425, opened 2019-08-12
Fix slow FindNextHashes

earlbread:
This fixes a problem where `BlockChain.FinxNextHashes()` method is slow when there are many blocks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@ To be released.
  -  `BaseIndex.ContainsKey()` method became `abstract`.  [[#390]]
  -  `BlockDownloadState.TotalBlockCount` and `BlockDownloadState.ReceivedBlockCount`
     became to `Int64` type.  [[#396], [#399]]
+ -  `IStore.IterateIndex()` method became to receive `offset` and `limit`
+    parameters.  [[#425]]
  -  Added `IStore.GetCanonicalNamespace()` method.  [[#426]]
  -  Added `IStore.SetCanonicalNamespace()` method.  [[#426]]
  -  Removed `IRandom.NextDouble()` method, because [floating-point arithmetics,
--- a/Libplanet.Tests/Store/StoreTracker.cs
+++ b/Libplanet.Tests/Store/StoreTracker.cs
@@ -118,10 +118,13 @@ namespace Libplanet.Tests.Store
             return _store.IterateBlockHashes();
         }
 
-        public IEnumerable<HashDigest<SHA256>> IterateIndex(string @namespace)
+        public IEnumerable<HashDigest<SHA256>> IterateIndex(
+            string @namespace,
+            int offset,
+            int? limit)
         {
             _logs.Add((nameof(IterateIndex), @namespace, null));
-            return _store.IterateIndex(@namespace);
+            return _store.IterateIndex(@namespace, offset, limit);
         }
 
         public IEnumerable<TxId> IterateStagedTransactionIds(bool toBroadcast)
--- a/Libplanet.Tests/Store/StoreTracker.cs
+++ b/Libplanet.Tests/Store/StoreTracker.cs
@@ -123,8 +123,8 @@ namespace Libplanet.Tests.Store
             int offset,
             int? limit)
         {
-            _logs.Add((nameof(IterateIndex), @namespace, null));
-            return _store.IterateIndex(@namespace, offset, limit);
+             _logs.Add((nameof(IterateIndex), @namespace, (offset, limit)));
+             return _store.IterateIndex(@namespace, offset, limit);
         }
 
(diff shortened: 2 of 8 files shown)


Libplanet/Store/IStore.cs
```diff
@@ -38,9 +38,14 @@ public interface IStore
...
         /// <param name="namespace">The namespace of the index that contains block hashes to
         /// iterate.</param>
+        /// <param name="skip">The number of block hashes to skip.</param>
+        /// <param name="limit">The maximum number of block hashes to get.</param>
         /// <returns>Block hashes in the index of the <paramref name="namespace"/>, in ascending
         /// order; the genesis block goes first, and the tip block goes last.</returns>
-        IEnumerable<HashDigest<SHA256>> IterateIndex(string @namespace);
+        IEnumerable<HashDigest<SHA256>> IterateIndex(
+            string @namespace,
+            int skip = 0,
```
dahlia: How about naming it `offset` instead of `skip`?

CHANGES.md
```diff
@@ -20,6 +20,8 @@ To be released.
  -  `BaseIndex.ContainsKey()` method became `abstract`.  [[#390]]
  -  `BlockDownloadState.TotalBlockCount` and `BlockDownloadState.ReceivedBlockCount`
     became to `Int64` type.  [[#396], [#399]]
+ -  `IStore.IterateIndex()` method became to receive `skip` and `limit`
+    parameter.  [[#425]]
```
```suggestion
    parameters.  [[#425]]
```

Libplanet/Store/IStore.cs
```diff
@@ -38,9 +38,14 @@ public interface IStore
...
         /// iterate.</param>
+        /// <param name="offset">The starting point to return block hashes.</param>
+        /// <param name="limit">The maximum number of block hashes to get.</param>
         /// <returns>Block hashes in the index of the <paramref name="namespace"/>, in ascending
         /// order; the genesis block goes first, and the tip block goes last.</returns>
-        IEnumerable<HashDigest<SHA256>> IterateIndex(string @namespace);
+        IEnumerable<HashDigest<SHA256>> IterateIndex(
+            string @namespace,
+            int offset = 0,
+            int limit = int.MaxValue);
```
dahlia: Defaulting it to `int.MaxValue` seems not that robust to me.  How about making it required, or changing its type to `int?` and defaulting it to `null`?

earlbread: Since the type is `int`, changing the default doesn't seem to make much difference. Or should it be a `long` type?

dahlia: My concern was that the current way is difficult to distinguish the intention of “unlimiting” from “limiting up to a kind of very large number, which is the maximum possible value of 32-bit integer.”  Also I thought it's a C-ish trick to me.

In the current way, if we're willing to make a store backend on SQLite, we need to determine if it's intended to be unlimited, by comparing it to `int.MaxValue`, in order to avoid SQL like `OFFSET 0 LIMIT 2147483647` and simply put `OFFSET 0`, which differ each other.

My suggestion is simple: how about `int? limit = null`?

earlbread: Ok, got it. Fixed!

Libplanet.Tests/Store/StoreTracker.cs
```diff
@@ -118,10 +118,13 @@ public IEnumerable<HashDigest<SHA256>> IterateBlockHashes()
             return _store.IterateBlockHashes();
         }
 
-        public IEnumerable<HashDigest<SHA256>> IterateIndex(string @namespace)
+        public IEnumerable<HashDigest<SHA256>> IterateIndex(
+            string @namespace,
+            int offset,
+            int? limit)
         {
             _logs.Add((nameof(IterateIndex), @namespace, null));
```
dahlia: Shouldn't this be changed either?  Although slots are not enough…

```suggestion
             _logs.Add((nameof(IterateIndex), @namespace, (offset, limit)));
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -736,39 +736,15 @@ internal HashDigest<SHA256> FindBranchPoint(BlockLocator locator)
...
-                IEnumerable<HashDigest<SHA256>> indexes = Store.IterateIndex(
-                    Id.ToString()
-                );
-
-                ImmutableArray<HashDigest<SHA256>> hashes = locator.ToImmutableArray();
-                var hashesIndex = hashes.Length - 1;
-
-                // Supposes the indexes are sorted in the order of the chain.
-                foreach (HashDigest<SHA256> index in indexes)
+                foreach (var hash in locator)
```
```suggestion
                foreach (HashDigest<SHA256> hash in locator)
```

PR #428, opened 2019-08-13
Fix receivedBlockCount to handle retries

earlbread:
This fixes a problem where `receivedBlockCount` is reset when an exception occurs during `FillBlocksAsync`.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -1378,18 +1378,24 @@ namespace Libplanet.Net
             _logger.Debug("Trying to fill up previous blocks...");
 
             int retry = 3;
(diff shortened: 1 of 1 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -1379,7 +1385,7 @@ CancellationToken cancellationToken
                         {
                             _logger.Error(
                                 e,
-                                "FillBlockAsync() failed. retrying..."
+                                $"{nameof(FillBlocksAsync)}() failed: {e}\nRetrying..."
```
dahlia: As usually `e` takes lines long, this would be more readable:

```suggestion
                                $"{nameof(FillBlocksAsync)}() failed; retrying...: {e}"
```

PR #433, opened 2019-08-13
Fixed a FileStoreFixture bug

### LINKED ISSUE #432: FileStoreFixture cause IOException during unit test
In windows, if file is created at `FileStoreFixture.Dispose()`'s target directory while deleting recursively, test will fail with `IOException`.

limebell:
This PR resolves #432

### DIFF
--- a/Libplanet.Tests/Store/FileStoreFixture.cs
+++ b/Libplanet.Tests/Store/FileStoreFixture.cs
@@ -1,6 +1,8 @@
 using System;
 using System.IO;
+using System.Threading;
 using Libplanet.Store;
+using Serilog;
 
 namespace Libplanet.Tests.Store
 {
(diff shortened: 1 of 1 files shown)


Libplanet.Tests/Store/FileStoreFixture.cs
```diff
@@ -18,9 +20,18 @@ public FileStoreFixture()
...
+                try
+                {
+                    Directory.Delete(Path, true);
+                }
+                catch (IOException)
+                {
+                    Log.Debug("IOException occurred during dispose()");
+                }
+
+                Thread.Sleep(0);
```
dahlia: Is this line necessary?

earlbread:
[changelog skip]  is needed.

Libplanet.Tests/Store/FileStoreFixture.cs
limebell: This line causes context switching, which is needed to prevent exceptions.

PR #434, opened 2019-08-13
Add readOnly option to LiteDBStore

longfin:


dahlia:
Is it need for running an explorer instance?

longfin:
> Is it need for running an explorer instance?

Yes, it's intended to be used by read-only instance like block explorer.

PR #438, opened 2019-08-15
Sanity check in PrivateKey constructor using bytes

### LINKED ISSUE #358: Sanity check on new PrivateKey()
Currently, there is no validation of `privateKey` parameter on `Libplanet.Crypto.PrivateKey`'s constructor. so illegal bytes can be passed to `new PrivateKey()` and occurs error when being used. it would be good if we catch this error in the constructor.

ipdae:
resolved #358


CHANGES.md
```diff
@@ -82,6 +82,7 @@ To be released.
     the `Transaction<T>` more than once.  [[#413]]
  -  `BlockChain<T>.Swap()` became to omit common block finding when `render` is
     `false`.  [[#423]]
+ - `PrivateKey(bytes)` became to valid check. [[#438]]
```
```suggestion
 -  `PrivateKey(bytes)` became to valid check.  [[#438]]
```

CHANGES.md
```diff
@@ -82,6 +82,7 @@ To be released.
     the `Transaction<T>` more than once.  [[#413]]
  -  `BlockChain<T>.Swap()` became to omit common block finding when `render` is
     `false`.  [[#423]]
+ -  `PrivateKey(bytes)` became to valid check.  [[#438]]
```
```suggestion
 -  `PrivateKey(byte[])` constructor became to check validity.  [[#438]]
```

Libplanet/Crypto/PrivateKey.cs
```diff
@@ -269,6 +265,29 @@ private static ECPrivateKeyParameters GenerateKeyParam()
...
+            );
+
+            var key = new PrivateKey(param);
+            try
+            {
+                var publicKey = key.PublicKey;
+            }
+            catch (ArgumentException)
+            {
+                throw new InvalidOperationException(
```
dahlia: I believe `ArgumentException` would be more suitable here.

```suggestion
                throw new ArgumentException(
```

Libplanet/Crypto/PrivateKey.cs
```diff
@@ -269,6 +265,19 @@ private static ECPrivateKeyParameters GenerateKeyParam()
...
+        private static ECPrivateKeyParameters GenerateKeyFromBytes(byte[] privateKey)
+        {
+            var param = new ECPrivateKeyParameters(
+                "ECDSA",
+                new BigInteger(1, privateKey),
+                GetECParameters()
+            );
+
+            var key = new PrivateKey(param);
+            var publicKey = key.PublicKey;
```
longfin: This statement also seems unnatural, so it would be better if there is some comment about it.

longfin:
```suggestion
            var _ = key.PublicKey;
```

PR #439, opened 2019-08-15
Implement deterministic Guid

### LINKED ISSUE #371: Deterministic UUID (System.Guid) generator for distributed actions
Since actions are executed in a distributed network, these have to be deterministic so that they have the same result regardless of node.  Therefore, we've provided APIs like [`IActionContext.Random`][1] to replace .NET's `System.Random`.  (Disclaimer: Saying random numbers should be deterministic here could sound contradictory, but it technically means “random numbers should be unpredictable until a block is mined, and they should not change and be consistent between all nodes in a distributed network after a block is once mined.”)

However, for UUIDs (`System.Guid` values) we currently provide no correspondence to `IActionContext.Random` for this.  Although library users could implement their own UUID generator built on `IActionContext.Random`, it's not only unhandy but also leads to make a mistake that a library user thoughtlessly use the `System.Guid.NewGuid()` method.

So I suggest to have an API to generate a `System.Guid` value in the same way `IActionContext.Random` generates a random number!

See also: https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Guid.Unix.cs#L12-L35

[1]: https://docs.libplanet.io/0.4.1/api/Libplanet.Action.IActionContext.html#Libplanet_Action_IActionContext_Random

ipdae:
resolved #371

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,7 @@ To be released.
  -  Removed `IRandom.NextDouble()` method, because [floating-point arithmetics,
     which is underspecified, likely introduce
     indeterminism][floating-point determinism].  [[#410], [#419]]
+ -  Added `IActionContext.Guid` property.  [[#371], [#439]]
 
 ### Added interfaces
 
@@ -118,6 +119,7 @@ To be released.
 [#366]: https://github.com/planetarium/libplanet/pull/366
 [#367]: https://github.com/planetarium/libplanet/pull/367
 [#369]: https://github.com/planetarium/libplanet/pull/369
+[#371]: https://github.com/planetarium/libplanet/issues/371
 [#375]: https://github.com/planetarium/libplanet/pull/375
 [#377]: https://github.com/planetarium/libplanet/issues/377
 [#378]: https://github.com/planetarium/libplanet/pull/378
@@ -146,6 +148,7 @@ To be released.
 [#426]: https://github.com/planetarium/libplanet/pull/426
 [#434]: https://github.com/planetarium/libplanet/pull/434
 [#438]: https://github.com/planetarium/libplanet/pull/438
+[#439]: https://github.com/planetarium/libplanet/pull/439
 [LiteDB #1268]: https://github.com/mbdavid/LiteDB/issues/1268
 [floating-point determinism]: https://wp.me/p1fTCO-kT
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,7 +29,7 @@ To be released.
  -  Removed `IRandom.NextDouble()` method, because [floating-point arithmetics,
     which is underspecified, likely introduce
     indeterminism][floating-point determinism].  [[#410], [#419]]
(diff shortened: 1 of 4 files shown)


Libplanet.Tests/Action/ActionContextTest.cs
```diff
@@ -29,6 +30,28 @@ public void RandomShouldBeDeterministic()
...
         }
 
+        [Fact]
+        public void GuidShouldBeDeterministic()
+        {
+            (int, Guid)[] testCases =
+            {
+                (0, new Guid("6f460c1a-755d-d8e4-ad67-65d5f519dbc8")),
+                (1, new Guid("8286d046-9740-a3e4-95cf-ff46699c73c4")),
+            };
```
earlbread: How about using `InlineData`?

Libplanet/Action/ActionContext.cs
```diff
@@ -30,5 +32,15 @@ internal class ActionContext : IActionContext
         public IAccountStateDelta PreviousStates { get; }
 
         public IRandom Random { get; }
+
+        public Guid Guid
+        {
+            get
+            {
+                var b = new byte[16];
```
longfin: I think that we should create GUID based [rfc4122](https://www.ietf.org/rfc/rfc4122.txt) like [dotnet-core](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Guid.Unix.cs#L12-L35).

Libplanet/Action/IActionContext.cs
```diff
@@ -53,5 +55,7 @@ public interface IActionContext
         /// <returns>A random object that shares interface mostly equivalent
         /// to <see cref="System.Random"/>.</returns>
         IRandom Random { get; }
+
+        Guid Guid { get; }
```
earlbread: Could you please add documentation?

Libplanet/Action/ActionContext.cs
ipdae: I think implementing rfc4122 takes a lot of time, so it's hard to apply it in this PR.

Libplanet/Action/IActionContext.cs
```diff
@@ -53,5 +55,14 @@ public interface IActionContext
         /// <returns>A random object that shares interface mostly equivalent
         /// to <see cref="System.Random"/>.</returns>
         IRandom Random { get; }
+
+        /// <summary>
+        /// Creates new Guid.  Its seed (state)
```
longfin:
```suggestion
        /// Creates new Guid. Its seed (state)
```

CHANGES.md
```diff
@@ -29,6 +29,7 @@ To be released.
  -  Removed `IRandom.NextDouble()` method, because [floating-point arithmetics,
     which is underspecified, likely introduce
     indeterminism][floating-point determinism].  [[#410], [#419]]
+ -  Added `IActionContext.Guid` property.  [[#371], [#439]]
```
dahlia: Isn't it `NewGuid()` method?

PR #440, opened 2019-08-15
Check Preload completeness

earlbread:
Currently, `_blockChain.Swap` is performed when an exception occurred while preloading. We need to check the completeness of preloading before swapping.


earlbread:
I rebased on the current master. Please review this again.

PR #441, opened 2019-08-15
Less stateful RawTransaction

### LINKED ISSUE #172: Less stateful RawTransaction
Context:

> [Maybe it's better to change the type of `RawTransaction.Actions`... 😕][1]

> [Currently `RawTransaction` consists of many mutable types like `byte[]` and `Dictionary<K, V>`.  If we want `RawTransaction` to be freer from state I believe we need to adjust other fields too.  Seems like it should be a separated patch.][2]

[1]: https://github.com/planetarium/libplanet/pull/169#discussion_r271580759
[2]: https://github.com/planetarium/libplanet/pull/169#discussion_r271595495

minhoryang:


minhoryang:
TestFailed Diagnosis:
- `Linux_NETCore`, `macOS_NETCore` : flaky maybe.
- `macOS_Unity` : Retrying at local with following cmd:
    ```
    TURN_SERVER_URL=turn://localhost/ ../StandaloneOSX.app/Contents/MacOS/StandaloneOSX "$(pwd)"/./Libplanet.Tests/bin/Debug/net461/Libplanet.Tests.dll
    ```

Libplanet/Serialization/BencodexFormatter.cs
```diff
@@ -192,6 +193,8 @@ private IValue ToBencodexValue(object o)
                     return new Integer(bi);
                 case byte[] byteArray:
                     return ToBencodexKey(byteArray);
+                case ImmutableArray<byte> immutableByteArray:
```
longfin: I think it would be better if this conversion logic goes to `RawTransaction.GetObjectData()` because `byte[]` → `ImmutableArray<byte>` was covered in `RawTranscation(SerializationInfo info, StreamingContext context)`.

Libplanet/Tx/Transaction.cs
```diff
@@ -106,7 +106,7 @@ public class Transaction<T> : ISerializable, IEquatable<Transaction<T>>
         internal Transaction(RawTransaction rawTx)
             : this(
                 rawTx.Nonce,
-                new Address(rawTx.Signer),
+                new Address(rawTx.Signer.ToArray()),  // FIXME(minhoryang): Propagate?
```
dahlia: Note this this is covered by #443.

Libplanet/Serialization/BencodexFormatter.cs
minhoryang: Serializing at `ISerializable.GetObjectData(SerializationInfo info, StreamingContext context)` and Deserializing at `RawTranscation(SerializationInfo info, StreamingContext context)`.
Now I get it.

Libplanet.Tests/Blocks/BlockTest.cs
```diff
@@ -360,12 +360,12 @@ public void EvaluateInvalidTxPublicKey()
...
             RawTransaction rawTxWithoutSig = new RawTransaction(
                 0,
-                new PrivateKey().PublicKey.ToAddress().ToByteArray(),
-                new byte[][] { },
-                _fx.TxFixture.PublicKey1.Format(false),
+                new PrivateKey().PublicKey.ToAddress().ByteArray,
+                ImmutableArray<ImmutableArray<byte>>.Empty,
+                _fx.TxFixture.PublicKey1.Format(false).ToImmutableArray(),
                 DateTimeOffset.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.ffffffZ"),
-                new IDictionary<string, object>[0],
```
minhoryang: Is this differ with `IImmutableDictionary<string, object>.Empty`?

dahlia: It's an empty *array* of dictionaries, not an empty dictionary.

CHANGES.md
```diff
@@ -32,6 +32,13 @@ To be released.
  -  Added `IActionContext.NewGuId()` method.  [[#371], [#439]]
  -  `Address(byte[])` became to throw `ArgumentNullException`
     instead of `NullReferenceException`.  [[#443]]
+ -  `RawTransaction.Signer` became to `ImmutableArray<byte>` type.  [[#172], [#441]]
+ -  `RawTransaction.PublicKey` became to `ImmutableArray<byte>` type.  [[#172], [#441]]
+ -  `RawTransaction.Signature` became to `ImmutableArray<byte>` type.  [[#172], [#441]]
+ -  `RawTransaction.UpdatedAddresses` became to
+    `ImmutableArray<ImmutableArray<byte>>` type.  [[#172], [#441]]
+ -  `RawTransaction.Actions` became to
+    `IEnumerable<IImmutableDictionary<string, object>>` type.  [[#172], [#441]]
```
dahlia: I believe these changelogs don't have to be recorded because `RawTransaction` is an `internal` struct.  You could bypass the changelog checker by adding `[changelog skip]` in your commit message (any commit in this pull request).

Libplanet.Tests/Blocks/BlockTest.cs
minhoryang: Oh! What about List<ImmutableDictionary<string, object>>.Empty ?

CHANGES.md
minhoryang: I got it! I’ll bypass the changelog ci.

Libplanet.Tests/Blocks/BlockTest.cs
dahlia: As far as I know, there's no `.Empty` static property to `List<T>`.  The reason why `ImmutableDictionary<TKey, TValue>` has `.Empty` static property is it cannot utilize [C#'s collection initializer syntax][1], which requires a data structure to be *mutable*.

[1]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers#collection-initializers

minhoryang: Sorry for bothering and delivering effortless comment. I wasn't tried to test 
 `List<ImmutableDictionary<string, object>>.Empty`. It's my mistake.

The main reason why I concern this is I mixed up two things in this PR: `Immutable<T>.Empty` and `T[0]`. And It makes me confused.

What if change this:`new IDictionary<string, object>[0]` to that:`ImmutableList<IImmutableDictionary<string, object>>.Empty` for achieving more stateless? (but I didn't check the performance.)

dahlia: To me that would be fine.  As it's in a unit test, and only a single call, we don't have to think much about performance here.

Libplanet/Tx/Transaction.cs
```diff
@@ -107,7 +107,7 @@ internal Transaction(RawTransaction rawTx)
             : this(
                 rawTx.Nonce,
                 new Address(rawTx.Signer),
-                new PublicKey(rawTx.PublicKey),
+                new PublicKey(rawTx.PublicKey.ToArray()),  // FIXME: propagate?
```
longfin: Does `FIXME: propagate?` means that `PublicKey`'s constructor also should take `ImutableArray<byte>`?

Libplanet/Tx/RawTransaction.cs
```diff
@@ -180,14 +197,15 @@ private static T[][] To2dArray<T>(T[] array, int chunk)
             }
 
             int resultLength = array.Length / chunk;
-            T[][] result = new T[resultLength][];
+            ImmutableArray<T>[] result = new ImmutableArray<T>[resultLength];
             for (int i = 0; i < resultLength; i++)
             {
-                result[i] = new T[chunk];
-                Array.Copy(array, i * chunk, result[i], 0, chunk);
+                T[] partialResult = new T[chunk];
```
longfin: We're using `var` when variable's type is obvious. (e.g. `var x = new T();`)

```suggestion
                var partialResult = new T[chunk];
```

minhoryang: Oh! I got this, and I will check all changed codes for this.

Libplanet/Tx/RawTransaction.cs
```diff
@@ -180,14 +197,15 @@ private static T[][] To2dArray<T>(T[] array, int chunk)
             }
 
             int resultLength = array.Length / chunk;
-            T[][] result = new T[resultLength][];
+            ImmutableArray<T>[] result = new ImmutableArray<T>[resultLength];
```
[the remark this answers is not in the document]
minhoryang: Oh! I got this, and I will check all changed codes for this.

Libplanet/Tx/Transaction.cs
minhoryang: Yes, I thought that `PublicKey`'s constructor also should take `ImmutableArray<byte>`, but this PR doesn't cover this. So I'll remove this comment. (And others too.)

earlbread:
Could you rebase it on the current master not merge to avoid cross-merging between upstream and downstream? @minhoryang

minhoryang:
@earlbread Yes, I'll rebase this branch and force-push it.

longfin:
@minhoryang Thanks for the patch! 🎉

minhoryang:
Thanks all for kindly reviewing my patchset and supporting at the #pyconkr19 sprint!
It's honor to be here with you guys!

PR #444, opened 2019-08-16
Move BlockChain<T>.FindNextHashesChunkSize to Swarm<T>.FindNextHashesChunkSize

### LINKED ISSUE #407: Make FindNextHashesChunkSize configurable
The internal property `BlockChain<T>.FindNextHashesChunkSize` made to test `BlockDownloadState` in `Swarm<T>.PreloadAsync()`. But it is hard to understand the purpose of the property and not configurable. So it would be better to make it configurable in `Swarm<T>`.

kanghyojun:
As mentioned in issue, `BlockChain<T>.FindNextHashesChunkSize`  is used only for tests. So It's unneceesary to configure `FindNextHashesChunkSize` in `Swarm<T>` constructor.

So @dahlia suggest to make an property on `Swarm<T>`, and I applied it. cc @earlbread 

Resolve #407

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -254,6 +254,8 @@ namespace Libplanet.Net
 
         internal ICollection<Peer> Peers => _peers.Keys;
 
+        internal int FindNextHashesChunkSize { get; set; }
+
         /// <summary>
         /// Waits until this <see cref="Swarm{T}"/> instance gets started to run.
         /// </summary>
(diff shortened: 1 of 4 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -254,6 +254,8 @@ private set
 
         internal ICollection<Peer> Peers => _peers.Keys;
 
+        internal int FindNextHashesChunkSize { get; set; }
```
longfin: It doesn't seem to be used anywhere.

kanghyojun: Oops, I made a mistake.

limebell: The value should have its default value (500)

limebell:
```suggestion
        internal int FindNextHashesChunkSize { get; set; } = 500;
```

earlbread:
Replaced with #464.

PR #445, opened 2019-08-16
Correct GetBlockHashes chunk size after first phase

### LINKED ISSUE #408: Correct GetBlockHashes chunk size after first phase
When a node received `GetBlockHashes` message, it sends block hashes as many as`FindNextHashesChunkSize`. However, after the first step, it seems to send block hashes as many as `FindNextHashesChunkSize - 1`:
 
When there are 2000 blocks to send,
 - Expected : 500, 500, 500, 500
 - Actual : 500, 499, 499, 499, 3

gurrpi:
This PR fixes behavior of peer—which responses `getBlockHashes()`—to contains one more higher block hash for satisfy `FindNextHashesChunkSize` (Since the peer which performs `PreloadAsync()` drops the lowest block from the response of `getBlockHashes()` except initial phase). 

Resolves #408.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -78,6 +78,7 @@ To be released.
     instead of the window size of a chunk (i.e., 500).  [[#396], [#399]]
  -  `Swarm<T>.PreloadAsync()` became to get the first parameter, `progress`,
     which accepts `IProgress<PreloadState>`.  [[#397], [#400]]
+ -  `BlockHashes` messages became to contain one more higher hash. [[#408], [#445]]
  -  `Swarm<T>.PreloadAsync()` became safe from data corruption even
     if a preloading process suddenly gets shutdown.  [[#417]]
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -78,6 +78,7 @@ To be released.
     instead of the window size of a chunk (i.e., 500).  [[#396], [#399]]
  -  `Swarm<T>.PreloadAsync()` became to get the first parameter, `progress`,
     which accepts `IProgress<PreloadState>`.  [[#397], [#400]]
+ -  Response of `getBlockHashes()` now contains one more higher hash. [[#408]]
```
dahlia: Could you adjust this into like below?

```suggestion
 -  `BlockHashes` messages became to contain one more higher hash. [[#408], [#445]]
```

Also need to a link reference below:

```markdown
[#445]: https://github.com/planetarium/libplanet/pull/445
```

gurrpi: I changed changelog. 📃

PR #468, opened 2019-08-27
Fix FindBranchPoint

earlbread:
This fixes an issue where `FindBranchPoint()` returns a branch point if it exists in the `BlockSet` even if the block does not exist in the chain.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -738,7 +738,9 @@ namespace Libplanet.Blockchain
 
                 foreach (HashDigest<SHA256> hash in locator)
                 {
-                    if (Blocks.ContainsKey(hash))
+                    if (Blocks.ContainsKey(hash)
+                        && Blocks[hash] is Block<T> block
+                        && hash.Equals(Store.IndexBlockHash(Id.ToString(), block.Index)))
                     {
                         return hash;
                     }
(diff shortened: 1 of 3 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -735,7 +735,9 @@ internal HashDigest<SHA256> FindBranchPoint(BlockLocator locator)
 
                 foreach (HashDigest<SHA256> hash in locator)
                 {
-                    if (Blocks.ContainsKey(hash))
+                    if (Blocks.ContainsKey(hash)
+                        && Blocks[hash] is Block<T> block
```
longfin: Isn't this check redundant with the above line?

earlbread: I checked the hash first because `Blocks[hash]` throws `KeyNotFoundException` if the hash doesn't exists. Or catching the exception is better?

longfin: 👌  I've misled. this line is not only a check but also binding `block`.

dahlia: I guess it's it's for avoiding catching an exception?

earlbread:
I added changelog and changed the base branch to 0.5-maintenance.

earlbread:
I changed `block.Equals(this[block.Index])` to `hash.Equals(Store.IndexBlockHash(Id.ToString(), block.Index)))` to avoid `ArgumentOutOfRangeException`. Please take a look.

PR #469, opened 2019-08-27
Mute IOException during AcceptRelayedStreamAsync()

longfin:
This PR fixes unrecoverable `IOException` during `TurnClient.AcceptRelayedStreamAsync()` by muting it.

### DIFF
--- a/Libplanet.Stun/Stun/TurnClient.cs
+++ b/Libplanet.Stun/Stun/TurnClient.cs
@@ -143,7 +143,7 @@ namespace Libplanet.Stun
                 }
                 catch (IOException e)
                 {
-                    Log.Warning(e, "The connection seems to disconnect before parsing. ignored.");
+                    Log.Warning(e, "The connection seems to disconnect before parsing; ignored.");
                     continue;
                 }
             }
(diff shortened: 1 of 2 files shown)


Libplanet.Stun/Stun/TurnClient.cs
```diff
@@ -124,26 +124,28 @@ public class TurnClient : IStunContext, IDisposable
...
+                        _relayedClients.Add(relayedClient);
+                        return relayedStream;
+                    }
+
+                    throw new TurnClientException("ConnectionBind failed.", bindResponse);
+                }
+                catch (IOException e)
+                {
+                    Log.Warning(e, "The connection seems to disconnect before parsing. ignored.");
+                    continue;
```
limebell: If disconnected, shouldn't we `break` instead of `continue`?

longfin: It's intended because we should contiue to handle other new connections even if we caught an exception from previous connectio .

Libplanet.Stun/Stun/TurnClient.cs
```diff
@@ -124,26 +124,28 @@ public class TurnClient : IStunContext, IDisposable
...
+                    {
+                        _relayedClients.Add(relayedClient);
+                        return relayedStream;
+                    }
+
+                    throw new TurnClientException("ConnectionBind failed.", bindResponse);
+                }
+                catch (IOException e)
+                {
+                    Log.Warning(e, "The connection seems to disconnect before parsing. ignored.");
```
dahlia:
```suggestion
                    Log.Warning(e, "The connection seems to disconnect before parsing; ignored.");
```

PR #472, opened 2019-08-28
Ready to 0.5.2

earlbread:


CHANGES.md
```diff
@@ -1,6 +1,20 @@
...
+
+To be released.
+
+### Backward-incompatible interface changes
+
+### Added interfaces
+
+### Behavioral changes
+
+### Bug fixes
```
dahlia: We don't need these sections other than bug fixes, because patch releases can have only bug fixes.

earlbread: Removed!

PR #476, opened 2019-08-29
Fix Swarm<T>.BroadcastTxAsync() an exception from IStore on some race conditions

dahlia:
I believe this fixes <https://github.com/planetarium/libplanet/issues/352>, but I haven't found any concise way to reproduce this in unit tests…

### DIFF
--- a/.azure-pipelines/mono.yml
+++ b/.azure-pipelines/mono.yml
@@ -2,6 +2,7 @@ parameters:
   configuration: Debug
   testDisplayName: xunit.console.exe *.Tests.dll
   testCommand: "mono xunit.runner.console.*/tools/net461/xunit.console.exe"
+  testArguments: -verbose
 
 steps:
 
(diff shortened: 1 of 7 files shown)


longfin:
It's a little weird because [this scenario](https://github.com/planetarium/libplanet/issues/352#issuecomment-526045343) didn't seem a race condition. (`9e66F6faF87109163BFB7cdBfB62685DfbE1Fcea` and `925219D488Ee73c42EB308ac178Ea1D6D55288eC` are using separated store each other)

FYI I've seen `NullReferenceException` when `LiteDBStore` that already disposed has been accessed.

dahlia:
@longfin Made `SwarmTest.Dispose()` method to dispose things in the correct order: 3dc86f9e39f3dfb1a855bb50da7b1a888c6e69ef.

dahlia:
The reason Linux/Mono builds failed was not due to this patch, but `PreloadAsyncCancellation()` mining fixture blocks take long time on Linux/Mono (for unknown reason).  I fixed this by making fixture blocks reused.

PR #484, opened 2019-09-02
Ignore containing block hashes

### LINKED ISSUE #461: Ignore BlockHashes or TxIds which already stored
By current implementation, when `Swarm<T>` received `TxIds` or `BlockHashes` from other peers, it will try to receive payload even if it already has that transaction or block.

limebell:
This resolves #461

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,7 @@ To be released.
 
 ### Behavioral changes
 
+ -  `Swarm<T>` now ignores block hashes which already exists.  [[#461], [#484]]
  -  `Swarm<T>.PreloadAsync()` method became to download precalculated states
     of blocks from a likely branchpoint instead of a genesis block from
     a trusted peer (i.e., `trustedStateValidators`) where there are branches
@@ -80,9 +81,11 @@ To be released.
 [#353]: https://github.com/planetarium/libplanet/pull/353
 [#420]: https://github.com/planetarium/libplanet/pull/420
 [#450]: https://github.com/planetarium/libplanet/pull/450
+[#461]: https://github.com/planetarium/libplanet/issues/461
 [#470]: https://github.com/planetarium/libplanet/pull/470
 [#481]: https://github.com/planetarium/libplanet/pull/481
 [#483]: https://github.com/planetarium/libplanet/issues/483
+[#484]: https://github.com/planetarium/libplanet/pull/484
 [#486]: https://github.com/planetarium/libplanet/pull/486
 [Kademlia]: https://en.wikipedia.org/wiki/Kademlia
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -42,6 +42,7 @@ To be released.
 
 ### Behavioral changes
 
+ -  `Swarm<T>` now ignores `BlockHash`es which already exists.  [[#461], [#484]]
```
dahlia: Isn't the message name `BlockHashes`, not `BlockHash`?

limebell: I intended block hash instead of message, but it seems like 'block hash' is better than '`BlockHash`'.

Libplanet/Net/Swarm.cs
```diff
@@ -1401,19 +1401,36 @@ private void BroadcastTxIds(IEnumerable<TxId> txIds)
                 return;
             }
 
+            ImmutableList<HashDigest<SHA256>> newHashes = message.Hashes
+                .Where(hash => !_blockChain.Blocks.ContainsKey(hash))
+                .ToImmutableList();
+
+            if (!newHashes.Any())
+            {
+                _logger.Debug("No blocks to require. Ignore BlockHashes.");
```
```suggestion
                _logger.Debug($"No blocks to require. Ignore {nameof(BlockHashes)}.");
```

Libplanet/Net/Swarm.cs
```diff
@@ -1401,19 +1401,36 @@ private void BroadcastTxIds(IEnumerable<TxId> txIds)
...
+                return;
+            }
+
             _logger.Debug(
                 $"Trying to GetBlocksAsync() " +
```
```suggestion
                 $"Trying to {nameof(GetBlocksAsync)}() " +
```

Libplanet/Net/Swarm.cs
```diff
@@ -1401,19 +1401,36 @@ private void BroadcastTxIds(IEnumerable<TxId> txIds)
...
-                $"(using {message.Hashes.Count()} hashes)");
+                $"(using {newHashes.Count()} hashes)");
```
```suggestion
                "using {0} hashes.",
                newHashes.Count());
```

Libplanet/Net/Swarm.cs
```diff
@@ -1401,19 +1401,36 @@ private void BroadcastTxIds(IEnumerable<TxId> txIds)
...
             IAsyncEnumerable<Block<T>> fetched = GetBlocksAsync(
                 peer,
-                message.Hashes
+                newHashes
             );
 
             List<Block<T>> blocks = await fetched.ToListAsync(
                 cancellationToken
             );
             _logger.Debug("GetBlocksAsync() complete.");
```
```suggestion
             _logger.Debug($"{nameof(GetBlocksAsync)}() complete.");
```

dahlia:
@limebell Could we have some tests for the behavior added by this patch?

PR #508, opened 2019-09-10
Remove `IActionContext.NewGuid()` & Add `RandomExtension.GenerateRandomGuid()`

Unengine:
I think getting a random `Guid` in `IActionContext` is a bad usage because dummy objects (for testing, debugging or etc ...) using random `Guid` in games don't need `IActionContext`. `IRandom`s are enough to acquire the same thing.

So I removed `IActionContext.NewGuid()`.
Instead I added an extension class, `RandomExtension` that includes a extension method, `GenerateRandomGuid()`.
Since `RandomExtension` is an extension class, I implemented generating a random `Guid`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,8 @@ To be released.
     `Swarm<T>.BootstrapAsync()` method instead.  [[#353]]
  -  `Peer` with endpoints should be typed as `BoundPeer` which is inherited from
     `Peer`.  [[#353]]
+ -  Removed `IActionContext.NewGuid()` method. To get a randomly generated [Guid][Guid],
+    use `RandomExtension.GenerateRandomGuid()` instead. [[#508]]
 
 ### Added interfaces
 
@@ -71,6 +73,8 @@ To be released.
     both a `Block<T>.Hash` and a `Transaction<T>.Id`, so that signers cannot
     predict the order of transactions in a block before it's mined.
     [[#244], [#355]]
+ -  Added `RandomExtension.GenerateRandomGuid()` method that generates a random
+    [Guid][Guid]. [[#508]]
 
 ### Behavioral changes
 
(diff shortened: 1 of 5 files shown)


Libplanet/Action/RandomExtension.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// Generates a random <see cref="Guid"/>.
+        /// </summary>
+        /// <param name="random"> <see cref="IRandom"/> to generate
+        /// a random <see cref="Guid"/>.</param>
+        /// <returns> Generated random <see cref="Guid"/>.
+        /// </returns>
+        /// <seealso cref="IRandom"/>
+        public static Guid GenerateRandomGuid(this IRandom random)
+        {
+            var bytes = new byte[16];
```
longfin: It would be better if there is mention as [this](https://github.com/planetarium/libplanet/blob/714e241/Libplanet/Action/ActionContext.cs#L38).

Unengine: Both are fixed! Thanks!

dahlia: @Unengine Still not addressed I think?

Unengine: Sorry, I forgot to commit. I'll also commit your suggestions at once!

Unengine: Job's done!

Unengine:
> I think you should fix some tests which uses `NewGuid()`.

You're right. Job's done!

CHANGES.md
```diff
@@ -71,6 +73,8 @@ To be released.
     both a `Block<T>.Hash` and a `Transaction<T>.Id`, so that signers cannot
     predict the order of transactions in a block before it's mined.
     [[#244], [#355]]
+ -  Added `RandomExtension.GenerateRandomGuid()` method that generates a random
```
```suggestion
 -  Added `RandomExtension` static class.  [[#508]]
```

CHANGES.md
```diff
@@ -71,6 +73,8 @@ To be released.
...
+    [Guid][Guid]. [[#508]]
```
```suggestion
```

PR #513, opened 2019-09-16
Implement RFC 4122

earlbread:
This changes `GenerateRandomGuid` implement RFC 4122.


Libplanet/Action/RandomExtension.cs
```diff
@@ -10,6 +10,7 @@ public static class RandomExtension
     {
         /// <summary>
         /// Generates a random <see cref="Guid"/>.
```
dahlia:
```suggestion
         /// Generates a UUID version 4, i.e., a random <see cref="Guid"/>.
```

PR #519, opened 2019-09-18
Hash signature before feeding as seed

longfin:
This PR fixes a bug that `IActionContext.Random` had generated the same result in different `IAction.Execute()` against the same `Block<T>`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -107,6 +107,9 @@ To be released.
 
  -  Fixed a bug that `Swarm<T>` hadn't released its TURN related resources on
     `Swarm<T>.StopAsync()`.  [[#450]]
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -102,6 +102,8 @@ To be released.
 
  -  Fixed a bug that `Swarm<T>` hadn't released its TURN related resources on
     `Swarm<T>.StopAsync()`.  [[#450]]
+ -  Fixed a bug that `IActionContext.Random` had generated the same result in
+    different `IAction.Execute()` against the same `Block<T>`.  [[#519]]
```

> Fixed a bug that `IActionContext.Random` had been possible to generated equivalent results between actions of different transactions in a `Block<T>`.

PR #520, opened 2019-09-18
Make the order of Block<T>.Transactions aware of their nonce

dahlia:
See also the changelog and comments in the `Block<T>()` constructor.

### DIFF
--- a/Libplanet/Blocks/Block.cs
+++ b/Libplanet/Blocks/Block.cs
@@ -25,6 +25,28 @@ namespace Libplanet.Blocks
         private static readonly TimeSpan TimestampThreshold =
             TimeSpan.FromSeconds(900);
 
+        /// <summary>
+        /// Creates a <see cref="Block{T}"/> instance by manually filling all field values.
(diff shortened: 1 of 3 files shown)


Libplanet/Blocks/Block.cs
```diff
@@ -43,11 +65,39 @@ public class Block<T> : ISerializable
...
+            // As the order of transactions should be unpredictable until a block is mined,
+            // the sorter key should be derived from both a block hash and a txid.
             var hashInteger = new BigInteger(Hash.ToByteArray());
-            Transactions =
-                Transactions
-                    .OrderBy(tx =>
-                        new BigInteger(tx.Id.ToByteArray()) ^ hashInteger).ToArray();
+
+            // If there are multiple transactions for the same signer these should be ordered by
+            // their tx nonces.  So transations of the same signer should have the same sort key.
```
limebell:
```suggestion
            // their tx nonces.  So transactions of the same signer should have the same sort key.
```

dahlia: 👀

PR #521, opened 2019-09-19
Use in-memory LiteDBStore in tests

earlbread:
This improves the test speed by using in-memory LiteDBStore in tests and restores test timeout to 10 minutes.

### DIFF
--- a/Libplanet/Store/LiteDBStore.cs
+++ b/Libplanet/Store/LiteDBStore.cs
@@ -34,6 +34,8 @@ namespace Libplanet.Store
 
         private readonly ILogger _logger;
 
+        private readonly MemoryStream _memoryStream;
+
         private readonly LiteDatabase _db;
 
         private readonly ReaderWriterLockSlim _txLock;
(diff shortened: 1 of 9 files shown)


Libplanet/Store/LiteDBStore.cs
```diff
@@ -84,7 +86,10 @@ public class LiteDBStore : BaseStore, IDisposable
                 connectionString.Mode = LiteDB.FileMode.Exclusive;
             }
 
-            _db = new LiteDatabase(connectionString);
+            _db = memory
+                ? new LiteDatabase(new MemoryStream())
```
longfin: Is it okay we don't dispose of this `MemoryStream()` at `LiteDBStore.Dipose()`?

Libplanet/Store/LiteDBStore.cs
```diff
@@ -48,14 +48,16 @@ public class LiteDBStore : BaseStore, IDisposable
         /// <param name="cacheSize">Max number of pages in the cache.</param>
         /// <param name="flush">Writes data direct to disk avoiding OS cache.  Turned on by default.
         /// </param>
-        /// <param name="readOnly">Opens database readonly mode. Turned off by default.
+        /// <param name="readOnly">Opens database readonly mode. Turned off by default.</param>
+        /// <param name="memory">Creates database in memory using <see cref="MemoryStream"/>.
         /// </param>
         public LiteDBStore(
             string path,
```
dahlia: Is this parameter ignored when `memory: true`?  If two parameters are mutually exclusive how about representing memory mode by passing `null` to the `path` parameter?

earlbread:
I added the code to dispose `MemoryStream` and changed to create the database if the `path` parameter is `null`.

PR #525, opened 2019-09-23
Fix KademliaProtocol<T>.RebuildConnectionAsync

### LINKED ISSUE #524: Fix KademliaProtocol<T>.RebuildConnectionAsync to work more precisely
Currently, `KademliaProtocol<T>.RebuildConnectionAsync()` works just as `KademliaProtocol<T>.BootstrapAsync`, just searches for specified depth. Since the method works in background and has no pending, it should behave to search all peers in the network.

limebell:
Unlike `KademliaProtocol<T>.BootstrapAsync`, `KademliaProtocol<T>.RebuildConnectionAsync` now searches for all peers on network. Additionally, it runs on `Swarm<T>.StartAsync` since `KademliaProtocol<T>.BootstrapAsync` does not ensures to find all peers on network.

This resolves #524.

### DIFF
--- a/Libplanet/Net/Protocols/KademliaProtocol.cs
+++ b/Libplanet/Net/Protocols/KademliaProtocol.cs
@@ -151,6 +151,15 @@ namespace Libplanet.Net.Protocols
             }*/
         }
 
+        /// <summary>
+        /// Periodically checks whether <see cref="Peer"/>s in <see cref="RoutingTable"/> is
+        /// online by sending <see cref="Ping"/>.
+        /// </summary>
+        /// <param name="period">The cycle in which the operation is executed. If null,
+        /// <see cref="IdleRefreshInterval"/> will be used.</param>
+        /// <param name="cancellationToken">A cancellation token used to propagate notification
+        /// that this operation should be canceled.</param>
+        /// <returns>An awaitable task without value.</returns>
         public async Task RefreshTableAsync(
             TimeSpan? period,
             CancellationToken cancellationToken)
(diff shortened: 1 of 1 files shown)


Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -196,19 +194,21 @@ public ImmutableList<BoundPeer> PeersToBroadcast
                     tasks.Add(FindPeerAsync(
                         new Address(buffer),
                         null,
-                        0,
+                        -1,
```
longfin: What does means `-1`?

limebell: It indicates that the operation is not being limited by `depth`. Please take a look at <a href="https://github.com/planetarium/libplanet/pull/525/files#diff-ae527d5afcdf41da4a57b594cab3cfefR528">line 528</a>.

dahlia: It would be good if such behavior is documented.

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -179,14 +188,21 @@ public ImmutableList<BoundPeer> PeersToBroadcast
             }
         }
 
+        /// <summary>
+        /// Reconstructs network connection between peers on network. It runs operation once
+        /// right after called, repeated every <see cref="TimeSpan"/> of<paramref name="period"/>.
```
moreal:
```suggestion
        /// right after called, repeated every <see cref="TimeSpan"/> of <paramref name="period"/>.
```

PR #530, opened 2019-09-25
Fix mono version to 6.0.0.334

limebell:
After mono version upgrade to 6.4.0.198, some tests seem to fail. So temporary roll back mono version to 6.0.0.334.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -87,6 +87,7 @@ jobs:
       turnServerUrl: $(turnServerUrl)
   timeoutInMinutes: 30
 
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -94,7 +94,7 @@ jobs:
   - task: CmdLine@2
     displayName: choco install mono
     inputs:
-      script: choco install mono --yes
+      script: choco install mono --version 6.0.0.334 --yes
```
longfin: It would be great if there is `FIXME` comment about this pinning.

Issue #523, opened 2019-09-19
Avoid re-sorting transactions


limebell:
Currently, transaction is sorted by its nonce when a block is created. Though, sorting is re-done in `BlockChain<T>.Append()`.

moreal:
It's closed by #542.

PR #542, opened 2019-09-26
Avoids re-sorting transactions when appending the already sorted block to the blockchain

RozzaysRed:
[changelog skip]

PR Should Address #523

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -604,8 +604,7 @@ namespace Libplanet.Blockchain
                 HashDigest<SHA256>? tip = Store.IndexBlockHash(Id, -1);
 
                 var nonceDeltas = new Dictionary<Address, long>();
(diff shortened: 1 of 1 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -604,8 +604,7 @@ bool renderActions
                 HashDigest<SHA256>? tip = Store.IndexBlockHash(Id, -1);
 
                 var nonceDeltas = new Dictionary<Address, long>();
-                var orderedTransactions = block.Transactions.OrderBy(tx => tx.Nonce);
-                foreach (Transaction<T> tx1 in orderedTransactions)
```
dahlia: It would be better it there's a comment that mentions `blcok.Transactions` are already sorted in the tx nonce order.

RozzaysRed: I rebased and added comment

longfin:
@RozzaysRed Thanks for the contribution!

PR #547, opened 2019-10-01
Fix NullReferenceException in AppendBlocksAsync

earlbread:
This fixes `NullReferenceException` in `AppendBlocksAsync()` and adds `ToString()` to `Peer`.
I skipped the changelog because `Peer` is changed at 0.6.0.

### DIFF
--- a/Libplanet/Net/Peer.cs
+++ b/Libplanet/Net/Peer.cs
@@ -77,5 +77,11 @@ namespace Libplanet.Net
             info.AddValue("app_protocol_version", AppProtocolVersion);
             info.AddValue("public_ip_address", PublicIPAddress?.ToString());
         }
+
+        /// <inheritdoc/>
+        public override string ToString()
+        {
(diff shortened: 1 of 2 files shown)


Libplanet/Net/Peer.cs
```diff
@@ -77,5 +77,11 @@ public virtual void GetObjectData(SerializationInfo info, StreamingContext conte
             info.AddValue("app_protocol_version", AppProtocolVersion);
             info.AddValue("public_ip_address", PublicIPAddress?.ToString());
         }
+
+        /// <inheritdoc/>
+        public override string ToString()
+        {
+            return $"{Address}";
```
dahlia: How about this?

```suggestion
            return $"{nameof(Peer)} {{ {nameof(Address)} = {Address} }}";
```

earlbread: I amended the commit.

PR #566, opened 2019-10-08
Avoid infinite loop when a peer is evicted from routing table and add cached peer

limebell:
If a bucket is full, later peers who requested connection were cached into `Bucket.ReplacementCache`. If a peer is removed from a bucket then cached peer will be added automatically after liveness checking.

Previously, live check had a bug so check procedure repeats infinitely. This patch fixes such bug and added two test cases for the case and fixed one test case.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,8 @@ To be released.
 
 ### Behavioral changes
 
+ -  Changed to send `Pong` before updating the message sender to the routing
+    table when `Ping` is received.  [[#566]]
  -  Improved performance of `StoreExtension.LookupStateReference<T>()` method.
     [[#447], [#545]]
  -  Added .NET Core 2.2 as a targeted framework.  [[#209], [#561]]
@@ -39,6 +41,8 @@ To be released.
 
 ### Bug fixes
 
+ -  Fixed a bug where `KademliaProtocol<T>.UpdateAsync()` runs infinitely when
+    a peer is evicted from `Bucket` and replacement cache is used.  [[#566]]
  -  Fixed a bug where `Swarm<T>.AppendBlocksAsync()` re-requests blocks that
     already received when blockchain is empty.  [[#550], [#562]]
  -  Fixed a bug that `Swarm<T>` had thrown `SocketException` with a message
@@ -48,6 +52,7 @@ To be released.
  -  Fixed a bug that `Swarm<T>` had thrown `InvalidBlockIndexException` during
     synchronizing with other reorganized peer.  [[#528], [#576]]
 
+[#209]: https://github.com/planetarium/libplanet/issues/209
 [#405]: https://github.com/planetarium/libplanet/issues/405
 [#447]: https://github.com/planetarium/libplanet/issues/447
 [#462]: https://github.com/planetarium/libplanet/issues/462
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -16,17 +16,22 @@ To be released.
...
- -  Reworked `BlockChain<T>.GetStates()` into `GetState()` which takes only one `Address` instead of `IEnumerable<Address>`.  [[#510], [#563]]
+ -  Reworked `BlockChain<T>.GetStates()` into `GetState()` which takes only
+    one `Address` instead of `IEnumerable<Address>`.  [[#510], [#563]]
 
 ### Added interfaces
 
 ### Behavioral changes
 
+ -  `KademliaProtocol<T>.ReceivePingAsync()` method became to reply `Pong`
+    before update message sender to routing table.  [[#566]]
```
earlbread: `ReceivePingAsync` is a private method. How about this?
```
Changed to send `Pong` before updating the message sender to the routing table when `Ping` is received.
```

CHANGES.md
```diff
@@ -16,17 +16,22 @@ To be released.
...
  -  Improved performance of `StoreExtension.LookupStateReference<T>()` method.
     [[#447], [#545]]
 
 ### Bug fixes
 
+ -  Fixed a bug where `KademliaProtocol<T>.UpdateAsync()` runs infinitely when
+    peer is evicted from `Bucket` and replacement cache is used.  [[#566]]
```
```suggestion
    a peer is evicted from `Bucket` and replacement cache is used.  [[#566]]
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -13,38 +13,38 @@ namespace Libplanet.Net.Protocols
...
 
         private readonly ILogger _logger;
 
         public KademliaProtocol(
             Swarm<T> swarm,
             Address address,
             int appProtocolVersion,
-            ILogger logger)
+            ILogger logger,
+            int? tableSize,
```
dahlia: How about giving it a default value instead of allowing it to be `null`?

```suggestion
            int tableSize = Kademlia.TableSize,
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -13,38 +13,38 @@ namespace Libplanet.Net.Protocols
...
+            int? bucketSize,
```
dahlia: How about giving it a default value instead of allowing it to be `null`?

```suggestion
            int bucketSize = Kademlia.BucketSize,
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -263,12 +262,12 @@ public string Trace()
...
                 trace += $"**Bucket {i}**\n";
                 trace = _routing.BucketOf(i).Peers.Aggregate(trace, (current, peer) =>
-                    current + $"{count++} : [{peer.Address.ToHex()}]\n");
+                    current + $"{++count} : [{peer.Address.ToHex()}]\n");
 
                 trace = trace.TrimEnd(' ', ',');
             }
 
-            return trace.Trim('\n');
+            return "Total peer count : " + count + "\n" + trace.Trim('\n');
```
dahlia:
```suggestion
            return $"Total peer count: {count}\n{trace.Trim('\n')}";
```

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: If default value is given, cyclic constructor call will happen since public method without those property is exactly same with internal method.

It can be one way to just expose those parameters as public I guess...

PR #575, opened 2019-10-11
Access only blocks to related to render

earlbread:
This fixes a bug that accessed all blocks from the genesis block when a swap occurred.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -114,30 +114,7 @@ namespace Libplanet.Blockchain
         /// All <see cref="Block{T}.Hash"/>es in the current index.  The genesis block's hash goes
         /// first, and the tip goes last.
         /// </summary>
-        public IEnumerable<HashDigest<SHA256>> BlockHashes
-        {
-            get
-            {
-                try
-                {
-                    _rwlock.EnterUpgradeableReadLock();
-
-                    IEnumerable<HashDigest<SHA256>> indices = Store.IterateIndexes(Id);
-
-                    // NOTE: The reason why this does not simply return indices, but iterates over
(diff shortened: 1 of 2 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -171,6 +171,14 @@ public IEnumerable<HashDigest<SHA256>> BlockHashes
             }
         }
 
+        /// <summary>
+        /// Returns an <see cref="long"/> that represents the number of elements in the
```
dahlia:
```suggestion
        /// Returns a <see cref="long"/> integer that represents the number of elements in the
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1030,12 +1038,13 @@ internal void Swap(BlockChain<T> other, bool render)
...
-                IEnumerable<Block<T>> blocksToRender =
-                    topmostCommon is Block<T> branchPoint
-                        ? this.SkipWhile(b => b.Index <= branchPoint.Index)
-                        : this;
-                foreach (Block<T> b in blocksToRender)
+                long startToRenderIndex = topmostCommon is Block<T> branchPoint
+                    ? branchPoint.Index + 1
+                    : 0;
+
+                for (var i = startToRenderIndex; i <= Tip.Index; i++)
```
```suggestion
                for (long i = startToRenderIndex; i <= Tip.Index; i++)
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1030,12 +1038,13 @@ internal void Swap(BlockChain<T> other, bool render)
...
                 {
+                    Block<T> b = this[i];
```
dahlia: Isn't random accessing over `BlockChain<T>` for forward iteration too slow?

earlbread:
I fixed docs and random access when rendering.

Issue #570, opened 2019-10-08
Implement FIND_PEER in KademliaProtocol


limebell:
Currently `KademliaProtocol<T>.FindPeerAsync` searches peers close to target peer except target itself. It will be useful if there exists a method to find specific peer only.

riemannulus:
It looks like resolved by #580.

limebell:
Closed by #580

PR #580, opened 2019-10-14
Added search for a specific peer given the address

RozzaysRed:
Addresses #570 

Let me know if there is a better way to approach this.  I initially had it setup as having a target address and a specific address but switched it just having a flag being set to say whether they want a list of neighbors to the target or the BoundPeer for the target address.

### DIFF
--- a/Libplanet/Net/Protocols/KademliaProtocol.cs
+++ b/Libplanet/Net/Protocols/KademliaProtocol.cs
@@ -277,6 +277,63 @@ namespace Libplanet.Net.Protocols
             return $"Total peer count: {count}\n{trace.Trim('\n')}";
         }
 
+        /// <summary>
+        /// Send <see cref="FindNeighbors"/> messages to <paramref name="viaPeer"/>
+        /// to find <see cref="Peer"/>s near <paramref name="target"/>
+        /// to see if the <see cref="Peer"/>s have a specific address.
+        /// </summary>
+        /// <param name="history">The <see cref="Peer"/> that searched.</param>
+        /// <param name="target">The <see cref="Address"/> to find.</param>
+        /// <param name="viaPeer">The target <see cref="Peer"/> to send <see cref="FindNeighbors"/>
+        /// message. If null, selects 3 <see cref="Peer"/>s from <see cref="RoutingTable"/> of
+        /// self.</param>
+        /// <param name="depth">Target depth of recursive operation.</param>
+        /// <param name="searchTarget">The address to search for among peers.</param>
+        /// <param name="timeout"><see cref="TimeSpan"/> for waiting reply of
+        /// <see cref="FindNeighbors"/>.</param>
+        /// <param name="cancellationToken">A cancellation token used to propagate notification
+        /// that this operation should be canceled.</param>
+        /// <returns>BoundPeer containing the <paramref name="searchTarget"/>'s address.</returns>
+        public async Task<BoundPeer> FindSpecificPeerAsync(
+            ConcurrentBag<BoundPeer> history,
+            Address target,
+            BoundPeer viaPeer,
+            int depth,
+            Address searchTarget,
+            TimeSpan? timeout,
+            CancellationToken cancellationToken)
+        {
+            _logger.Debug(
+                $"{nameof(FindSpecificPeerAsync)}() with {{Target}} to {{Peer}}. " +
+                "(depth: {Depth})",
+                target,
+                viaPeer,
+                depth);
+
+            IEnumerable<BoundPeer> found;
+            if (viaPeer is null)
+            {
+                found = await QueryNeighborsAsync(history, target, timeout, cancellationToken);
+            }
+            else
+            {
+                found = await GetNeighbors(viaPeer, target, timeout, cancellationToken);
+                history.Add(viaPeer);
+            }
+
+            BoundPeer specificPeerFound = await ProcessFoundForSpecificAsync(
+                history,
+                found,
+                target,
+                depth,
+                timeout,
+                cancellationToken,
+                searchTarget
+            );
+
+            return specificPeerFound;
+        }
+
         internal async Task PingAsync(
             BoundPeer target,
             TimeSpan? timeout,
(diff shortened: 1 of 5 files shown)


Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -166,6 +166,19 @@ public ICollection<BoundPeer> Neighbors(Address target, int k)
...
+        {
+            var sorted = _buckets
+                .Where(b => !b.IsEmpty())
+                .SelectMany(b => b.Peers)
+                .ToList();
+
+            var peers = new List<BoundPeer>();
+            peers.Add(sorted.Where(peer => peer.Address.Equals(specificPeer)).FirstOrDefault());
+
+            return peers;
```
limebell: I recommand you to use `RoutingTable.GetBucketIndexOf()` method to find specific peer.

RozzaysRed: Just so I completely understanding what you're looking for.  We are provided a peer to look for and we want to find where it resides?  Instead of what I implemented of, we are provided an address and returning a peer with that address.

limebell: Your understanding is correct—we have to find the peer with the specific address. But it's implementation is not accurate so it may fail to find peer with the address even it is on the network. 
Since Kademlia is one of protocol to implement [DHT (Distribute Hash Table)](https://en.wikipedia.org/wiki/Distributed_hash_table), a peer may not contain a specific peer even it is on network. So we have to recursively find the specific peer by asking closest peer to target address which peer-to-look has in each table. Since Kademlia saves closes peer more specifically, it can finally reach the target address if it is on the network.
Feel free to ask us if you have more questions.

RozzaysRed:
not stale

kijun:
Hey @RozzaysRed ! Have you had time to look into this issue?

RozzaysRed:
@kijun we can close this for now if you would like.  I've been trying to find time.  If I managed to come up with a solution to it using the recommendations and resources provided by @limebell before someone else is able to resolve it then I can submit another PR?

kijun:
We're happy that you're working on this! Just wanted to make sure that you didn't have any stumble blocks.

Libplanet/Net/Protocols/RoutingTable.cs
RozzaysRed: Okay, I've tried a different approach.  Could you please review the PR and let me know your thoughts on this?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -277,6 +277,67 @@ public string Trace()
...
+            CancellationToken cancellationToken)
+        {
+            _logger.Debug(
+                $"{nameof(FindSpecificPeerAsync)}() with {{Target}} to {{Peer}}. " +
+                "(depth: {Depth})",
+                target,
+                viaPeer,
+                depth);
+
+            if (depth == 0)
```
moreal: Now, if depth is 0, it returns null and exit `FindSpecificPeerAsync()`. So it looks like some breakpoint, right?
If then, IMO it will not work as well. Because `depth` will stay away from 0 by the line (i.e. [KademliaProtocol.cs#L757](https://github.com/planetarium/libplanet/pull/580/files#diff-ae527d5afcdf41da4a57b594cab3cfefR757)).

limebell: Actually, `depth` became to decrease every step since #665. But this file does not behave as it.

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -617,6 +678,111 @@ private void ReceivePing(Ping ping)
...
+                if (findNeighboursTasks.All(findPeerTask => findPeerTask.IsFaulted))
+                {
+                    throw new TimeoutException(
+                        "Timeout exception occurred during ProcessFoundForSpecificAsync().");
+                }
+            }
+
+            if (foundSpecificPeer.Count > 0)
+            {
+                peerFound = foundSpecificPeer[0];
```
limebell: `foundSpecificPeer[0]` may not have same address with `target` because it may not exists on the network. How about compare with `target` before this statement and if it's different, just return `null` instead?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -617,6 +678,111 @@ private void ReceivePing(Ping ping)
...
+                    string.CompareOrdinal(
+                        Kademlia.CalculateDistance(peers[i].Address, target).ToHex(),
+                        Kademlia.CalculateDistance(closestKnown.Address, target).ToHex()
+                    ) < 1)
+                {
+                    findNeighboursTasks.Add(FindSpecificPeerAsync(
+                        history,
+                        target,
+                        peers[i],
+                        (depth == -1) ? depth : depth + 1,
```
limebell:
```suggestion
                        (depth == -1) ? depth : depth - 1,
```
Your logic in [KademliaProtocol.cs#L313](https://github.com/planetarium/libplanet/pull/580/files#diff-ae527d5afcdf41da4a57b594cab3cfefR313) expects this.

longfin:
>   Net/SwarmTest.cs(2501,1): error MEN005: File SwarmTest.cs must be no longer than 2500 lines (now 2550). [/home/vsts/work/1/s/Libplanet.Tests/Libplanet.Tests.csproj]

https://dev.azure.com/planetarium/libplanet/_build/results?buildId=2445&view=logs&jobId=eff0806a-f871-53fb-98ec-4f04cba4fef8&taskId=40d54594-fa31-5411-db1e-dbb432feffa2&lineStart=1274&lineEnd=1275&colStart=1&colEnd=1

Builds seem to fail due to file length of `SwarmTest.cs` 😢 

@RozzaysRed Could you increase the value of `MaxFileLines` in `Menees.Analyzers.Settings.xml`?

limebell:
We added patch to stabilize tests recently. The tests would be less fail if you rebase on our latest master 🙇‍♂️

limebell:
This resolves #570.

moreal:
@RozzaysRed It seems to need rebase.

dahlia:
Thanks for your contribution! 👍

PR #591, opened 2019-10-17
Execute actions before appending block

earlbread:
Fixed a bug that actions are executed after appending a block.


Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -419,6 +419,25 @@ public void AppendWithoutEvaluateActions()
             }
         }
 
+        [Fact]
+        public async Task AppendWhenActionEvaluationFailed()
```
limebell:
```suggestion
        public async Task AppendFailWhenActionEvaluationFailed()
```

CHANGES.md
```diff
@@ -68,6 +68,7 @@ To be released.
  -  Fixed a bug where `Swarm<T>` renders actions in same block multiple times
     when reorg happens.  [[#579]]
  -  `LiteDBStore` became to guarantee atomicity of storing blocks. [[#584]]
+ -  Fixed a bug that actions are executed after appending a block.  [[#591]]
```
```suggestion
 -  Fixed a bug that `BlockChain<T>` had appended a block even if fails to evaluate.  [[#591]]
```

Libplanet.Tests/Common/Action/ThrowException.cs
```diff
@@ -13,14 +12,15 @@ public ThrowException()
         {
         }
 
-        public bool ThrowOnRehearsal { get; set; } = false;
+        public bool ThrowOnRehearsal { get; set; }
 
-        public bool ThrowOnExecution { get; set; } = false;
+        public bool ThrowOnExecution { get; set; }
```
dahlia: Removing these values does affect nothing, right?

earlbread: Yeah, this does affect nothing.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -708,6 +708,11 @@ bool renderActions
                     nonceDeltas[txSigner] = nonceDelta + 1;
                 }
 
+                if (evaluateActions)
+                {
+                    ExecuteActions(block, renderActions);
```
dahlia: This causes inconsistency between states and a blockchain, since there would be a moment that states made from a block not appended yet are visible through `GetState()` method.

earlbread: `GetState()` will be okay for now because the `Tip` is not changed yet. But we need to address the inconsistency between states and a blockchain as you said.

earlbread:
Appended a commit to separate executing and rendering actions. https://github.com/planetarium/libplanet/pull/591/commits/2a33b18c491a7a086eec67524e103ba89951b099

Libplanet.Tests/Common/Action/ThrowException.cs
```diff
@@ -13,14 +12,18 @@ public ThrowException()
...
+
+        public bool ThrowOnRendering { get; set; }
 
         public IValue PlainValue =>
-            new Bencodex.Types.Dictionary(new Dictionary<IKey, IValue>
+            new Dictionary(new Dictionary<IKey, IValue>
```
dahlia: IMHO for this one line the previous code was better because it's hard to distinguish if it's from Bencodex or .NET Framework at a first glance.

earlbread: Agree. Added a commit https://github.com/planetarium/libplanet/pull/591/commits/e225a0af78ae4176d7795aba17b1b805ddfe32cf.

PR #596, opened 2019-10-21
Fix docs build

earlbread:
Fixed a problem where documents API could not be built after a7b5c157da399ada711f17ab178bafa0b8fedd2b.


earlbread:
Fixed to fail GitHub Actions if `_site/api/Libplanet.html` does not exist.

PR #608, opened 2019-10-23
Separate cache checking with removing

limebell:
This patch separates `KademliaProtocol<T>.CheckReplacementCacheAsync()` from `KademliaProtocol<T>.RemovePeerAsync()`.

### DIFF
--- a/Libplanet/Net/Protocols/KademliaProtocol.cs
+++ b/Libplanet/Net/Protocols/KademliaProtocol.cs
@@ -216,8 +216,9 @@ namespace Libplanet.Net.Protocols
         }
 
         /// <summary>
-        /// Checks peers in ReplacementCache in each <see cref="KBucket"/> of
-        /// <see cref="RoutingTable"/>.
+        /// Checks the <see cref="KBucket"/> in the <see cref="RoutingTable"/> and if
+        /// there is an empty <see cref="KBucket"/>, fill it with <see cref="Peer"/>s
+        /// in the <see cref="KBucket.ReplacementCache"/>.
         /// </summary>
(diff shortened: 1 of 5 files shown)


Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -215,6 +215,42 @@ public async Task RebuildConnectionAsync(CancellationToken cancellationToken)
             }
         }
 
+        /// <summary>
+        /// Checks peers in ReplacementCache in each <see cref="KBucket"/> of
+        /// <see cref="RoutingTable"/>.
```
earlbread: How about this?
```
Checks the <see cref="KBucket"/> in the <see cref="RoutingTable"/> and
if there is an empty <see cref="KBucket"/>, fill it with <see cref="Peer"/>s
in the <see cref="KBucket.ReplacementCache"/>.
```

Libplanet/Net/Swarm.cs
```diff
@@ -2217,6 +2218,25 @@ private void ValidateSender(Peer peer)
...
+        private async Task CheckReplacementCacheAsync(
+            TimeSpan period,
+            CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    await Task.Delay(period, cancellationToken);
+                    await Protocol.CheckReplacementCacheAsync(cancellationToken);
```
earlbread: How about running this after removing some peers in `Protocol.RefreshTableAsync` instead of running it regularly?

PR #612, opened 2019-10-24
Retry request when an exception occurs in ProcessRuntime

earlbread:
This fixes `ProcessRuntime` to retry when an exception occurs while processing the request.
Because NetMQ Runtime is introduced in 0.7.0, I skipped the changelog.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -2053,60 +2053,78 @@ namespace Libplanet.Net
             while (!cancellationToken.IsCancellationRequested)
             {
                 _logger.Verbose("Waiting for new request...");
-                var req = await _requests.TakeAsync(cancellationToken);
-                _logger.Verbose("Request {RequestId} taken.", req.Id);
+                MessageRequest req = await _requests.TakeAsync(cancellationToken);
 
-                using (var dealer = new DealerSocket(ToNetMQAddress(req.Peer)))
+                try
                 {
-                    dealer.Options.Linger = Timeout.InfiniteTimeSpan;
+                    await ProcessRequest(req, cancellationToken);
+                }
+                catch (Exception e)
+                {
                     _logger.Debug(
-                        "Trying to send {Message} to {PeerAddress}...",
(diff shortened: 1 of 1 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -2053,59 +2053,76 @@ private void DoReply(object sender, NetMQQueueEventArgs<Message> e)
...
 
-                using (var dealer = new DealerSocket(ToNetMQAddress(req.Peer)))
+                try
+                {
+                    await ProcessRequest(req, cancellationToken);
+
+                    // Delaying dealer disposing to avoid ObjectDisposedException on NetMQPoller
+                    await Task.Delay(100, cancellationToken);
+                }
+                catch (Exception e)
```
longfin: Some tests fail. it seems to need bypass some exceptions on some environments(i.e. mono).

earlbread: I misplaced the code that delays disposing of the dealer socket and fixed it.

Libplanet/Net/Swarm.cs
```diff
@@ -2053,60 +2053,78 @@ private void DoReply(object sender, NetMQQueueEventArgs<Message> e)
...
-                    dealer.Options.Linger = Timeout.InfiniteTimeSpan;
                     _logger.Debug(
-                        "Trying to send {Message} to {PeerAddress}...",
-                        req.Message,
-                        req.Peer.Address
+                        "Unexpected exception occurred during {ProcessRuntime}(). {e}",
+                        nameof(ProcessRuntime),
+                        e);
+                    await _requests.AddAsync(req, cancellationToken);
+                    await Task.Delay(100, cancellationToken);
```
limebell: How about using `final` instead of delay in both `ProcessRequest()` and `catch` block?

earlbread: `Task.Delay` in `ProcessRequest()` is for delaying dealer disposing to avoid `ObjectDisposedException` on `NetMQPoller`. So they have different purpose.

PR #622, opened 2019-10-26
Fix IceServerTest hanging

longfin:
This PR fixes URLs in `IceServerTest`'s fixture to avoid unexpected hanging.

### DIFF
--- a/Libplanet.Tests/Net/IceServerTest.cs
+++ b/Libplanet.Tests/Net/IceServerTest.cs
@@ -19,8 +19,8 @@ namespace Libplanet.Tests.Net
             var userInfo = turnUri.UserInfo.Split(':');
             var servers = new List<IceServer>()
             {
-                new IceServer(new[] { "turn:turn.does-not-exists.org" }),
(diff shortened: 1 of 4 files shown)


limebell:
It seems a test in Linux_Mono keep fails.

Libplanet.Tests/Net/IceServerTest.cs
```diff
@@ -19,8 +19,8 @@ public async Task CreateTurnClient()
             var userInfo = turnUri.UserInfo.Split(':');
             var servers = new List<IceServer>()
             {
-                new IceServer(new[] { "turn:turn.does-not-exists.org" }),
-                new IceServer(new[] { "stun:stun.l.google.com:19302" }),
+                new IceServer(new[] { "turn://turn.does-not-exists.org" }),
+                new IceServer(new[] { "stun://stun.l.google.com:19302" }),
```
dahlia: How about making the `IceServer(string[])` constructor to throw an `ArgumentException` when the given URLs are unacceptable?

PR #629, opened 2019-10-29
Fixed a bug peer getter in KBucket throws ArgumentOutOfRangeException

limebell:
Peer getter in `KBucket` was throwing `ArgumentOutOfRangeException` when list `_peers` is empty.


longfin:
Can we get a test case related to it?

PR #637, opened 2019-10-30
Fix GetNextTxNonce when transactions with old nonce are staged

earlbread:
Fixed a bug where `BlockChain<T>.GetNextTxNonce` only returned the same nonce when transactions with an old nonce were staged.

### DIFF
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -1289,6 +1289,34 @@ namespace Libplanet.Tests.Blockchain
         }
 
         [Fact]
+        public async Task GetNextTxNonceWithStaleTx()
+        {
+            var privateKey = new PrivateKey();
+            var address = privateKey.PublicKey.ToAddress();
+            var actions = new[] { new DumbAction(address, "foo") };
+
+            Transaction<DumbAction>[] txs =
+            {
+                _fx.MakeTransaction(actions, privateKey: privateKey),
+                _fx.MakeTransaction(actions, privateKey: privateKey, nonce: 1),
+            };
+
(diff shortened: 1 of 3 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -526,26 +526,34 @@ public void UnstageTransactions(ISet<Transaction<T>> transactions)
...
+            try
             {
-                if (n != prevNonce + 1)
+                long nonce = Store.GetTxNonce(Id, address);
+                var stagedTxNonces = Store.IterateStagedTransactionIds()
+                    .Select(Store.GetTransaction<T>)
+                    .Where(tx => tx.Signer.Equals(address) && tx.Nonce >= nonce)
+                    .Select(tx => tx.Nonce)
+                    .OrderBy(n => n)
+                    .ToList();
```
longfin: I guess it's okay that we can skip `.ToList()` because we use `stagedTxNonces` only once.

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -1358,6 +1386,31 @@ public void MakeTransaction()
...
+        public void MakeTransactionConcurrency()
+        {
+            var privateKey = new PrivateKey();
+            Address address = privateKey.PublicKey.ToAddress();
+            var actions = new[] { new DumbAction(address, "foo") };
+
+            var tasks = Enumerable.Range(0, 10)
+                .Select(_ => Task.Run(() => _blockChain.MakeTransaction(privateKey, actions)));
+
+            Task.WhenAll(tasks).Wait();
```
longfin: How about `await Task.WhenAll(tasks)` and marking this test as `async`?

Issue #634, opened 2019-10-30
Keep downloaded blocks during IBD


limebell:
During IBD, peer sends a stack of blocks (which is now 500 blocks). If preloading is aborted, all blocks are gone even only one stack is remained. I guess block should be stored every time finished to fetch each stack of blocks.

PR #644, opened 2019-10-31
SyncPreviousBlocksAsync deletes chains only created during FillBlocksAsync

limebell:
Fixed a bug that chain is reset when IBD fails.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -102,8 +102,8 @@ To be released.
     through the TurnClient.
  -  Fixed a bug where `BlockChain<T>.GetNextTxNonce` only returned the same
     nonce when transactions with an old nonce were staged.  [[#637]]
- -  Fixed a bug that blockchain reset when `Swarm<T>.PreloadAsync()` fails.
-    [[#644]]
+ -  Fixed a bug that `BlockChain<T>` had reset when `Swarm<T>.PreloadAsync()`
+    fails.  [[#644]]
 
 [#209]: https://github.com/planetarium/libplanet/issues/209
 [#405]: https://github.com/planetarium/libplanet/issues/405
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -102,6 +102,8 @@ To be released.
     through the TurnClient.
  -  Fixed a bug where `BlockChain<T>.GetNextTxNonce` only returned the same
     nonce when transactions with an old nonce were staged.  [[#637]]
+ -  Fixed a bug that blockchain reset when `Swarm<T>.PreloadAsync()` fails.
+    [[#644]]
 
 [#209]: https://github.com/planetarium/libplanet/issues/209
 [#405]: https://github.com/planetarium/libplanet/issues/405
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -98,6 +98,8 @@ To be released.
     connection through a TURN relay server. [[#453], [#599]]
  -  Fixed a bug that `KeyNotFoundException` occurred when sending a message
     through the TurnClient.
+ -  Fixed a bug that blockchain reset when `Swarm<T>.PreloadAsync()` fails.
```
```suggestion
 -  Fixed a bug that `BlockChain<T>` had reset when `Swarm<T>.PreloadAsync()` fails.
```

Libplanet/Net/Swarm.cs
```diff
@@ -1650,107 +1641,129 @@ CancellationToken cancellationToken
         )
         {
             BlockChain<T> workspace = blockChain;
+            List<Guid> scope = new List<Guid>();
```
longfin: Trivial, just for consistency.

```suggestion
            var scope = new List<Guid>();
```

Libplanet/Net/Swarm.cs
```diff
@@ -1572,15 +1572,6 @@ CancellationToken cancellationToken
                         blockChain.Swap(synced, evaluateActions);
                     }
                 }
-
-                IStore store = BlockChain.Store;
-                foreach (Guid chainId in store.ListChainIds().ToList())
-                {
-                    if (!chainId.Equals(blockChain.Id))
```
earlbread: How about change this condition like below?
 
`!(chainId.Equals(blockChain.Id) || chainId.Equals(BlockChainId)`

limebell: Do you mean `BlockChain.Id`?

CHANGES.md
```diff
@@ -98,6 +98,8 @@ To be released.
...
+    [[#644]]
```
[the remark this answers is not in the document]
limebell: I thought this PR address different issue with #634 because this patch is focusing on keeping blocks that existed before IBD.

Libplanet/Net/Swarm.cs
earlbread: Right, `BlockChain.Id`. There was a typo.

Libplanet/Net/Swarm.cs
```diff
@@ -1656,107 +1647,129 @@ CancellationToken cancellationToken
...
+                if (!(workspace is null))
+                {
+                    IStore store = blockChain.Store;
+                    store.DeleteChainId(workspace.Id);
+                }
+            }
+            finally
+            {
+                IStore store = blockChain.Store;
+                foreach (var id in scope.Where(guid => guid != workspace.Id))
```
dahlia: How about inserting an `if` statement in the `foreach` loop, instead of `Where()` method?

Libplanet/Net/Swarm.cs
limebell: This was intended to remove chains that created by `FillBlocksAsync()`, so removing chains in it is better IMO.

Libplanet/Net/Swarm.cs
limebell: Rider recommended to use this.

PR #645, opened 2019-10-31
Calculate state when evaluating an action if the block state doesn't exist

earlbread:
When there is no block state when evaluating an action, it is changed to calculate the state from the beginning to avoid `IncompleteBlockStatesException`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -74,6 +74,10 @@ To be released.
     not related with [Kademlia protocol][Kademlia].  [[#594], [#627]]
  -  `Swarm<T>` became not to check least recently used peer every time when
     new peer is fetched.  [[#627]]
+ -  `IAction` became guaranteed that the given
+    `IActionContext.PreviousStates.GetState()` never throws
+    `IncompleteBlockStatesException`. Instead, now it may calculate the
+    incomplete states from the beginning if necessary.  [[#645]]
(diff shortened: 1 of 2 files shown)


earlbread:
> As I know there's the reason why we want to calculate incomplete states. So how about manually turning on the `completeStates` option for the problematic case while maintaining the default value turned off, rather than making the `completeStates` option turned on by default?

To do that, we need to add a flag to `EvaluateActions`, `ExecuteActions`, and `Append`. I thought it is too complicated for the case.

earlbread:
Changed to calculate the state from the beginning when there is no block state when evaluating an action instead of changing the default value of `completeStates` parameter.

CHANGES.md
```diff
@@ -74,6 +74,8 @@ To be released.
     not related with [Kademlia protocol][Kademlia].  [[#594], [#627]]
  -  `Swarm<T>` became not to check least recently used peer every time when
     new peer is fetched.  [[#627]]
+ -  When there is no block state when evaluating an action, it is changed to
+    calculate the state from the beginning.  [[#645]]
```

> `IAction` became guaranteed that the given `IActionContext.PreviousStates.GetState()` never throws `IncompleteBlockStatesException`.  Instead, now it may calculate the incomplete states from the beginning if necessary.

Issue #616, opened 2019-10-25
Make PutBlock<T>() idempotent


earlbread:
When we call `LiteDBStore.PutBlock<T>()` with a block that already exists, the block is rewritten. However, if a block already exists, doing nothing is better for performance.

mahditakrim:
Hi @earlbread ,
I agree with you (good performance lover ;D) .
May I take a look ?

earlbread:
Sure, thanks! @mahditakrim

mahditakrim:
Dear @earlbread ,
I just sent a pull request but 1 check failed , it's changelog.

mahditakrim:
Dear @earlbread ,
I took a look at the codes, and I got questions:

1. How to know that a block is already exists, I mean I know that blocks got unique hash codes so how to compare a block hash code to other hash codes (if I'm right) ?

2. Is it better to have a method which can determine that a block exists **or** just put the if statement in the usage code ?

earlbread:
1. You can check whether a block already exists by using `_db.FileStorage.FindById()` method and `BlockFileId()` method. Please refer to this commit 19c613241a93a3c5d63a768bd3f78ed900bbf6ab.

2. I think the `if` statement is enough.

3. To build and test this project, please refer to this document. https://github.com/planetarium/libplanet/blob/master/CONTRIBUTING.md

mahditakrim:
> 1. You can check whether a block already exists by using `_db.FileStorage.FindById()` method and `BlockFileId()` method. Please refer to this commit [19c6132](https://github.com/planetarium/libplanet/commit/19c613241a93a3c5d63a768bd3f78ed900bbf6ab).
> 2. I think the `if` statement is enough.
> 3. To build and test this project, please refer to this document. https://github.com/planetarium/libplanet/blob/master/CONTRIBUTING.md
> 
> Please let me know if you have any questions.

Thanks

earlbread:
`FileStorage.FindById` is from LiteDB(https://github.com/mbdavid/LiteDB/wiki/FileStorage). If the file exists, it returns `LiteFileInfo`, and if it is not `null`, we don't need to write the block again.

See `LiteDBStore.PutTransaction` method to know how to use `_db.FileStorage.FindById() ` method.(https://github.com/planetarium/libplanet/blob/master/Libplanet/Store/LiteDBStore.cs#L315).

mahditakrim:
[...]

PR #647, opened 2019-11-01
PutBlock method checks whether a block exists. [skip changelog]

mahditakrim:
Resolving https://github.com/planetarium/libplanet/issues/616 ,
PutBlock method in LiteDBStore.cs checks whether a block exists and if the answer is yes it returns from the method without writing the block (performance upgrades).

### DIFF
--- a/Libplanet/Store/LiteDBStore.cs
+++ b/Libplanet/Store/LiteDBStore.cs
@@ -376,6 +376,12 @@ namespace Libplanet.Store
         /// <inheritdoc/>
         public override void PutBlock<T>(Block<T> block)
         {
+            // checks whether a block already exists or not.
+            if (!(_db.FileStorage.FindById(BlockFileId(block.Hash)) is null))
+            {
+                return;
+            }
+
             _blockLock.EnterWriteLock();
             try
             {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -74,6 +74,8 @@ To be released.
     not related with [Kademlia protocol][Kademlia].  [[#594], [#627]]
  -  `Swarm<T>` became not to check least recently used peer every time when
     new peer is fetched.  [[#627]]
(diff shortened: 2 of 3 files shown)


Libplanet/Store/LiteDBStore.cs
```diff
@@ -376,6 +376,12 @@ public override IEnumerable<HashDigest<SHA256>> IterateBlockHashes()
         /// <inheritdoc/>
         public override void PutBlock<T>(Block<T> block)
```
earlbread: Since the behavior of the method has changed, it should be written to the changelog and method documentation.

Could you add this to the changelog behavioral change section?
```
 -  `IStore.PutBlock<T>()` became to do nothing when it takes
    the `Block<T>` more than once.  [[#647]]
```

And could you add this to the `IStore.PutBlock<T>` documentation summary section?
```
If the same block already exists in the store it does nothing.
```

Libplanet/Store/LiteDBStore.cs
```diff
@@ -376,6 +376,12 @@ public override IEnumerable<HashDigest<SHA256>> IterateBlockHashes()
...
         {
+            // checks whether a block already exists or not.
+            if ((_db.FileStorage.FindById(BlockFileId(block.Hash)) is null) == false)
```
earlbread: How about this?
```suggestion
            if (!(_db.FileStorage.FindById(BlockFileId(block.Hash)) is null))
```

mahditakrim: I prefer simple and readable codes, but if you want it that way, that's alright.

Libplanet/Store/LiteDBStore.cs
mahditakrim: Ok

Issue #642, opened 2019-10-31
Key derivation using Scrypt


dahlia:
The previous patch #614 had implemented only minimum compliance:

> At the moment, it implements only minimum compliance according to the [spec][1], `pbkdf2` and `aes-128-ctr`:
>
> > PBKDF2 must be supported by all minimally-compliant implementations, denoted though:
> > - `kdf`: `pbkdf2`
>
> > All minimally-compliant implementations must support the AES-128-CTR algorithm, denoted through:
> > - `cipher`: `aes-128-ctr`

However, Geth, one of the most widely used Ehterum client, uses Scrypt instead of PBKDF2 to derive keys from user-input passphrases.  For better interoperability, we need to implement `"kdf": "pbkdf"` option as well.  Here's a sample keystore file generated by Geth:

```json
{
  "address": "8e5f4b9b8f84ff90c559c6a4deb3c1febe551f29",
  "crypto": {
    "cipher": "aes-128-ctr",
    "cipherparams": {
      "iv": "cc730222e304f4c230854c63f105ff5b"
    },
    "ciphertext": "45a5b80fd3ea8f4f0db824b9b56b42be2c1e3759e77c0197cfa3a136f18d5957",
    "kdf": "scrypt",
    "kdfparams": {
      "dklen": 32,
      "n": 262144,
      "p": 1,
      "r": 8,
      "salt": "3ada4bad7033fdaa65684ecf93dab3186e48d3f80965070d5fc9aac07811dc8f"
    },
    "mac": "445c0bbc6c310d1fcabf7d7456fd3c558e504c4d308ddb83eddb0b2b8f66f834"
  },
  "id": "29ee8172-710d-44b3-9fe9-f401f7b35ad5",
  "version": 3
}
```

Notes for who implement this:

- There's `IKdf` interface.  We need to add a new class named `Scrypt` which implements the interface.
- `Pbkdf2` class, which also implements `IKdf`, would be good to read first.
- For deserialization, a new `case` is also need for the `switch` statement in the `ProtectedPrivateKey.FromJson()` method.
- [Norgerman.Cryptography.Scrypt][2] looks decent.

[1]: https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition
[2]: https://www.nuget.org/packages/Norgerman.Cryptography.Scrypt/

minhoryang:
#654 handled this.

PR #654, opened 2019-11-06
[WIP] Added Key derivation using Scrypt.

minhoryang:
Related: #642

TODO:
- Code Style
- Comments
- Variable Naming
- Necessary Tests: ...?

Questions:
- [Scrypt require some criteria for parameters.](https://github.com/Norgerman/Scrypt/blob/master/src/Norgerman.Cryptography.Scrypt/ScryptUtil.cs#L16-L19) Which exception should I fire for this?


CHANGES.md
```diff
@@ -52,6 +52,7 @@ To be released.
  -  Added `BlockChain<T>.GetTransaction(TxId)` method.  [[#409], [#583]]
  -  Added `BlockChain<T>.Contains(TxId)` method.  [[#409], [#583]]
  -  Added `ByteUtil.Hex(ImmutableArray<byte>)` overloaded method.  [[#614]]
+ -  Added `Scrypt` class.  [[#642]]
```
dahlia: It would be better to be placed right after `Pbkdf2`.

Libplanet/KeyStore/Kdfs/Scrypt.cs
```diff
@@ -0,0 +1,211 @@
...
+            Salt = salt;
+            KeyLength = keyLength;
+            R = r;
+            P = p;
+        }
+
+        /// <summary>
+        /// CPU/memory cost parameter.
+        /// </summary>
+        public int N { get; }
```
dahlia: How about naming it in more human-readable way, like `Cost`?  As these letters are conventional, it would be good if we mention that in the docs like:

```xml
Corresponds to <c>N</c>.
```

minhoryang:
Q. Current implementation of [`ProtectedPrivateKey.Protect()`](https://github.com/planetarium/libplanet/blob/ef6506e2b666108c6ffaf7a4e85c1ae500aca589/Libplanet/KeyStore/ProtectedPrivateKey.cs#L105-L125) use Pbkdf2 only. Do we add the option or selector for Scrypt?

Libplanet/KeyStore/Kdfs/Scrypt.cs
```diff
@@ -0,0 +1,250 @@
...
+    /// <a href="https://en.wikipedia.org/wiki/Scrypt">Scrypt</a>.
+    /// </summary>
+    /// <typeparam name="T">PRF (pseudorandom function) to use, e.g.,
+    /// <see cref="Sha256Digest"/>.</typeparam>
+    [SuppressMessage(
+        "Microsoft.StyleCop.CSharp.ReadabilityRules",
+        "SA1402",
+        Justification = "There are just generic & non-generic versions of the same name classes.")]
+    [Pure]
+    public sealed class Scrypt<T> : IKdf
```
earlbread: It seems this type parameter `T` is not used in this class. Do we need this?

Libplanet/KeyStore/Kdfs/Scrypt.cs
```diff
@@ -0,0 +1,250 @@
...
+            int keyLength,
+            int parallelization,
+            int blockSize
+        )
+        {
+            if (cost < 2 || (cost & (cost - 1)) != 0)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(cost),
+                    $"Cost must be a power of 2 greater than 1!"
```
earlbread:
```suggestion
                    "Cost must be a power of 2 greater than 1!"
```

Libplanet/KeyStore/Kdfs/Scrypt.cs
```diff
@@ -0,0 +1,250 @@
...
+                );
+            }
+
+            if (cost > int.MaxValue / 128 / blockSize)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(cost),
+                    $"Parameter cost is too large!"
```
earlbread:
```suggestion
                    "Parameter cost is too large!"
```

Libplanet.Tests/KeyStore/ProtectedPrivateKeyTest.cs
```diff
@@ -694,6 +817,16 @@ public void WriteJson()
...
             Assert.Equal(AddressFixture, key.Unprotect(PassphraseFixture).PublicKey.ToAddress());
+
+            using (var stream = new MemoryStream())
+            {
+                Fixture2.WriteJson(stream);
+                json = Encoding.UTF8.GetString(stream.ToArray());
+            }
+
+            ProtectedPrivateKey key2 = ProtectedPrivateKey.FromJson(json);
+            Assert.Equal(AddressFixture, key.Address);
```
earlbread: Shouldn't this be like below?
```suggestion
            Assert.Equal(AddressFixture2, key2.Address);
```

Libplanet.Tests/KeyStore/ProtectedPrivateKeyTest.cs
```diff
@@ -694,6 +817,16 @@ public void WriteJson()
...
+            Assert.Equal(AddressFixture, key.Unprotect(PassphraseFixture).PublicKey.ToAddress());
```
earlbread: Shouldn't this be like below?
```suggestion
            Assert.Equal(AddressFixture2, key2.Unprotect(PassphraseFixture).PublicKey.ToAddress());
```

Libplanet.Tests/KeyStore/ProtectedPrivateKeyTest.cs
```diff
@@ -310,7 +406,7 @@ public void FromJsonInvalidCases()
 
             Assert.Throws<InvalidKeyJsonException>(() =>
                 load(@"{
-                    ""address"": ""d80d933db45cc0cf69e9632090f8aaff635dc8e50"",  // Invalid length
+                    ""address"": ""d80d933db45cc0cf69e9632090f8aaff635dc8e500"",  // Invalid length
```
earlbread: What is this change for?

Libplanet/KeyStore/Kdfs/Scrypt.cs
minhoryang: That's a breadcrumb when I copied and pasted from Pbkdf2, and Norgerman.Cryptography.Scrypt use SHA256 digest internally, so there are no reason to use <T> in this class. I'll remove it.

Libplanet.Tests/KeyStore/ProtectedPrivateKeyTest.cs
minhoryang: I tried to increase the coverage of [`ProtectedPrivateKey.FromJson()`'s address validation logic](https://github.com/planetarium/libplanet/blob/ef6506e2b666108c6ffaf7a4e85c1ae500aca589/Libplanet/KeyStore/ProtectedPrivateKey.cs#L262-L268).
- `GetHexProperty()` called `ByteUtil.ParseHex()` internally
- `ByteUtil.ParseHex()` checked the length of input are even, if not throw `ArgumentOutOfRangeException`
- when `GetHexProperty()` met `Exception`, rethrow it as `InvalidKeyJsonException`
- and that's now catched at `try {} catch (ArgumentException e) {}`

Libplanet.Tests/KeyStore/Kdfs/ScryptTest.cs
```diff
@@ -0,0 +1,286 @@
+using System;
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
+using Libplanet.KeyStore;
+using Libplanet.KeyStore.Kdfs;
+using Org.BouncyCastle.Crypto.Digests;
```
earlbread: This doesn't seem to be used.

Libplanet/KeyStore/Kdfs/Scrypt.cs
```diff
@@ -0,0 +1,244 @@
...
+
+namespace Libplanet.KeyStore.Kdfs
+{
+    /// <summary>
+    /// <a href="https://en.wikipedia.org/wiki/Scrypt">Scrypt</a>.
+    /// </summary>
+    [SuppressMessage(
+        "Microsoft.StyleCop.CSharp.ReadabilityRules",
+        "SA1402",
+        Justification = "There are just generic & non-generic versions of the same name classes.")]
```
earlbread: We don't need this now because there is only one `Scrypt` class.

Libplanet/KeyStore/Kdfs/Scrypt.cs
```diff
@@ -0,0 +1,244 @@
+using System;
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
```
earlbread: This can be removed.

Libplanet/KeyStore/Kdfs/Scrypt.cs
```diff
@@ -0,0 +1,244 @@
...
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Text.Json;
+using Org.BouncyCastle.Crypto.Digests;
```
earlbread: This doesn't seem to be used.

dahlia:
> Current implementation of `ProtectedPrivateKey.Protect()` use Pbkdf2 only. Do we add the option or selector for Scrypt?

That seems a good idea.  How about having an optional parameter `IKdf kdf`?  Also, the current `Pbkdf2` settings could be a good default IMHO.

Libplanet.Tests/KeyStore/Kdfs/ScryptTest.cs
minhoryang: Oh! I missed this when I purged <T> (Sha256Digest).
Is there any linter or vscode extension for preventing this anymore?

dahlia: I hope we have a such lint if there is anything. 🤔

minhoryang:
@dahlia I'm totally agreed to finish this PR without `IKdf kdf` parameter. Now I'm hesitating to implement that or not. Go for merge!

PR #656, opened 2019-11-06
Store benchmarks

dahlia:
Added some benchmarks on basic `IStore` operations.  Also wrote minimal docs on how to run benchmarks.  As benchmarks take too long time on the CI, `-j short` option is also added.

### DIFF
--- a/Libplanet.Benchmarks/Store.cs
+++ b/Libplanet.Benchmarks/Store.cs
@@ -1,0 +1,198 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using BenchmarkDotNet.Attributes;
+using Libplanet.Blocks;
+using Libplanet.Crypto;
+using Libplanet.Store;
+using Libplanet.Tests;
+using Libplanet.Tests.Common.Action;
+using Libplanet.Tests.Store;
+using Libplanet.Tx;
+
+namespace Libplanet.Benchmarks
+{
+    public class Store
+    {
+        private readonly ImmutableArray<Block<DumbAction>> Blocks = default;
+        private readonly int BlocksCount = default;
+        private readonly ImmutableArray<Transaction<DumbAction>> Txs = default;
+        private StoreFixture _fx = null;
+        private int TxsCount = default;
+        private IStore _store = null;
+
+        public Store()
+        {
+            var blocks = new List<Block<DumbAction>>();
+            var txs = new List<Transaction<DumbAction>>();
+            Block<DumbAction> block = TestUtils.MineGenesis<DumbAction>();
+            blocks.Add(block);
+            var key = new PrivateKey();
+            long nonce = 0;
+            for (int i = 0; i < 500; i++)
+            {
+                var blockTxs = new List<Transaction<DumbAction>>();
+                for (int j = 0; j < i % 5; j++)
+                {
+                    blockTxs.Add(Transaction<DumbAction>.Create(nonce++, key, new DumbAction[0]));
+                }
+                block = TestUtils.MineNext(block, blockTxs);
+                blocks.Add(block);
+                txs.AddRange(blockTxs);
+            }
+            Blocks = blocks.ToImmutableArray();
+            BlocksCount = Blocks.Length;
+            Txs = txs.ToImmutableArray();
+            TxsCount = Txs.Length;
+        }
+
+        [IterationSetup]
+        public void InitializeFixture()
+        {
+            _fx = new LiteDBStoreFixture();
+            _store = _fx.Store;
+        }
+
+        [IterationCleanup]
+        public void FinalizeFixture()
+        {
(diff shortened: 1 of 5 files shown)


Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        [IterationSetup(
+            Targets = new []
+            {
+                nameof(PutTxOnManyTxs),
+                nameof(GetTx),
+            }
+        )]
+        public void PutManyTxs()
```
earlbread: This looks the same as `PutManyBlocks()`. Is it intended?

Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,147 @@
...
+        )]
+        public void PutManyBlocks()
+        {
+            InitializeFixture();
+            int i = 0;
+            foreach (Block<DumbAction> block in Blocks)
+            {
+                _store.PutBlock(block);
+                i++;
+                if (i >= Blocks.Length - 1)
```
earlbread: What is this for?

Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        private static readonly ImmutableArray<Block<DumbAction>> Blocks;
+        private static readonly int BlockCount;
+        private static readonly ImmutableArray<Transaction<DumbAction>> Txs;
+        private static readonly int TxsCount;
+        private IStore _store = null;
+
+        static Store()
+        {
+            List<Block<DumbAction>> blocks = new List<Block<DumbAction>>();
```
longfin: Trivial, just for consistency.
```suggestion
            var blocks = new List<Block<DumbAction>>();
```

Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,147 @@
...
+            List<Transaction<DumbAction>> txs = new List<Transaction<DumbAction>>();
```
longfin: Trivial, just for consistency.
```suggestion
            var txs = new List<Transaction<DumbAction>>();
```

Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,147 @@
...
+            Block<DumbAction> block = TestUtils.MineGenesis<DumbAction>();
+            blocks.Add(block);
+            PrivateKey key = new PrivateKey();
```
longfin: Trivial, just for consistency.

```suggestion
            var key = new PrivateKey();
```

Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,147 @@
...
+            long nonce = 0;
+            for (int i = 0; i < 500; i++)
+            {
+                List<Transaction<DumbAction>> blockTxs = new List<Transaction<DumbAction>>();
```
longfin: Trivial, just for consistency.

```suggestion
                var blockTxs = new List<Transaction<DumbAction>>();
```

Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,147 @@
...
+            Targets = new []
+            {
+                nameof(PutFirstEmptyBlock),
+                nameof(PutFirstBlockWithTxs),
+                nameof(PutFirstTx),
+            }
+        )]
+        public void InitializeFixture()
+        {
+            var fx = new LiteDBStoreFixture();
```
longfin: Does `fx` will be disposed of automatically?

Libplanet.Benchmarks/Store.cs
dahlia: No, you're right.  I'm going to fix this.

Libplanet.Benchmarks/Store.cs
dahlia: It's intended to avoid to put the last block until this initialization phase, and the last one is for the benchmark method.

Libplanet.Benchmarks/Store.cs
dahlia: I guess it wouldn't.  I will add cleanup functions too.

Libplanet.Benchmarks/Store.cs
```diff
@@ -0,0 +1,183 @@
...
+            _fx = new LiteDBStoreFixture();
+            _store = _fx.Store;
+        }
+
+        [IterationCleanup]
+        public void FinalizeFixture()
+        {
+            if (_store is IDisposable disposableStore)
+            {
+                disposableStore.Dispose();
```
longfin: I guess it's okay that we only dispose of `_fx` because `LiteDBStoreFixture` has a responsibility to disposing of its own resources include `_fx.Store`.

PR #657, opened 2019-11-06
Update routing table after reply done

limebell:
Previous PR had made `KademliaProtocol<T>.ReceivePingAsync()` to update routing table after replied. This patch makes all type of protocol message to update routing table after replied. Thus, `KademliaProtocol<T>.ReceivePing()` and `KademliaProtocol<T>.ReceiveFindPeer()` are no more `async`.

### DIFF
--- a/Libplanet/Net/Protocols/KademliaProtocol.cs
+++ b/Libplanet/Net/Protocols/KademliaProtocol.cs
@@ -255,23 +255,21 @@ namespace Libplanet.Net.Protocols
             _logger.Debug("Replacement cache checked.");
         }
 
-#pragma warning disable CS4014
+#pragma warning disable CS4014 // To run UpdateAsync() without await.
         public void ReceiveMessage(Message message)
         {
             switch (message)
             {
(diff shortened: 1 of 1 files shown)


Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -255,25 +255,21 @@ public async Task CheckReplacementCacheAsync(CancellationToken cancellationToken
...
-                    ReceiveFindPeerAsync(findPeer);
-                    break;
-
-                default:
-                    UpdateAsync(message.Remote);
+                    ReceiveFindPeer(findPeer);
                     break;
             }
+
+            var t = UpdateAsync(message?.Remote);
```
longfin: if we don't want to `await` for `UpdateAsync()`, how about using `#pragma warning disable CS4014` and writing comment about it?

limebell: Ok, I'll revert changes.

Issue #664, opened 2019-11-11
Remove Sonar Analyzer suppressions


earlbread:
We need to remove [Sonar Analyzer](https://github.com/SonarSource/sonar-dotnet) suppressions.

- [x] S927: Rename parameter 'A' to 'B' to match the interface declaration.
- [x] S1066: Merge this if statement with the enclosing one.
- [x] S1118: Add a 'protected' constructor or the 'static' keyword to the class  declaration.
- [ ] S1125: Remove the unnecessary Boolean literal(s).
- [ ] S1450: Remove the field '_bucketSize' and declare it as a local variable in the relevant methods.
- [ ] S1481: Remove the unused local variable 'x'.
- [ ] S2234: Parameters to 'Equal' have the same names but not the same order as the method arguments.
- [ ] S3358: Extract this nested ternary operation into an independent statement.
- [ ] S3442: Change the visibility of this constructor to 'protected'.
- [ ] S4035: Seal class or implement 'IEqualityComparer<T>' instead.
- [ ] S4201: Remove this unnecessary null check; 'is' returns false for nulls.

hmc0105:
I want to do deleting suppression S927

dahlia:
We suppressed S1128 warnings too here and there in order to work around the bug <https://github.com/SonarSource/sonar-dotnet/issues/2694>, but now it seems fixed (<https://github.com/SonarSource/sonar-dotnet/pull/2798>).  So we should restore these suppressed warnings now.

Oaroomy:
I'm going to work on S1066

Oaroomy:
I'm going to work on S1118

stylesm:
I'm going to work on the remaining surpressions here.

OnedgeLee:
Seems to have been resolved already.
All checklists are marked except S1481, and when I opened `Libplanet.ruleset`, could find lines below.
```
    <!-- Remove the unused local variable 'x'. -->
    <Rule Id="S1481" Action="None" />
```

sindydwns:
✋

PR #667, opened 2019-11-12
Remove Sonar Analyzer suppression S927

hmc0105:
It's related with #664.


longfin:
@hmc0105 Thanks for your first contribution!

It had seemed to be out-of-date, so I merged `master` into your branch(`hmc0105:delete-suppression-s927`).

PR #672, opened 2019-11-12
add total difficulty in block

h2oche:


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -557,6 +557,7 @@ public long GetNextTxNonce(Address address)
                 .IterateStagedTransactionIds()
                 .Select(Store.GetTransaction<T>)
                 .Where(tx => tx.Nonce < GetNextTxNonce(tx.Signer));
+            Block<T> prevBlock = Store.GetBlock<T>(prevHash.Value);
```
limebell: Since `prevHash` is `nullable`, we have to prevent error that may caused by the case. I suggest as follow:
```suggestion
            Block<T> prevBlock = prevHash is null ? null : Store.GetBlock<T>(prevHash.Value);
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -576,6 +577,7 @@ void WatchTip(object target, TipChangedEventArgs args)
                     () => Block<T>.Mine(
                         index: index,
                         difficulty: difficulty,
+                        previousTotalDifficulty: prevBlock.TotalDifficulty,
```
limebell: Since prevBlock may not exist (which means this is the first block), the statement may cause error. I suggest:
```suggestion
                        previousTotalDifficulty: prevBlock?.TotalDifficulty ?? 0,
```

limebell:
And please sign CLA so that we can use your code, and add changelogs for what you did :) @h2oche

PR #675, opened 2019-11-14
Fix the canonical chain is deleted

earlbread:
This fixed a bug where the canonical chain would be deleted if block download failed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,8 +26,11 @@ To be released.
 ### Behavioral changes
 
 ### Bug fixes
+ -  Fixed a bug where the canonical chain would be deleted if block download
+    failed.  [[#675]]
 
 [#662]: https://github.com/planetarium/libplanet/pull/662
+[#675]: https://github.com/planetarium/libplanet/pull/675
 
 
 Version 0.7.0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,8 +26,8 @@ To be released.
 ### Behavioral changes
 
 ### Bug fixes
- -  Fixed a bug where the canonical chain would be deleted if block download
-    failed.  [[#675]]
+ -  Fixed a bug where the canonical chain could be deleted if `Swarm<T>` failed
+    to download blocks due to network connection.  [[#675]]
 
 [#662]: https://github.com/planetarium/libplanet/pull/662
 [#675]: https://github.com/planetarium/libplanet/pull/675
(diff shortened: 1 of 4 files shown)


Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2388,6 +2388,40 @@ public async Task CreateNewChainWhenBranchPointNotExist()
...
+            var block = await chainA.MineBlock(_fx1.Address1);
+
+            try
+            {
+                await StartAsync(swarmA);
+                await StartAsync(swarmB);
+                await BootstrapAsync(swarmA, swarmB.AsPeer);
+
+                swarmA.BroadcastBlocks(new[] { block });
+                await Task.Delay(100);
```
dahlia: This scenario seems likely to make the test flaky, because `swarmA.BroadcastBlocks(new[] { block })` could be faster than we assume… how about watching `chainB.TipChanged` event instead?

Libplanet/Net/Swarm.cs
```diff
@@ -1560,18 +1560,7 @@ CancellationToken cancellationToken
...
-                    // We can't recover with OperationCanceledException and
-                    // ObjectDisposedException. so just re-throw them.
-                    catch (ObjectDisposedException)
-                    {
-                        throw;
-                    }
-                    catch (OperationCanceledException)
-                    {
-                        throw;
-                    }
```
dahlia: Why did you get rid of these rethrowings?

CHANGES.md
```diff
@@ -26,8 +26,11 @@ To be released.
 ### Behavioral changes
 
 ### Bug fixes
+ -  Fixed a bug where the canonical chain would be deleted if block download
```
```suggestion
 -  Fixed a bug where the canonical chain could be deleted if `Swarm<T>` failed
```

CHANGES.md
```diff
@@ -26,8 +26,11 @@ To be released.
...
+    failed.  [[#675]]
```
```suggestion
    to download blocks due to network connection.  [[#675]]
```

Libplanet.Tests/Net/SwarmTest.cs
earlbread: Unfortunately, we can't use the `TipChanged` event because block downloading should be failed before block appending is complete.

Libplanet/Net/Swarm.cs
earlbread: I changed it to only catch `TimeoutException` here and throw all other exceptions.

Libplanet.Tests/Net/SwarmTest.cs
dahlia: Could we make a new event to test this then?

earlbread: I added `Swarm<T>. FillBlocksAsyncFailed` for this test case.

PR #676, opened 2019-11-14
Move BlockChain<T>.Contains(TxId) to IStore

longfin:
This PR moves `BlockChain<T>.Contains(TxId)` to `IStore` to separate the global transaction view from `BlockChain<T>`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@ To be released.
 ### Backward-incompatible API changes
 
  -  Removed `LiteDBStore` class.  Use `DefaultStore` instead.  [[#662]]
+ -  Removed `BlockChain<T>.Contains(TxId)` method.
+    Use `IStore.ContainsTransaction(TxId)` instead.  [[#676]]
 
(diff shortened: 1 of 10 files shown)


CHANGES.md
```diff
@@ -9,6 +9,8 @@ To be released.
 ### Backward-incompatible API changes
 
  -  Removed `LiteDBStore` class.  Use `DefaultStore` instead.  [[#662]]
+ -  Removed `BlockChain<T>.Contains(TxId)` method. Use
```
dahlia: It would be better if we split lines right after the period and start a new line with the beginning of the next sentence.

CHANGES.md
```diff
@@ -9,6 +9,8 @@ To be released.
...
+    `IStore.ContainsTransaction(TxId)` instead.  [[#TBD]]
```
earlbread: We need the link as well.
```suggestion
    `IStore.ContainsTransaction(TxId)` instead.  [[#676 ]]
```

PR #682, opened 2019-11-19
Reduce SwarmTest code size and test time

earlbread:
This reduces `SwarmTest` code size and test time.

- Add `SwarmTest.CreateSwarm()` method for `SwarmTest` to reduce code size.
- Remove unnecessary delay in `SwarmTest`.
- Make `Swarm<T>.BlockHashRecvTimeout` configurable.
- Reduce `Swarm<T>.StopAsync()` delay for SwarmTest from 1s to 10ms.

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -89,7 +89,7 @@ namespace Libplanet.Tests.Net
                 int i = 1;
                 foreach (Swarm<DumbAction> s in _swarms)
                 {
-                    s.StopAsync().Wait(DisposeTimeout);
+                    StopAsync(s).Wait(DisposeTimeout);
                     Log.Logger.Debug(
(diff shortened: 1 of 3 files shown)


Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2478,6 +2335,12 @@ public async Task DoNotDeleteCanonicalChainWhenBlockDownloadFailed()
             return task;
         }
 
+        private Task StopAsync<T>(Swarm<T> swarm)
+            where T : IAction, new()
+        {
+            return swarm.StopAsync(TimeSpan.FromMilliseconds(10));
```
dahlia: Is it okay if we give it such a short timeout?

earlbread: I tested several timeout parameters like 500, 50, 10, 1. The tests failed only if the timeout was 1ms.

PR #686, opened 2019-11-20
Fix the workspace chain is deleted during the `FillBlocksAsync` retry

earlbread:
This fixes a bug that the workspace chain is deleted during the `FillBlocksAsync` retry.
This fix revealed that `HandleReorgInSynchronizing` did not work properly. We need to fix the test to work deterministically.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -1554,6 +1554,7 @@ namespace Libplanet.Net
                         long currentTipIndex = blockChain.Tip?.Index ?? -1;
                         long receivedBlockCount = currentTipIndex - previousTipIndex;
 
+                        FillBlocksAsyncStarted.Set();
                         synced = await FillBlocksAsync(
                             peer,
                             blockChain,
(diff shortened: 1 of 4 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -1578,9 +1579,15 @@ CancellationToken cancellationToken
                         }
                         else
                         {
+                            FillBlocksAsyncFailed.Set();
```
longfin: I guess this statement isn't necessary because we'll set this event [here](https://github.com/planetarium/libplanet/pull/686/files#diff-9501da6c2e844a81d5e3ad3d3d5456ccR1588).

earlbread: I thought so, but the [catch block](https://github.com/planetarium/libplanet/pull/686/files#diff-9501da6c2e844a81d5e3ad3d3d5456ccR1586) doesn't seem to catch the re-thrown `TimeoutException`. 😢

longfin: 👌

PR #695, opened 2019-11-27
Block relay benchmark

limebell:
I added a benchmark for the case of block broadcasting relay through multiple swarms.

### DIFF
--- a/Libplanet.Benchmarks/SwarmBenchmark.cs
+++ b/Libplanet.Benchmarks/SwarmBenchmark.cs
@@ -49,6 +49,7 @@ namespace Libplanet.Benchmarks
             _blockChains = new BlockChain<DumbAction>[SwarmNumber];
             _swarms = new Swarm<DumbAction>[SwarmNumber];
 
+            // Private keys to make swarms broadcast blocks in relay.
             _keys[0] = new PrivateKey(new byte[]
             {
                 0x98, 0x66, 0x98, 0x50, 0x72, 0x8c, 0x6c, 0x41, 0x0b, 0xf4,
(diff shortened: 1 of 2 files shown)


Libplanet.Benchmarks/SwarmBenchmark.cs
```diff
@@ -0,0 +1,195 @@
...
+                0x75, 0xd3, 0xbc, 0xaa, 0xf0, 0xa2, 0xec, 0xdb, 0x05, 0xa5,
+                0x59, 0x79,
+            });
+            _keys[9] = new PrivateKey(new byte[]
+            {
+                0x3d, 0xae, 0x2d, 0x2f, 0x87, 0x1b, 0x11, 0xaa, 0x41, 0xbd,
+                0xec, 0x81, 0x4c, 0x4d, 0x27, 0xf3, 0xba, 0xd9, 0x1f, 0x61,
+                0xc3, 0x57, 0xab, 0x43, 0xa0, 0x0c, 0x63, 0x1b, 0x2b, 0x15,
+                0x13, 0xf2,
+            });
```
dahlia: Could we have comments on the above addresses (for example, purposes or a bucket that each private key is assigned to)?

Libplanet.Benchmarks/SwarmBenchmark.cs
```diff
@@ -0,0 +1,195 @@
...
+                    _blockChains[i],
+                    _keys[i],
+                    0,
+                    host: IPAddress.Loopback.ToString());
+                StartAsync(_swarms[i]).Wait(WaitTimeout);
+            }
+
+            for (int i = 0; i < SwarmNumber - 1; i++)
+            {
+                _swarms[i].AddPeersAsync(new[] { _swarms[i + 1].AsPeer }, null)
```
longfin: How about using `Task.WaitAll()`?

limebell: `Task.WaitAll()` can be applied to `StartAsync` of swarms, but sadly, `AddPeersAsync` hangs using `Task.WaitAll()`

longfin: 😢

PR #701, opened 2019-12-02
Improve ListAllStateReferences performance

earlbread:
This moves `StoreExtension.ListAllStateReferences` to `IStore.ListAllStateReferences` and improve the preformance of the method.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,7 @@ To be released.
  -  `Swarm<T>.PreloadAsync()` method and `Swarm<T>.StartAsync()` method became
     to take `preloadBlockDownloadFailed` event handler as an argument.
     [[#694]]
+ -  Removed `StoreExtension.ListAllStateReferences<T>()` method.  [[#701]]
 
 ### Backward-incompatible network protocol changes
 
@@ -41,6 +42,7 @@ To be released.
 ### Added APIs
 
  -  Added `DefaultStore` class to replace `LiteDBStore`.  [[#662]]
+ -  Added `IStore.ListAllStateReferences<T>()` method.  [[#701]]
 
 ### Behavioral changes
 
@@ -70,6 +72,7 @@ To be released.
 [#685]: https://github.com/planetarium/libplanet/pull/685
 [#692]: https://github.com/planetarium/libplanet/pull/692
 [#694]: https://github.com/planetarium/libplanet/pull/694
+[#701]: https://github.com/planetarium/libplanet/pull/701
 
 
 Version 0.7.0
--- a/Libplanet/Store/DefaultStore.cs
+++ b/Libplanet/Store/DefaultStore.cs
@@ -266,6 +266,44 @@ namespace Libplanet.Store
         }
 
         /// <inheritdoc/>
+        public override IImmutableDictionary<Address, IImmutableList<HashDigest<SHA256>>>
+            ListAllStateReferences(
+                Guid chainId,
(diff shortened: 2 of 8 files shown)


Libplanet/Store/StoreExtension.cs
```diff
@@ -1,6 +1,7 @@
 using System;
-using System.Collections.Generic;
+#pragma warning disable S1128 // Remove this unnecessary 'using'
 using System.Collections.Immutable;
+#pragma warning restore S1128 // Remove this unnecessary 'using'
```
limebell: Why not just remove this 'using' but ignore?

Libplanet/Store/DefaultStore.cs
```diff
@@ -265,6 +265,40 @@ public override IEnumerable<Address> ListAddresses(Guid chainId)
...
+                    : 0;
+            long highestIndex =
+                ignoreAfter is HashDigest<SHA256> tgt && GetBlockIndex(tgt) is long tgtIdx
+                    ? tgtIdx
+                    : long.MaxValue;
+
+            string collId = StateRefId(chainId);
+            LiteCollection<StateRefDoc> coll = _db.GetCollection<StateRefDoc>(collId);
+
+            var query = Query.And(
```
longfin:
```suggestion
            Query query = Query.And(
```

CHANGES.md
```diff
@@ -28,6 +28,8 @@ To be released.
  -  `Swarm<T>.PreloadAsync()` method and `Swarm<T>.StartAsync()` method became
     to take `preloadBlockDownloadFailed` event handler as an argument.
     [[#694]]
+ -  Removed `StoreExtension.ListAllStateReferences<T>()` method.  [[#701]]
```
longfin: How about merge with the below line?

Libplanet/Store/StoreExtension.cs
earlbread: This is to suppress a false error(https://github.com/SonarSource/sonar-dotnet/issues/2694). It is resolved(https://github.com/SonarSource/sonar-dotnet/pull/2798) but not released yet. We can remove this warning suppression after the release.

CHANGES.md
```diff
@@ -28,6 +28,8 @@ To be released.
  -  `Swarm<T>.PreloadAsync()` method and `Swarm<T>.StartAsync()` method became
     to take `preloadBlockDownloadFailed` event handler as an argument.
     [[#694]]
+ -  Moved `StoreExtension.ListAllStateReferences<T>()` method to
+    `IStore.ListAllStateReferences<T>()` method.  [[#701]]
```
dahlia: Isn't the destination `DefaultStore` (not `IStore`)?

earlbread: You're right. However, to add a `DefaultStore`, first, we need to add it to the `IStore` interface.

dahlia: Ah, okay. Then IMHO it's better to have two changes: removed an extension method, and added a new method into the interface.

earlbread: Added `Removed StoreExtension.ListAllStateReferences<T>() method.` to Backward-incompatible API changes and  `Added IStore.ListAllStateReferences<T>() method.` to Added APIs.  @longfin @dahlia

limebell: I guess the changes you mentioned did not applied.

earlbread: Oh, it's missed while rebasing. Fixed.

PR #704, opened 2019-12-04
Update state ref even if block state exists

earlbread:
Fixed a bug where the state reference would not be updated when forking a chain and adding existed blocks.


CHANGES.md
```diff
@@ -61,6 +61,8 @@ To be released.
  -  Fixed a bug where the canonical chain could be deleted if `Swarm<T>` failed
     to download blocks due to network connection.  [[#675]]
  -  Fixed bug that re-download block from scratch in preloading.  [[#685]]
+ -  Fixed a bug where the state reference would not be updated when forking a
+    chain and adding existed blocks.  [[#704]]
```

> Fixed a bug where state references become outdated if a chain is forked and then adding existing blocks into it.

earlbread:
I fixed `ForkStateReferences` not to copy state references of the genesis block.

PR #707, opened 2019-12-04
Preload blocks from multiple peers at a time

dahlia:
This patch makes IDL parallel by requesting chunks of blocks to multiple peers.  As complexity of IDL logic grows up, I added a separated class named `BlockCompletion<T>` which has only a pure logic of dealing with block demands and completion without actual I/Os, and made `Swarm<T>.PreloadAsync()` method to use it.

### DIFF
--- a/Libplanet/Blockchain/BlockLocator.cs
+++ b/Libplanet/Blockchain/BlockLocator.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Security.Cryptography;
@@ -7,6 +8,37 @@ namespace Libplanet.Blockchain
     internal class BlockLocator : IEnumerable<HashDigest<SHA256>>
     {
         private readonly IEnumerable<HashDigest<SHA256>> _impl;
+
+        public BlockLocator(
+            Func<long, HashDigest<SHA256>?> indexBlockHash,
+            Func<HashDigest<SHA256>, long> blockHashByIndex,
+            int sampleAfter = 10
+        )
+        {
+            HashDigest<SHA256>? current = indexBlockHash(-1);
+            long step = 1;
+            var hashes = new List<HashDigest<SHA256>>();
+            while (current is HashDigest<SHA256> hash)
+            {
+                hashes.Add(hash);
+                long currentBlockIndex = blockHashByIndex(hash);
+
+                if (currentBlockIndex == 0)
+                {
+                    break;
+                }
+
+                long nextIndex = Math.Max(currentBlockIndex - step, 0);
+                current = indexBlockHash(nextIndex);
+
+                if (hashes.Count >= sampleAfter)
+                {
+                    step *= 2;
+                }
+            }
+
+            _impl = hashes;
+        }
 
         public BlockLocator(IEnumerable<HashDigest<SHA256>> hashes)
         {
--- a/Libplanet/Blockchain/BlockLocator.cs
+++ b/Libplanet/Blockchain/BlockLocator.cs
@@ -11,7 +11,7 @@ namespace Libplanet.Blockchain
 
         public BlockLocator(
             Func<long, HashDigest<SHA256>?> indexBlockHash,
-            Func<HashDigest<SHA256>, long> blockHashByIndex,
+            Func<HashDigest<SHA256>, long> indexByBlockHash,
             int sampleAfter = 10
         )
         {
@@ -21,7 +21,7 @@ namespace Libplanet.Blockchain
             while (current is HashDigest<SHA256> hash)
             {
                 hashes.Add(hash);
-                long currentBlockIndex = blockHashByIndex(hash);
+                long currentBlockIndex = indexByBlockHash(hash);
 
                 if (currentBlockIndex == 0)
                 {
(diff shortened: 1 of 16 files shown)


longfin:
It seems builds fail. 😢

Libplanet/Net/Swarm.cs
```diff
@@ -700,59 +724,85 @@ await DialToExistingPeers(dialTimeout, cancellationToken)
...
-                var exceptions = new List<Exception>();
-                var blockDownloadComplete = false;
-                foreach ((BoundPeer Peer, long? Height) peerWithHeight in peersWithHeight)
-                {
-                    try
+                FillBlocksAsyncStarted.Set();
+
+                var blockCompletion = new BlockCompletion<BoundPeer, T>(
+                    completionPredicate: workspace.ContainsBlock,
+                    window: 100
```
limebell: How about define 100 as constant or variable?

Libplanet/Net/Messages/Message.cs
```diff
@@ -77,6 +72,11 @@ internal enum MessageType : byte
             /// Contains the calculated recent states and state references.
             /// </summary>
             RecentStates = 0x0c,
+
+            /// <summary>
+            /// Message containing demand block hashes with their index numbers.
+            /// </summary>
+            BlockHashes = 0x0d,
```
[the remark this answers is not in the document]
dahlia: As the message coding became incompatible with what it had been, I assigned the new message ID number to this so that two types (old and new) of `BlockHashes` don't be mixed or used at a time on the network.

Libplanet/Blockchain/BlockLocator.cs
```diff
@@ -8,6 +9,37 @@ internal class BlockLocator : IEnumerable<HashDigest<SHA256>>
     {
         private readonly IEnumerable<HashDigest<SHA256>> _impl;
 
+        public BlockLocator(
+            Func<long, HashDigest<SHA256>?> indexBlockHash,
+            Func<HashDigest<SHA256>, long> blockHashByIndex,
```
moreal: How about this? Because the function gets index by block hash.

```suggestion
            Func<HashDigest<SHA256>, long> indexByBlockHash,
```

limebell:
Is this PR resolves #714?

longfin:
> Is this PR resolves #714?

I guess it would affect #714. @dahlia any idea?

dahlia:
I believe this resolves #714 as well.

Libplanet/Blockchain/BlockLocator.cs
```diff
@@ -8,6 +9,37 @@ internal class BlockLocator : IEnumerable<HashDigest<SHA256>>
     {
         private readonly IEnumerable<HashDigest<SHA256>> _impl;
 
+        public BlockLocator(
+            Func<long, HashDigest<SHA256>?> indexBlockHash,
```
moreal:
```suggestion
            Func<long, HashDigest<SHA256>?> blockHashByIndex,
```

Libplanet/Blockchain/BlockLocator.cs
```diff
@@ -8,6 +9,37 @@ internal class BlockLocator : IEnumerable<HashDigest<SHA256>>
...
+            Func<HashDigest<SHA256>, long> indexByBlockHash,
+            int sampleAfter = 10
+        )
+        {
+            HashDigest<SHA256>? current = indexBlockHash(-1);
```
moreal:
```suggestion
            HashDigest<SHA256>? current = blockHashByIndex(-1);
```

Libplanet/Blockchain/BlockLocator.cs
```diff
@@ -8,6 +9,37 @@ internal class BlockLocator : IEnumerable<HashDigest<SHA256>>
...
+                hashes.Add(hash);
+                long currentBlockIndex = indexByBlockHash(hash);
+
+                if (currentBlockIndex == 0)
+                {
+                    break;
+                }
+
+                long nextIndex = Math.Max(currentBlockIndex - step, 0);
+                current = indexBlockHash(nextIndex);
```
moreal:
```suggestion
                current = blockHashByIndex(nextIndex);
```

Libplanet/Net/BlockCompletion.cs
```diff
@@ -0,0 +1,335 @@
...
+                        {
+                            _completions.Enqueue(peer);
+                            Interlocked.Increment(ref _finished);
+                        }
+                    },
+                    cancellationToken: cancellationToken
+                );
+            }
+
+            public async Task WaitAll(CancellationToken cancellationToken = default)
```
limebell:
```suggestion
            public async Task WaitAll(CancellationToken cancellationToken = default(CancellationToken))
```
Isn't this does not work in some environments?

Libplanet/Blockchain/BlockLocator.cs
dahlia: I believe current naming is more intuitive, because it's a function and natural to read it as “to index a block hash.”  Ditto for the same reviews below.

dahlia:
@planetarium/libplanet Rebased this on the current master.  Please review this again!  FYI only changelogs had been changed since then.

PR #712, opened 2019-12-09
Prevent redundant message broadcasting

limebell:
This patch modifies `KademliaProtocol.PeersToBroadcast` prevent message to be re-broadcast to  previous broadcaster peer.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -574,9 +574,7 @@ namespace Libplanet.Net
 
         public void BroadcastTxs(IEnumerable<Transaction<T>> txs)
         {
-            _logger.Debug("Broadcast Txs.");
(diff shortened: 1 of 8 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -1693,7 +1699,7 @@ CancellationToken cancellationToken
             }
         }
 
-        private async Task AppendBlocksAsync(
+        private async Task<bool> AppendBlocksAsync(
```
longfin: Is this change related to broadcasting?

limebell: Yes, I made mistake. I'll fix it.

limebell:
> Shouldn't we change [this](https://github.com/planetarium/libplanet/pull/712/files#diff-9501da6c2e844a81d5e3ad3d3d5456ccR1960) too?

Since transactions are not broadcasted, no.

PR #724, opened 2019-12-14
remove rule S1066

Oaroomy:
Merged 2 If statement


longfin:
@Oaroomy Thanks for your first contribution! :tada:

- Could you submit our CLA?
- We're using a changelog checker. but it doesn't seem to need any changelog for these changes because there is no interface change or bug fixing. so you need `skip changelog` on your commit message to bypass this check.

PR #730, opened 2019-12-14
Add a 'static' keyword to the class declaration.

### LINKED ISSUE #664: Remove Sonar Analyzer suppressions

Oaroomy:
Add a 'static' keyword to `Crc32` class

fix #664

### DIFF
--- a/Libplanet.ruleset
+++ b/Libplanet.ruleset
@@ -48,9 +48,6 @@
     <Rule Id="S112" Action="None" />
     <!-- Add the default parameter value defined in the overridden method. -->
     <Rule Id="S1006" Action="None" />
-    <!-- Add a 'protected' constructor or the 'static' keyword to the class
-    declaration. -->
-    <Rule Id="S1118" Action="None" />
     <!-- Add an explanation -->
     <Rule Id="S1123" Action="None" />
     <!-- Take the required action to fix the issue indicated by this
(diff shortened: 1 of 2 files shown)


Libplanet.ruleset
```diff
@@ -1,122 +1,54 @@
...
-    <!-- Extract this nested ternary operation into an independent
-    statement. -->
-    <Rule Id="S3358" Action="None" />
-    <!-- Seal class 'Transaction' or implement 'IEqualityComparer<T>'
-    instead. -->
     <Rule Id="S4035" Action="None" />
+    <Rule Id="S4457" Action="None" />
   </Rules>
-</RuleSet>
+</RuleSet>
```
longfin:
```suggestion
</RuleSet>

```

PR #732, opened 2019-12-17
Prevent to fill blocks when received continuous blocks

limebell:


Libplanet/Net/Swarm.cs
```diff
@@ -1726,9 +1726,18 @@ CancellationToken cancellationToken
...
             if (tip is null || latest.Index > tip.Index)
             {
+                List<Block<T>> blocksToAppend;
                 if (oldest.PreviousHash is null)
                 {
                     _logger.Debug("The oldest block[{block}] seems to be genesis.", oldest);
+                    blocksToAppend = blocks;
+                }
+                else if (!(tip is null) &&
+                         blocks.Any(block => block.PreviousHash.Equals(tip.Hash)))
```
longfin: I guess it's fine that we check only the first block of blocks because we assume `blocks` have been sorted.

limebell: I made this because received list of blocks may contains blocks with lower index than current tip, and also include blocks with higher index.

PR #736, opened 2019-12-18
Calculate state if received state height is shorter than Tip

earlbread:
This fixes a bug where the states were empty between the tip of the peer to receive the states and the tip of the downloaded block.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -106,6 +106,8 @@ To be released.
     some `CultureInfo.CurrentCulture` (e.g., `ar_SA`, `fr_FR`, `th_TH`)
     so that it had caused network protocol incompatibilities.
     [[#734]]
+ -  Fixed a bug where the states became empty between the tip of the peer to
+    receive the states and the tip of the downloaded block.  [[#736]]
 
 [#604]: https://github.com/planetarium/libplanet/issues/604
 [#613]: https://github.com/planetarium/libplanet/issues/613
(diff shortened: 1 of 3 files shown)


Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -1856,6 +1856,70 @@ public async Task PreloadWithBranchesAndTrustedPeers()
...
+
+                Assert.Equal(receiverChain.Count, receiverStateRefs.Count + 1);
+                Assert.Equal(minerStateRefs, receiverStateRefs);
+            }
+            finally
+            {
+                await StopAsync(minerSwarm);
+                await StopAsync(receiverSwarm);
+                fxMiner.Dispose();
+                fxReceiver.Dispose();
```
longfin: I guess we should `.Dispose` of `minerSwarm` and `receiverSwarm` here.

Libplanet/Net/Swarm.cs
```diff
@@ -787,44 +787,22 @@ await DialToExistingPeers(dialTimeout, cancellationToken)
                         .OrderByDescending(pair => pair.Item2)
                         .Select(pair => (pair.Item1, workspace[pair.Item2.Value].Hash));
 
-                bool received = await SyncRecentStatesFromTrustedPeersAsync(
+                long receivedStateHeight = await SyncRecentStatesFromTrustedPeersAsync(
```
longfin: How about using `long?` and `null` instead of `-1`? (it isn't a big deal)

CHANGES.md
```diff
@@ -106,6 +106,8 @@ To be released.
     some `CultureInfo.CurrentCulture` (e.g., `ar_SA`, `fr_FR`, `th_TH`)
     so that it had caused network protocol incompatibilities.
     [[#734]]
+ -  Fixed a bug where the states were empty between the tip of the peer to
```
```suggestion
 -  Fixed a bug where the states became empty between the tip of the peer to
```

PR #738, opened 2019-12-19
Sort peers to download blocks by height

earlbread:


moreal:
I also considered you did. But I didn't apply it and stay with the ordering from *Kademlia* because of possibility preloading can be concentrated on one node having the most blocks. How do you think about that?

Libplanet/Net/Swarm.cs
```diff
@@ -693,6 +693,8 @@ await DialToExistingPeers(dialTimeout, cancellationToken)
                 return;
             }
 
+            peersWithHeight = peersWithHeight.OrderByDescending(p => p.Item2).ToList();
```
dahlia: How about sorting it before it's become realized (i.e., before `.ToListAsync()` being called) rather than here?

earlbread:
> I also considered you did and it will help your chain keep up with the chain of network. But I didn't apply it and stay with the ordering from _Kademlia_ because of possibility preloading can be concentrated on one node having the most blocks. How do you think about that?

Makes sense. But in general, I think many peers have blocks of the same height. Don't they?

Libplanet/Net/Swarm.cs
earlbread: I tried to but couldn't. Maybe because `peersWithHeight` is `IAsyncEnumerable`.

moreal:
> > I also considered you did and it will help your chain keep up with the chain of network. But I didn't apply it and stay with the ordering from _Kademlia_ because of possibility preloading can be concentrated on one node having the most blocks. How do you think about that?
> 
> Makes sense. But in general, I think many peers have blocks of the same height. Don't they?

Right.

PR #744, opened 2019-12-27
Tune Swarm<T>

longfin:
This PR addresses some issues as below

- Fixed a bug where `Swarm<T>.StartAsync()` had thrown `NullReferenceException` when without using NAT and `host` parameter.
- To stabilize request processing, `Swarm<T>._runtimeProcessor` became a long-running task. (it's a dedicated thread.)
- `Swarm<T>` became spawn more `BindingProxies()` to handle TURN request concurrently.
  - We should parameterize its count in the future.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -120,7 +120,7 @@ To be released.
  -  Fixed a bug where the states became empty between the tip of the peer to
     receive the states and the tip of the downloaded block.  [[#736]]
  -  Fixed a bug where `Swarm<T>.StartAsync()` had thrown
-    `NullReferenceException` when without using NAT and `host` parameter.
+    `NullReferenceException` when `host` parameter is present on the outside of NAT.
     [[#744]]
  -  Fixed a bug where `Swarm<T>` had failed to request a TURN relay when it has
     an IPv6 address.  [[#744]]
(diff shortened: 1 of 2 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -170,30 +170,38 @@ static Swarm()
...
                     }
-                }
-                catch (NetMQException)
-                {
-                }
-                catch (ObjectDisposedException)
-                {
-                }
-            });
+                    catch (NetMQException)
```
earlbread: How about adding logs when these exceptions occur?

longfin: I added logs at e337696397b93310bce342be20740c6b4cae704b.

longfin:
I added a commit(b63489ebf339e3a0f74ee8f9fc51f998bdfa75b2) to address an issue that a peer can't request TURN relay properly when it has IPv6 address.

CHANGES.md
```diff
@@ -117,6 +119,11 @@ To be released.
     [[#734]]
  -  Fixed a bug where the states became empty between the tip of the peer to
     receive the states and the tip of the downloaded block.  [[#736]]
+ -  Fixed a bug where `Swarm<T>.StartAsync()` had thrown
+    `NullReferenceException` when without using NAT and `host` parameter.
```
```suggestion
    `NullReferenceException` when `host` parameter is present on the outside of NAT.
```

Issue #747, opened 2020-01-03
Optimize state size


longfin:
Currently, libplanet stores every new state associated with each block. This is a good structure for getting the state of a particular address quickly, but it will waste a lot of space if only a small part of the state changes.

To prevent this, we need to introduce some techniques such as reusing only the changes made in the previous state, as used in [Etheruem's MPT](https://github.com/ethereum/wiki/wiki/Patricia-Tree).

dahlia:
Hasn't this been achieved by MPT (#939, #950, #968, #966, & #978) @moreal recently introduced?

PR #751, opened 2020-01-08
Remove serializer in Block<T>, RawBlock<T>, Transaction<T>, RawTransaction<T>

limebell:
This patch removes serializer in `Block<T>`, `RawBlock<T>`, `Transaction<T>` and `RawTransaction<T>`. `Block<T>` and `Transaction<T>`'s `ToBencodex()` method and `FromBencodex()` method will to the task.
Additionally `RawBlock<T>.Transactions` is now `IEnumerable<byte[]>` instead of `IEnumerable`.


CHANGES.md
```diff
@@ -39,6 +39,8 @@ To be released.
     from `Swarm<T>`.  [[#705], [#725]]
  -  Added `workers` optional parameter into `Swarm<T>()` constructor.
     [[#613], [#727]]
+ -  `Block<T>` class became not to implement `ISerializable`.  [[#747]]
```
longfin: #747 doesn't seem this PR but seems a different issue.

CHANGES.md
```diff
@@ -163,6 +167,7 @@ To be released.
 [#739]: https://github.com/planetarium/libplanet/pull/739
 [#744]: https://github.com/planetarium/libplanet/pull/744
 [#746]: https://github.com/planetarium/libplanet/pull/746
+[#747]: https://github.com/planetarium/libplanet/pull/747
```
earlbread: We need to add this link to this PR.

Libplanet/Store/DefaultStore.cs
```diff
@@ -817,11 +816,12 @@ public void Dispose()
...
             try
             {
-                var formatter = new BencodexFormatter<RawBlock>();
                 using (Stream stream = _blocks.OpenFile(
                     path, System.IO.FileMode.Open, FileAccess.Read, FileShare.Read))
+                using (MemoryStream ms = new MemoryStream())
                 {
-                    rawBlock = (RawBlock)formatter.Deserialize(stream);
+                    stream.CopyTo(ms);
+                    rawBlock = new RawBlock(ms.ToArray());
```
longfin: We also can use [`File.ReadAllBytes()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readallbytes?view=netstandard-2.0) instead of streams.

Libplanet/Blocks/Block.cs
```diff
@@ -101,8 +100,13 @@ public class Block<T> : ISerializable
...
         }
 
-        protected Block(SerializationInfo info, StreamingContext context)
-            : this(new RawBlock(info, context))
+        /// <summary>
+        /// Creates a <see cref="Block{T}"/> instance from its serialization.
+        /// </summary>
+        /// <param name="bytes">The serialization of block instance.
+        /// </param>
+        public Block(byte[] bytes)
```
dahlia: As encapsulation-wise `IValue` is a more semantic representation than `byte[]`, IMHO it's unnatural that a function for `IValue` → `Block<T>` is a static method whereas a function for `byte[]` → `Block<T>` is a constructor.  Aren't they inside out?  To me it would be more natural if there are `Block<T>(IValue)` constructor and `Block<T>.Deserialize(byte[])` static method.

Libplanet.Tests/Common/Action/DumbAction.cs
```diff
@@ -69,17 +69,17 @@ public IValue PlainValue
...
 
                 if (Idempotent)
                 {
                     plainValue =
-                        (Dictionary)plainValue.Add(
-                            (Text)"idempotent",
-                            new Bencodex.Types.Boolean(Idempotent));
+                        plainValue.Add(
+                            "idempotent",
+                            (IValue)new Bencodex.Types.Boolean(Idempotent));
```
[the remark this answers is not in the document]
limebell: `Bencodex.Types.Dictionary` does not have interface for `.Add(string, Bencodex.Types.Boolean)`.

Libplanet/Tx/RawTransaction.cs
```diff
@@ -70,26 +53,26 @@ ImmutableArray<byte> signature
             Signature = signature;
         }
 
-        public RawTransaction(Dictionary<string, object> dict)
+        public RawTransaction(IValue value)
```
longfin: How about restricting the type of `value` here to `Dictionary`?

limebell: If so, should we restrict return type of `ToBencodex()` to `Bencodex.Types.Dictionary` either?

longfin: Maybe yes. if we were implementing for other types than `Transaction<T>` or `Block<T>`, it can be quite weird that `ToBencodex()` returns `Dictionary` instead of `IValue`. but in this case, it seems that we only implement `ToBencodex()` for `Transaction<T>` and `Block<T>` and they can be treat as complicated composite types. so IMHO it's okay that we also restrict `ToBencodex`'s return type to `Dictionary`.

limebell: Great, I've applied your suggestions :)

Libplanet/Blocks/Block.cs
```diff
@@ -229,31 +233,36 @@ private Block(RawBlock rb)
...
-            using (var stream = new MemoryStream())
-            {
-                serializer.Serialize(stream, this);
-                return stream.ToArray();
-            }
+            var codec = new Codec();
+            return codec.Encode(ToBencodex(hash, transactionData));
         }
 
+        public Bencodex.Types.Dictionary ToBencodex(bool hash, bool transactionData) =>
```
dahlia: You wrote, in the changelog, that this method was gone, but as the method of the same name still exists, it's better to be adjusted.

Libplanet/Blocks/RawBlock.cs
```diff
@@ -47,21 +50,27 @@ namespace Libplanet.Blocks
...
         {
+            Func<string, byte[]> b = Encoding.ASCII.GetBytes;
+            Index = dict.GetValue<Integer>(b("index"));
+            Timestamp = dict.GetValue<Text>(b("timestamp"));
+            Difficulty = dict.GetValue<Integer>(b("difficulty"));
+            Transactions = dict.GetValue<Bencodex.Types.List>(b("transactions"))
+                .Select(tx => (byte[])(Binary)tx);
+            Nonce = dict.GetValue<Binary>(b("nonce"));
+
+            Miner = dict.ContainsKey((Binary)b("reward_beneficiary"))
```
dahlia: We might be better to define constants for these keys, e.g.:

```csharp
public static Binary RewardBeneficiaryKey = (Binary)b("reward_beneficiary");
```

@moreal Also we'd better to define an alternative constructor for `Bencodex.Types.Binary` to take a `string` and an `Encoding`, so that we can make a `Binary` value by calling `new Binary("reward_beneficiary", Encoding.ASCII)`?

Libplanet/Tx/RawTransaction.cs
```diff
@@ -3,33 +3,36 @@
...
-                    Address.Size),
-                timestamp: info.GetString("timestamp"),
-                signature: info.GetValue<byte[]>("signature").ToImmutableArray(),
-                actions: info.GetValue<List>(
-                    "actions")
-            )
-        {
-        }
+        public static readonly byte[] NonceKey =
+            (Binary)Encoding.ASCII.GetBytes("nonce");
```
moreal:
```suggestion
            Encoding.ASCII.GetBytes("nonce");
```

Libplanet/Tx/RawTransaction.cs
```diff
@@ -3,33 +3,36 @@
...
+
+        public static readonly byte[] SignerKey =
+            (Binary)Encoding.ASCII.GetBytes("signer");
```
moreal:
```suggestion
            Encoding.ASCII.GetBytes("signer");
```

Libplanet/Tx/RawTransaction.cs
```diff
@@ -3,33 +3,36 @@
...
+
+        public static readonly byte[] UpdatedAddressesKey =
+            (Binary)Encoding.ASCII.GetBytes("updated_addresses");
```
moreal:
```suggestion
            Encoding.ASCII.GetBytes("updated_addresses");
```

Libplanet/Tx/RawTransaction.cs
```diff
@@ -3,33 +3,36 @@
...
+
+        public static readonly byte[] PublicKeyKey =
+            (Binary)Encoding.ASCII.GetBytes("public_key");
```
moreal:
```suggestion
            Encoding.ASCII.GetBytes("public_key");
```

Libplanet/Tx/RawTransaction.cs
```diff
@@ -3,33 +3,36 @@
...
+
+        public static readonly byte[] TimestampKey =
+            (Binary)Encoding.ASCII.GetBytes("timestamp");
```
moreal:
```suggestion
            Encoding.ASCII.GetBytes("timestamp");
```

Libplanet/Tx/RawTransaction.cs
```diff
@@ -3,33 +3,36 @@
...
+
+        public static readonly byte[] ActionsKey =
+            (Binary)Encoding.ASCII.GetBytes("actions");
```
moreal:
```suggestion
            Encoding.ASCII.GetBytes("actions");
```

Libplanet/Tx/RawTransaction.cs
```diff
@@ -3,33 +3,36 @@
...
+
+        public static readonly byte[] SignatureKey =
+            (Binary)Encoding.ASCII.GetBytes("signature");
```
moreal:
```suggestion
            Encoding.ASCII.GetBytes("signature");
```

Libplanet/Blocks/RawBlock.cs
dahlia: @limebell @moreal Sent a pull request to add a new constructor: <https://github.com/planetarium/bencodex.net/pull/14>.

PR #757, opened 2020-01-13
Implement block header

### LINKED ISSUE #749: Seperate header from block
Now, `Block` includes all properties (e.g. `Index`, `Difficulty`, `Transactions`, etc...).
So we must read or transfer all including `Transactions` when we want to know or want to transfer the block's information. It can be so heavy task. But if we seperated it, it will have less cost than before. Further more, it may help us formulate more effecient strategies. And the format seperating header and body, have been working well in some popular blockchain core (e.g. ethereum, bitcoin) already.

limebell:
Previously `RawBlock` can be in two different states which are:
- Contains all transactions that block has.
- Contains only transaction id of that block has.

To be clear, separated `BlockDigest` which contains only transaction ids from `RawBlock`. Additionally  added a structure named `BlockHeader` which summarizes block information except transactions.
To sum it up, there are now 4 block related objects:
- `Block` : A class represents block. This contains all block parameters and transactions.
- `RawBlock` : A serialization used in network transportation. This contains serialized transactions `BlockHeader`.
- `BlockDigest` : A serialization used in storage. This contains transaction ids and `BlockHeader`.
- `BlockHeader` : A serialization describes block information. This does not neither transactions nor transaction ids.

Caution: This patch is based on #751 
This closes #749.


Libplanet/Blocks/Block.cs
```diff
@@ -481,33 +490,47 @@ internal void Validate(DateTimeOffset currentTime)
...
+
+        internal RawBlock ToRawBlock()
+        {
+            // For consistency, order transactions by its id.
+            return new RawBlock(
+                header: GetBlockHeader(),
+                transactions: Transactions.OrderBy(tx => tx.Id).Select(tx => tx.Serialize(true)));
+        }
+
+        private byte[] ComputeHash()
```
longfin: Actually, this method doesn't seem to calculate hash digest yet. 😕

Libplanet/Blocks/BlockDigest.cs
```diff
@@ -0,0 +1,58 @@
...
+                .Select(txId => (byte[])(Binary)txId);
+        }
+
+        public BlockHeader Header { get; }
+
+        public IEnumerable<byte[]> TxIds { get; }
+
+        public static BlockDigest Deserialize(byte[] bytes)
+        {
+            var value = new Codec().Decode(bytes);
```
longfin:
```suggestion
            IValue value = new Codec().Decode(bytes);
```

Libplanet/Store/BaseStore.cs
```diff
@@ -105,7 +105,7 @@ public Block<T> GetBlock<T>(HashDigest<SHA256> blockHash)
         /// <inheritdoc/>
         public long? GetBlockIndex(HashDigest<SHA256> blockHash)
         {
-            return GetRawBlock(blockHash)?.Index;
+            return GetBlockDigest(blockHash)?.Header.Index;
```
longfin: Is it okay that we omit `?` on `.Header`?

```suggestion
            return GetBlockDigest(blockHash)?.Header?.Index;
```

limebell: Yes, because `Header` always exists if `GetBlockDigest(blockHash)` isn't null.

Libplanet/Blocks/Block.cs
limebell: Great. I renamed it to `SerializeForHash`.

limebell:
This closes #749.

limebell:
I made `BlockHeader` and `BlockDigest` structs readonly.

PR #761, opened 2020-01-15
Unsure to merge: Compress blocks and txs as well

dahlia:
This patch is continued from #753.  Besides block states, this compresses blocks and txs as well.  Unfortunately, the compression rate is almost the same to #753 (5.1G → 1.6G).

### DIFF
--- a/Libplanet/Store/DefaultStore.cs
+++ b/Libplanet/Store/DefaultStore.cs
@@ -385,6 +385,14 @@ namespace Libplanet.Store
                 return null;
             }
 
+            if (_compress)
+            {
+                using (var inputBuffer = new MemoryStream(bytes))
+                {
+                    bytes = Decompress(inputBuffer);
+                }
+            }
+
             Transaction<T> tx = Transaction<T>.Deserialize(bytes);
             _txCache.AddOrUpdate(txid, tx);
             return tx;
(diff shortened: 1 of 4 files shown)


longfin:
I guess it's okay that we contain this feature and can also drop out when it is obvious that it is not necessary or harmful to features.

Libplanet/Store/DefaultStore.cs
```diff
@@ -819,28 +827,73 @@ public void Dispose()
...
+        private static byte[] CompressBytes(byte[] input)
+        {
+            using (var buffer = new MemoryStream())
+            using (var deflate = new DeflateStream(buffer, CompressionLevel.Fastest, true))
+            {
+                deflate.Write(input, 0, input.Length);
+                deflate.Flush();
+                buffer.Flush();
+                int length = (int)buffer.Position;
+                var output = new byte[length];
```
longfin: Can we use `buffer.ToArray()` instead of temp variable(`output`)?

Libplanet/Store/DefaultStore.cs
```diff
@@ -819,28 +827,73 @@ public void Dispose()
...
+        private static byte[] Decompress(Stream inputBuffer)
+        {
+            using (var outputBuffer = new MemoryStream())
+            using (var deflate = new DeflateStream(inputBuffer, CompressionMode.Decompress))
+            {
+                deflate.CopyTo(outputBuffer);
+                deflate.Flush();
+                outputBuffer.Flush();
+                int length = (int)outputBuffer.Position;
+                var output = new byte[length];
```
longfin: Can we use `outputBuffer.ToArray()` instead of temp variable(`output`)?

dahlia:
As this patch does not solve the problem and is likely to make the code more complicated, I decided to do not merge this.

PR #767, opened 2020-01-20
Guarantee at least 10 peers to broadcast

### LINKED ISSUE #765: Guarantee minimum amount of peer to broadcast
Now select 1 peer from each bucket in order to broadcast, it may be insufficient.

limebell:
If the number of total peers in routing table is less then 10, broadcast messages to all peers.
If not, pick one from each bucket, and pick additional peers which were not selected in first step until total peers count reach 10.

Closes #765.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -123,6 +123,11 @@ To be released.
     serialization.  [[#762]]
  -  `Swarm<T>` became to ignore broadcasted block that has lower index than
     the current tip.  [[#764]]
+ -  The way `Swarm<T>` chose peers to spread messages has changed.  [[#765], [#767]]
+     -  If there are less than 10 peers in the routing table, select all peers.
+     -  If there are more than 10 peers in the routing table,
+        choose one from each bucket, and if the number is less than 10,
+        then select an additional peers so that the total is 10.
 
 ### Bug fixes
 
@@ -207,6 +212,8 @@ To be released.
 [#762]: https://github.com/planetarium/libplanet/pull/762
 [#763]: https://github.com/planetarium/libplanet/pull/763
 [#764]: https://github.com/planetarium/libplanet/pull/764
+[#765]: https://github.com/planetarium/libplanet/issues/765
+[#767]: https://github.com/planetarium/libplanet/pull/767
 
 
 Version 0.7.0
--- a/Libplanet.Tests/Net/Protocols/TestTransport.cs
+++ b/Libplanet.Tests/Net/Protocols/TestTransport.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Net;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Libplanet.Crypto;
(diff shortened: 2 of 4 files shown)


Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -7,6 +7,9 @@ namespace Libplanet.Net.Protocols
 {
     internal class RoutingTable
     {
+        // FIXME: This would be configurable.
+        private const int _minPeersToBroadcast = 8;
```
earlbread: The minimum number of peers to broadcast in the description is 10. Is this okay?

limebell: Oh, I'll fix it :(

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -80,9 +83,18 @@ private IEnumerable<KBucket> NonEmptyBuckets
...
-            return NonEmptyBuckets
+            var peers = NonEmptyBuckets
                 .Select(bucket => bucket.GetRandomPeer(except))
-                .Where(peer => !(peer is null));
+                .Where(peer => !(peer is null)).ToList();
+            var count = peers.Count;
+            if (count < _minPeersToBroadcast)
+            {
+                peers.AddRange(Peers
+                    .Where(peer => except is null || !peer.Address.Equals(except.Value))
```
longfin: Can we rewrite as below?

```suggestion
                    .Where(peer => !except?.Equals(peer.Address)
```

limebell: It's `System.Nullable<bool>` so can't be used in `where` phrase.

CHANGES.md
```diff
@@ -123,6 +123,11 @@ To be released.
     serialization.  [[#762]]
  -  `Swarm<T>` became to ignore broadcasted block that has lower index than
     the current tip.  [[#764]]
+ -  The way `Swarm<T>` chose peers to spread messages has changed.  [[#765], [#767]]
+     -  If there are less than 10 peers in the routing table, select all peers.
+     -  If there are more than 10 peers in the routing table,
+        choose one from each bucket, and if the number is less than 10,
+        then select an additional peers so that the total is 10.
```
dahlia: As it's not a manual or docs, rather than explaining its behavioral details, we'd better to write a short descriptive guarantee and its intention.

Libplanet.Tests/Net/Protocols/TestTransport.cs
```diff
@@ -248,10 +252,14 @@ public void BroadcastTestMessage(Address? except, string data)
         public void BroadcastMessage(Address? except, Message message)
         {
             var peers = Protocol.PeersToBroadcast(except).ToList();
+            var peersString = peers.Aggregate(
+                new StringBuilder(),
+                (builder, peer) => builder.Append($"{peer.Address.ToHex()}, ")).ToString();
```
dahlia: Can't the same thing be easily achieved using [`string.Join()` method][1]?

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.string.join?view=netstandard-2.0

Issue #368, opened 2019-07-24
Get rid of `Transaction<T>.UpdatedAddresses` and “rehearsal mode”, and more


dahlia:
Historically, at the very beginning, we had `Transaction.Sender`–`Transaction.Recipient` model, which was made after Bitcoin's or Ethereum's cryptocurrency model.  In 0.2 release, we became to need “multiple recipients” because Libplanet is not for cryptocurrencies but games, so we renamed the terminology to `Transaction.Signer`–`Transaction.UpdatedAddresses` (#121).  However, it's still unhandy to manually specify the list of state addresses that every action in a transaction will mutate, so we had “rehearsal mode” as well in the end (#131), which turns out overkill so that everybody (including me who invented it) dislikes. 🤷🏻‍♂️

In Ethereum, the concept of states are associated with accounts, so keys of the global state map are account addresses.  This model works well with the concept of smart contract, which is a kind of accounts hence has an address for each in Ethereum; every smart contract has its own code and state, and there's kind of *access control*, mutating a contract's state can be done by the code of the same contract (though anyone can read any contract's state).  It's basically cryptocurrency and *assets*, and if you want users to be able to upload some code (i.e., smart contracts), code must not steel others' private assets without owners' authorization.

On the other hand, Libplanet is for games and has no concept of smarts contracts, but follows [application-specific blockchain][1] instead.  Executed code in the network is part of consensus, and code with no consensus is never executed in the network.  Hence we don't need access control unlike Ethereum.

So here's a conclusion: we don't need `Transaction.UpdatedAddresses` property at all, hance, no “rehearsal mode” either.

Although it's further than the topic, I've recently started to believe keys of the global state map don't have to be (and shouldn't be) addresses, but arbitrary strings instead, which is more easier to understand to existing programmers who are familiar to Redis or memcached.

Any opinions?

[1]: https://blog.cosmos.network/why-application-specific-blockchains-make-sense-32f2073bfb37

longfin:
That sounds good.

dahlia:
We will never get rid of `Transaction.UpdatedAddresses` field, because it already affected already signed transactions.  If we remove the field their signatures and IDs would change too.

However, we still can remove the guarantees that each transaction can change only the states listed in its `UpdateAddresses` at most, (and indeed such guarantee had been silently a dead letter since <https://github.com/planetarium/libplanet/pull/1389>,) and further get rid of the “rehearsal mode” to automatically list `UpdatedAddresses` as well.

[...]

PR #774, opened 2020-01-22
Make IStore to represent state keys as string (was Address)

dahlia:
It's the first step to refer to states by arbitrary `string` keys instead of `Address`es, as I suggested in the issue #368:

> Although it's further than the topic, I've recently started to believe keys of the global state map don't have to be (and shouldn't be) addresses, but arbitrary strings instead, which is more easier to understand to existing programmers who are familiar to Redis or memcached.

I am going to send more patches to make state keys arbitrary `string`s instead of `Address`es in the whole of Libplanet.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,21 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  The internal representation for state keys became `string` (was `Address`).
+    [[#368], [#774]]
+     -  The return type of `IStore.GetBlockStates()` method became
+        `IImmutableDictionary<string, IValue>` (was `AddressStateMap`,
+        which was removed too).  [[#368], [#774]]
+     -  The type of the second parameter of `IStore.SetBlockStates()` method
+        became `IImmutableDictionary<string, IValue>` (was `AddressStateMap`,
+        which was removed too).  [[#368], [#774]]
+     -  The second parameter of `IStore.IterateStateReferences()` method became
+        `string key` (was `Address address`).  [[#368], [#774]]
+     -  The second parameter of `IStore.StoreStateReference()` method became
+        `IImmutableSet<string> keys` (was `IImmutableSet<Address> addresses`).
+        [[#368], [#774]]
+     -  `IStore.ListAddresses()` method was replaced by `IStore.ListStateKeys()`
+        method. [[#368], [#774]]
  -  Added `Swarm<T>.FindSpecificPeer()` method to find a specific peer given
     the address.  [[#570], [#580]]
  -  Removed `LiteDBStore` class.  Use `DefaultStore` instead.  [[#662]]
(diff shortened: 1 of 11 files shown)


CHANGES.md
```diff
@@ -8,6 +8,21 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  The internal representation for state keys became `string` (was `Address`).
+    [[#368], [#774]]
+     -  The return type of `IStore.GetBlockStates()` method became
+        `IImmutableDictionary<string, IValue>`
+        (was `ImmutableDictionary<Address, IValue>)`. [[#368], [#774]]
```
limebell:
```suggestion
        (was `IImmutableDictionary<Address, IValue>)`. [[#368], [#774]]
```

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -587,7 +587,10 @@ public void ExecuteActions()
                 expectedStates.Count,
                 _blockChain.Store.GetBlockStates(block1.Hash).Count);
             Assert.Equal(
-                expectedStates.ToImmutableDictionary(),
+                expectedStates.ToImmutableDictionary(
+                    kv => kv.Key.ToHex().ToLowerInvariant(),
```
longfin: How about using `.ToString()` instead of `.ToHex()`? it adds `0x` as prefix so we can determine easily.

dahlia: That might a good idea, but we eventually remove all these `Address` → `string` conversions in this month.

CHANGES.md
dahlia: Fixed and amended the commit.

Libplanet.Tests/Blockchain/BlockChainTest.cs
longfin: 👌

dahlia:
@planetarium/libplanet I rebased this on the current master and fixed broken tests.

dahlia:
@earlbread @limebell @moreal Could you review this patch?

dahlia:
@planetarium/libplanet I amended the last commit due to some mistakes on the changelog.  Please review this again.

Libplanet/Store/IStore.cs
```diff
@@ -217,42 +216,53 @@ void PutBlock<T>(Block<T> block)
...
             HashDigest<SHA256> blockHash,
-            IImmutableDictionary<Address, IValue> states
+            IImmutableDictionary<string, IValue> states
         );
 
-#pragma warning disable MEN002
         /// <summary>
         /// Looks up a state reference, which is a block's <see cref="Block{T}.Hash"/> that contains
-        /// an action mutating the <paramref name="address"/>' state.
+        /// an action mutating the <paramref name="key"/>' state.
```
moreal: I'm not sure, but in my thought, it seems gone the reason to omit *s*.

```suggestion
        /// an action mutating the <paramref name="key"/>'s state.
```

dahlia: Fixed and amended the commit: 7e100f863e404cda983c05f0baa90905d2a8ecdd.

dahlia:
@planetarium/libplanet I amended the commit due to a typo.  Please review this again.

PR #777, opened 2020-01-28
Store states as files in DefaultStore

longfin:
This PR separates states from `index.ldb` to help maintain with large states. 

also, I skipped changelog because 0.8.0 introducing `DefaultStore` wasn't released yet. 😕

### DIFF
--- a/Libplanet/Store/DefaultStore.cs
+++ b/Libplanet/Store/DefaultStore.cs
@@ -35,12 +35,14 @@ namespace Libplanet.Store
 
         private static readonly UPath TxRootPath = UPath.Root / "tx";
         private static readonly UPath BlockRootPath = UPath.Root / "block";
+        private static readonly UPath StateRootPath = UPath.Root / "state";
 
         private readonly ILogger _logger;
 
         private readonly IFileSystem _root;
         private readonly SubFileSystem _txs;
         private readonly SubFileSystem _blocks;
+        private readonly SubFileSystem _states;
         private readonly bool _compress;
         private readonly LruCache<TxId, object> _txCache;
         private readonly LruCache<HashDigest<SHA256>, BlockDigest> _blockCache;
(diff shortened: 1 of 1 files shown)


Libplanet/Store/DefaultStore.cs
```diff
@@ -551,7 +571,6 @@ HashDigest<SHA256> blockHash
             HashDigest<SHA256> blockHash,
             IImmutableDictionary<Address, IValue> states)
         {
-            _statesCache.AddOrUpdate(blockHash, states);
```
moreal: It seems caching after calling `GetBlockStates()` at least once, is it intended?
If then, I wonder why `SetBlockStates()` doesn't set cache. 🤔

moreal: Oh sorry, it was moved to line 603. It was my mistake.

Libplanet/Store/DefaultStore.cs
```diff
@@ -524,20 +528,36 @@ HashDigest<SHA256> blockHash
...
+                    }
+                }
+                else
+                {
+                    file.CopyTo(buffer);
+                }
 
-                var deserialized = (Bencodex.Types.Dictionary)_codec.Decode(stream);
-                ImmutableDictionary<Address, IValue> states = deserialized.ToImmutableDictionary(
+                IValue value = new Codec().Decode(buffer.ToArray());
```
dahlia: `Bencodex.Codec` has an overload `Decode(Stream)` too.  You could reuse the `buffer` object by `buffer.Seek(0)` without allocating a new array.

Libplanet/Store/DefaultStore.cs
```diff
@@ -560,16 +579,28 @@ HashDigest<SHA256> blockHash
                     )
                 )
             );
-            using (var stream = new MemoryStream())
+
+            UPath path = StatePath(blockHash);
+            UPath dirPath = path.GetDirectory();
+            CreateDirectoryRecursively(_states, dirPath);
+
+            using (var buffer = new MemoryStream(new Codec().Encode(serialized)))
```
dahlia: As `Bencodex.Codec` has an overload `Encode(IValue, Stream)` as well, you don't have a `buffer` but could directly write to `deflate` or `file` instead, e.g., `new Codec().Encode(serialized, deflate)`.

PR #779, opened 2020-01-28
Add breadcrumb trail to messages

### LINKED ISSUE #771: Breadcrumb metadata for broadcasted messages (blocks and txs)
In order to investigate the path each message had come from so that we can be aware of each node's stay time.  This could be added as a field to the message header, like message signature, and the field should encode a sequence of (node id, departed timestamp) pairs.  For the precision, the departed timestamp would be better to include milliseconds or even microseconds.

limebell:
This patch adds metadata for breadcrumb trail to messages.

Todo:
- ~Add unit tests~
- ~Add viewer~

Closes #771


Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2682,6 +2682,141 @@ public async Task FindSpecificPeerAsync()
...
+                swarmC.Dispose();
+                swarmD.Dispose();
+            }
+        }
+
+        [Fact(Timeout = Timeout)]
+        public async Task BreadcrumbTrail()
+        {
+            var keys = new PrivateKey[5];
+            var fxs = new DefaultStoreFixture[5];
```
dahlia:
```suggestion
            var fxs = new DefaultStoreFixture[keys.Length];
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2682,6 +2682,141 @@ public async Task FindSpecificPeerAsync()
...
+            var blockChains = new BlockChain<DumbAction>[5];
+            var swarms = new Swarm<DumbAction>[5];
```
dahlia:
```suggestion
            var swarms = new Swarm<DumbAction>[fxs.Length];
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2682,6 +2682,141 @@ public async Task FindSpecificPeerAsync()
...
+                0x55, 0xd7, 0x38, 0xa3, 0x67, 0x48, 0x73, 0x08, 0xcd, 0x74,
+                0x3c, 0x4b,
+            });
+
+            var genesisBlock = BlockChain<DumbAction>.MakeGenesisBlock();
+
+            try
+            {
+                var tasks = new List<Task>();
+                for (int i = 0; i < 5; i++)
```
dahlia:
```suggestion
                for (int i = 0; i < fxs.Length; i++)
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2682,6 +2682,141 @@ public async Task FindSpecificPeerAsync()
...
+                swarmD.Dispose();
+            }
+        }
+
+        [Fact(Timeout = Timeout)]
+        public async Task BreadcrumbTrail()
+        {
+            var keys = new PrivateKey[5];
+            var fxs = new DefaultStoreFixture[5];
+            var blockChains = new BlockChain<DumbAction>[5];
```
dahlia:
```suggestion
            var blockChains = new BlockChain<DumbAction>[keys.Length];
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2682,6 +2682,141 @@ public async Task FindSpecificPeerAsync()
...
+                        blockChains[i],
+                        keys[i],
+                        0,
+                        host: IPAddress.Loopback.ToString());
+                    tasks.Add(StartAsync(swarms[i]));
+                }
+
+                await Task.WhenAll(tasks.ToArray());
+
+                for (int i = 0; i < 4; i++)
```
dahlia:
```suggestion
                for (int i = 0; i < swams.Length - 1; i++)
```

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2682,6 +2682,141 @@ public async Task FindSpecificPeerAsync()
...
+                expected += $"[{receivedTrail}:{swarms[4].Address}:" +
+                            $"{header.Trail.Trails[count].Timestamp}]";
+
+                Assert.Equal(
+                    expected,
+                    header.Trail.ToString());
+            }
+            finally
+            {
+                for (int i = 0; i < 5; i++)
```
dahlia:
```suggestion
                for (int i = 0; i < swarms.Length; i++)
```

Libplanet/FixedSizedQueue.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Libplanet
+{
+    /// <summary>
+    /// Implements a <see cref="ConcurrentQueue{T}"/> of a fixed size.
+    /// </summary>
+    /// <typeparam name="T">Specifies the type of elements in the queue.</typeparam>
+#pragma warning disable MEN002
+    // This class referenced the implementation of the following.
+    // https://github.com/couchcoding/Logbert/blob/bbed28fabf9493310f4b51324f8af88c14551c9d/src/Logbert/Helper/FixedSizedQueue.cs
```
dahlia: Rather than suppressing the `MEN002` warning, how about shortening the link using [gio.io](https://git.io/), GitHub's official shortener?

Libplanet/FixedSizedQueue.cs
```diff
@@ -0,0 +1,55 @@
...
+#pragma warning restore MEN002
+    internal class FixedSizedQueue<T> : ConcurrentQueue<T>
+    {
+        /// <summary>
+        /// Simple object for thread synchronization.
```
dahlia:
```suggestion
        /// Equivalent to <see cref="ConcurrentQueue{T}"/>, except this does not accept more than
        /// the specified maximum size.
```

limebell: Maybe review of line 6?

dahlia: Yeah, it was misplaced.

Libplanet.Tests/TestUtils.cs
```diff
@@ -21,6 +21,82 @@ public class TestUtils
...
+                0x59, 0x79,
+            }),
+            new PrivateKey(new byte[]
+            {
+                0x3d, 0xae, 0x2d, 0x2f, 0x87, 0x1b, 0x11, 0xaa, 0x41, 0xbd,
+                0xec, 0x81, 0x4c, 0x4d, 0x27, 0xf3, 0xba, 0xd9, 0x1f, 0x61,
+                0xc3, 0x57, 0xab, 0x43, 0xa0, 0x0c, 0x63, 0x1b, 0x2b, 0x15,
+                0x13, 0xf2,
+            }),
+        };
```
dahlia: :+1:

Libplanet/Net/Messages/BreadcrumbTrail.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        /// <inheritdoc/>
+        public override string ToString()
+        {
+            return Record
+                ? _trails.Aggregate(
+                    string.Empty,
+                    (s, trail) => s + $"[{trail}]/",
+                    s => s.TrimEnd('/'))
```
dahlia: Can't this use `string.Join()` method instead?

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2757,6 +2757,105 @@ public async Task FindSpecificPeerAsync()
...
+                swarmB.Dispose();
+                swarmC.Dispose();
+                swarmD.Dispose();
+            }
+        }
+
+        [Fact(Timeout = Timeout)]
+        public async Task BreadcrumbTrail()
+        {
+            int swarmCount = 5;
```
moreal: It seems better to be const variable because `swarmCount` does not change.

```suggestion
            const int swarmCount = 5;
```

Libplanet/Net/Messages/GetRecentStates.cs
```diff
@@ -12,19 +12,19 @@ internal class GetRecentStates : Message
...
         }
 
         public GetRecentStates(NetMQFrame[] frames)
-            : this(
-                new BlockLocator(frames.Skip(2).Select(f => new HashDigest<SHA256>(f.Buffer))),
-                new HashDigest<SHA256>(frames[1].Buffer),
-                frames[0].ConvertToInt64()
-            )
         {
+            BaseLocator =
```
longfin: Is this change necessary?

Libplanet/Net/Messages/Message.cs
```diff
@@ -11,6 +11,19 @@ namespace Libplanet.Net.Messages
 {
     internal abstract class Message
     {
+        internal const int HeaderSize = 4;
+
+        protected const bool LeaveTrail = true;
```
longfin: I guess it's okay that we omit this constant and use `true` or `false` directly. if we want to show its meaning, `base(trail: true)` is enough.

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -816,12 +816,9 @@ private void ReceiveFindPeer(FindNeighbors findNeighbors)
         {
             IEnumerable<BoundPeer> found = _routing.Neighbors(findNeighbors.Target, _bucketSize);
 
-            Neighbors neighbors = new Neighbors(found)
-            {
-                Identity = findNeighbors.Identity,
-            };
+            Neighbors neighbors = new Neighbors(found);
```
longfin:
```suggestion
            var neighbors = new Neighbors(found);
```

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -91,7 +91,9 @@ public IEnumerable<BoundPeer> PeersToBroadcast(Address? except)
             if (count < _minPeersToBroadcast)
             {
                 peers.AddRange(Peers
-                    .Where(peer => except is null || !peer.Address.Equals(except.Value))
+                    .Where(peer =>
+                        !peers.Contains(peer) &&
```
longfin: It doesn't seem related to adding breadcrumb. Is it intended? (if so, how about writing about it to changelogs?)

Libplanet/Net/Messages/GetRecentStates.cs
limebell: Yes, because message generated with `NetMQFrame`s should keep previous breadcrumb trail.

Libplanet/Net/Protocols/RoutingTable.cs
limebell: First time I thought this should be in a separated PR, but without this new test case fails. I'll add this change to the changelog.

Libplanet/Net/Messages/GetRecentStates.cs
longfin: Oh, I see. how about reusing `this()` constructor?

limebell:
Since broadcasting logic has changed, this patch is unnecessary.

PR #787, opened 2020-02-06
Introduce message history for testing

limebell:
This patch introduces message history queue for `NetMQTransport`. Since all related changes are internal, skipped changelog.

### DIFF
--- a/Libplanet/FixedSizedQueue.cs
+++ b/Libplanet/FixedSizedQueue.cs
@@ -1,0 +1,54 @@
+using System.Collections.Concurrent;
+
+namespace Libplanet
+{
+    /// <summary>
+    /// Equivalent to <see cref="ConcurrentQueue{T}"/>, except this does not accept more than
+    /// the specified maximum size.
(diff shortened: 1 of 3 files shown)


Libplanet/FixedSizedQueue.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Libplanet
+{
+    /// <summary>
+    /// Implements a <see cref="ConcurrentQueue{T}"/> of a fixed size.
+    /// </summary>
+    /// <typeparam name="T">Specifies the type of elements in the queue.</typeparam>
+#pragma warning disable MEN002
+    // This class referenced the implementation of the following.
+    // https://github.com/couchcoding/Logbert/blob/bbed28fabf9493310f4b51324f8af88c14551c9d/src/Logbert/Helper/FixedSizedQueue.cs
```
earlbread: How about shortening the URL like this and removing suppression warning?

```suggestion
    // https://git.io/Jvs1f
```

PR #798, opened 2020-02-21
Restore BlockCompletion<T> (#707) and more

dahlia:
The first commit is a commit squashed from the PR #707, except for some adjustments to resolve conflicts and make the build to pass.

This also introduces C# 8's new [async streams syntax][1] to replace Dasync's *[AsyncEnumerator]* package with it.  Although this removed the dependency on *AsyncEnumerator* package and introduce the new syntax into *Libplanet* project, note *Libplanet.Tests* project still uses C# 7.2 and the third-party package, which is intended.

Instead of *AsyncEnumerator*, *Libplanet* project has a new dependency on the *[System.Linq.Async]* package, which provides several LINQ extension methods laid on `IAsyncEnumerator<T>`.

The key changes over the reverted PR #707 in this patch are of course fixing #707's bugs and more `PreloadState`s.  See also each commit message.

[1]: https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/generate-consume-asynchronous-stream
[AsyncEnumerator]: https://www.nuget.org/packages/AsyncEnumerator/
[System.Linq.Async]: https://www.nuget.org/packages/System.Linq.Async/

### DIFF
--- a/Libplanet.Tests/Libplanet.Tests.csproj
+++ b/Libplanet.Tests/Libplanet.Tests.csproj
@@ -33,6 +33,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AsyncEnumerator" Version="4.0.2" />
     <PackageReference Include="Menees.Analyzers.2017" Version="2.0.3">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
--- a/Libplanet/Net/BlockCompletion.cs
+++ b/Libplanet/Net/BlockCompletion.cs
@@ -1,0 +1,335 @@
+using System;
+using System.Collections.Async;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Threading;
+using System.Threading.Tasks;
+using Libplanet.Action;
+using Libplanet.Blocks;
+using Serilog;
+
+namespace Libplanet.Net
+{
+    internal class BlockCompletion<TPeer, TAction>
+        where TAction : IAction, new()
+    {
+        private readonly ILogger _logger;
+        private readonly Func<HashDigest<SHA256>, bool> _completionPredicate;
+        private readonly int _window;
+        private readonly ConcurrentDictionary<HashDigest<SHA256>, bool> _satisfiedBlocks;
+        private readonly ConcurrentQueue<HashDigest<SHA256>> _demands;
+        private readonly SemaphoreSlim _demandEnqueued;
+        private bool _started;
+
+        public BlockCompletion(
+            Func<HashDigest<SHA256>, bool> completionPredicate = null,
+            int window = 100
+        )
+        {
+            _logger = Log.ForContext<BlockCompletion<TPeer, TAction>>();
+            _completionPredicate = completionPredicate;
+            _window = window;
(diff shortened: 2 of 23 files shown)


longfin:
Tests seem failed. 😢  

@dahlia Is it still in progress?

dahlia:
@longfin Fixed the broken tests.

Libplanet.Tests/Libplanet.Tests.csproj
```diff
@@ -33,6 +33,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AsyncEnumerator" Version="4.0.2" />
```
moreal: Why is it needed? `AsyncEnumerator` seemed to be replaced with `System.Linq.Async` on [the line](https://github.com/planetarium/libplanet/pull/798/files#diff-3bf6e3fb30349a8654faf74cc804ad61R96)

dahlia: This is only for tests:

> Although this removed the dependency on *AsyncEnumerator* package and introduce the new syntax into *Libplanet* project, note *Libplanet.Tests* project still uses C# 7.2 and the third-party package, which is intended.

Libplanet/Net/Swarm.cs
```diff
@@ -634,131 +846,168 @@ await DialToExistingPeers(dialTimeout, cancellationToken)
...
+                if (blockHashes.StartIndex is long idx)
+                {
+                    HashDigest<SHA256>[] hashes = blockHashes.Hashes.ToArray();
+                    _logger.Debug(
+                        $"Received a {nameof(BlockHashes)} message with an offset index " +
+                        "{OffsetIndex} (total {HashesLength} hashes).",
+                        idx,
+                        hashes.LongLength
+                    );
+                    foreach (var hash in hashes)
```
```suggestion
                    foreach (HashDigest<SHA256> hash in hashes)
```

Libplanet/Net/BlockCompletion.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+
+            PeerPool pool = new PeerPool(peers);
+            var queue = new AsyncProducerConsumerQueue<Tuple<Block<TAction>, TPeer>>();
+            var completion =
+                new ConcurrentDictionary<HashDigest<SHA256>, bool>(_satisfiedBlocks);
+
+            await foreach (var hashes in EnumerateChunks(cancellationToken))
+            {
+                foreach (HashDigest<SHA256> hash in hashes)
```
earlbread: Shouldn't [this](https://github.com/planetarium/libplanet/pull/798/files#diff-a155979ce9e3dfbb6ba0ea031a5b7a44R214) be done before here and use the `hashDigests` instead of `hashs`?

Libplanet/Net/BlockCompletion.cs
```diff
@@ -0,0 +1,466 @@
...
+            {
+                throw new ArgumentException("The list of peers must not be empty.", nameof(peers));
+            }
+
+            PeerPool pool = new PeerPool(peers);
+            var queue = new AsyncProducerConsumerQueue<Tuple<Block<TAction>, TPeer>>();
+            var completion =
+                new ConcurrentDictionary<HashDigest<SHA256>, bool>(_satisfiedBlocks);
+
+            await foreach (var hashes in EnumerateChunks(cancellationToken))
```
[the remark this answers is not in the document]
dahlia: I tried in that way first, but turned that into this way because the type made the line too long (> 100 columns). 😩

PR #801, opened 2020-02-27
Deploy Libplanet.RocksDBStore to NuGet

earlbread:
This deploys Libplanet.RocksDBStore to NuGet with the same package version as Libplanet.

### DIFF
--- a/.github/bin/dist-nuget.sh
+++ b/.github/bin/dist-nuget.sh
@@ -3,7 +3,7 @@
 # Note that this script is intended to be run by GitHub Actions.
 set -e
 
-project=Libplanet
+projects="Libplanet Libplanet.RocksDBStore"
 configuration=Release
 
 if [ ! -f obj/package_version.txt ]; then
@@ -30,7 +30,10 @@ if [ "$GITHUB_REPOSITORY" != "planetarium/libplanet" ] || (
 fi
 
 package_version="$(cat obj/package_version.txt)"
-dotnet nuget push \
-  "./$project/bin/$configuration/$project.$package_version.nupkg" \
-  --api-key "$NUGET_API_KEY" \
-  --source https://api.nuget.org/v3/index.json
+
+for project in $projects; do
+  dotnet nuget push \
+    "./$project/bin/$configuration/$project.$package_version.nupkg" \
+    --api-key "$NUGET_API_KEY" \
+    --source https://api.nuget.org/v3/index.json
+done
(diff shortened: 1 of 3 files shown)


.github/bin/dist-nuget.sh
```diff
@@ -3,7 +3,7 @@
 # Note that this script is intended to be run by GitHub Actions.
 set -e
 
-project=Libplanet
+projects="Libplanet Libplanet.RocksDBStore"
```
dahlia: How about using Bash's proper array:

```suggestion
projects=( "Libplanet" "Libplanet.RocksDBStore" )
```

.github/bin/dist-nuget.sh
```diff
@@ -30,7 +30,10 @@ if [ "$GITHUB_REPOSITORY" != "planetarium/libplanet" ] || (
 fi
 
 package_version="$(cat obj/package_version.txt)"
-dotnet nuget push \
-  "./$project/bin/$configuration/$project.$package_version.nupkg" \
-  --api-key "$NUGET_API_KEY" \
-  --source https://api.nuget.org/v3/index.json
+
+for project in $projects; do
```
dahlia: Using the proper array:

```suggestion
for project in "${projects[@]}"; do
```

.github/bin/dist-pack.sh
```diff
@@ -3,29 +3,31 @@
 # Note that this script is intended to be run by GitHub Actions.
 set -e
 
-project=Libplanet
+projects="Libplanet Libplanet.RocksDBStore"
```
dahlia:
```suggestion
projects=( "Libplanet" "Libplanet.RocksDBStore" )
```

.github/bin/dist-pack.sh
```diff
@@ -3,29 +3,31 @@
...
 configuration=Release
 
 if ! (env | grep '^GITHUB_'); then
   echo "This script is intended to be run by GitHub Actions." > /dev/stderr
   exit 1
 fi
 
-rm -rf "./$project/bin/$configuration/"
+for project in $projects; do
```
dahlia:
```suggestion
for project in "${projects[@]}"; do
```

.github/bin/dist-nuget.sh
earlbread: We're using POSIX sh which is not supporting arrays. Would it be better to change to bash instead?

earlbread: I added a commit to use bash array. 1f6aa35cbd66b5095b46bc1014b590c8974a5511

PR #803, opened 2020-02-28
Replace to use RocksDB for Tx Nonce and Canonical Chain ID 

earlbread:
This replaces to use RocksDB for Tx Nonce and Canonical Chain ID in `Libplanet.RocksDBStore`. The remainder of using LiteDB will be replaced in the next PRs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,18 @@ To be released.
 
  -  `BaseStore` class became to implement `IDisposable`.  [[#789]]
  -  Removed `IStore.DeleteIndex(Guid, HashDigest<SHA256>)` method.  [[#802]]
+ -  Extension classes have been renamed. However, it is not affected
+    if you do not call it directly.  [[#803]]
+     -  Renamed `StunAddressExtension` class to `StunAddressExtensions`.
+     -  Renamed `BytesConvertExtension` class to `BytesConvertExtensions`.
+     -  Renamed `RandomExtension` class to `RandomExtensions`.
+     -  Renamed `AddressExtension` class to `AddressExtensions`.
+     -  Renamed `HashDigestExtension` class to `HashDigestExtensions`.
+     -  Renamed `NetMQFrameExtension` class to `NetMQFrameExtensions`.
+     -  Renamed `NetMQSocketExtension` class to `NetMQSocketExtensions`.
+     -  Renamed `SerializationInfoExtension` class to
+        `SerializationInfoExtensions`.
+     -  Renamed `StoreExtension` class to `StoreExtensions`.
 
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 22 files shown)


CHANGES.md
```diff
@@ -31,6 +31,7 @@ To be released.
  -  Added `BlockHeader` struct.  [[#785]]
  -  Added `IStore.GetBlockDigest(HashDigest<SHA256>)` method.  [[#785]]
  -  Added `Block<T>.ToBlockDigest()` method.  [[#785]]
+ -  Added `ByteArrayExtension` class.  [[#803]]
```
longfin: We should add `#803` to the footnotes too.

Libplanet/ByteArrayExtension.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// true if <paramref name="prefix"/> matches the beginning of <paramref name="bytes"/>;
+        /// otherwise, false.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="bytes"/> or <paramref name="prefix"/> is null.
+        /// </exception>
+        [Pure]
+        public static bool StartsWith(this byte[] bytes, byte[] prefix)
+        {
+            if (bytes == null)
```
longfin:
```suggestion
            if (bytes is null)
```

Libplanet/ByteArrayExtension.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                throw new ArgumentNullException(nameof(bytes));
+            }
+
+            if (prefix == null)
```
longfin:
```suggestion
            if (prefix is null)
```

Libplanet/ByteArrayExtension.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Diagnostics.Contracts;
+
+namespace Libplanet
+{
+    /// <summary>
+    /// This extension class enables some convenient methods to deal with byte array.
+    /// </summary>
+    /// <seealso cref="Address"/>
+    public static class ByteArrayExtension
```
dahlia: How about renaming it to `ByteArrayExtensions`?   The convention seems to prefer plural over singular.

earlbread: Renamed every extension class to plural.

CHANGES.md
```diff
@@ -10,6 +10,18 @@ To be released.
 
  -  `BaseStore` class became to implement `IDisposable`.  [[#789]]
  -  Removed `IStore.DeleteIndex(Guid, HashDigest<SHA256>)` method.  [[#802]]
+ -  Extension classes have been renamed. However, it is not affected
+    if you do not call it directly.  [[#803]]
```
```suggestion
 -  Extension classes was renamed. However, it would not be affected
    if you have called it by using instance method syntax.  [[#803]]
```

dahlia: This seems reverted?

earlbread: My mistake. Restored it.

PR #805, opened 2020-03-02
Replace to use RocksDB for Chain Index

earlbread:
This replaces to use RocksDB for chain indexes in Libplanet.RocksDBStore.

### DIFF
--- a/Libplanet.RocksDBStore/RocksDBStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBStore.cs
@@ -26,13 +26,13 @@ namespace Libplanet.RocksDBStore
     /// <seealso cref="IStore"/>
     public class RocksDBStore : BaseStore
     {
-        private const string IndexColPrefix = "index_";
-
         private const string StateRefIdPrefix = "stateref_";
 
         private const string BlockDbName = "blockdb";
 
         private const string ChainDbName = "chaindb";
+
+        private static readonly byte[] IndexKeyPrefix = { (byte)'I' };
 
         private static readonly byte[] BlockKeyPrefix = { (byte)'B' };
 
@@ -41,6 +41,8 @@ namespace Libplanet.RocksDBStore
         private static readonly byte[] TxKeyPrefix = { (byte)'T' };
 
         private static readonly byte[] TxNonceKeyPrefix = { (byte)'N' };
+
+        private static readonly byte[] IndexCountKey = { (byte)'c' };
 
         private static readonly byte[] CanonicalChainIdIdKey = { (byte)'C' };
 
@@ -55,6 +57,8 @@ namespace Libplanet.RocksDBStore
             _lastStateRefCaches;
 
         private readonly LiteDatabase _liteDb;
+        private readonly DbOptions _options;
+        private readonly string _path;
         private readonly RocksDb _blockDb;
         private readonly RocksDb _chainDb;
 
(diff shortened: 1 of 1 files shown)


Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -198,9 +221,28 @@ public override long CountIndex(Guid chainId)
...
             int? limit)
         {
-            return IndexCollection(chainId)
-                .Find(Query.All(), offset, limit ?? int.MaxValue)
-                .Select(i => i.Hash);
+            int skip = 0;
+            int count = 0;
+            byte[] prefix = IndexKeyPrefix;
+
+            foreach (Iterator it in IterateDb(_chainDb, prefix, chainId))
```
moreal: Is there reason to skip as offset manually, instead `IEnumerable.Skip()`?

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -788,9 +849,25 @@ private byte[] TxNonceKey(Address address)
                 .ToArray();
         }
 
-        private ColumnFamilyHandle GetColumnFamilyFromChainId(Guid chainId)
+        private IEnumerable<Iterator> IterateDb(RocksDb db, byte[] prefix, Guid? chainId = null)
         {
+            ColumnFamilyHandle cf = GetColumnFamily(chainId);
+            Iterator it = db.NewIterator(cf);
+            for (it.Seek(prefix); it.Valid() && it.Key().StartsWith(prefix); it.Next())
```
moreal: It's just a question. Now it checks prefix to match one more time because of bloom filter. Then does it guarantee values not matching the prefix come in the last? If not, it can't guarantee to iterate all values matching the prefix. doesn't it?

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -153,15 +159,28 @@ public class RocksDBStore : BaseStore
...
-                .Select(name => ParseChainId(name.Substring(IndexColPrefix.Length)));
+            string path = Path.Combine(_path, ChainDbName);
+
+            foreach (string name in RocksDb.ListColumnFamilies(_options, path))
+            {
+                Guid guid;
+
+                try
+                {
+                    guid = new Guid(name);
```
dahlia: According to [a micro benchmark conducted by someone at Stack Overflow][1], `Guid.Parse(string)` is slighly faster than `new Guid(string)`.

```suggestion
                    guid = Guid.Parse(name);
```

[1]: https://stackoverflow.com/a/17570268/383405

Libplanet.RocksDBStore/RocksDBStore.cs
earlbread: No, I'll fix it.

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -216,13 +258,45 @@ public override long CountIndex(Guid chainId)
...
 
-            return IndexCollection(chainId).FindById(index + 1)?.Hash;
+            ColumnFamilyHandle cf = GetColumnFamily(chainId);
+            byte[] indexBytes = BitConverter.GetBytes(index);
+
+            // Use Big-endian to order index lexicographically.
+            if (BitConverter.IsLittleEndian)
+            {
+                Array.Reverse(indexBytes);
+            }
```
dahlia: Instead of this kind of branch, how about using `NetworkOrderBitsConverter` provided by NetMQ? Some classes in the `Libplanet.Net.Messages` namespace have been already used this.

```suggestion
            byte[] indexBytes = NetMQ.NetworkOrderBitsConverter.GetBytes(index);
```

Libplanet.RocksDBStore/RocksDBStore.cs
dahlia: @moreal +1. I believe his question has a valid point.  If the result iterator can emit either matched keys and mismatched keys and they are mixed between each other (e.g., `a1`, `b1`, `a2`, `b2`, … where prefix = `a`) the current code could omit every matched key appeared later than the first unmatched key.

earlbread: Yes, it is guaranteed that mismatched keys come after the matched key because the returned key is ordered byte-wise.

dahlia: Fair enough.

earlbread:
I applied suggestions and added commits. ce3050e, b3235e3, be8f2d3

PR #808, opened 2020-03-03
Replace State Reference to use RocksDB

earlbread:
This replaces to use RocksDB for state references in Libplanet.RocksDBStore.

### DIFF
--- a/Libplanet.RocksDBStore/RocksDBStoreBitConverter.cs
+++ b/Libplanet.RocksDBStore/RocksDBStoreBitConverter.cs
@@ -1,0 +1,71 @@
+using System;
+using System.Text;
+
+namespace Libplanet.RocksDBStore
+{
+    /// <summary>
+    /// Collection of wrapper methods to convert to byte-arrays from various types to store
+    /// <see cref="RocksDBStore"/>.  Every integer data is converted to Big-endian, or
+    /// "Network Byte Order" to order index lexicographically.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


Libplanet.RocksDBStore/RocksDBStoreBitConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+    internal static class RocksDBStoreBitConverter
+    {
+        /// <summary>
+        /// Get <c>long</c> representation of the <paramref name="value"/>.
+        /// </summary>
+        /// <param name="value">The Big-endian byte-array value to convert to <c>long</c>.</param>
+        /// <returns>The <c>long</c> representation of the <paramref name="value"/>.</returns>
+        public static long ToInt64(byte[] value)
+        {
+            return NetMQ.NetworkOrderBitsConverter.ToInt64(value);
```
dahlia: As we've decided to have this our own utility class, it would be better if it's independent from NetMQ.  You've once implemented this without NetMQ already, right?

earlbread: You're right, I'll fix it.

earlbread: Changed implementation and amended the commit. https://github.com/planetarium/libplanet/pull/808/commits/7efba51a500a0e9a241136d149d6b4b701b6ac02

PR #811, opened 2020-03-05
All-in-one bundles

dahlia:
For development environments that cannot leverage NuGet like Unity, distribute all-in-one bundles that include all referenced assemblies including Libplanet in itself.

Bundles are an archive file which consists of all required assemblies (*\*.dll*) for a target framework.  Currently 4 target frameworks are supported:

<https://github.com/planetarium/libplanet/blob/4db8cd603325f75d57445a417d568cb409f07382/.github/workflows/main.yml#L107>

There are 2 types of bundle packages: *Libplanet* & *Libplanet.RocksDBStore*.  A *Libplanet.RocksDBStore* bundle consists of all assemblies in a *Libplanet* bundle with RocksDB-related assemblies.

For every push (and pull request), 2 packages × 4 target frameworks = 8 bundles are made in GitHub Actions, and these are uploaded as a single artifact named *bundles*.  You could check how bundles look like in the following GitHub Actions artifact:

<https://github.com/dahlia/libplanet/suites/499499235/artifacts/2529016>

For tag pushes, these bundles are also uploaded to GitHub Actions besides *\*.nupkg* files:

### DIFF
--- a/.github/bin/constants.sh
+++ b/.github/bin/constants.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# shellcheck disable=SC2034
+projects=("Libplanet" "Libplanet.RocksDBStore")
+configuration=Release
+
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,10 +31,16 @@ jobs:
         name: dist-obj
         path: obj/
     - run: .github/bin/dist-pack.sh
+    - run: |
+        . .github/bin/constants.sh
+        mkdir -p /tmp/dist-bin/
+        for project in "${projects[@]}"; do
+          cp -r "$project/bin/$configuration"/* /tmp/dist-bin/
+        done
     - uses: actions/upload-artifact@master
       with:
         name: dist-bin
-        path: Libplanet/bin/Release/
+        path: /tmp/dist-bin/
     - if: >-
         github.event_name != 'pull_request' &&
         startsWith(github.ref, 'refs/tags/')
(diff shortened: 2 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -80,3 +86,49 @@ jobs:
...
+    - uses: actions/checkout@master
+      if: github.event_name == 'pull_request'
+      with:
+        ref: ${{ github.pull_request.head.sha }}
+    - uses: actions/download-artifact@master
+      with:
+        name: dist-bin
+        path: /tmp/nupkg
+    - run: |
+        targets=(netcoreapp3.1 net472 net462 net461)
```
earlbread: Shouldn't we add netstandard2.0?

.github/bin/constants.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# shellcheck disable=SC2034
+projects=("Libplanet" "Libplanet.RocksDBStore")
+configuration=Release
```
earlbread: How about export these?

dahlia: Arrays cannot be exported. 😇

earlbread: Oh, I didn't know that. :cry:

.github/workflows/main.yml
dahlia: It would be better if we can support `netstandard2.0`, but a console app seems unable to be `netstandard2.0`…

PR #815, opened 2020-03-06
Signed app protocol version & new callback interface to negotiate versions

dahlia:
- App protocol versions had been an `Int32` before, but now is an `AppProtocolVersion`, which is a distinct value type.
- App protocol versions became signed whether other peers trust them or not.
- `Swarm<T>` now have an option named `trustedAppProtocolVersionSigners`.
- Events to deal with encountered different versions became replaced by a new delegate interface.
- As *SwarmTest.cs* is too long, I separated them into few files using `partial class`.


Libplanet/Net/AppProtocolVersion.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// Verifies whether the claim is certainly signed by the <see cref="Signer"/>.
+        /// </summary>
+        /// <param name="publicKey">A public key of the <see cref="Signer"/>.</param>
+        /// <returns><c>true</c> if and only if the given <paramref name="publicKey"/> is
+        /// <see cref="Signer"/>'s and the <see cref="Signature"/> is certainly signed by
+        /// the <see cref="Signer"/>.</returns>
+        [Pure]
+        public bool Verify(PublicKey publicKey) =>
+            Signer.Equals(new Address(publicKey)) &&
+            publicKey.Verify(GetMessage(), Signature.ToBuilder().ToArray());
```
earlbread: Is `ToBuilder()` needed?

Libplanet/Net/NetMQTransport.cs
```diff
@@ -378,9 +382,24 @@ public async Task RunAsync(CancellationToken cancellationToken)
...
             }
-            catch (DifferentAppProtocolVersionException)
+            catch (DifferentAppProtocolVersionException e)
             {
-                _logger.Debug($"Different version encountered during {nameof(AddPeersAsync)}().");
+                AppProtocolVersion expected = e.ExpectedVersion, actual = e.ActualVersion;
+                _logger.Debug(
+                    $"Different version encountered during {nameof(AddPeersAsync)}().\n" +
+                    "Expected version: {ExpectedVersion} ({ExpectedVersionExtra}) " +
+                    "[{ExpectedSignature}; {ExpectedSigner}]\n" +
```
earlbread: I don't know how much it will be used, but how about adding a `ToString()` method to `AppProtocolVersion`?

dahlia: `AppProtocolVersion` already has `ToString()` method, but for the sake of convenience, I made it to omit signature and signer that lengths. Wouldn't it be inconvenient even if every time its string representation is longer than ~250 characters?

Libplanet/Net/AppProtocolVersion.cs
dahlia: Without `ToBuild()` the array is built through LINQ which is way general, and I thought building an array through `ImmutableArray<T>.Builder`, which is specific and aware of the implementation details of `ImmutableArray<T>`, would be more efficient, but it's gut feeling. 🙄

Libplanet/Net/NetMQTransport.cs
earlbread: Okay, make sense.

Libplanet.Tests/Net/PeerTest.cs
```diff
@@ -12,6 +12,8 @@ public class PeerTest
     {
         public static IEnumerable<object[]> GetPeers()
         {
+            var signer = new PrivateKey();
+            var ver = new AppProtocolVersion(signer, 1);
```
longfin: I guess it would be better if we provide a factory method like `Block<T>.Mine()` instead of constructer.

Libplanet/Net/Peer.cs
```diff
@@ -70,11 +85,39 @@ protected Peer(SerializationInfo info, StreamingContext context)
...
+            if (AppProtocolVersion.Equals(localVersion))
+            {
+                return true;
+            }
+
+            if (trustedSigners is null || trustedSigners.Any(s => AppProtocolVersion.Verify(s)))
+            {
+                return recognizer is null
+                    ? false
+                    : recognizer(this, AppProtocolVersion, localVersion);
```
longfin: It may be a little complicated to run the callback in this predicate... 😕

Libplanet.Tests/Net/PeerTest.cs
dahlia: You mean it would be better if we can omit the first argument, `signer` here?

dahlia: Ah, I understood.  You mean *this* overload of the constructor misleads code readers into considering the given `PrivateKey` will be actually stored as a part of `AppProtocolVersion`'s internal state.  Fair enough.  I will turn this into a static method instead of a constructor.

Libplanet/Net/Peer.cs
dahlia: Indeed, but I haven't found any better approach yet… 😇

Libplanet/Net/DifferentAppProtocolVersionException.cs
```diff
@@ -23,10 +23,13 @@ public sealed class DifferentAppProtocolVersionException : Exception
...
-            int expectedVersion,
-            int actualVersion)
-            : base($"Expected Version: {expectedVersion}, " +
-                   $"Actual Version: {actualVersion}, {message}")
+            AppProtocolVersion expectedVersion,
+            AppProtocolVersion actualVersion)
+            : base($"{message}\n" +
+                   $"Expected Version: {expectedVersion} ({expectedVersion.Extra}) " +
+                   $"[{ByteUtil.Hex(expectedVersion.Signature)} by {expectedVersion.Signer}]\n" +
+                   $"Actual Version: {actualVersion} ({actualVersion.Extra}) " +
```
moreal:
```suggestion
                   $"Expected Version: {expectedVersion} " +
                   $"[{ByteUtil.Hex(expectedVersion.Signature)} by {expectedVersion.Signer}]\n" +
                   $"Actual Version: {actualVersion} " +
```

If it had `Extra` value, `AppProtocolVersion.ToString()` will return `{Version} {Extra}`.
So it doesn't need to insert `Extra` field one more, does it?

dahlia: You're right. I'll fix this too.

dahlia:
@planetarium/libplanet I've addressed reviews you left.  Please review this again.

PR #821, opened 2020-03-13
Fix RocksDBStore ForkBlockIndex

earlbread:
This fixes `RocksDBStore.ForkBlockIndex()` method and adds `StoreTest.ForkBlockIndex`.

### DIFF
--- a/Libplanet.RocksDBStore/RocksDBStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBStore.cs
@@ -248,14 +248,21 @@ namespace Libplanet.RocksDBStore
             Guid destinationChainId,
             HashDigest<SHA256> branchPoint)
         {
-            foreach (HashDigest<SHA256> hash in IterateIndexes(sourceChainId, 0, null))
+            var genesisHash = IterateIndexes(sourceChainId, 0, 1).First();
+
+            if (branchPoint.Equals(genesisHash))
             {
+                return;
+            }
+
+            foreach (HashDigest<SHA256> hash in IterateIndexes(sourceChainId, 1, null))
+            {
+                AppendIndex(destinationChainId, hash);
+
                 if (hash.Equals(branchPoint))
                 {
                     break;
                 }
-
-                AppendIndex(destinationChainId, hash);
             }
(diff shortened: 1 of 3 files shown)


Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -248,14 +248,21 @@ public override long AppendIndex(Guid chainId, HashDigest<SHA256> hash)
             Guid destinationChainId,
             HashDigest<SHA256> branchPoint)
         {
-            foreach (HashDigest<SHA256> hash in IterateIndexes(sourceChainId, 0, null))
+            var genesisHash = IterateIndexes(sourceChainId, 0, 1).First();
```
longfin:
```suggestion
            HashDigest<SHA256> genesisHash = IterateIndexes(sourceChainId, 0, 1).First();
```

longfin: It would be better if there is an empty check for `IterateIndexes(sourceChainId, 0, 1)`.

dahlia: It has never checked until now, even in `DefaultStore`: 😇

https://github.com/planetarium/libplanet/blob/946dc8cd610e7d9ae2f22b7e94ab16b74a2565cb/Libplanet/Store/DefaultStore.cs#L259

earlbread: Should we check it, even there is always a genesis block?

dahlia: I believe that assumption is applied only to `BlockChain<T>`. 🤔 It would be better if we *define* the behavior for an empty source (and it would probably be raising a certain exception) and mention that in the docs of `IStore.ForkBlockIndexes()`.

earlbread: Okay, I agree. I'll fix it.

earlbread: Added commit 959b897.

PR #822, opened 2020-03-13
Do not run upstream specific workfolows in forks

gurrpi:
This PR changes the scheduled jobs and publishing artifact to be allowed only upstream. It was motivated by nightly builds on my forked repo that trigger email notifications for failed runs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,7 @@ name: main
 
 jobs:
   build:
+    if: github.event_name != 'schedule' || github.repository == 'planetarium/libplanet'
     name: dist
     runs-on: ubuntu-18.04
     steps:
@@ -47,12 +48,13 @@ jobs:
       run: .github/bin/dist-github-release.sh
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    - if: github.event_name != 'pull_request'
+    - if: github.event_name != 'pull_request' && env.NUGET_API_KEY != ''
       run: .github/bin/dist-nuget.sh
       env:
         NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
 
   docs:
+    if: github.event_name != 'schedule' || github.repository == 'planetarium/libplanet'
     name: docs
     runs-on: ubuntu-18.04
     steps:
@@ -85,7 +87,7 @@ jobs:
           --state=success
           --url="$(cat Docs/obj/url.txt)"
           --token="$GH_CHECK_STATUS_TOKEN"
-      if: github.event_name != 'pull_request'
+      if: github.event_name != 'pull_request' && env.GH_CHECK_STATUS_TOKEN != ''
 
   bundle:
     name: bundle


.github/workflows/main.yml
```diff
@@ -69,11 +76,16 @@ jobs:
...
+        github.repository == 'planetarium/libplanet'
+      run: Docs/publish.sh
       env:
         GHPAGES_SSH_KEY: ${{ secrets.GHPAGES_SSH_KEY }}
-      if: github.event_name != 'pull_request'
-    - uses: shawnbot/action-status@master
+    - if: >-
+        github.event_name != 'pull_request' &&
+        github.repository == 'planetarium/libplanet'
+      uses: shawnbot/action-status@master
```
dahlia: I believe these two changes are unnecessary, because it's enough that the `docs` job in itself is under the same condition.  These changes also affect other active contributors' workflow as well because we sometimes check how things we've added look like in the docs, before we submit a pull request to the upstream.

.github/workflows/main.yml
```diff
@@ -47,12 +51,15 @@ jobs:
       run: .github/bin/dist-github-release.sh
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    - if: github.event_name != 'pull_request'
+    - if: >-
+        github.event_name != 'pull_request' &&
+        github.repository == 'planetarium/libplanet'
```
dahlia: I believe this change is unnecessary, because it's enough that the `build` job in itself is under the same condition.

.github/workflows/main.yml
```diff
@@ -47,12 +51,15 @@ jobs:
...
       run: .github/bin/dist-nuget.sh
       env:
         NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
 
   docs:
+    if: (github.event_name == 'schedule' && github.repository == 'planetarium/libplanet') || (github.event_name != 'schedule')
```
dahlia: Since it's logically equivalent to the `build` job's `if` condition, and the hard-coded `'planetarium/libplanet'` is repeated, it would be better if the dup is eliminated using an [anchor].  I recommend you to read [this SO answer][1] if you aren't aware of YAML's this feature.

[anchor]: https://yaml.org/spec/1.2/spec.html#id2760569
[1]: https://stackoverflow.com/a/5954748/383405

gurrpi: I tried to using anchor but I figured out that [GitHub Actions currently didn't support anchors](https://github.com/gurrpi/libplanet/runs/511133814). 
```
### ERRORED 13:39:08Z

- Your workflow file was invalid: The pipeline is not valid. .github/workflows/main.yml: Anchors are not currently supported. Remove the anchor 'condition'
```

.github/workflows/main.yml
gurrpi: I found that the [Run .github/bin/dist-nuget.sh](https://github.com/gurrpi/libplanet/runs/511138491?check_suite_focus=true) step failed on the forked repository when triggered by `on: push`.

```
Run .github/bin/dist-nuget.sh
This script requires NUGET_API_KEY envrionment variable
##[error]Process completed with exit code 1.
```

.github/workflows/main.yml
dahlia: Okay, it can't be helped. 😇 Could you adjust the code style then, so that two equivalent `if` conditions look exactly same?

.github/workflows/main.yml
dahlia: That's because [it should be configured in the repository settings.][1]

[1]: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets

.github/workflows/main.yml
gurrpi: I found that the [Run shawnbot/action-status@master](https://github.com/gurrpi/libplanet/runs/511138510?check_suite_focus=true) step failed on the forked repository when triggered by `on: push`.
```
cat: Docs/obj/url.txt: No such file or directory
8
{ HttpError: Bad credentials
9
    at response.text.then.message (/action-status/node_modules/@octokit/request/lib/request.js:55:27)
10
    at process._tickCallback (internal/process/next_tick.js:68:7)
11
  name: 'HttpError',
12
  status: 401,

<below is intentionally omitted. See above link to see full message.>
```

gurrpi: I understand that the ` Docs/publish.sh` step should not have condition.

.github/workflows/main.yml
```diff
@@ -73,7 +81,10 @@ jobs:
       env:
         GHPAGES_SSH_KEY: ${{ secrets.GHPAGES_SSH_KEY }}
       if: github.event_name != 'pull_request'
-    - uses: shawnbot/action-status@master
+    - if: >-
+        github.event_name != 'pull_request' &&
+        github.repository == 'planetarium/libplanet'
+      uses: shawnbot/action-status@master
```
gurrpi: `L84:86` remain because I couldn't find to silent this step on fork repo.

.github/workflows/main.yml
```diff
@@ -11,6 +11,10 @@ name: main
 
 jobs:
   build:
+    if: >-
+      (github.event_name != 'schedule') ||
+      (github.event_name == 'schedule' &&
+      github.repository == 'planetarium/libplanet')
```
dahlia: Couldn't this be simplified like below?:

```suggestion
    if: github.event_name != 'schedule' || github.repository == 'planetarium/libplanet'
```

.github/workflows/main.yml
dahlia: The error you'd encountered was this, right?:

    cat: Docs/obj/url.txt: No such file or directory

The reason why that file does not exist is that the previous step (`Docs/publish.sh`) hasn't executed (`if: github.event_name != 'pull_request'`).  Therefore, it would be find if these two steps have the same condition.

gurrpi: The (`Docs/publish.sh`) step has executed but the (`hawnbot/action-status@master`) step failed due to empty token (`GH_CHECK_STATUS_TOKEN`). 

I replaced above condition (`github.event_name != 'pull_request' && github.repository == 'planetarium/libplanet') to check empty token (`github.event_name != 'pull_request' && env.GH_CHECK_STATUS_TOKEN != ''`)

gurrpi:
1. The condition of each job(`dist`. `docs`) simplified through reflecting @dahlia 's comment.
2. I introduced empty token check for `GH_CHECK_STATUS_TOKEN` and `NUGET_API_KEY` to prevent failed build.

.github/workflows/main.yml
```diff
@@ -47,12 +48,13 @@ jobs:
       run: .github/bin/dist-github-release.sh
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    - if: github.event_name != 'pull_request'
+    - if: github.event_name != 'pull_request' && env.NUGET_API_KEY != ''
```
dahlia: Does it fail without this change?  I believe the *dist-nuget.sh* script already has addressed this case:

https://github.com/planetarium/libplanet/blob/e65b5e7161478a393e115b8cbe60f6bfb91aa0e0/.github/bin/dist-nuget.sh#L25-L36

gurrpi: I'm not sure but maybe this change might be better than following the below guide:
> It would be silent if you configure a secret named NUGET_API_KEY with an empty string in your fork repo.

dahlia: Fair enough.

PR #825, opened 2020-03-17
AppProtocolVersion tokens & PrivateKey.ToAddress()

dahlia:
In order to make `AppProtocolVersion` values easy to copy and paste, the concept of tokens is added to `AppProtocolVersion`.  Tokens consist of ASCII characters so that they can be pasted without escaping into command arguments or strings in shells and JSON/YAML etc.

As a bonus, I added the `PrivateKey.ToAddress()` extension method as well.  As this change affected many files, I recommend you to review each single commit.

### DIFF
--- a/Libplanet/Net/AppProtocolVersion.cs
+++ b/Libplanet/Net/AppProtocolVersion.cs
@@ -66,6 +66,38 @@ namespace Libplanet.Net
             Signer = signer;
         }
 
+        /// <summary>
+        /// A token string which serializes an <see cref="AppProtocolVersion"/>.
+        /// <para>As this is designed to be easy to copy and paste, the format consists of only
+        /// printable characters in the ASCII character set.</para>
+        /// <para>A token can be deserialized into an <see cref="AppProtocolVersion"/> through
+        /// <see cref="FromToken(string)"/> method.</para>
+        /// </summary>
+        /// <seealso cref="FromToken(string)"/>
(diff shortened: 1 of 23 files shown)


Libplanet/Net/AppProtocolVersion.cs
```diff
@@ -66,6 +66,38 @@ Address signer
...
+        /// <seealso cref="FromToken(string)"/>
+        [Pure]
+        public string Token
+        {
+            get
+            {
+                string sig = Convert.ToBase64String(
+                    Signature.ToBuilder().ToArray(),
+                    Base64FormattingOptions.None
+                ).Replace('/', '.');
```
longfin: How about using `;` or `.` as delimiter instead of `/`? then it seems can be omitted.

Libplanet/Net/Swarm.cs
```diff
@@ -1692,7 +1692,7 @@ CancellationToken cancellationToken
                     {
                         _logger.Debug(
                             "Peer [{0}] didn't return any hashes; ignored.",
-                            peer.PublicKey.ToAddress().ToHex()
+                            peer.ToAddress().ToHex()
```
longfin: `BoundPeer` doesn't have `.ToAddress()`. it seems to break builds. 😢

dahlia:
@earlbread @moreal Could you review this?

PR #827, opened 2020-03-18
Add IBlockPolicy.IsTransactionValid

earlbread:
This adds `IBlockPolicy.IsValidTransaction` method which is a method to determine if a transaction is valid.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,8 +51,8 @@ To be released.
     that receives only one transaction.  [[#820]]
  -  Replaced `BlockChain<T>.UnstageTransactions()` with `.UnstageTransaction()`
     that receives only one transaction.  [[#820]]
- -  Added `IBlockPolicy.IsTransactionValid` method which is a method to
-    determine if a transaction is valid.  [[#827]]
+ -  Added `IBlockPolicy.DoesTransactionFollowPolicy()` method which is a method to
+    determine if a transaction follows the block policy.  [[#827]]
 
 ### Backward-incompatible network protocol changes
 
@@ -85,7 +85,7 @@ To be released.
  -  Added `Swarm<T>.TrustedAppProtocolVersionSigners` property.
     [[#266], [#815]]
  -  Added `Peer.IsCompatibleWith()` method.  [[#266], [#815]]
- -  Added `InvalidTxByBlockPolicyException` class.  [[#827]]
+ -  Added `TxViolatingBlockPolicyException` class.  [[#827]]
 
 ### Behavioral changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,7 +26,7 @@ To be released.
     represented as an `Int32` became retyped to `AppProtocolVersion`.
     [[#266], [#815]]
      -  `Swarm()` constructor's parameter `appProtocolVersion` became
-        `AppProtocolVersion` (was `Int32`). 
+        `AppProtocolVersion` (was `Int32`).
      -  `Peer()` and `BoundPeer()` constructors' parameter `appProtocolVersion`
         became `AppProtocolVersion` (was `Int32`).
      -  `Peer.AppProtocolVersion` property became `AppProtocolVersion`
@@ -51,6 +51,8 @@ To be released.
     that receives only one transaction.  [[#820]]
  -  Replaced `BlockChain<T>.UnstageTransactions()` with `.UnstageTransaction()`
     that receives only one transaction.  [[#820]]
+ -  Added `IBlockPolicy.IsTransactionValid` method which is a method to
+    determine if a transaction is valid.  [[#827]]
 
 ### Backward-incompatible network protocol changes
 
@@ -83,6 +85,7 @@ To be released.
  -  Added `Swarm<T>.TrustedAppProtocolVersionSigners` property.
     [[#266], [#815]]
  -  Added `Peer.IsCompatibleWith()` method.  [[#266], [#815]]
+ -  Added `InvalidTxByBlockPolicyException` class.  [[#827]]
 
 ### Behavioral changes
 
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -621,11 +621,24 @@ namespace Libplanet.Blockchain
             long index = Store.CountIndex(Id);
             long difficulty = Policy.GetNextBlockDifficulty(this);
             HashDigest<SHA256>? prevHash = Store.IndexBlockHash(Id, index - 1);
-            IEnumerable<Transaction<T>> transactions = Store
+            IEnumerable<Transaction<T>> stagedTransactions = Store
                 .IterateStagedTransactionIds()
-                .Select(Store.GetTransaction<T>)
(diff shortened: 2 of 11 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -621,11 +621,24 @@ public long GetNextTxNonce(Address address)
...
+            IEnumerable<Transaction<T>> stagedTransactions = Store
                 .IterateStagedTransactionIds()
-                .Select(Store.GetTransaction<T>)
-                .Where(tx => Store.GetTxNonce(Id, tx.Signer) <= tx.Nonce
-                             && tx.Nonce < GetNextTxNonce(tx.Signer));
+                .Select(Store.GetTransaction<T>);
+
+            var transactionsToMine = new List<Transaction<T>>();
+
+            foreach (var tx in stagedTransactions)
```
```suggestion
            foreach (Transaction<T> tx in stagedTransactions)
```

Libplanet/Blockchain/Policies/IBlockPolicy.cs
```diff
@@ -18,6 +20,11 @@ public interface IBlockPolicy<T>
         /// </summary>
         IAction BlockAction { get; }
 
+        /// <summary>
+        /// A method that determines if the transaction is valid.
+        /// </summary>
+        Func<Transaction<T>, bool> IsValidTransaction { get; }
```
longfin: How about changing `IsValidTransaction()` to simple predicate instead of returning `Func`?

```csharp
bool IsValidTransaction(Transaction<T> transaction)
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -748,6 +761,13 @@ bool renderActions
                 // the tx nounce order when the block was created
                 foreach (Transaction<T> tx1 in block.Transactions)
                 {
+                    if (!Policy.IsValidTransaction(tx1))
+                    {
+                        throw new InvalidTxException(
```
longfin: It may be better if we add some specific exception types instead of using concrete `InvalidTxException.`

Libplanet/Blockchain/Policies/IBlockPolicy.cs
earlbread: Changed to a normal method and renamed to `IsTransactionValid`.

Libplanet/Blockchain/BlockChain.cs
earlbread: Added `InvalidTxByBlockPolicyException`.

CHANGES.md
```diff
@@ -51,6 +51,8 @@ To be released.
     that receives only one transaction.  [[#820]]
  -  Replaced `BlockChain<T>.UnstageTransactions()` with `.UnstageTransaction()`
     that receives only one transaction.  [[#820]]
+ -  Added `IBlockPolicy.IsTransactionValid` method which is a method to
```
dahlia: A trivial adjustment:

```suggestion
 -  Added `IBlockPolicy.IsTransactionValid()` method which is a method to
```

CHANGES.md
```diff
@@ -83,6 +85,7 @@ To be released.
  -  Added `Swarm<T>.TrustedAppProtocolVersionSigners` property.
     [[#266], [#815]]
  -  Added `Peer.IsCompatibleWith()` method.  [[#266], [#815]]
+ -  Added `InvalidTxByBlockPolicyException` class.  [[#827]]
```
dahlia: How about adding a new term for this case which is distinct from “invalid tx” instead of calling both cases “invalid tx”?

earlbread: I agree. Let's think about the more proper term.

Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -13,6 +14,8 @@ namespace Libplanet.Blockchain.Policies
     public class BlockPolicy<T> : IBlockPolicy<T>
         where T : IAction, new()
     {
+        private readonly Predicate<Transaction<T>> _isTransactionValid;
```
moreal: `IsTransactionInvalid` method just calls `_isTransactionsValid` and pass its return value. so how about using property? 

```suggestion
        public Predicate<Transaction<T>> IsTransactionValid { get; }
```

dahlia: @moreal I believe the way you suggested does not satisfy the interface anyway, because in the IL level delegate properties and methods are merely distinct each other; the same looking syntax between delegate properties and methods in C# is just a coincidence.

moreal: @dahlia If I had understood correctly, it means that method `bool IsTransactionValid(Transaction<T>)` and delegate property `Predicate<Transaction<T>>` are distinct each other so it will not be achieved.

If then, it was what I really wanted to suggest to replace `IBlockPolicy><T>.IsTransactionValid(Transaction<T>)` with `Predicate<Transaction<T>>`. As code, it will be like below.

```csharp
public interface IBlockPolicy<T>
    where T : IAction, new()
{
    // bool IsTransactionValid(Transaction<T> transaction);
    Predicate<Transaction<T>> IsTransactionValid { get; }
}
```

dahlia: @moreal That was how this patch looked before it's amended, but then it became like the current way in a reason: .NET interfaces say nothing about constructors, so an interface anyway cannot enforce its implementations to take a delegate. On the other way, it's unclear what's benefit of having a predicate as a delegate property over a method, which is more common and natural.

moreal: Okay, I understood. Thanks for your good explanation.

CHANGES.md
```diff
@@ -51,6 +51,8 @@ To be released.
     that receives only one transaction.  [[#820]]
  -  Replaced `BlockChain<T>.UnstageTransactions()` with `.UnstageTransaction()`
     that receives only one transaction.  [[#820]]
+ -  Added `IBlockPolicy.DoesTransactionFollowPolicy` method which is a method to
```
dahlia: A trivial adjustment: 🙄 

```suggestion
 -  Added `IBlockPolicy.DoesTransactionFollowPolicy()` method which is a method to
```

CHANGES.md
earlbread: I changed to use the term <q>follow policy</q> and <q>policy violation</q>. Please take a look.

Libplanet/Tx/BlockPolicyViolationTxException.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+
+namespace Libplanet.Tx
+{
+    /// <summary>
+    /// The exception that is thrown when a <see cref="Transaction{T}"/> does not follow the
+    /// block policy.
+    /// </summary>
+    [Serializable]
+    public class BlockPolicyViolationTxException : InvalidTxException
```
dahlia: How about `TxViolatingBlockPolicyException`? 🤔

earlbread: Changed to `TxViolatingBlockPolicyException`.

Issue #458, opened 2019-08-26
Introduce NotNullAttribute and NullGuard.Fody


dahlia:
Although C# 8 will have [non-nullable reference types][1], it will requires [.NET Standard 2.1][4] (note that we currently depend on .NET Standard 2.0), which is unlikely to supported by Unity in the near future.

Instead, I suggest to introduce [JetBrain's `NotNullAttribute`][2] until Unity becomes to support .NET Standard 2.1.  There's a Roslyn Analyzer implementation named [NullGuard.Fody][3] which does `null`-checks according to `NotNullAttribute` at compile time.

[1]: https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references
[2]: https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#NotNullAttribute
[3]: https://github.com/Fody/NullGuard
[4]: https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/

dahlia:
If we decide to use C# 8.0 with .NET Standard 2.0 (see #780) this approach is probably not that useful.

dahlia:
> If we decide to use C# 8.0 with .NET Standard 2.0 (see #780) this approach is probably not that useful.

We decided to use C# 8.0 with .NET Standard 2.0, not for null checks, but async streams.  Anyway we would be able to use null checks as well, if we introduce C# 8.0 soon.

dahlia:
Update: although we still haven't enabled null checker project-wide, we've already enabled it in several files (the recently added files for the most part).  For gradual progression, I suggest the @libplanet team the methodology to adopt null checker with the minimum pain (and actually it is what we've recently done in few months indeed):

- When we add a new file, it should enable null checker in that file.  E.g.:

   <https://github.com/planetarium/libplanet/blob/8d99392e90be39ee56d23eadf4c4cfbe13eb4cfc/Libplanet/Currency.cs#L1>

- When we change a file of not-that-long lines, let's try to enable null checker in that file.  However, if it seems going deeper, retreat and wait a chance for next time.

- If we feel more than 2/3 in the code base (count only *Libplanet* project) let's try to enable null checker project-wide.

@libplanet Opinions?

dahlia:
The current status:

```console
$ find Libplanet/ -name '*.cs' | wc -l
193
$ git grep '#nullable enable' Libplanet/ | wc -l
37
```

dahlia:
The current status:

```
[...]

PR #831, opened 2020-03-23
IKeyStore & Web3KeyStore

dahlia:
This patch finally implements the complete [Web3 Secret Storage][1].

I also experimented with C# 8's [nullable reference types][2].  We could gradually adopt it by enabling [nullable context][3] per file (instead of the whole project, i.e., #458).

[1]: https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition
[2]: https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references
[3]: https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-contexts

### DIFF
--- a/Libplanet/KeyStore/IKeyStore.cs
+++ b/Libplanet/KeyStore/IKeyStore.cs
@@ -1,0 +1,53 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+
+namespace Libplanet.KeyStore
+{
+    /// <summary>
+    /// The interface to store <see cref="ProtectedPrivateKey"/>s.  An appropriate implementation
+    /// should be used according to a running platform.
+    /// </summary>
+    public interface IKeyStore
+    {
+        /// <summary>
+        /// Lists all keys IDs in the key store.
+        /// </summary>
+        /// <returns>All keys IDs in the key store.  The order is undefined and not deterministic.
+        /// </returns>
+        public IEnumerable<Guid> ListIds();
+
+        /// <summary>
+        /// Lists all keys in the key store.
+        /// </summary>
+        /// <returns>All keys in the key store.  The order is undefined and not deterministic.
+        /// </returns>
+        public IEnumerable<Tuple<Guid, ProtectedPrivateKey>> List();
+
+        /// <summary>
+        /// Looks for a key having the requested <paramref name="id"/> in the key store.
+        /// </summary>
+        /// <param name="id">The key ID to look for.</param>
+        /// <returns>The found key.</returns>
+        /// <exception cref="NoKeyException">Thrown when there are no key of the given
+        /// <paramref name="id"/>.</exception>
+        public ProtectedPrivateKey Get(Guid id);
+
+        /// <summary>
+        /// Adds a new <paramref name="key"/> into the key store.
+        /// </summary>
+        /// <param name="key">A key to add.</param>
+        /// <returns>The id of the added <paramref name="key"/>.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when <c>null</c> is passed to
+        /// <paramref name="key"/>.</exception>
+        public Guid Add(ProtectedPrivateKey key);
+
+        /// <summary>
+        /// Removes a key having the given <pramref name="id"/> from the key store.
+        /// </summary>
+        /// <param name="id">The key ID to remove.</param>
+        /// <exception cref="NoKeyException">Thrown when there is no key having
+        /// the given <paramref name="id"/>.</exception>
+        public void Remove(Guid id);
+    }
+}
--- a/Libplanet/KeyStore/Web3KeyStore.cs
+++ b/Libplanet/KeyStore/Web3KeyStore.cs
@@ -1,0 +1,187 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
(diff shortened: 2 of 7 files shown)


Libplanet/KeyStore/Web3KeyStore.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// <inheritdoc/>
+        public ProtectedPrivateKey Get(Guid id)
+        {
+            string name;
+            try
+            {
+                (_, name) = ListFiles().First(pair => pair.Item1.Equals(id));
+            }
+            catch (InvalidOperationException)
```
longfin: IMHO, `InvalidOperationException` can be too broad to guess `ListFiles().Any(pair => pair.Item1.Equals(id))` because it can be thrown in `ListFiles()`...

Libplanet/KeyStore/Web3KeyStore.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                if (System.IO.Path.GetFileName(keyPath) is string f)
+                {
+                    Match m = NamePattern.Match(f);
+                    if (m.Success)
+                    {
+                        Guid id;
+                        try
+                        {
+                            id = Guid.Parse(m.Groups[1].Value);
```
longfin: How about using [`.TryParse()`](https://docs.microsoft.com/ko-kr/dotnet/api/system.guid.tryparse?view=netstandard-2.1)?

Libplanet/KeyStore/Web3KeyStore.cs
```diff
@@ -0,0 +1,185 @@
...
+                }
+            }
+        }
+
+        private ProtectedPrivateKey Get(string name)
+        {
+            using (StreamReader reader = new StreamReader(System.IO.Path.Combine(Path, name)))
+            {
+                return ProtectedPrivateKey.FromJson(reader.ReadToEnd());
+            }
```
longfin: Could we use `using var` syntax here?

```suggestion
            using var reader = new StreamReader(System.IO.Path.Combine(Path, name)))
            return ProtectedPrivateKey.FromJson(reader.ReadToEnd());
```

Libplanet/KeyStore/Web3KeyStore.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+}
+#nullable restore
```
earlbread: I think it's okay to remove this.

Libplanet/KeyStore/IKeyStore.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <summary>
+        /// Removes a key having the given <pramref name="id"/> from the key store.
+        /// </summary>
+        /// <param name="id">The key ID to remove.</param>
+        /// <exception cref="NoKeyException">Thrown when there is no key having
+        /// the given <paramref name="id"/>.</exception>
+        public void Remove(Guid id);
+    }
+}
+#nullable restore
```
earlbread: I think we can remove this.

Libplanet/KeyStore/Web3KeyStore.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            Guid keyId = Guid.NewGuid();
+            string filename = string.Format(
+                CultureInfo.InvariantCulture,
+                NameFormat,
+                DateTimeOffset.UtcNow,
+                keyId
+            );
+            var keyPath = System.IO.Path.Combine(Path, filename);
```
moreal: How about using `Path` directly without namespace?

```suggestion
            var keyPath = Path.Combine(Path, filename);
```

Libplanet/KeyStore/Web3KeyStore.cs
dahlia: Due to [Sonar Analyzer's bug][1], that syntax cause a false alarm.  Anyway that bug was fixed in the [version 8.3], but again it introduces another bug…  😇  I hope the next version of Sonar Analyzer to eliminate all of these bugs.

[1]: https://community.sonarsource.com/t/dispose-with-using-triggers-s2930-c-8-0/13740
[2]: https://github.com/SonarSource/sonar-dotnet/releases/tag/8.3.0.14607

Libplanet/KeyStore/IKeyStore.cs
dahlia: Is `#nullable` context per file?  Could you give any docs to address it?

Libplanet/KeyStore/Web3KeyStore.cs
dahlia: Without `System.IO.` qualification the compiler recognizes this to the property `Web3KeyStore.Path` instead of the class `System.IO.Path` due to ambiguity.

Libplanet/KeyStore/IKeyStore.cs
earlbread: It is difficult to find, but it is written here.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/nullable-reference-types-specification#nullable-contexts

> A directive sets the context(s) it controls for subsequent lines of code, until another directive overrides it, or until the end of the source file.

dahlia:
@earlbread Removed unnecessary `#nullable restore` directives.

@planetarium/libplanet Sorry guys, I amended it again.  Please review this again.

PR #837, opened 2020-03-30
Improve performance of `Swarm<T>`

longfin:
This PR improves the performance of `Swarm<T>` by preventing unnecessary task creation.


longfin:
@dahlia @earlbread @moreal I've missed some points, so I also amended commits.

PR #838, opened 2020-03-31
Remove/hide unused public interfaces

dahlia:
See also the commit messages and changelogs.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -57,6 +57,31 @@ namespace Libplanet.Net
             }
         }
 
+        /// <summary>
+        /// Creates a <see cref="Swarm{T}"/>.  This constructor in only itself does not start
+        /// any communication with the network.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -57,14 +57,36 @@ static Swarm()
...
+        /// <summary>
+        /// Creates a <see cref="Swarm{T}"/>.  This constructor in only itself does not start
+        /// any communication with the network.
+        /// </summary>
+        /// <param name="blockChain">A blockchain to publicize on the network.</param>
+        /// <param name="privateKey">A private key to sign messages.  The public part of
+        /// this key become a part of its end address for being pointed by peers.</param>
+        /// <param name="appProtocolVersion">An app protocol to comply.</param>
+        /// <param name="workers">The number of background workers (i.e., threads).</param>
+        /// <param name="host">A hostname to listen.</param>
```
longfin: Actually, we aren't using `host` as an argument for traditional server socket but use for indicating its public endpoint that can be connected by other peers. 😕

PR #842, opened 2020-04-06
Increase state transmission chunk size

earlbread:


earlbread:
> Does this change pass the tests?

The tests passed in my local. Let's see the CI.

PR #859, opened 2020-04-29
Separate _poller in NetMQTransport

longfin:
This PR divides `_poller` by `_routerPoller` and `_broadcastPoller` to avoid unexpected blocking and improve throughput.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,10 +16,15 @@ To be released.
 
 ### Behavioral changes
 
+ -  Improved performance of `Swarm<T>` by multiplexing response and
+    broadcast.  [[#858], [#859]]
+
 ### Bug fixes
 
 ### CLI tools
 
+[#858]: https://github.com/planetarium/libplanet/issues/858
+[#859]: https://github.com/planetarium/libplanet/pull/859
 
 Version 0.9.0
 -------------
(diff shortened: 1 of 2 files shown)


Libplanet/Net/NetMQTransport.cs
```diff
@@ -992,6 +983,32 @@ private void ValidateSender(Peer peer)
...
+                {
+                    // Ignore NetMQ related exceptions during NetMQPoller.Run() to stabilize
+                    // tests.
+                    try
+                    {
+                        poller.Run();
+                    }
+                    catch (TerminatingException)
+                    {
+                        _logger.Error($"TerminatingException occurred in {nameof(poller)}");
```
dahlia, line 998: Why `"poller"` should be written as `nameof(poller)` instead?

CHANGES.md
```diff
@@ -16,10 +16,15 @@ To be released.
 
 ### Behavioral changes
 
+ -  Improved performance of `Swarm<T>` by multiplexing respond and
+    broadcast.  [[#858], [#859]]
```
dahlia, line 20:
```suggestion
 -  Improved performance of `Swarm<T>` by multiplexing response and
    broadcast.  [[#858], [#859]]
```

Libplanet/Net/NetMQTransport.cs
longfin: I guess there is no intention and it seems moved automatically. I'll fix.

PR #869, opened 2020-05-15
Add timeout create permission

### LINKED ISSUE #611: Add timeout for CreatePermission
Timeout is required in `Swarm<T>.CreatePermission()` to prevent blocking.

riemannulus:
It should be resolve #611.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,16 +23,12 @@ To be released.
 
 ### Bug fixes
 
-- `NetMQTransport.CreatePermission()` became to have timeout when locked for 2.5 seconds. [[#611], [#868]]
-
 ### CLI tools
 
-[#611]: https://github.com/planetarium/libplanet/issues/611
 [#756]: https://github.com/planetarium/libplanet/issues/756
 [#858]: https://github.com/planetarium/libplanet/issues/858
 [#859]: https://github.com/planetarium/libplanet/pull/859
 [#868]: https://github.com/planetarium/libplanet/pull/868
-[#868]: https://github.com/planetarium/libplanet/pull/869
 
 
 Version 0.9.1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,12 +23,16 @@ To be released.
 
 ### Bug fixes
 
+- `NetMQTransport.CreatePermission()` became to have timeout when locked for 2.5 seconds. [[#611], [#868]]
+
 ### CLI tools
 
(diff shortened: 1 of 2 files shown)


Libplanet/Net/NetMQTransport.cs
```diff
@@ -857,7 +857,11 @@ private string ToNetMQAddress(BoundPeer peer)
 
         private async Task CreatePermission(BoundPeer peer)
         {
+            var cts = new CancellationTokenSource();
             IPAddress[] ips;
+
+            // Cancellation After 2.5 sec
+            cts.CancelAfter(2500);
```
earlbread, line 864: Can we have a test case for this?

I don't think it will work as intended because the `Token` from `cts` is not used anywhere.

Libplanet/Net/NetMQTransport.cs
```diff
@@ -857,7 +857,11 @@ private string ToNetMQAddress(BoundPeer peer)
 
         private async Task CreatePermission(BoundPeer peer)
         {
+            var cts = new CancellationTokenSource();
```
moreal, line 860: How about using its `CancellationTokenSource.Token`? We can pass cancellationToken to `TurnClient.CreatePermissionAsync()` and `TurnClient.GetMappedAddressAsync()` in these lines
- https://github.com/planetarium/libplanet/pull/869/files#diff-9045fd925fccb8bca88db49a0ab95d0cR896
- https://github.com/planetarium/libplanet/pull/869/files#diff-9045fd925fccb8bca88db49a0ab95d0cR890

moreal: It seems helpful link. https://docs.microsoft.com/en-us/dotnet/standard/threading/cancellation-in-managed-threads .

Libplanet/Net/NetMQTransport.cs
riemannulus: Still we need a test case for this?

Now `GetMappedAddressAsync` and `CreatePermissionAsync` using `Token` from `cts`.

CHANGES.md
```diff
@@ -23,12 +23,16 @@ To be released.
 
 ### Bug fixes
 
+- `NetMQTransport.CreatePermission()` became to have timeout when locked for 2.5 seconds. [[#611], [#869]]
```
longfin, line 26: I guess we don't need to mention about `NetMQTransport.CreatePermission()` because it hadn't opened to Libplanet users. 😕

PR #870, opened 2020-05-19
Dispose iterator after use

earlbread:


earlbread:
I changed the base to 0.9-maintenance and added a changelog.

CHANGES.md
```diff
@@ -1,6 +1,16 @@
 Libplanet changelog
 ===================
 
+Version 0.9.2
+-------------
+
+To be released.
+
+ -  Fixed a memory leak bug in Libplanet.RocksDBStore.  [[#870]]
```
dahlia, line 9: @planetarium/libplanet How about adding a prefix like this for changes in other NuGet packages than *Libplanet*?

```suggestion
 -  (Libplanet.RocksDBStore) Fixed a memory leak bug in `RocksDBStore`.  [[#870]]
```

PR #878, opened 2020-05-22
Add genesis hash to transaction

### LINKED ISSUE #796: Add the genesis hash to Transaction<T>
In order to prevent inapplicable transactions (made for a different network) polluting the network, I suggest to add `Transaction<T>.GenesisHash` property.

limebell:
This patch adds nullable genesis hash field to transaction, which resolves #796.


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -518,6 +522,15 @@ public Transaction<T> GetTransaction(TxId txId)
         /// <paramref name="transaction"/> is invalid.</exception>
         public void StageTransaction(Transaction<T> transaction)
         {
+            if (!transaction.GenesisHash.Equals(Genesis.Hash))
+            {
+                throw new InvalidTxGenesisHashException(
+                    transaction.Id,
+                    Genesis.Hash,
+                    transaction.GenesisHash,
+                    "Genesis of the transaction is invalid.");
```
dahlia, line 531:
```suggestion
                    "GenesisHash of the transaction is not compatible with the BlockChain<T>.Genesis.Hash.");
```

Libplanet/Tx/RawTransaction.cs
```diff
@@ -12,6 +12,8 @@ namespace Libplanet.Tx
 
         public static readonly byte[] SignerKey = { 0x73 }; // 's'
 
+        public static readonly byte[] GenesisHashKey = { 0x67 }; // 'g'
```
dahlia, line 15: This change should be added to the <q>Backward-incompatible storage format changes</q> section.

PR #884, opened 2020-05-28
Prevent Swarm<T>.GetDemandBlockHashes() from being hanged when a peer's chain is during reorg

dahlia:
This fixes the bug #880.

- Make `Swarm<T>.GetDemandBlockHashes()` method `internal` (was `private`) and add unit tests on it.
- Prevent `Swarm<T>.GetDemandBlockHashes()` method from being hanged when a peer's chain is during reorg.

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.Preload.cs
+++ b/Libplanet.Tests/Net/SwarmTest.Preload.cs
@@ -1207,5 +1207,57 @@ namespace Libplanet.Tests.Net
                 .Select(b => (b.Index, b.Hash));
             Assert.Equal(expectedBlocks, demands);
         }
+
+        [Fact(Timeout = Timeout)]
+        public async Task GetDemandBlockHashesDuringReorg()
+        {
+            Swarm<DumbAction> minerSwarm = _swarms[0];
+            Swarm<DumbAction> receiverSwarm = _swarms[1];
+            Log.Logger.Information("Miner:    {0}", minerSwarm.Address);
+            Log.Logger.Information("Receiver: {0}", receiverSwarm.Address);
+
+            BlockChain<DumbAction> minerChain = _blockchains[0];
+            BlockChain<DumbAction> receiverChain = _blockchains[1];
+
+            Block<DumbAction>[] blocks =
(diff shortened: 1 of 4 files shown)


Libplanet.Tests/Net/SwarmTest.Preload.cs
```diff
@@ -1167,5 +1167,97 @@ public async Task PreloadAsyncCancellation(int cancelAfter)
                 );
             }
         }
+
+        [Fact(Timeout = Timeout)]
+        public async void GetDemandBlockHashes()
```
longfin, line 1172: `async void` isn't traceable since it doesn't return `Task`.

```suggestion
        public async Task GetDemandBlockHashes()
```

Libplanet.Tests/Net/SwarmTest.Preload.cs
```diff
@@ -1167,5 +1167,97 @@ public async Task PreloadAsyncCancellation(int cancelAfter)
...
+            ).ToArrayAsync();
+
+            IEnumerable<(long, HashDigest<SHA256>)> expectedBlocks = minerChain.IterateBlocks()
+                .Where(b => b.Index >= receiverChain.Count)
+                .Select(b => (b.Index, b.Hash));
+            Assert.Equal(expectedBlocks, demands);
+        }
+
+        [Fact(Timeout = Timeout)]
+        public async void GetDemandBlockHashesDuringReorg()
```
longfin, line 1212:
```suggestion
        public async Task GetDemandBlockHashesDuringReorg()
```

Libplanet.Tests/Net/SwarmTest.Preload.cs
dahlia: Fixed in 25e887051f8b20fb9f72c80bf00dc11a319d95aa (amended).

Libplanet.Tests/Net/SwarmTest.Preload.cs
dahlia: Fixed in 5efe719923cce1445c58c7804bf04b6b292c620b (amended).

Issue #871, opened 2020-05-19
Block propagation is too slow


longfin:
Blocks sometimes propagate after tens of seconds. if the target time for block creation is short(10~20s), this is a sufficient bottleneck.

longfin:
I guess another bottleneck seemed left. I'll look into this issue again.

longfin:
I found some clues from the old MacBook. it seemed to skip some block hashes and waiting for next block hashes.

longfin:
We've deployed #890 and reported symptom seems to be gone.

Issue #404, opened 2019-08-07
System.IO.IOException: Unable to read data from the transport connection


longfin:
```
System.IO.IOException: Unable to read data from the transport connection: 현재 연결은 원격 호스트에 의해 강제로 끊겼습니다.

. ---> System.Net.Sockets.SocketException: 현재 연결은 원격 호스트에 의해 강제로 끊겼습니다.



  at System.Net.Sockets.Socket.EndReceive (System.IAsyncResult asyncResult) [0x00012] in <c9226b2f33b7421e85bb391f94075e49>:0 

  at System.Net.Sockets.NetworkStream.EndRead (System.IAsyncResult asyncResult) [0x00057] in <c9226b2f33b7421e85bb391f94075e49>:0 

   --- End of inner exception stack trace ---

  at System.Net.Sockets.NetworkStream.EndRead (System.IAsyncResult asyncResult) [0x0009b] in <c9226b2f33b7421e85bb391f94075e49>:0 

  at System.IO.Stream+<>c.<BeginEndReadAsync>b__43_1 (System.IO.Stream stream, System.IAsyncResult asyncResult) [0x00000] in <1f0c1ef1ad524c38bbc5536809c46b48>:0 

  at System.Threading.Tasks.TaskFactory`1+FromAsyncTrimPromise`1[TResult,TInstance].Complete (TInstance thisRef, System.Func`3[T1,T2,TResult] endMethod, System.IAsyncResult asyncResult, System.Boolean requiresSynchronization) [0x00000] in <1f0c1ef1ad524c38bbc5536809c46b48>:0 

--- End of stack trace from previous location where exception was thrown ---

  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in <1f0c1ef1ad524c38bbc5536809c46b48>:0 

  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x0003e] in <1f0c1ef1ad524c38bbc5536809c46b48>:0 

  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00028] in <1f0c1ef1ad524c38bbc5536809c46b48>:0 

[...]

PR #890, opened 2020-05-29
Fix message dropping

longfin:
This PR addresses #871 and #404. 

The main reason is, a relayed connection had closed before reading by `DealerSocket.Dispose()`.  To resolve it, I've tried found the way to guess the exact timing of the connection made, but I couldn't. so I added a delay as timeout as a workaround.

p.s. since the environment that occurred was 9C CBT, I was primarily testing in the environment of the 9c-beta branch. So this PR is based on 9c-beta for convenience. but please let me know if we need to get into 9c-maintenance.


CHANGES.md
```diff
@@ -33,15 +33,21 @@ To be released.
 
 ### Bug fixes
 
+ -  Fixed a bug that `Swarm<T>` had failed to receive a request from TURN relay
+    connections  [[#404], [#871], [#890]]
```
moreal, line 37:
```suggestion
    connections.  [[#404], [#871], [#890]]
```

riemannulus:
Does making the variable for Task delay time work for this problem?

longfin:
I changed the approach due to continuous test breaking 😓 .

- Instead of an unnatural timeout to hold `DealerSocket` in `ProcessRuntimes`, I made `NetMQTransport._dealers` as the holder for `DealerSocket`s to broadcasting. 
- `NetMQTransport._dealers` will be reused for the same peer(by `Peer.Address`). it's okay for broadcasting but isn't for request/response. so I also separated broadcasting codes from `NetMQTransport.ProcessRequest()` to distinguish it.
  - it seems to have a limit on throughput but it doesn't matter much as `DealerSocket.Send()` works almost asynchronously and is handled by NetMQ's IO thread.

Libplanet/Net/NetMQTransport.cs
```diff
@@ -934,6 +932,17 @@ private void ValidateSender(Peer peer)
                     await Task.Delay(period, cancellationToken);
                     await Protocol.RefreshTableAsync(maxAge, cancellationToken);
                     await Protocol.CheckReplacementCacheAsync(cancellationToken);
+
+                    ImmutableHashSet<Address> peerAddreses =
```
earlbread, line 936:
```suggestion
                    ImmutableHashSet<Address> peerAddresses =
```

Libplanet/Net/NetMQTransport.cs
```diff
@@ -934,6 +932,17 @@ private void ValidateSender(Peer peer)
...
+                        Peers.Select(p => p.Address).ToImmutableHashSet();
+                    foreach (Address address in _dealers.Keys)
+                    {
+                        if (!peerAddreses.Contains(address) &&
```
earlbread, line 940:
```suggestion
                        if (!peerAddresses.Contains(address) &&
```

Libplanet/Net/NetMQTransport.cs
```diff
@@ -569,56 +577,58 @@ public async Task CheckAllPeersAsync(CancellationToken cancellationToken, TimeSp
...
-                    ValidateSender(message.Remote);
+                    _logger.Debug("Ignore message from peer with different version.");
                 }
-
-                try
+                catch (InvalidMessageException ex)
                 {
-                    Protocol.ReceiveMessage(message);
-                    ProcessMessageHandler?.Invoke(this, message);
+                    _logger.Error(ex, $"Could not parse NetMQMessage properly; ignore: {ex}");
```
dahlia, line 622:
```suggestion
                    _logger.Error(ex, $"Could not parse NetMQMessage properly; ignore: {{0}}", ex);
```

Libplanet/Net/NetMQTransport.cs
```diff
@@ -569,56 +577,58 @@ public async Task CheckAllPeersAsync(CancellationToken cancellationToken, TimeSp
...
-                catch (Exception exc)
+                catch (Exception ex)
                 {
                     _logger.Error(
-                        exc,
-                        "Something went wrong during message parsing: {0}",
-                        exc);
-                    throw;
+                        ex,
+                        $"An unexpected exception occurred during {nameof(ReceiveMessage)}(): {ex}"
```
dahlia, line 628:
```suggestion
                        $"An unexpected exception occurred during {nameof(ReceiveMessage)}(): {{0}}",
                        ex
```

PR #898, opened 2020-06-03
Ignore invalid header immediately

longfin:
This PR fixes `Swarm<T>` to ignore invalid block header immediately.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,6 +41,7 @@ To be released.
  -  Added `Transaction<T>.GenesisHash` property.  [[#796], [#878]]
  -  `Swarm<T>` became to ignore received transaction with different
     genesis hash.  [[#796], [#878]]
+ -  `Swarm<T>` became to ignore invalid block header immediately.  [[#898]]
 
 ### Bug fixes
 
@@ -62,6 +63,7 @@ To be released.
 [#879]: https://github.com/planetarium/libplanet/pull/879
 [#883]: https://github.com/planetarium/libplanet/pull/883
 [#883]: https://github.com/planetarium/libplanet/pull/890
+[#898]: https://github.com/planetarium/libplanet/pull/898
 
 
 Version 0.9.4
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,7 +62,7 @@ To be released.
 [#878]: https://github.com/planetarium/libplanet/pull/878
 [#879]: https://github.com/planetarium/libplanet/pull/879
 [#883]: https://github.com/planetarium/libplanet/pull/883
-[#883]: https://github.com/planetarium/libplanet/pull/890
+[#890]: https://github.com/planetarium/libplanet/pull/890
 [#898]: https://github.com/planetarium/libplanet/pull/898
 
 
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -41,6 +41,7 @@ To be released.
  -  Added `Transaction<T>.GenesisHash` property.  [[#796], [#878]]
  -  `Swarm<T>` became to ignore received transaction with different
     genesis hash.  [[#796], [#878]]
+ -  `Swarm<T>` became to ignore invalid block header immediately.  [[#898]]
```
dahlia, line 44:
```suggestion
 -  `Swarm<T>` became to ignore invalid `BlockHeader`s immediately.  [[#898]]
```

Libplanet.Tests/Blocks/BlockHeaderTest.cs
```diff
@@ -0,0 +1,147 @@
+using System;
+using System.Collections.Immutable;
+using System.Globalization;
+using Libplanet.Blocks;
+using Xunit;
+
+namespace Libplanet.Tests.Blocks
+{
+    public class BlockHeaderTest : IClassFixture<BlockFixture>
```
dahlia, line 9: Haven't we had such `Validate()` tests for `Block` before?

longfin: Yes, and we still have tests for `Block<T>.Validate()`.

I've added these tests to have confidence for separated codes while working because it's more direct than `Block<T>.Validate()`'s tests.

Libplanet/Blocks/Block.cs
```diff
@@ -431,7 +426,9 @@ internal BlockHeader GetBlockHeader()
             // FIXME: When hash is not assigned, should throw an exception.
             return new BlockHeader(
                 index: Index,
-                timestamp: Timestamp.ToString(TimestampFormat, CultureInfo.InvariantCulture),
+#pragma warning disable MEN002 // line is too long
+                timestamp: Timestamp.ToString(BlockHeader.TimestampFormat, CultureInfo.InvariantCulture),
```
earlbread, line 430: How about using local variables instead of disabling the warning?

Libplanet.Tests/Blocks/BlockHeaderTest.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        private BlockFixture _fx;
+
+        public BlockHeaderTest(BlockFixture fixture) => _fx = fixture;
+
+        [Fact]
+        public void ValidateTimestamp()
+        {
+            DateTimeOffset now = DateTimeOffset.UtcNow;
+            string future = (now + TimeSpan.FromSeconds(901))
```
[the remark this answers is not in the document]
longfin: That can make sense. but I prefer specifying detail directly as a spec in test case instead of reusing constant from target project. (it isn't strong opinion 😅 )

Libplanet/Blocks/Block.cs
longfin: 🆗 I'll fix it.

Issue #861, opened 2020-05-04
States API specialized for assets


dahlia:
Besides the existing free-state API (i.e., states free to set/update to arbitrary values), we need a different kind of states specialized for assets.  This kind of states (“asset states” API) would be free to get (in the same way to the “free-state” API), but should provide only restricted APIs to update (like `Transfer()`).

In the future, the internal representation behind this kind of “asset states” API could be implemented using UTXO.

dahlia:
I've roughly ideated the assets API as follows.

First of all, requirements I've tried to address are:

 -  Assets must not be copied, but can be only moved (transferred).
 -  There could be more than one kind of assets (e.g., currencies).
 -  Some kinds of assets are fungible, and some are not.  However, fungible assets and non-fungible assets are never mixed.

`IAccountStateDelta`, which has had methods like `GetState(Address)` and `SetState(Address, IValue)`, becomes to have the following new methods as well:

 - `IAccountStateDelta Mint(Address recipient, Currency currency, BigInteger amount)`
 - `IAccountStateDelta Transfer(Address sender, Address recipient, Currency currency, BigInteger amount)`
 - `IAccountStateDelta Burn(Address owner, Currency currency, BigInteger amount)`
 - `(IAccountStateDelta NextStates, Guid NonFungibleAsset) Construct(Address recipient)`
 - `IAccountStateDelta Transfer(Address sender, Address recipient, Guid nonFungibleAsset)`
 - `IAccountStateDelta Destruct(Address recipient, Guid nonFungibleAsset)`

The `Currency` types looks like:

~~~~ csharp
public readonly struct Currency
{
  public readonly string Ticker;
  public readonly Address Minter;

  // A deterministic digest derived from other fields.
  public readonly HashDigest<SHA1> Hash => ...;
}
~~~~

Here's a sample code to deal with fungible assets in an action:

~~~~ csharp
static readonly Currency NCG = new Currency
{
  Ticker = "NCG",
  Minter = new Address(...),
};

IAccountStateDelta Execute(IActionContext context)
{
  IAccountStateDelta states = context.PreviousStates;
  
  // Mint new 1,000 NCG and give "A" it.
  // Note that this will fail unless the context.Signer equals to NCG.Minter.
  states = states.Mint(A, NCG, 1000);
  
  // Transfer 500 NCG from "A" to "B".
  // Note that this will fail due to insufficient balance when 500 is replaced by 1,001.
  states = states.Transfer(A, B, NCG, 500);

[...]

PR #900, opened 2020-06-07
Fungible asset states

dahlia:
This implements fungible asset states APIs, the first half of the proposal #861.  As non-fungible asset states APIs will be orthogonal to this, I am going to send a separated pull request for that later.

The usage can be found in the proposal issue #861 (please read comments I left as well).

### DIFF
--- a/Libplanet/Action/AccountBalanceGetter.cs
+++ b/Libplanet/Action/AccountBalanceGetter.cs
@@ -1,0 +1,19 @@
+#nullable enable
+using System.Numerics;
+
+namespace Libplanet.Action
+{
+    /// <summary>
+    /// An delegate to provide read-only view of account balances.
+    /// <para>Gets <paramref name="address"/>' balance of the given <paramref name="currency"/>.
+    /// </para>
+    /// </summary>
+    /// <param name="address">An address of the account to query its balance of
+    /// the <paramref name="currency"/>.</param>
+    /// <param name="currency">The currency type to query <paramref name="address"/>' balance.
+    /// </param>
+    /// <returns>
+    /// The <paramref name="address"/>'s balance of the <paramref name="currency"/>.
+    /// </returns>
+    public delegate BigInteger AccountBalanceGetter(Address address, Currency currency);
+}
--- a/Libplanet/Action/AccountStateDeltaImpl.cs
+++ b/Libplanet/Action/AccountStateDeltaImpl.cs
@@ -1,5 +1,10 @@
+#nullable enable
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Numerics;
 using Bencodex.Types;
 
 namespace Libplanet.Action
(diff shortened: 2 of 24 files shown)


Libplanet/Action/AccountBalanceGetter.cs
```diff
@@ -0,0 +1,19 @@
+#nullable enable
```
longfin, line 1: Is this statement needed?

Libplanet/Action/ActionEvaluation.cs
```diff
@@ -138,6 +138,8 @@ int randomSeed
                     string msg;
                     if (!rehearsal)
                     {
+                        // FIXME: The below "exc" object seems never used; Hong Minhee believes
```
longfin, line 141: `exc` seems used [for `new ActionEvaluation()`](https://github.com/planetarium/libplanet/pull/900/files#diff-37d0c58de3562a6f55cfb92ca89c8e9dR169-R174)

Libplanet/Currency.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// <summary>
+        /// Returns <c>true</c> if and only if the given <paramref name="address"/> is allowed
+        /// to mint or burn assets of this currency.
+        /// </summary>
+        /// <param name="address">The account address to test.</param>
+        /// <returns><c>true</c> if and only if the given <paramref name="address"/> is allowed to
+        /// mint or burn assets of this currency.</returns>
+        [Pure]
+        public bool AllowsToMint(Address address) =>
+            Minters is IImmutableSet<Address> s ? s.Contains(address) : true;
```
earlbread, line 112: How about this?
```suggestion
            Minters is null || Minters.Contains(address);
```

Libplanet/Action/AccountStateDeltaImpl.cs
```diff
@@ -1,53 +1,198 @@
...
+            }
+
+            BigInteger senderBalance = GetBalance(sender, currency),
+                       recipientBalance = GetBalance(recipient, currency);
+
+            if (!allowNegativeBalance && senderBalance < amount)
+            {
+                var msg = $"The account {sender}'s balance of {currency} is insufficient to " +
+                          $"transfer: {senderBalance} {currency} < {amount} {currency}.";
+                throw new InsufficientBalanceException(sender, currency, 0, msg);
```
earlbread, line 134: Shouldn't this be the sender's balance?
```suggestion
                throw new InsufficientBalanceException(sender, currency, senderBalance, msg);
```

Libplanet/Action/AccountStateDeltaImpl.cs
```diff
@@ -1,53 +1,198 @@
...
+                throw new CurrencyPermissionException(_signer, currency, msg);
+            }
+
+            BigInteger balance = GetBalance(owner, currency);
+
+            if (balance < amount)
+            {
+                msg = $"The account {owner}'s balance of {currency} is insufficient to burn: " +
+                      $"{balance} {currency} < {amount} {currency}.";
+                throw new InsufficientBalanceException(owner, currency, 0, msg);
```
earlbread, line 170: Shouldn't this be the `balance`?
```suggestion
                throw new InsufficientBalanceException(owner, currency, balance, msg);
```

Libplanet/Action/AccountBalanceGetter.cs
dahlia: We're going to eventually configure `<Nullable>enable</Nullable>` in project-level, so it's a transition step.

Libplanet/Action/ActionEvaluation.cs
dahlia: No, this one is unused, because the same variable is always overwritten later:

https://github.com/planetarium/libplanet/blob/1ff9ffeeea623cf7375673521adfacd8690af74b/Libplanet/Action/ActionEvaluation.cs#L160-L162

longfin: 🆗

Libplanet/Action/AccountBalanceGetter.cs
longfin: I see. then, is it okay we skip `#nullable restore` at the end of the file?

dahlia: Yeah, see also @earlbread's early review on the other PR: <https://github.com/planetarium/libplanet/pull/831#discussion_r396854992>.

Libplanet/Blocks/Block.cs
```diff
@@ -296,15 +300,22 @@ public byte[] Serialize()
...
+            AccountStateGetter accountStateGetter = null,
+            AccountBalanceGetter accountBalanceGetter = null
+        )
         {
-            IAccountStateDelta delta =
-                new AccountStateDeltaImpl(
-                    accountStateGetter ?? (a => null)
-                );
+            accountStateGetter ??= a => null;
+            accountBalanceGetter ??= (a, c) => 0;
```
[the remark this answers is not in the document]
dahlia: I am going to write tests for this specific case.  Thanks!

PR #901, opened 2020-06-08
Ignore TaskCanceledException in CreatePermission

earlbread:
This fixes a bug where `Swarm<T>` was terminated when a timeout occurred in `CreatePermission`.

### DIFF
--- a/Libplanet/Net/NetMQTransport.cs
+++ b/Libplanet/Net/NetMQTransport.cs
@@ -891,21 +891,35 @@ namespace Libplanet.Net
                 ips = new[] { peer.PublicIPAddress };
             }
 
-            foreach (IPAddress ip in ips)
+            try
             {
-                var ep = new IPEndPoint(ip, peer.EndPoint.Port);
-                if (IPAddress.IsLoopback(ip))
+                foreach (IPAddress ip in ips)
                 {
-                    // This translation is only used in test case because a
-                    // seed node exposes loopback address as public address to
-                    // other node in test case
-                    ep = await _turnClient.GetMappedAddressAsync(cts.Token);
-                }
+                    var ep = new IPEndPoint(ip, peer.EndPoint.Port);
+                    if (IPAddress.IsLoopback(ip))
+                    {
+                        // This translation is only used in test case because a
+                        // seed node exposes loopback address as public address to
+                        // other node in test case
+                        ep = await _turnClient.GetMappedAddressAsync(cts.Token);
+                    }
 
-                // FIXME Can we really ignore IPv6 case?
-                if (ip.AddressFamily.Equals(AddressFamily.InterNetwork))
+                    // FIXME Can we really ignore IPv6 case?
+                    if (ip.AddressFamily.Equals(AddressFamily.InterNetwork))
+                    {
+                        await _turnClient.CreatePermissionAsync(ep, cts.Token);
+                    }
+                }
+            }
+            catch (TaskCanceledException tce)
+            {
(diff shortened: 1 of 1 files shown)


Libplanet/Net/NetMQTransport.cs
```diff
@@ -893,19 +893,26 @@ private async Task CreatePermission(BoundPeer peer)
...
+                    if (ip.AddressFamily.Equals(AddressFamily.InterNetwork))
+                    {
+                        await _turnClient.CreatePermissionAsync(ep, cts.Token);
+                    }
+                }
+                catch (TaskCanceledException)
                 {
-                    await _turnClient.CreatePermissionAsync(ep, cts.Token);
+                    _logger.Debug($"Task is cancelled during {nameof(CreatePermission)}");
                 }
```
dahlia, line 916: How about separately wrapping those two `await` calls with `try`–`catch` and logging what specific task (`GetMappedAddressAsync()` or `CreatePermissionAsync()`) was cancelled?

Libplanet/Net/NetMQTransport.cs
```diff
@@ -893,19 +893,26 @@ private async Task CreatePermission(BoundPeer peer)
...
+                    // FIXME Can we really ignore IPv6 case?
+                    if (ip.AddressFamily.Equals(AddressFamily.InterNetwork))
+                    {
+                        await _turnClient.CreatePermissionAsync(ep, cts.Token);
+                    }
+                }
+                catch (TaskCanceledException)
                 {
-                    await _turnClient.CreatePermissionAsync(ep, cts.Token);
+                    _logger.Debug($"Task is cancelled during {nameof(CreatePermission)}");
```
longfin, line 915: I guess it would be better we add the filter judges the exception from `cts` like as below...

```csharp
if (cts.IsCancellationRequested)
{
    _logger.Debug($"Task is cancelled during {nameof(CreatePermission)}");
}

throw;
```

Libplanet/Net/NetMQTransport.cs
earlbread: How about adding an exception to log to prevent continuing execution when `GetMappedAddressAsync` is canceled?

dahlia: That would be better! (1. logging what specific task was cancelled 2. logging why [= what exception does make] that task was cancelled).

earlbread:
```
[16:29:01 DBG] Timeout occurred during CreatePermission: System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Libplanet.Stun.TurnClient.SendMessageAsync(NetworkStream stream, StunMessage message, CancellationToken cancellationToken)
   at Libplanet.Stun.TurnClient.CreatePermissionAsync(IPEndPoint peerAddress, CancellationToken cancellationToken)
   at Libplanet.Net.NetMQTransport.CreatePermission(BoundPeer peer)
```
How about like this?

dahlia: - Can the message `Timeout occurred during CreatePermission` be `Timeout occurred during GetMappedAddressAsync` as well?
- How about this: `CreatePermissionAsync() has timed out`?

earlbread: `CreatePermission` is the name of the method name that calls `GetMappedAddressAsync` and `CreatePermissionAsync`. We can check `GetMappedAddressAsync` and `CreatePermissionAsync` in the stack trace.

PR #912, opened 2020-06-26
Change StateReference key to string

earlbread:
This changes `StateReference` key to string to fix a bug in which pre-computed state delivery had failed when a state key is not an `Address` when preloading.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,12 @@ To be released.
 
 ### Backward-incompatible network protocol changes
 
+ -  `RecentStates.StateReferences` became to
+    `IImmutableDictionary<string, IImmutableList<HashDigest<SHA256>>>` from
+    `IImmutableDictionary<Address, IImmutableList<HashDigest<SHA256>>>`.  [[#912]]
+ -  The existing `RecentStates` message type (with the type number `0x0f`) was
+    replaced by a new `RecentStates` message type (with type number `0x13`).  [[#912]]
+
 ### Backward-incompatible storage format changes
 
  -  Added `RawTransaction<T>.GenesisHash` property.  [[#796], [#878]]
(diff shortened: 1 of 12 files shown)


Libplanet/Net/Messages/Message.cs
```diff
@@ -72,7 +72,7 @@ internal enum MessageType : byte
             /// A reply to <see cref="GetRecentStates"/>.
             /// Contains the calculated recent states and state references.
             /// </summary>
-            RecentStates = 0x0f,
+            RecentStates = 0x13,
```
longfin, line 75: I guess `0x0g` is safer than `0x13` because `0x13` can be used for another message type in the older version.

CHANGES.md
```diff
@@ -30,6 +30,12 @@ To be released.
 
 ### Backward-incompatible network protocol changes
 
+ -  `RecentStates.StateReferences` became to
```
dahlia, line 33:
```suggestion
 -  The message field `RecentStates.StateReferences` became to
```

CHANGES.md
```diff
@@ -30,6 +30,12 @@ To be released.
...
+    `IImmutableDictionary<string, IImmutableList<HashDigest<SHA256>>>` from
+    `IImmutableDictionary<Address, IImmutableList<HashDigest<SHA256>>>`.  [[#912]]
+ -  The existing `RecentStates` message type (with the type number `0x0f`) was
+    replaced by a new `RecentStates` message type (with type number `0x13`).  [[#912]]
```
dahlia, line 37:
```suggestion
    replaced by a new `RecentStates` message type
    (with the type number `0x13`).  [[#912]]
```

Libplanet/Net/Messages/Message.cs
dahlia: @longfin Do you mean `0x14`?

earlbread: I tried it but realized that the hexadecimal number is up to `f`. :cry:

longfin: Ah, I've confused. nevermind.

PR #891, opened 2020-06-02
Make RocksDBStore.DeleteChainId() idempotent

dahlia:
According to `IStore.DeleteChainId()` method's XML comment:

> If there is no such `chainId` it does nothing.

Howver, `RocksDBStore.DeleteChainId()` hadn't been like that.  This fixed the bug.

PR #918, opened 2020-07-08
Fix for deleting chain ID

earlbread:
This fixes an issue that attempted to delete the chain multiple times. Also, in order to facilitate debugging, the deletion log of  `_chainDb` and the `_stateRefDb`  are separated.

### DIFF
--- a/Libplanet.RocksDBStore/RocksDBStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBStore.cs
@@ -141,18 +141,28 @@ namespace Libplanet.RocksDBStore
         /// <inheritdoc/>
         public override void DeleteChainId(Guid chainId)
         {
+            _logger.Debug($"Deleting chainID: {chainId}.");
+            _lastStateRefCaches.Remove(chainId);
+
+            var cfName = chainId.ToString();
             try
             {
-                _lastStateRefCaches.Remove(chainId);
-
-                var cfName = chainId.ToString();
                 _chainDb.DropColumnFamily(cfName);
+            }
+            catch (KeyNotFoundException)
+            {
+                // Do nothing according to the specification: DeleteChainId() should be idempotent.
+                _logger.Debug($"No such chain ID in _chainDb: {cfName}.", cfName);
+            }
+
+            try
+            {
                 _stateRefDb.DropColumnFamily(cfName);
             }
             catch (KeyNotFoundException)
             {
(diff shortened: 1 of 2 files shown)


longfin:
Should we add a mention about this fix to `CHANGES.md` ?

earlbread:
> Should we add a mention about this fix to `CHANGES.md` ?

I didn't add it to the changelog because the actual behavior except the log output is not changed. Should we add it?

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -141,18 +141,28 @@ public override IEnumerable<Guid> ListChainIds()
...
             try
             {
-                _lastStateRefCaches.Remove(chainId);
-
-                var cfName = chainId.ToString();
                 _chainDb.DropColumnFamily(cfName);
+            }
+            catch (KeyNotFoundException)
+            {
+                // Do nothing according to the specification: DeleteChainId() should be idempotent.
```
dahlia, line 154: It would be better if we can test if a `IStore` implementation actually satisfies this rule.

longfin:
> > Should we add a mention about this fix to `CHANGES.md` ?
> 
> I didn't add it to the changelog because the actual behavior except the log output is not changed. Should we add it?

[this](https://github.com/planetarium/libplanet/pull/918/files#diff-9501da6c2e844a81d5e3ad3d3d5456ccL495) change doesn't affect behavior? (just curious)

earlbread:
> > > Should we add a mention about this fix to `CHANGES.md` ?
> > 
> > 
> > I didn't add it to the changelog because the actual behavior except the log output is not changed. Should we add it?
> 
> [this](https://github.com/planetarium/libplanet/pull/918/files#diff-9501da6c2e844a81d5e3ad3d3d5456ccL495) change doesn't affect behavior? (just curious)

In `BlockChain<T>.Swap()`, `BlockChain.Id` is removed and `Store.DeleteChainId()` ignores the second attempt, so the behavior does not change.

Libplanet.RocksDBStore/RocksDBStore.cs
earlbread: The test seems to have already been added [here](https://github.com/planetarium/libplanet/pull/891/files#diff-5a20150ec899716c4ff9efea02fdf6a5R76).

PR #919, opened 2020-07-10
Promoting the most difficult chain as canonical chain

### LINKED ISSUE #459: Promote the most difficult chain as a canonical chain instead of the longest chain
A longest chain can be easily made by mining many blocks with low difficulties.  In order to be more fair, a canonical chain should be chosen by the total difficulty of a block chain instead of its height.

limebell:
Includes:
- Select the peer with the most difficult chain to preload.
- Demand only more difficult chain than current.

Closes #459.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,10 @@ To be released.
     constructor.  [[#666], [#917]]
  -  Added `BigInteger`-typed `previousTotalDifficulty` parameter to
     `Block<T>.Mine()` static method.  [[#666], [#917]]
+ -  `Pong()` constructor has changed:  [[459], [#919]]
+     -  Added `Pong()` constructor with no any parameter.
+     -  Removed `Pong(long?)` constructor.
+     -  Added `Pong(long, BigInteger)` constructor.
 
 ### Backward-incompatible network protocol changes
 
@@ -41,6 +45,9 @@ To be released.
     replaced by a new `RecentStates` message type
     (with the type number `0x13`).  [[#912]]
  -  Added `BlockHeader.TotalDifficulty` property.  [[#666], [#917]]
+ -  The existing `Pong` message type (with the type number `0x02`) was
+    replaced by a new `Pong` message type
+    (with the type number `0x14`).  [[#459]. [#919]]
 
 ### Backward-incompatible storage format changes
 
@@ -78,6 +85,10 @@ To be released.
     [[#902]]
  -  `BlockPolicy<T>` became to validate `Block<T>.TotalDifficulty` property
     of a `Block<T>`.  [[#666], [#917]]
+ -  `Swarm<T>` became to preload from peer that has the most difficult chain.
+    [[#459], [#919]]
+ -  `Swarm<T>` became to promote the most difficult chain as a canonical chain
+    instead of the longest chain.  [[#459], [#919]]
 
 ### Bug fixes
 
@@ -97,6 +108,7 @@ To be released.
 ### CLI tools
 
 [#404]: https://github.com/planetarium/libplanet/issues/404
+[#459]: https://github.com/planetarium/libplanet/issues/459
 [#666]: https://github.com/planetarium/libplanet/issues/666
 [#756]: https://github.com/planetarium/libplanet/issues/756
 [#796]: https://github.com/planetarium/libplanet/issues/796
@@ -119,6 +131,7 @@ To be released.
 [#913]: https://github.com/planetarium/libplanet/pull/913
 [#916]: https://github.com/planetarium/libplanet/pull/916
 [#917]: https://github.com/planetarium/libplanet/pull/917
+[#919]: https://github.com/planetarium/libplanet/pull/919
 [sleep mode]: https://en.wikipedia.org/wiki/Sleep_mode
 
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,10 +31,6 @@ To be released.
     constructor.  [[#666], [#917]]
  -  Added `BigInteger`-typed `previousTotalDifficulty` parameter to
     `Block<T>.Mine()` static method.  [[#666], [#917]]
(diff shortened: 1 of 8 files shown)


CHANGES.md
```diff
@@ -31,6 +31,10 @@ To be released.
     constructor.  [[#666], [#917]]
  -  Added `BigInteger`-typed `previousTotalDifficulty` parameter to
     `Block<T>.Mine()` static method.  [[#666], [#917]]
+ -  `Pong()` constructor has changed:  [[459], [#919]]
```
longfin, line 34: I guess this mention may not be needed because `Pong` isn't exposed to Libplanet users.

Libplanet/Net/Swarm.cs
```diff
@@ -45,7 +45,7 @@ public class Swarm<T> : IDisposable
         private CancellationTokenSource _workerCancellationTokenSource;
         private CancellationToken _cancellationToken;
 
-        private (long, BoundPeer, HashDigest<SHA256>)? _demandBlockHash;
+        private (BlockHeader, BoundPeer, HashDigest<SHA256>)? _demandBlockHash;
```
longfin, line 48: Maybe it is better to make it a separate type... 🤔

Libplanet/Net/Swarm.cs
```diff
@@ -460,18 +460,23 @@ public string TraceTable()
                 BlockChain.Tip.Hash
             );
 
-            var peersWithHeight = (await DialToExistingPeers(dialTimeout, cancellationToken))
-                .Where(pp => pp.Item2.TipIndex > (initialTip?.Index ?? -1))
-                .Select(pp => (pp.Item1, pp.Item2.TipIndex))
+            var peeers = await DialToExistingPeers(dialTimeout, cancellationToken);
```
longfin, line 463: 1. How about specifying the type of `peeers`.
2. I guess `peeers` is typo.

Libplanet/Net/Swarm.cs
```diff
@@ -1642,12 +1649,12 @@ private void ProcessMessageHandler(object target, Message message)
 
             using (await _blockSyncMutex.LockAsync(cancellationToken))
             {
-                // FIXME: this should be changed into total difficulty.
-                if (header.Index > BlockChain.Tip.Index &&
-                    (_demandBlockHash is null || _demandBlockHash.Value.Item1 < header.Index))
+                if (header.TotalDifficulty > BlockChain.Tip.TotalDifficulty &&
+                    (_demandBlockHash is null ||
+                     _demandBlockHash.Value.Item1.TotalDifficulty < header.TotalDifficulty))
```
longfin, line 1654: How about abstracting this comparison as an external method?

Libplanet/Net/Swarm.cs
limebell: This was included by accident. :(   I'll fix it!

CHANGES.md
dahlia: +1, except message types should remain as they are part of the network protocol.

Libplanet/Net/Swarm.cs
limebell: Do you mean by tuple? Or `BlockHeader`?

longfin: Tuple. 😅

Libplanet/Net/Messages/Pong.cs
```diff
@@ -1,13 +1,19 @@
 using System.Collections.Generic;
+using System.Numerics;
 using NetMQ;
 
 namespace Libplanet.Net.Messages
 {
     internal class Pong : Message
     {
-        public Pong(long? tipIndex)
+        public Pong()
```
moreal, line 9: IMHO, I believe we should avoid using constructor with no parameters in this case. Because if someone uses the constructor, it must set fields (e.g., `Pong.Difficulty`, `Pong.TipIndex`), but sadly there are no tools to guarantee it at compile-time (e.g., static code analyzer). Of course, I believe the libplanet team, reviewers of this pull-request, and watchers of this project's pull-requests will fill the fields correctly. But new contributors, sometimes some of them, can have a mistake about it. And it will be critical when they weren't filled because it's the part of the network protocol.

In my thought, there are two ways to resolve the problem.

1. Avoid using a constructor with no parameters.
2. Make a static code analyzer with [Roslyn Analyzer][roslyn-analyzer] for *Libplanet*.

It's just my opinion.

[roslyn-analyzer]: https://github.com/dotnet/roslyn

limebell: I've made this because `Pong` message is used in two ways. First case is the one which used in protocol, just for handshaking. Second case is the one used in `Swarm`, dial for preloading.

I guess adding new type of message for dial is required... Then this empty constructor will be removed naturally :)

moreal: Okay, I hope so soon :)

dahlia: Can't agree more. Although initializing through default constructor with [object initializers][1] has been treated as the best practice for C#, the practice leads…:

1. `null` fields by mistake
2. enforcing every property/field to be nullable, even though some of them should be required business-logic-wise

I prefer to simply take all the required field values through constructor parameters with good names.

[1]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers#object-initializers

moreal: > I've made this because `Pong` message is used in two ways. First case is the one which used in protocol, just for handshaking. Second case is the one used in `Swarm`, dial for preloading.
> 
> I guess adding new type of message for dial is required... Then this empty constructor will be removed naturally :)

How about opening an issue about this if there is no issue yet?

limebell: @moreal #920 😊

PR #923, opened 2020-07-13
Make CLI tools installable through npm

dahlia:
To make Libplanet CLI tools easily installable without .NET Core SDK, I packaged it using `npm`: [@planetarium/cli](https://www.npmjs.com/package/@planetarium/cli).  The `npm` package in itself actually does not contain any executable binary/code for CLI tools, but includes downloader/installer for the `planet` command.  FYI [Elm's npm installer][1] inspired this approach.

Here's a demo session:

[asciicast](https://asciinema.org/a/347262)

[1]: https://github.com/elm/compiler/tree/master/installers/npm

### DIFF
--- a/.github/bin/dist-npm.sh
+++ b/.github/bin/dist-npm.sh
@@ -12,16 +12,8 @@ if [ "$NODE_AUTH_TOKEN" = "" ]; then
   exit 1
 fi
 
-# Publish a package only if the repository is upstream (planetarium/libplanet)
-# and the branch is for releases (master or *-maintenance or 9c-*).
-# shellcheck disable=SC2235
-if [ "$GITHUB_REPOSITORY" = "planetarium/libplanet" ] && [[ \
-    "$GITHUB_REF" = refs/tags/* || \
-    "$GITHUB_REF" = refs/heads/master || \
-    "$GITHUB_REF" = refs/heads/*-maintenance || \
-    "$GITHUB_REF" = refs/heads/9c-* \
-  ]]; then
-  dry_run=
+if [ "$publish_package" = "" ]; then
+  dry_run=--dry-run
 else
   dry_run=--dry-run
 fi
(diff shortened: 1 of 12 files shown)


.github/bin/dist-npm.sh
```diff
@@ -0,0 +1,36 @@
...
+if [ "$NODE_AUTH_TOKEN" = "" ]; then
+  echo "This script requires NODE_AUTH_TOKEN envrionment variable." \
+    > /dev/stderr
+  exit 1
+fi
+
+# Publish a package only if the repository is upstream (planetarium/libplanet)
+# and the branch is for releases (master or *-maintenance or 9c-*).
+# shellcheck disable=SC2235
+if [ "$GITHUB_REPOSITORY" = "planetarium/libplanet" ] && [[ \
```
longfin, line 18: this check seems duplicated with [dist-nuget.sh's](https://github.com/planetarium/libplanet/blob/b5dcfbe/.github/bin/dist-nuget.sh#L25-L29)... Can we reuse it?

Libplanet.Tools/bin/planet
```diff
@@ -0,0 +1,20 @@
...
+
+const exePath = path.join(BIN_PATH, BIN_NAME);
+
+function runExe() {
+  child_process.spawn(exePath, process.argv.slice(2), { stdio: "inherit" })
+    .on("exit", process.exit);
+}
+
+fs.stat(exePath, (err, stats) => {
+  if (!err && stats.isFile() && stats.size > 64 * 1024) return runExe();
```
longfin, line 16: Do we need to check the file's size?

dahlia: This heuristics purposes to determine if the file is a node.js script file (< 64 KiB) or an actual executable binary (> 64 Kib), but indeed it's hardly understandable its intention.  I am going to adjust this code so that it's more easily understandable.

.github/bin/dist-npm.sh
dahlia: 🙄 Yeah, I did copy-and-pasted this piece from the *dist-nuget.sh* script… I am going to merge two check conditionals to a common script (probably *constants.sh*?).

Issue #922, opened 2020-07-13
Adjust threshold for future block


longfin:
Related issue: #65 

Libplanet provides a static value(900 seconds) as a time threshhold for accepting future blocks. however, it can be a problem because other block mining can be delayed if a block has 900 seconds later value, 

So, I guess it would be better if we provide dynamic value linked to block target time(`BlockPolicy.BlockInterval`) instead of static.

PR #925, opened 2020-07-14
Fix TimestampThreshold to 15s

earlbread:
This fixes `BlockHeader.TimestampThreshold` to 15s to mitigate #922. It should be changed later to be the same as `BlockPolicy.BlockInterval`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,8 @@ To be released.
  -  The existing `Pong` message type (with the type number `0x02`) was
     replaced by a new `Pong` message type
     (with the type number `0x14`).  [[#459]. [#919]]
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -44,6 +44,8 @@ To be released.
  -  The existing `Pong` message type (with the type number `0x02`) was
     replaced by a new `Pong` message type
     (with the type number `0x14`).  [[#459]. [#919]]
+ -  The `TimestampThreshold` for future block has been changed from 900s to 15s.
+    [[#922], [#925]]
```
dahlia, line 48:
```suggestion
 -  The `TimestampThreshold` between `Block<T>`s was changed from 15 minutes to
    15 seconds.  [[#922], [#925]]
```

PR #926, opened 2020-07-14
Separate timeouts in Swarm<T> to SwarmOptions

earlbread:
Separate timeouts in `Swarm<T>` to `SwarmOptions` to be configurable.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -32,7 +32,7 @@ namespace Libplanet.Net
         private static readonly TimeSpan MaxTimeout = TimeSpan.FromSeconds(150);
         private static readonly TimeSpan BlockRecvTimeout = TimeSpan.FromSeconds(15);
         private static readonly TimeSpan TxRecvTimeout = TimeSpan.FromSeconds(3);
-        private static readonly TimeSpan RecentStateRecvTimeout = TimeSpan.FromSeconds(150);
+        private static readonly TimeSpan RecentStateRecvTimeout = TimeSpan.FromSeconds(90);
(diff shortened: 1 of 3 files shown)


Libplanet/Net/SwarmOptions.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using Libplanet.Blocks;
+using Libplanet.Tx;
+
+namespace Libplanet.Net
+{
+    public class SwarmOptions
```
longfin, line 7: How about using `struct` instead of `class`?

earlbread: I made it to a `class` to use initializers. Would it be better to change it to `struct`?

Libplanet/Net/Swarm.cs
```diff
@@ -81,6 +78,7 @@ static Swarm()
         /// <param name="trustedAppProtocolVersionSigners"><see cref="PublicKey"/>s of parties
         /// to trust <see cref="AppProtocolVersion"/>s they signed.  To trust any party, pass
         /// <c>null</c>, which is default.</param>
+        /// <param name="swarmOptions">Options for <see cref="Swarm{T}"/>.</param>
```
dahlia, line 81: As it's of course swarm's options, how about simply renaming it to `options`?

earlbread: Changed it.

Libplanet/Net/SwarmOptions.cs
longfin: Maybe... because it seems just data class. but on the other way, it isn't a big deal.

PR #930, opened 2020-07-21
Separate Ping/Pong message

### LINKED ISSUE #920: Distinguish usage of Ping/Pong message
Currently, `Ping`/`Pong` messages are used in two places:
 1. Dial during preloading in `Swarm<T>`.
 2. Handshake and liveness checking in protocol.

I suggest to make new type of message for second case.

limebell:
`Ping` and `Pong` messages were used in both health checks and chain status checking. Thus, added new messages `GetChainStatus` and `ChainStatus` for the case of chain status checking. Now `Ping` and `Pong` used in health checks only.

Closes #920.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,9 +45,12 @@ To be released.
  -  Added `BlockHeader.TotalDifficulty` property.  [[#666], [#917]]
  -  The existing `Pong` message type (with the type number `0x02`) was
     replaced by a new `Pong` message type
-    (with the type number `0x14`).  [[#459]. [#919]]
+    (with the type number `0x15`).  [[#459], [#919], [#920], [#930]]
  -  The `TimestampThreshold` between `Block<T>`s was changed from 15 minutes to
     15 seconds.  [[#922], [#925]]
+ -  Added new message types:
+     -  Added `GetChainStatus` message.  [[#920], [#930]]
+     -  Added `ChainStatus` message.  [[#920], [#930]]
 
 ### Backward-incompatible storage format changes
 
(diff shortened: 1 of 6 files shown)


Libplanet/Net/Messages/Message.cs
```diff
@@ -21,7 +21,7 @@ internal enum MessageType : byte
             /// <summary>
             /// A reply to <see cref="Ping"/>.
             /// </summary>
-            Pong = 0x14,
+            Pong = 0x15,
```
dahlia, line 24: Why do we need to bump this type id? If I remember correctly, we've never released `0x14` yet.

CHANGES.md
```diff
@@ -45,9 +45,12 @@ To be released.
  -  Added `BlockHeader.TotalDifficulty` property.  [[#666], [#917]]
  -  The existing `Pong` message type (with the type number `0x02`) was
     replaced by a new `Pong` message type
-    (with the type number `0x14`).  [[#459]. [#919]]
+    (with the type number `0x15`).  [[#459], [#919], [#920], [#930]]
  -  The `TimestampThreshold` between `Block<T>`s was changed from 15 minutes to
     15 seconds.  [[#922], [#925]]
+ -  Added new message types:
+     -  Added `GetChainStatus` message.  [[#920], [#930]]
+     -  Added `ChainStatus` message.  [[#920], [#930]]
```
riemannulus, line 53: Do we need this? The interface does not seem to have changed.

limebell: I've added these lines because this patch will lead to network inconsistency between swarms.

CHANGES.md
```diff
@@ -45,9 +45,12 @@ To be released.
  -  Added `BlockHeader.TotalDifficulty` property.  [[#666], [#917]]
  -  The existing `Pong` message type (with the type number `0x02`) was
     replaced by a new `Pong` message type
-    (with the type number `0x14`).  [[#459]. [#919]]
+    (with the type number `0x14`).  [[#459], [#919], [#920], [#930]]
  -  The `TimestampThreshold` between `Block<T>`s was changed from 15 minutes to
     15 seconds.  [[#922], [#925]]
+ -  Added new message types:
+     -  Added `GetChainStatus` message.  [[#920], [#930]]
+     -  Added `ChainStatus` message.  [[#920], [#930]]
```
moreal, line 53:
```suggestion
     -  Added `GetChainStatus` (`0x20`) message.  [[#920], [#930]]
     -  Added `ChainStatus` (`0x21`) message.  [[#920], [#930]]
```

It's just suggestion, how about adding the message type value like the below lines in *CHANGES.md*, or as another format.

```
 -  `Swarm<T>` became to have two more message types: `GetRecentStates` (`0x0b`)
    and `RecentStates` (`0x0c`).  [[#272], [#343]]
```

https://github.com/planetarium/libplanet/blame/7723a3deb39ea49400e76862318ec482a2a179f9/CHANGES.md#L1074-L1075

PR #933, opened 2020-07-23
Don't throw exception when connecting to all neighbors fails

earlbread:
This PR changes to not throw an exception in `Swarm.BootstrapAsync()` when connecting to all neighbors fails.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -98,8 +98,8 @@ To be released.
     [[#459], [#919]]
  -  `Swarm<T>` became to promote the most difficult chain as a canonical chain
     instead of the longest chain.  [[#459], [#919]]
- -  Changed to not throw an exception in `Swarm.BootstrapAsync()` when
-    connecting to all neighbors fails.  [[#933]]
+ -  `Swarm<T>.BootstrapAsync()` method became not to throw `TimeoutException` when
+    it fails to connect to all neighbors.  [[#933]]
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -98,6 +98,8 @@ To be released.
     [[#459], [#919]]
  -  `Swarm<T>` became to promote the most difficult chain as a canonical chain
     instead of the longest chain.  [[#459], [#919]]
+ -  Changed to not throw an exception in `Swarm.BootstrapAsync()` when
+    connecting to all neighbors fails.  [[#933]]
```
dahlia, line 102:
```suggestion
 -  `Swarm.BootstrapAsync()` method became not to throw `TimeoutException` when
    it fails to connect to all neighbors.  [[#933]]
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -127,15 +127,6 @@ private readonly DifferentAppProtocolVersionEncountered
             }
             catch (Exception e)
             {
-                if (findPeerTasks.All(task =>
-                    task.IsFaulted &&
-                    !(task.Exception is null) &&
-                    task.Exception.InnerExceptions.All(ex => ex is TimeoutException)))
-                {
-                    throw new TimeoutException(
```
dahlia, line 135: Do we log inner exceptions in other places?  If not it might better to log them here. 🤔

earlbread: I thought the inner exceptions are logged here. https://github.com/planetarium/libplanet/pull/933/files#diff-ae527d5afcdf41da4a57b594cab3cfefL655

earlbread:
I removed another exception https://github.com/planetarium/libplanet/pull/933/files#diff-ae527d5afcdf41da4a57b594cab3cfefL698

Issue #929, opened 2020-07-21
Using trusted peers to fill states during `BlockChain<T>.GetRawState()`


longfin:
Currently, `BlockChain<T>.GetRawState()` tries to calculate and fill whole state on the fly if `completeStates` is `true`. it may take a very long time with longer chain.

To shorten it, querying to trusted peers when encountered missing states can be a possible solution.

PR #934, opened 2020-07-23
More flexible state complement

dahlia:
This changes is a prior work to address the issue #929.  The existing option `bool completeStates` on `BlockChain<T>.GetState()` and `BlockChain<T>.GetBalance()` methods had provided only two ways to handle *incomplete block states*:

- `completeStates: true` recalculates actions from scratch and fills the incomplete block states with the recalculated states.  Of course this probably takes a long time.
- `completeStates: false` rejects inquiry on a state value and throws `IncompleteBlockStatesException` instead.  This prevents high latency, but also prevents to *use* 🙄 at the same time.

However, in order to achieve the idea of #929, we need to add one more option, and make `BlockChain<T>` to depend on `Swarm<T>`, which causes circular dependency (`BlockChain<T>` → `Swarm<T>` → `BlockChain<T>`).  To avoid such circular dependency, the new way turns the `bool`ean flag into the callback named `stateCompleter`.

There are two types of delegates for `GetState()` and `GetBalance()` methods, respectively:

- `IValue StateCompleter<T>(BlockChain<T>, HashDigest<SHA256>, Address)`
- `BigInteger FungibleAssetStateCompleter<T>(BlockChain<T>, HashDigest<SHA256>, Address, Currency)`

Other higher methods than `GetState()` and `GetBalance()` need both of them, so a new `StateCompleterSet<T>` struct is introduced.  Currently two predefined built-in completer sets are provided, and they corresponds to the existing `completeStates: true` and `false` options:

- `StateCompleterSet<T>.Recalculate`
- `StateCompleterSet<T>.Reject`

I am going to submit a separated patch later, which will introduce a new completer set: `Swarm<T>.GetTrustedStateCompleter()`.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -660,7 +660,7 @@ namespace Libplanet.Blockchain
                 throw new OperationCanceledException(cancellationToken);
             }
 
-            var actionEvaluations = EvaluateActions(block);
+            var actionEvaluations = EvaluateActions(block, StateCompleterSet<T>.Recalculate);
 
             block = new Block<T>(block, ActionEvaluationsToHash(actionEvaluations));
 
@@ -853,12 +853,18 @@ namespace Libplanet.Blockchain
         /// Render actions from block index of <paramref name="offset"/>.
         /// </summary>
         /// <param name="offset">Index of the block to start rendering from.</param>
-        internal void RenderBlocks(long offset)
+        /// <param name="stateCompleters">The strategy to complement incomplete block states.
+        /// <see cref="StateCompleterSet{T}.Recalculate"/> by default.</param>
+        internal void RenderBlocks(long offset, StateCompleterSet<T>? stateCompleters = null)
         {
+            // Since rendering process requires every step's states, if required block states
+            // are incomplete they are complemented anyway:
+            stateCompleters ??= StateCompleterSet<T>.Recalculate;
+
             // FIXME: We should consider the case where block count is larger than int.MaxSize.
             foreach (var block in IterateBlocks((int)offset))
             {
-                RenderBlock(null, block);
+                RenderBlock(null, block, stateCompleters);
             }
         }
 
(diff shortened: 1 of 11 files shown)


Libplanet/Blockchain/StateCompleters.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Security.Cryptography;
+using Bencodex.Types;
+using Libplanet.Action;
+
+namespace Libplanet.Blockchain
+{
+    /// <summary>
+    /// Predefined built-in state completers that statisfy <see cref="StateCompleter{T}"/> delegate.
```
earlbread, line 9:
```suggestion
    /// Predefined built-in state completers that satisfy <see cref="StateCompleter{T}"/> delegate.
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -845,12 +853,18 @@ bool renderActions
         /// Render actions from block index of <paramref name="offset"/>.
         /// </summary>
         /// <param name="offset">Index of the block to start rendering from.</param>
-        internal void RenderBlocks(long offset)
+        /// <param name="stateCompleters">The strategy to complemnet incomplete block states.
```
earlbread, line 856:
```suggestion
        /// <param name="stateCompleters">The strategy to complement incomplete block states.
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -860,13 +874,23 @@ internal void RenderBlocks(long offset)
         /// <param name="evaluations"><see cref="ActionEvaluation"/>s of the block.  If it is
         /// <c>null</c>, evaluate actions of the <paramref name="block"/> again.</param>
         /// <param name="block"><see cref="Block{T}"/> to render actions.</param>
-        internal void RenderBlock(IReadOnlyList<ActionEvaluation> evaluations, Block<T> block)
+        /// <param name="stateCompleters">The strategy to complemnet incomplete block states.
```
earlbread, line 877:
```suggestion
        /// <param name="stateCompleters">The strategy to complement incomplete block states.
```

PR #936, opened 2020-07-27
Add PeerChainState

earlbread:
This adds `PeerChainStatus` struct and `Swarm<T>.GetPeerChainStatus()` method to check the chain status of the connected peers.

### DIFF
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -2023,8 +2023,9 @@ namespace Libplanet.Tests.Net
             Swarm<DumbAction> swarm2 = _swarms[1];
             Swarm<DumbAction> swarm3 = _swarms[2];
 
-            var peerStatus = await swarm1.GetPeerChainStatusAsync(TimeSpan.FromSeconds(1), default);
-            Assert.Empty(peerStatus);
+            var peerChainState = await swarm1.GetPeerChainStateAsync(
+                TimeSpan.FromSeconds(1), default);
+            Assert.Empty(peerChainState);
 
             try
             {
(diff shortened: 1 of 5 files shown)


earlbread:
I changed an `Array` to a `HashSet` to fix the flaky test. 945d6d3

Libplanet/Net/Swarm.cs
```diff
@@ -404,6 +404,26 @@ public string TraceTable()
...
+        /// <summary>
+        /// Gets the <see cref="PeerChainStatus"/> of the connected <see cref="Peers"/>.
+        /// </summary>
+        /// <param name="dialTimeout">A timeout value for dialing.</param>
+        /// <param name="cancellationToken">
+        /// A cancellation token used to propagate notification that this
+        /// operation should be canceled.
+        /// </param>
+        /// <returns><see cref="PeerChainStatus"/> of the connected <see cref="Peers"/>.</returns>
+        public async Task<IEnumerable<PeerChainStatus>> GetPeerChainStatus(
```
dahlia, line 416: The name should end with `Async`.

earlbread: Renamed `GetPeerChainStatus` to `GetPeerChainStatusAsync`. ba45953

earlbread:
> How about renaming `PeerChainStatus` to `PeerChainState`? Cf. https://english.stackexchange.com/questions/12958/status-vs-state. /cc @planetarium/libplanet

I think that's good. I'll rename it.

limebell:
Currently, `Swarm<T>.DialToExistingPeers` returns `Task<(BoundPeer, ChainStatus)>` as its return type. How about merge `PeerChainState` struct and `ChainStatus` (and make it public)? Then we can just use `Task<PeerChainState>` as the method's return type.

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -2016,6 +2016,58 @@ public async Task DoNotFillMultipleTimes()
             }
         }
 
+        [Fact(Timeout = Timeout)]
+        public async Task GetPeerChainStatusAsync()
```
dahlia, line 2020:
```suggestion
        public async Task GetPeerChainStateAsync()
```

earlbread:
> Currently, `Swarm<T>.DialToExistingPeers` returns `Task<(BoundPeer, ChainStatus)>` as its return type. How about merge `PeerChainState` struct and `ChainStatus` (and make it public)? Then we can just use `Task<PeerChainState>` as the method's return type.

I thought about it at first, but to do that, the current internal `Message` class should be public. Is it okay?

limebell:
> > Currently, `Swarm<T>.DialToExistingPeers` returns `Task<(BoundPeer, ChainStatus)>` as its return type. How about merge `PeerChainState` struct and `ChainStatus` (and make it public)? Then we can just use `Task<PeerChainState>` as the method's return type.
> 
> I thought about it at first, but to do that, the current internal `Message` class should be public. Is it okay?

OH I made mistake, you're right.

PR #941, opened 2020-07-30
Fix NullReferenceException in TransferRecentStates

earlbread:
This fixes `NullReferenceException` in `TransferRecentStates` when `GetRecentStates.TargetBlockHash` doesn't exist in the chain. Currently, the purpose of `ReorgWhilePreloadAsync` test case is to check if a timeout occurs and should be fixed in the following PR.

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -120,7 +120,7 @@ namespace Libplanet.Net
             IEnumerable<IceServer> iceServers = null,
             DifferentAppProtocolVersionEncountered differentAppProtocolVersionEncountered = null,
             IEnumerable<PublicKey> trustedAppProtocolVersionSigners = null,
-            SwarmOptions swarmOptions = null)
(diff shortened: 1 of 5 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -120,7 +120,7 @@ static Swarm()
             IEnumerable<IceServer> iceServers = null,
             DifferentAppProtocolVersionEncountered differentAppProtocolVersionEncountered = null,
             IEnumerable<PublicKey> trustedAppProtocolVersionSigners = null,
-            SwarmOptions swarmOptions = null)
+            SwarmOptions options = null)
```
longfin, line 123: I guess we may need to describe this change to CHANGES.md.

earlbread: I didn't add it to the changelog because the constructor is internal.

longfin: 🆗

PR #942, opened 2020-07-30
Implement .Equals for Currency

longfin:
The title says it all.


dahlia:
Can't we use Equals.Fody for this?

longfin:
> Can't we use Equals.Fody for this?

I've tried but it didn't work as expected.

PR #943, opened 2020-07-31
Retry preload when RecentState target is missing

earlbread:
This replaces #941.

This fixes NullReferenceException in TransferRecentStates when `GetRecentStates.TargetBlockHash` doesn't exist in the chain. Also, this makes retry preload when `GetRecentStates.TargetBlockHash` is missing.

### DIFF
--- a/Libplanet/Net/InvalidStateTargetException.cs
+++ b/Libplanet/Net/InvalidStateTargetException.cs
@@ -1,0 +1,28 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace Libplanet.Net
+{
+    [Serializable]
+    public class InvalidStateTargetException : SwarmException
+    {
+        public InvalidStateTargetException()
+        {
+        }
+
+        public InvalidStateTargetException(string message)
+            : base(message)
+        {
+        }
+
+        public InvalidStateTargetException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+        }
+
+        protected InvalidStateTargetException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        {
+        }
+    }
+}
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -213,8 +213,6 @@ namespace Libplanet.Net
         // FIXME: Should have a unit test.
         internal AsyncAutoResetEvent BlockAppended { get; }
 
-        internal TimeSpan BlockHashRecvTimeout { get; set; } = TimeSpan.FromSeconds(30);
-
         // FIXME: We need some sort of configuration method for it.
         internal int FindNextHashesChunkSize { get; set; } = 500;
 
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -31,7 +31,6 @@ namespace Libplanet.Net
         private const int InitialBlockDownloadWindow = 100;
         private readonly PrivateKey _privateKey;
         private readonly AppProtocolVersion _appProtocolVersion;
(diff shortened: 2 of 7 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -573,230 +639,172 @@ await foreach (var pair in demandBlockHashes)
...
-
-                tipCandidate = tempTip;
-
-                if (tipCandidate is null)
-                {
-                    // If there is no blocks in the network (or no consensus at least)
-                    // it doesn't need to receive states from other peers at all.
-                    return;
-                }
+                        // If there is no blocks in the network (or no consensus at least)
```
longfin, line 690: How about logging here?

Libplanet/Net/Swarm.cs
```diff
@@ -573,230 +639,172 @@ await foreach (var pair in demandBlockHashes)
...
+                            progress,
+                            trustedPeersWithTip.ToImmutableList(),
+                            initialLocator,
+                            cancellationToken
                         );
+                        break;
+                    }
+                    catch (InvalidStateTargetException e)
+                    {
+                        Log.Error(e.ToString());
```
dahlia, line 803:
```suggestion
                        Log.Error(e, e.ToString());
```

Libplanet/Net/Swarm.cs
```diff
@@ -1360,8 +1390,14 @@ CancellationToken cancellationToken
                         break;
                     }
 
-                    if (reply is RecentStates recentStates && !recentStates.Missing)
+                    if (reply is RecentStates recentStates)
                     {
+                        if (recentStates.Missing)
+                        {
+                            throw new InvalidStateTargetException(
+                                $"The target block hash is invalid: {request.TargetBlockHash}");
```
dahlia, line 1398:
```suggestion
                                $"The {nameof(RecentStates)}.{nameof(RecentStates.TargetBlockHash)}"
                                + $" is invalid: {request.TargetBlockHash}"
                            );
```

Libplanet/Net/InvalidStateTargetException.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace Libplanet.Net
+{
+    [Serializable]
+    internal class InvalidStateTargetException : SwarmException
```
dahlia, line 7: As what “state target” means here is not obvious (out of context), how about making this an inner class of `GetRecentStates` and renaming it `GetRecentStates.InvalidTargetException`? This would be easier to get it because the term “invalid target” is common between <code>GetRecentStates.<em>InvalidTarget</em>Exception</code> class and <code>GetRecentStates.<em>InvalidTarget</em></code> property.

dahlia: Also it would be better if this class has a property named `InvalidTarget` (or more simply `Target`) property.

moreal:
@earlbread The build seems failed on CI envrionment with below message:

```text
Libplanet/Net/Swarm.cs(1398,83): Error CS0117: 'RecentStates' does not contain a definition for 'TargetBlockHash'
```

Libplanet/Net/Messages/GetRecentStates.cs
```diff
@@ -47,5 +48,17 @@ protected override IEnumerable<NetMQFrame> DataFrames
                 }
             }
         }
+
+        [Serializable]
+        internal class InvalidTargetException : SwarmException
```
longfin, line 53: I guess we need a protected constructor for [`Serializable`].

earlbread: Added a constructor for serialization.

PR #945, opened 2020-08-05
Introduce Swarm<T>.Reorged event

riemannulus:


CHANGES.md
```diff
@@ -103,6 +103,7 @@ To be released.
      -  Added `StateCompleterSet<T>` struct.
      -  Added `StateCompleters<T>` static class.
      -  Added `FungibleAssetStateCompleters<T>` static class.
+ -  Added `Reorged` event on `BlockChain<T>`.  [[#945]]
```
earlbread, line 106: `ReorgedEventArgs` should be added as well.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -176,6 +176,11 @@ bool render
         /// </summary>
         public event EventHandler<TipChangedEventArgs> TipChanged;
 
+        /// <summary>
+        /// An event which is invoked when chain is <see cref="Swap"/>ed.
```
moreal, line 180:
```suggestion
        /// An event which is invoked when chain is <see cref="Swap"/>ped.
```

CHANGES.md
riemannulus: Rebase https://github.com/planetarium/libplanet/pull/945/commits/56c9ce1b2b3e56e32b974e28bb6a3615d72ec85e

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2173,6 +2173,37 @@ void TipChangedHandler(object target, BlockChain<DumbAction>.TipChangedEventArgs
...
+            _blockChain.Reorged += ReorgedHandler;
+            var branchpoint = _blockChain.Tip;
+            var fork = _blockChain.Fork(_blockChain.Tip.Hash);
+            await fork.MineBlock(_fx.Address1);
+            await fork.MineBlock(_fx.Address2);
+            await _blockChain.MineBlock(_fx.Address3);
+
+            var oldTip = _blockChain.Tip;
+            var newTip = fork.Tip;
+
```
dahlia, line 2199: Need to ensure if the event has never been triggered.

```suggestion
            Assert.Null(actualOldTip.Hash);
            Assert.Null(actualNewTip.Hash);
            Assert.Null(actualBranchpoint.Hash); 

```

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2173,6 +2173,37 @@ void TipChangedHandler(object target, BlockChain<DumbAction>.TipChangedEventArgs
...
+        {
+            Block<DumbAction> actualOldTip = default;
+            Block<DumbAction> actualNewTip = default;
+            Block<DumbAction> actualBranchpoint = default;
+
+            void ReorgedHandler(object target, BlockChain<DumbAction>.ReorgedEventArgs args)
+            {
+                actualOldTip = args.OldTip;
+                actualNewTip = args.NewTip;
+                actualBranchpoint = args.Branchpoint;
```
dahlia, line 2187: One thing we need to test here is how many times the event is triggered: it should be only once.  How about declaring a counter variable and increment it here?  We also need to assert the counter should be 0 until `Fork()` is called and only 1 in the end.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1624,5 +1638,26 @@ public class TipChangedEventArgs : EventArgs
...
+        /// </summary>
+        public class ReorgedEventArgs : EventArgs
+        {
+            /// <summary>
+            /// The <see cref="Block{T}"/> before chain swapped.
+            /// </summary>
+            public Block<T> OldTip { get; set; }
+
+            /// <summary>
+            /// The <see cref="Block{T}"/> after chain swapped.
```
moreal, line 1653: I want to hear your opinion about the expression, `The Block<T>`. Because it just describes as `The Block<T>` without the description of the tip, so it is hard to know what the referenced block is among blocks after or before swapped when we looked only the comment. Of course, we can know it from the name of the property, but IMO it will be better to describe that the block is the tip block. If you agree with my opinion, I suggest referencing `BlockChain<T>.Tip` instead of `Block<T>`?

```suggestion
            /// The <see cref="BlockChain{T}.Tip"/> after chain swapped.
```

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2173,6 +2173,37 @@ void TipChangedHandler(object target, BlockChain<DumbAction>.TipChangedEventArgs
...
+            }
+
+            _blockChain.Reorged += ReorgedHandler;
```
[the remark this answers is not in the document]
riemannulus: you mean like this. right?
```suggestion
            _blockChain.Reorged += (target, args) =>
            {
                actualOldTip = args.OldTip;
                actualNewTip = args.NewTip;
                actualBranchpoint = args.Branchpoint;
            }
```

Libplanet/Blockchain/BlockChain.cs
dahlia: @moreal :+1: Seems much better.

One more thing I want to suggest @riemannulus is that how about adjusting the term “swap” here.  Actually, we've never set the consistent terminology for reog/fork/swap etc yet, I believe it's good chance to fix this and give consistent nuance to these similar terms.  IMHO I feel the word “reorg” is way more appropriate here (i.e., “after the chain is reorged”).  /cc @libplanet

Libplanet.Tests/Blockchain/BlockChainTest.cs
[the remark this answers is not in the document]
dahlia: @riemannulus Oh yes, you're right. :+1:

Libplanet/Blockchain/BlockChain.cs
riemannulus: @moreal I highly agreed with your opinion.

@dahlia It is a much better way to represent the meaning. I agree with that.

riemannulus: change it https://github.com/planetarium/libplanet/pull/945/commits/5f413117f96112ae0336dcf72a461dd5b55209a1

Libplanet.Tests/Blockchain/BlockChainTest.cs
riemannulus: Add test by https://github.com/planetarium/libplanet/pull/945/commits/cd1d870e19835906ab1b5b04b6623e9e97d6a80e

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1624,5 +1638,26 @@ public class TipChangedEventArgs : EventArgs
...
             public HashDigest<SHA256> Hash { get; set; }
         }
+
+        /// <summary>
+        /// Provides data for the <see cref="BlockChain{T}.Reorged"/> event.
+        /// </summary>
+        public class ReorgedEventArgs : EventArgs
+        {
+            /// <summary>
+            /// The <see cref="BlockChain{T}.Tip"/> before chain reorged.
```
dahlia, line 1648:
```suggestion
            /// The <see cref="BlockChain{T}.Tip"/> before the chain is reorged.
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1624,5 +1638,26 @@ public class TipChangedEventArgs : EventArgs
...
+            /// </summary>
+            public Block<T> OldTip { get; set; }
+
+            /// <summary>
+            /// The <see cref="BlockChain{T}.Tip"/> after chain reorged.
```
dahlia, line 1653:
```suggestion
            /// The <see cref="BlockChain{T}.Tip"/> after the chain is reorged.
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1624,5 +1638,26 @@ public class TipChangedEventArgs : EventArgs
...
+            /// </summary>
+            public Block<T> NewTip { get; set; }
+
+            /// <summary>
+            /// The <see cref="Block{T}"/> point of chain was branch off.
```
dahlia, line 1658:
```suggestion
            /// The <see cref="Block{T}"/> point of the chain branches off.
```

riemannulus: rebased https://github.com/planetarium/libplanet/pull/945/commits/1e3a5b6374bcd2882969c6bf35abe89b9c4067a9

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -176,6 +176,11 @@ bool render
         /// </summary>
         public event EventHandler<TipChangedEventArgs> TipChanged;
 
+        /// <summary>
+        /// An event which is invoked when chain is reorged.
```
dahlia, line 180:
```suggestion
        /// An event which is invoked when the chain is reorged.
```

riemannulus: amend https://github.com/planetarium/libplanet/pull/945/commits/617b615628cfe69506a498fbd69cf076522d9793

PR #935, opened 2020-07-24
EvaluationDigest on Block and BlockHeader

riemannulus:
Closes #931 

- This PR adds the field `EvaluationDigest` and `PreEvaluationHash` to `Block<T>` and `BlockHeader`.

Issue #931, opened 2020-07-21
Inserting a state hash into the `BlockHeader`


riemannulus:
Now it's a structure where different result states can come from the same block hash, and it doesn't have to be a problem right now, but assuming the same result status when the `BlockHash` is the same can be synchronized faster.

PR #946, opened 2020-08-06
Trusted state completer

dahlia:
This patch is continued from the previous patch #934 and addresses the idea sketched in the issue #929.

**However, in general I became to believe this approach is not appropriate.**  Here I elaborate:

- Rendering and unrendering whole actions between two tips on reorg anyway requires *all states of every step*, and it means the idea of cherry-picking only recent states is pointless for this case.

- Even if we decide to fetch (trust) all states from peers for *every step* between two tips on reorg, it rather takes longer time than recalculating all states.

- We are slowly moving towards chain-interior-states (from chain-exterior-states, the status quo): #931, #935, and #939.  Eventually, we won't need any *trust* for reusing states calculated by peers, after all.

Hence, although I prepared this patch for a quite long time, I doubt this works well for the problem addressed by the issue #929.  🤔


dahlia:

Libplanet/Net/Swarm.TrustedStateCompleter.cs
```diff
@@ -0,0 +1,147 @@
...
+                        else if (blockStates.States is null)
+                        {
+                            string msg =
+                                "A trusted peer {Peer} does not have the block {BlockHash}'s " +
+                                "states.";
+                            _logger.Error(msg, peer, blockHash);
+                            continue;
+                        }
+
+                        BlockChain.Store.SetBlockStates(blockStates.BlockHash, blockStates.States);
```
longfin, line 115: Is it fine without any of the mutex?

Libplanet/Net/Swarm.TrustedStateCompleter.cs
```diff
@@ -0,0 +1,147 @@
...
+                        _logger.Error(msg, peer, reply);
+                    }
+                }
+            }
+
+            return new StateCompleterSet<T>
+            {
+                StateCompleter = (blockChain, blockHash, address) =>
+                {
+                    FillTrustedBlockStates(blockHash).Wait(cancellationToken);
```
longfin, line 131: It can be a matter if `StateCompleter` is called in `async` method... but it doesn't seem translatable to `async` because `StateCompleter` and related features are all synchronous. 🤔

Libplanet.Tests/TestUtils.cs
```diff
@@ -171,11 +171,11 @@ public static byte[] GetRandomBytes(int size)
 
         public static Block<T> MineGenesis<T>(
             Address? miner = null,
-            IEnumerable<Transaction<T>> transactions = null
+            IEnumerable<Transaction<T>> transactions = null,
+            DateTimeOffset? timestamp = null
         )
             where T : IAction, new()
         {
-            var timestamp = new DateTimeOffset(2018, 11, 29, 0, 0, 0, TimeSpan.Zero);
```
riemannulus, line 178: How about
```
timestamp ??= new DateTimeOffset(2018, 11, 29, 0, 0, 0, TimeSpan.Zero);
```
like this?

dahlia: @riemannulus We don't use C# 8 yet in the test projects… 😢

riemannulus: Ah, I forgot about that. 😭

Libplanet/Net/Swarm.TrustedStateCompleter.cs
dahlia: Rather it does not work if we try to lock the `BlockChain._rwlock` here, because `BlockChain<T>.GetState()`/`GetBalance()` methods guarantee their `stateCompleter` to be called only in the read lock:

https://github.com/planetarium/libplanet/blob/ed9b1f86532d3440176511b68da55b795ed6fca9/Libplanet/Blockchain/BlockChain.cs#L1680-L1702

Libplanet/Net/Swarm.TrustedStateCompleter.cs
dahlia: That's what I'd grappled with, but I gave up and decided to request for comments about this problem…

Libplanet/Net/Swarm.TrustedStateCompleter.cs
```diff
@@ -0,0 +1,147 @@
...
+        )
+        {
+            if (trustedStateValidators is null)
+            {
+                throw new ArgumentNullException(nameof(trustedStateValidators));
+            }
+
+            IReadOnlyList<BoundPeer> trustedPeers =
+                (await DialToExistingPeers(dialTimeout, cancellationToken))
+                    .Select(pair => pair.Peer)
```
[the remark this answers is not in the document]
dahlia: Oh, I missed that.  Thanks!  🙏🏼

Libplanet/Net/Swarm.TrustedStateCompleter.cs
longfin: I see... then it seems that we should upgrade `_rwlock` before calling `rawStateCompleter` in `BlockChain<T>.GetRawState()`.

dahlia: Not all state completers update the blockchain, but only `Recalculate` completers do.  Those implementations have already entered the write lock by themselves.

<https://github.com/planetarium/libplanet/blob/ed9b1f86532d3440176511b68da55b795ed6fca9/Libplanet/Blockchain/BlockChain.cs#L1638-L1647>

longfin: Then, shouldn't these state completers upgrade `_rwlocks`? (like Recalculate)

dahlia: @longfin I'm going to address this problem in the next separated pull request!

Libplanet/Net/Swarm.TrustedStateCompleter.cs
dahlia: @longfin I'm going to address this problem in the next separated pull request!

dahlia:
@planetarium/libplanet Please review this again, folks!

PR #948, opened 2020-08-07
Fix exception message properly

earlbread:
This fixes the wrong exception message properly. 😢

### DIFF
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -801,7 +801,7 @@ namespace Libplanet.Net
                     }
                     catch (InvalidStateTargetException e)
                     {
-                        Log.Error(e.ToString());
+                        Log.Error(e, e.ToString());
                         retryCount++;
                     }
                 }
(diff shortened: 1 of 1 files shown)


Libplanet/Net/Swarm.cs
```diff
@@ -1395,7 +1395,11 @@ CancellationToken cancellationToken
                     {
                         if (recentStates.Missing)
                         {
-                            throw new InvalidStateTargetException("Fuck");
+                            throw new InvalidStateTargetException(
+                                $"The {nameof(GetRecentStates)}."
+                                + $"{nameof(GetRecentStates.TargetBlockHash)}"
+                                + $" is invalid: {request.TargetBlockHash}"
```
dahlia, line 1401: How about this?

```suggestion
                                $"The peer {peer} lacks the block states for {blockHash}."
```

earlbread: In this situation, the peer doesn't have the block itself as well as block states.
```
$"The peer {peer} lacks the target block: {blockHash}."
```
Would this be better? @dahlia

dahlia: That looks better!

earlbread: Fixed it and amended the commit 33c3ab8.

PR #949, opened 2020-08-10
Introduce Version property to messages

limebell:
Removed `Peer.AppProtocolVersion` which were used to check version between swarms. Instead, added `AppProtocolVersion`-typed `Version` property to `Message` and use it to check validity of messages.

If received message has different app protocol version, `NetMQTransport` will reply a `DifferentVersion` message regardless of received message's type and does not process anything.

### DIFF
--- a/Libplanet/Net/BoundPeer.cs
+++ b/Libplanet/Net/BoundPeer.cs
@@ -17,22 +17,18 @@ namespace Libplanet.Net
         /// <see cref="Peer"/>.</param>
         /// <param name="endPoint">A <see cref="DnsEndPoint"/> consisting of the
         /// host and port of the <see cref="Peer"/>.</param>
-        /// <param name="appProtocolVersion">An application protocol version
-        /// that the <see cref="Peer"/> is using.</param>
         public BoundPeer(
             PublicKey publicKey,
-            DnsEndPoint endPoint,
-            AppProtocolVersion appProtocolVersion)
-        : this(publicKey, endPoint, appProtocolVersion, null)
+            DnsEndPoint endPoint)
+        : this(publicKey, endPoint, null)
         {
         }
 
         internal BoundPeer(
             PublicKey publicKey,
             DnsEndPoint endPoint,
-            AppProtocolVersion appProtocolVersion,
             IPAddress publicIPAddress)
-        : base(publicKey, appProtocolVersion, publicIPAddress)
+        : base(publicKey, publicIPAddress)
         {
             EndPoint = endPoint ?? throw new ArgumentNullException(nameof(endPoint));
         }
@@ -71,7 +67,7 @@ namespace Libplanet.Net
         /// <inheritdoc/>
         public override string ToString()
         {
-            return $"{Address}.{EndPoint}.{AppProtocolVersion}";
+            return $"{Address}.{EndPoint}.{PublicIPAddress}";
         }
     }
 }
--- a/Libplanet/Net/DifferentAppProtocolVersionException.cs
+++ b/Libplanet/Net/DifferentAppProtocolVersionException.cs
@@ -1,12 +1,13 @@
 #pragma warning disable S3871
 using System;
+using Libplanet.Net.Messages;
 
 namespace Libplanet.Net
 {
     /// <summary>
     /// The exception that is thrown when the version of the
-    /// <see cref="Peer" /> that <see cref="Swarm{T}" /> is trying to connect
-    /// to is different.
+    /// <see cref="Message" /> that <see cref="Swarm{T}" /> received
+    /// is different.
     /// </summary>
     [Serializable]
     internal sealed class DifferentAppProtocolVersionException : Exception
@@ -15,6 +16,8 @@ namespace Libplanet.Net
         /// Initializes a new instance of the
         /// <see cref="DifferentAppProtocolVersionException"/> class.
         /// </summary>
+        /// <param name="identity">The identity of the message received. Will have empty
+        /// string if the message is a reply.</param>
         /// <param name="expectedVersion">The protocol version of the current
         /// <see cref="Swarm{T}"/>.</param>
         /// <param name="actualVersion">The protocol version of the
@@ -24,6 +27,7 @@ namespace Libplanet.Net
         /// </param>
         public DifferentAppProtocolVersionException(
             string message,
+            byte[] identity,
             AppProtocolVersion expectedVersion,
             AppProtocolVersion actualVersion)
             : base($"{message}\n" +
@@ -32,9 +36,16 @@ namespace Libplanet.Net
                    $"Actual Version: {actualVersion} " +
                    $"[{ByteUtil.Hex(actualVersion.Signature)} by {actualVersion.Signer}]")
         {
+            Identity = identity;
             ExpectedVersion = expectedVersion;
             ActualVersion = actualVersion;
         }
+
+        /// <summary>
+        /// The identity of the message received.
+        /// Will have <c>null</c> if the message is a reply.
+        /// </summary>
+        public byte[] Identity { get; }
 
         /// <summary>
         /// The protocol version of the current <see cref="Swarm{T}"/>.
--- a/Libplanet/Net/Messages/Message.cs
+++ b/Libplanet/Net/Messages/Message.cs
@@ -116,6 +116,8 @@ namespace Libplanet.Net.Messages
 
         public byte[] Identity { get; set; }
 
+        public AppProtocolVersion Version { get; set; }
+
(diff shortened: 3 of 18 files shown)


Libplanet/Net/Peer.cs
```diff
@@ -90,53 +61,10 @@ protected Peer(SerializationInfo info, StreamingContext context)
...
-        /// to trust no one, so that this method always returns <c>false</c>.</em>  Conversely,
-        /// <c>null</c> means to <em>trust anyone</em>.  Be aware of the difference of those two.
-        /// </param>
-        /// <param name="recognizer">A callback to determine an encountered peer's
-        /// <see cref="AppProtocolVersion"/> is compatible with the currently running
-        /// application, in the case of the peer's version signature is trustworthy but its
-        /// version number is not the same.</param>
-        /// <returns><c>true</c> if the peer is compatible with the given criteria.
-        /// Otherwise <c>false</c>.</returns>
-        public bool IsCompatibleWith(
```
earlbread, line 107: This should be added to the changelog.

Libplanet/Net/Peer.cs
```diff
@@ -67,14 +46,6 @@ protected Peer(SerializationInfo info, StreamingContext context)
         [Pure]
         public PublicKey PublicKey { get; }
 
-        /// <summary>
-        /// The corresponding application protocol version of this peer.
-        /// </summary>
-        // FIXME: This would be better to be nullable...
-        [IgnoreDuringEquals]
-        [Pure]
-        public AppProtocolVersion AppProtocolVersion { get; }
```
earlbread, line 76: This should be added to the changelog.

Libplanet/Net/Peer.cs
```diff
@@ -20,39 +15,23 @@ namespace Libplanet.Net
     [Equals]
     public class Peer : ISerializable
     {
-        private static readonly Codec _codec = new Codec();
-
-        public Peer(
-            PublicKey publicKey,
-            AppProtocolVersion appProtocolVersion)
```
earlbread, line 27: This should be added to the changelog.

Libplanet/Net/NetMQTransport.cs
```diff
@@ -947,19 +956,29 @@ private async Task InitializeTurnClient()
...
+                    return false;
+                }
+                else
+                {
+                    return _differentAppProtocolVersionEncountered(
+                        message.Remote, remoteVersion, _appProtocolVersion);
+                }
             }
+
+            return false;
```
earlbread, line 981: How about this?

```suggestion
            if (!(_trustedAppProtocolVersionSigners is null) &&
                !_trustedAppProtocolVersionSigners.Any(s => remoteVersion.Verify(s)))
            {
                return false;
            }

            if (_differentAppProtocolVersionEncountered is null)
            {
                return false;
            }

            return _differentAppProtocolVersionEncountered(
                message.Remote, remoteVersion, _appProtocolVersion);
```

Libplanet/Net/BoundPeer.cs
```diff
@@ -17,22 +17,18 @@ public sealed class BoundPeer : Peer
         /// <see cref="Peer"/>.</param>
         /// <param name="endPoint">A <see cref="DnsEndPoint"/> consisting of the
         /// host and port of the <see cref="Peer"/>.</param>
-        /// <param name="appProtocolVersion">An application protocol version
-        /// that the <see cref="Peer"/> is using.</param>
         public BoundPeer(
             PublicKey publicKey,
-            DnsEndPoint endPoint,
-            AppProtocolVersion appProtocolVersion)
```
earlbread, line 25: This should be added to the changelog.

Libplanet/Net/NetMQTransport.cs
```diff
@@ -947,19 +956,27 @@ private async Task InitializeTurnClient()
             }
         }
 
-        // FIXME: This method should be in Swarm<T>
-        private void ValidateSender(Peer peer)
+        private bool IsMessageValid(Message message)
```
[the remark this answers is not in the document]
limebell: According to [dotnet naming guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-properties), the member with boolean type is recommended to name that can be answered with "yes" or "no". So how about name it as `IsAppProtocolVersionValid()`?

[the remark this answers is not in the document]
longfin: That looks also good. 👍

limebell:
I moved version check progress to `Message.Parse()`, which will throw `DifferentAppProtocolVersionException` even the frame is changed a lot.

Libplanet.Tests/Libplanet.Tests.csproj
```diff
@@ -18,13 +16,11 @@
     <AdditionalFiles Include="..\stylecop.json" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'!='Mono' And
-                             '$(MSBuildVersion)'&gt;='16.3.0'">
+  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'!='Mono' And&#xD;&#xA;                             '$(MSBuildVersion)'&gt;='16.3.0'">
```
longfin, line 19: is this change needed?

Libplanet/Net/DifferentAppProtocolVersionException.cs
```diff
@@ -24,6 +27,7 @@ internal sealed class DifferentAppProtocolVersionException : Exception
         /// </param>
         public DifferentAppProtocolVersionException(
             string message,
+            byte[] identity,
```
longfin, line 30: I guess `identity` isn't useful because it's a temporary value for `NetMQRouter`.

limebell: I've added this parameter just for use in [here](https://github.com/planetarium/libplanet/pull/949/files#diff-9045fd925fccb8bca88db49a0ab95d0cR600-R608).

Libplanet.Tests/Libplanet.Tests.csproj
limebell: Added by mistake 😥. I'll fix it.

Libplanet/Net/DifferentAppProtocolVersionException.cs
longfin: 🆗

Libplanet/Net/Messages/Message.cs
```diff
@@ -94,26 +97,67 @@ internal enum MessageType : byte
...
                 throw new ArgumentException("Can't parse empty NetMQMessage.");
             }
 
-            // (reply == true)  [type, sign, peer, frames...]
-            // (reply == false) [identity, type, sign, peer, frames...]
-            int headerCount = reply ? 3 : 4;
+            // (reply == true)  [version, type, sign, peer, frames...]
+            // (reply == false) [identity, version, type, sign, peer, frames...]
+            int headerCount = reply ? CommonFrames : CommonFrames + 1;
+            var versionToken = raw[headerCount - 4].ConvertToString();
```
riemannulus, line 134: IMO If we use a constant here, it will be easy to make mistakes. if changed `CommonFrames`, easily forgot about change const number. Can we change like that?

```suggestion
            var versionToken = raw[headerCount - CommonFrames].ConvertToString();
```
or something more semantic variable name.

limebell: Then these lines:

https://github.com/planetarium/libplanet/blob/d46118902d7876c7ea20f25c7eedccf33677d9fa/Libplanet/Net/Messages/Message.cs#L161-L163

should be like this as well?

```
 var rawType = (MessageType)raw[headerCount - (CommonFrames - 1)].ConvertToInt32(); 
 var peer = raw[headerCount - (CommonFrames - 2)].ToByteArray(); 
 byte[] signature = raw[headerCount - (CommonFrames - 3)].ToByteArray(); 
```

riemannulus: I think it seems OK. any opinion? @libplanet

longfin: Or, it would be more readable if we slice the `raw` array after checking 'reply'.

```charp
byte[] remains  = (reply) ? raw.Skip(1).ToArray() : raw;

// using remains instead of raw
```

limebell: I've added `MessageFrame` enum in [`e8eacb83f1d6f9a96d363f791b38752bc781884d`](https://github.com/planetarium/libplanet/pull/949/files#diff-4008fb0720717f5ed3e957339adebf4dR120).

Issue #944, opened 2020-08-03
Make fungible asset values to support exponent (minor units)


dahlia:
In order to support [minor currency units (i.e., *exponent*)][1], fungible asset values should be represented as [`Decimal`][2] instead of [`BigInteger`][3].

[1]: https://en.wikipedia.org/wiki/ISO_4217#Treatment_of_minor_currency_units_(the_%22exponent%22)
[2]: https://docs.microsoft.com/en-us/dotnet/api/system.decimal
[3]: https://docs.microsoft.com/en-us/dotnet/api/system.numerics.biginteger

dahlia:
[.NET's `Decimal` apparently does not support arbitrary precision…][1] It might be better to define our own type (say `AssetValue`) which consists of [two fields (`BigInteger` & scale)][2].

[1]: https://stackoverflow.com/q/4523741/383405
[2]: https://stackoverflow.com/a/4524254/383405

PR #954, opened 2020-08-12
Represent fungible asset values as a distinct type and support minor units

dahlia:
This pull request adds `Currency.DecimalPlaces` property and `FungibleAssetValue` type to address the issue #944.

From this changes, fungible asset values are no more represented as `BigInteger`, which does not hold any decimal places, but `FungibleAssetValue` instead.  The introduced `FungibleAssetValue` type is essentially a pair of `(Currency, BigInteger)`.  It cannot mix heterogeneous currencies.  `FungibleAssetValue`s can be made like below:

```csharp
var gold = new Currency(ticker: "GOLD", decimalPlaces: 2, minter: null);
FungibleAssetValue value;

value = new FungibleAssetValue(gold, majorUnit: 123, minorUnit: 0);
// Or alternatively:
value = 123 * gold;
// Or:
value = gold * 123;
```

This patch also introduces a new property named `DecimalPlaces` to `Currency`.  This defines how precise can currency values be, like 2 decimal places for cents of US Dollar.  Continued from the above example, you can represent 123.45 GOLD like below:

```csharp
var v = new FungibleAssetValue(gold, 123, 45);
```

On the other hand, it does not allow more decimal places than it's specified in `DecimalPlaces`.  For example, the following code causes `ArgumentException`:

```csharp
new FungibleAssetValue(gold, 123, 456);
// ArgumentException: Since the currency GOLD (dfc2eebfd2bfb1fc31d921d665d9210b9ea8ebb3) allows upto 2 decimal places, the given minor unit 456 is too big.
```

`FungibleAssetValue` type mimics other integral types like `BigInteger`; it provides `+`, `-`, `*`, `%`, `.DivRem()`, `.Abs()`, `.Sign` and so on, except for division operator (`/`) which is bug-prone IMHO.  For the rationale, see also the XML comment on `FungibleAssetValue` struct.

In order to print the value, simply use `.ToString()` method or if you want only quantity without currency use `.GetQuantityString()` method.

Besides, this patch adds a new namespace `Libplanet.Assets` and `Currency` is also moved into that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ To be released.
  -  Added `IAction.RenderError()` and `IAction.UnrenderError()` methods.
     [[#860], [#875]]
  -  Added methods related fungible asset states to `IAccountStateDelta`:
-    [[#861], [#900]]
+    [[#861], [#900], [#954]]
      -  `UpdatedFungibleAssetsAccounts` property
      -  `MintAsset(Address, Currency, BigInteger)` method
      -  `TransferAsset(Address, Address, Currency, BigInteger)` method
@@ -118,12 +118,13 @@ To be released.
 ### Added APIs
 
  -  Added `Currency` struct.  [[#861], [#900], [#954]]
- -  Added `AccountBalanceGetter` delegate.  [[#861], [#900]]
+ -  Added `FungibleAssetValue` struct.  [[#861], [#944], [#954]]
+ -  Added `AccountBalanceGetter` delegate.  [[#861], [#900], [#954]]
  -  Added `TurnClient.BindProxies()` method. [[#756], [#868]]
  -  Added `ActionEvaluation.Exception` property.  [[#860], [[#875]]]
  -  Added `InvalidTxGenesisHashException` class.  [[#796], [#878]]
  -  Added `CurrencyPermissionException` class.  [[#861], [#900]]
- -  Added `InsufficientBalanceException` class.  [[#861], [#900]]
+ -  Added `InsufficientBalanceException` class.  [[#861], [#900], [#954]]
  -  Added `BlockChain<T>.GetBalance()` method.  [[#861], [#900]]
  -  Added `Block<T>.TotalDifficulty` property.  [[#666], [#917]]
  -  Added `SwarmOptions` class.  [[#926]]
@@ -135,12 +136,19 @@ To be released.
  -  Added `BlockHeader.PreEvaluationHash` property.  [[#931], [#935]]
  -  Added `HashDigest(ImmutableArray<byte>)` constructor.  [[#931], [#935]]
  -  Incomplete block states became able to be handled in more flexible way.
-    [[#929], [#934], [#946]]
+    [[#929], [#934], [#946], [#954]]
      -  Replaced `BlockChain<T>.GetState(Address, HashDigest<SHA256>?, bool)`
         method with `GetState(Address, HashDigest<SHA256>?, StateCompleter<T>)`
         method.  Specifying `completeStates: true` and `false` can be replaced
         by `stateCompleter: StateCompleters<T>.Recalculate` and
         `StateCompleters<T>.Reject`, respectively.
+     -  Replaced
+        `BlockChain<T>.GetBalance(Address, Currency, HashDigest<SHA256>?, bool)`
+        method with
+        `GetState(Address, Currency, HashDigest<SHA256>?, StateCompleter<T>)`
+        method.  Specifying `completeStates: true` and `false` can be replaced
+        by `stateCompleter: FungibleAssetStateCompleters<T>.Recalculate` and
+        `FungibleAssetStateCompleters<T>.Reject`, respectively.
      -  Added `StateCompleter<T>` delegate.
      -  Added `FungibleAssetStateCompleter<T>` delegate.
      -  Added `StateCompleterSet<T>` struct.
@@ -246,6 +254,7 @@ To be released.
 [#936]: https://github.com/planetarium/libplanet/pull/936
 [#940]: https://github.com/planetarium/libplanet/pull/940
 [#941]: https://github.com/planetarium/libplanet/pull/941
+[#944]: https://github.com/planetarium/libplanet/issues/944
 [#945]: https://github.com/planetarium/libplanet/pull/945
 [#946]: https://github.com/planetarium/libplanet/pull/946
 [#949]: https://github.com/planetarium/libplanet/pull/949
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -117,7 +117,7 @@ To be released.
 
 ### Added APIs
 
- -  Added `Currency` struct.  [[#861], [#900]]
+ -  Added `Currency` struct.  [[#861], [#900], [#954]]
  -  Added `AccountBalanceGetter` delegate.  [[#861], [#900]]
  -  Added `TurnClient.BindProxies()` method. [[#756], [#868]]
  -  Added `ActionEvaluation.Exception` property.  [[#860], [[#875]]]
@@ -250,6 +250,7 @@ To be released.
 [#946]: https://github.com/planetarium/libplanet/pull/946
 [#949]: https://github.com/planetarium/libplanet/pull/949
 [#950]: https://github.com/planetarium/libplanet/pull/950
+[#954]: https://github.com/planetarium/libplanet/pull/954
 [sleep mode]: https://en.wikipedia.org/wiki/Sleep_mode
 
 
--- a/Libplanet.Tests/Assets/FungibleAssetValueTest.cs
+++ b/Libplanet.Tests/Assets/FungibleAssetValueTest.cs
@@ -1,0 +1,306 @@
+#pragma warning disable S1764
+using System;
+using System.IO;
+using System.Runtime.Serialization.Formatters.Binary;
+using Libplanet.Assets;
+using Xunit;
+
+namespace Libplanet.Tests.Assets
+{
+    public class FungibleAssetValueTest
+    {
+        private static readonly Currency FOO = new Currency("FOO", minter: null);
+        private static readonly Currency BAR = new Currency("BAR", minter: null);
+
+        [Fact]
+        public void Equality()
+        {
+            FungibleAssetValue foo100a = new FungibleAssetValue(FOO, 100);
+            FungibleAssetValue foo100b = new FungibleAssetValue(FOO, 100);
+            FungibleAssetValue foo200a = new FungibleAssetValue(FOO, 200);
+            FungibleAssetValue foo200b = new FungibleAssetValue(FOO, 200);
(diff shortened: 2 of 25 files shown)


Libplanet/Assets/FungibleAssetValue.cs
```diff
@@ -0,0 +1,530 @@
...
+                );
+            }
+        }
+
+        /// <summary>
+        /// Creates a value of the <paramref name="currency"/> from the given <paramref
+        /// name="sign"/>, <paramref name="majorUnit"/> and <paramref name="minorUnit"/>.
+        /// </summary>
+        /// <param name="currency">The currency to create a value.</param>
+        /// <param name="sign">Indicates the sign (negative, positive, or zero) of the value.
```
longfin, line 75: We may need new enum for this.

Libplanet/Assets/FungibleAssetValue.cs
```diff
@@ -0,0 +1,530 @@
...
+        /// <param name="remainder">When this method returns (without any exception), the remainder
+        /// after dividing the value by <paramref name="divisor"/>.  This parameter is passed
+        /// uninitialized.</param>
+        /// <returns>The quotient of the division.</returns>
+        /// <exception cref="ArgumentException">Thrown when two values do not have the same
+        /// <see cref="Currency"/>.</exception>
+        /// <exception cref="DivideByZeroException">Thrown when the <paramref name="divisor"/> is
+        /// zero.</exception>
+        [Pure]
+        public BigInteger DivRem(FungibleAssetValue divisor, out FungibleAssetValue remainder)
```
longfin, line 417: Is this operation(`someFAV.DivRem(someOtherFAV, out FAV rem)`) really needed?

Libplanet/Assets/FungibleAssetValue.cs
```diff
@@ -0,0 +1,530 @@
...
+    /// <para>It behaves like numbers except for division operator (<c>/</c>) to prevent to forget
+    /// to handle its remainder; use <see cref="DivRem(FungibleAssetValue)"/> and <see
+    /// cref="DivRem(BigInteger)"/> methods instead.</para>
+    /// </summary>
+    [Serializable]
+    public readonly struct FungibleAssetValue :
+        IEquatable<FungibleAssetValue>,
+        IComparable<FungibleAssetValue>,
+        IComparable,
+        ISerializable
```
longfin, line 22: It would be better if we also adding [`IForamttable`](https://docs.microsoft.com/en-us/dotnet/api/system.iformattable?view=netstandard-2.0). (even next PR).

Libplanet/Blockchain/FungibleAssetStateCompleters.cs
```diff
@@ -40,6 +41,9 @@ public static class FungibleAssetStateCompleters<T>
             Currency currency
         ) =>
             (blockChain, hash) =>
-                (Bencodex.Types.Integer)stateCompleter(blockChain, hash, address, currency);
+            {
+                FungibleAssetValue balance = stateCompleter(blockChain, hash, address, currency);
+                return (Bencodex.Types.Integer)balance.RawValue;
```
longfin, line 46: Is it okay returning only `.RawValue` without `currency`?

Libplanet.Tests/Assets/FungibleAssetValueTest.cs
```diff
@@ -0,0 +1,306 @@
...
+        [Fact]
+        public void Negate()
+        {
+            FungibleAssetValue foo_3 = new FungibleAssetValue(FOO, -3);
+            FungibleAssetValue foo0 = new FungibleAssetValue(FOO);
+            FungibleAssetValue foo3 = new FungibleAssetValue(FOO, 3);
+
+            Assert.Equal(foo_3, -foo3);
+            Assert.Equal(foo3, -foo_3);
+            Assert.Equal(foo0, -foo0);
```
riemannulus, line 198: Have we offer negative? I wonder when it can be negative.

Libplanet/Assets/FungibleAssetValue.cs
dahlia: Avoiding a new enum type was intended, because:

- [`BigInteger.Sign` property][1] does the same thing.
- It is much handier than enum values when adding a sign to values, i.e., you can simply do that by `v.Sign * anotherValue` instead of `v.Sign == Sign.Negative ? -anotherValue : anotherValue`.

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.sign?view=netstandard-2.0

Libplanet/Assets/FungibleAssetValue.cs
dahlia: It's intended to be parallel with `BigInteger`… a overloaded method which returns a remainder through a tuple instead `out` parameter does not exist in `BigInteger` but I added here for convenience' sake.

Libplanet/Assets/FungibleAssetValue.cs
dahlia: I also wanted to implement `IFormattable` in this patch, but its specification seemed quite complex… I am going to implement it next time

Libplanet/Blockchain/FungibleAssetStateCompleters.cs
dahlia: Yes, raw state completer (`Func<BlockChain<T>, HashDigest<SHA256>, IValue>`) used by only `BlockChain<T>.GetRawState()` method, which is `internal`, takes `IValue` instead of `FungibleAssetValue`.

Libplanet.Tests/Assets/FungibleAssetValueTest.cs
dahlia: Sorry, I don't get it well what you mean… but if you mean “what does mean negative fungible asset values?” negative values purpose to represent debts.

CHANGES.md
```diff
@@ -11,7 +11,7 @@ To be released.
  -  Added `IAction.RenderError()` and `IAction.UnrenderError()` methods.
     [[#860], [#875]]
  -  Added methods related fungible asset states to `IAccountStateDelta`:
-    [[#861], [#900]]
+    [[#861], [#900], [#952]]
```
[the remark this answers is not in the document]
dahlia: Amended!

CHANGES.md
```diff
@@ -117,13 +117,14 @@ To be released.
...
- -  Added `AccountBalanceGetter` delegate.  [[#861], [#900]]
+ -  Added `Currency` struct.  [[#861], [#900], [#954]]
+ -  Added `FungibleAssetValue` struct.  [[#861], [#944], [#954]]
+ -  Added `AccountBalanceGetter` delegate.  [[#861], [#900], [#954]]
  -  Added `TurnClient.BindProxies()` method. [[#756], [#868]]
  -  Added `ActionEvaluation.Exception` property.  [[#860], [[#875]]]
  -  Added `InvalidTxGenesisHashException` class.  [[#796], [#878]]
  -  Added `CurrencyPermissionException` class.  [[#861], [#900]]
- -  Added `InsufficientBalanceException` class.  [[#861], [#900]]
+ -  Added `InsufficientBalanceException` class.  [[#861], [#900], [#952]]
```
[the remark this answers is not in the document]
dahlia: Amended!

CHANGES.md
```diff
@@ -135,12 +136,19 @@ To be released.
  -  Added `BlockHeader.PreEvaluationHash` property.  [[#931], [#935]]
  -  Added `HashDigest(ImmutableArray<byte>)` constructor.  [[#931], [#935]]
  -  Incomplete block states became able to be handled in more flexible way.
-    [[#929], [#934], [#946]]
+    [[#929], [#934], [#946], [#952]]
```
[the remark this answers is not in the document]
dahlia: Amended!

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1532,7 +1535,9 @@ bool buildStateReferences
                             new KeyValuePair<string, IValue>(
                                 ToFungibleAssetKey(pair),
                                 new Bencodex.Types.Integer(
-                                    lastStates?.GetBalance(pair.Item1, pair.Item2) ?? 0
+                                    lastStates is null
+                                        ? 0
+                                        : lastStates.GetBalance(pair.Item1, pair.Item2).RawValue
```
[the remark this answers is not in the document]
dahlia: Amended!

Libplanet/Action/IAccountStateDelta.cs
```diff
@@ -103,75 +103,67 @@ public interface IAccountStateDelta
         /// The <paramref name="address"/>'s balance of the <paramref name="currency"/>.
         /// </returns>
         [Pure]
-        BigInteger GetBalance(Address address, Currency currency);
+        FungibleAssetValue GetBalance(Address address, Currency currency);
 
         /// <summary>
-        /// Mints the <paramref name="amount"/> of a new <paramref name="currency"/>, fungible asset
-        /// (i.e., in-game monetary), and give it to the <paramref name="recipient"/>.
+        /// Mints thefungible asset <paramref name="value"/> (i.e., in-game monetary),
```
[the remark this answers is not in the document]
dahlia: Amended!

Libplanet/Assets/FungibleAssetValue.cs
longfin: I was wondering there are situations in which one FAV needs to be divided by another FAV (not a number). I guess we can cover most of cases by `.DivRem(BigInteger, out FungibleAssetValue)`.

Libplanet/Assets/Currency.cs
```diff
@@ -100,6 +128,38 @@ private Currency(SerializationInfo info, StreamingContext context)
...
+        /// <param name="quantity">The major unit of the fungible asset value,
+        /// i.e., digits <em>before</em> the decimal separator.</param>
+        /// <param name="currency">The currency to get a value.</param>
+        /// <returns>A fungible asset value with the given <paramref name="quantity"/> of the
+        /// specified <paramref name="currency"/>.</returns>
+        /// <remarks>This cannot specify <see cref="FungibleAssetValue.MinorUnit"/> but only
+        /// <see cref="FungibleAssetValue.MajorUnit"/>.  For more precision, directly use <see
+        /// cref="FungibleAssetValue"/>'s constructors instead.</remarks>
+        [Pure]
+        public static FungibleAssetValue operator *(BigInteger quantity, Currency currency) =>
```
limebell, line 160: Can you please explain why this returns `FungibleAssetValue` instead of `Currency`?

dahlia: Even though this diverts the multiplying operator (`*`), it's not an actual multiplication, but a shortcut for creating a new `FungibleAssetValue`: `123 * foo` is equivalent to `new FungibleAssetValue(currency: foo, major: 1, minor: 0)`.

PR #960, opened 2020-08-18
FungibleAssetValue.Parse() method

dahlia:
As the `FungibleAssetValue` type has never released yet, I omitted changelog for this.


dahlia:
> Docstring seems missing.

Fixed and amended!

PR #961, opened 2020-08-19
Fix the outdated branch name in the docs build script

dahlia:
Because of this, <https://docs.libplanet.io/main/> hadn't updated for several days.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,7 +104,7 @@ on GitHub consists of several projects:
 [Libplanet.RocksDBStore]: https://www.nuget.org/packages/Libplanet.RocksDBStore/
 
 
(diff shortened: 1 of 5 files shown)


moreal (approved):

limebell (approved):

longfin (dismissed):

longfin:
I guess we should update other things too. 

e.g. https://github.com/planetarium/libplanet/blob/e687b8d542e2c4bf9c60bd33505c52c1f01f0473/Libplanet.Tools/package.json#L47

longfin:
> I guess we should update other things too.
> 
> e.g.
> 
> https://github.com/planetarium/libplanet/blob/e687b8d542e2c4bf9c60bd33505c52c1f01f0473/Libplanet.Tools/package.json#L47

Fortunately this seems to be redirecting. but I think it would be better to fix it.

longfin (approved):

moreal (approved):

Issue #967, opened 2020-08-26
Let IRenderer<T> turn on and off evaluating actions


dahlia:
`BlockChain<T>`'s `render` option, replaced by `renderers` option (in the patch #963), purposed to prevent unnecessary action evaluation if rendering is not needed.  Although passing `renderers: null` or `renderers: Enumerable.Empty<IRenderer<T>>()` shares the same behavior to the previous usage `render: false`, two ways do not have the same performance.  Further, even if actions do not need to be rendered at all and only new tips need to be rendered (i.e., only `RenderBlock()` is used), `BlockChain<T>` anyway evaluates actions because it cannot be aware whether `{Render,Unrender}Action{,Error}()` methods are empty or has any code…

`IRenderer<T>` should be able to entirely turn off evaluating actions when it needs.  Here are my ideas:

1. Add `bool IRenderer<T>.RendersAction()` method.  However, it probably is prone to cause bugs like making `RendersAction()` returns `false` by mistake while filling out `{Render,Unrender}Action{,Error}()` methods.

2. Extract `IRenderer<T>.{Render,Unrender}Action{,Error}()` methods into a new interface like `IActionRenderer<T>`, and add `ActionRenderer` property to `IRenderer<T>`.  If `IRenderer<T>.ActionRenderer` is `null` it means actions entirely do not need to be evaluated at all.  The downside is that it makes API more complex than now.

@planetarium/libplanet What's your preference? Or any ideas?

longfin:
> Extract IRenderer<T>.{Render,Unrender}Action{,Error}() methods into a new interface like IActionRenderer<T>, and add ActionRenderer property to IRenderer<T>. If IRenderer<T>.ActionRenderer is null it means actions entirely do not need to be evaluated at all. The downside is that it makes API more complex than now.

I think it's better to separate action rendering and block events. also, I'm not sure if we need a hierarchy...

dahlia:
> I'm not sure if we need a hierarchy...

You mean to make `IRenderer<T>` interface to have only block/reorg render methods, and add a new interface something like `IActionRenderer<T>` which inherits `IRenderer<T>` and has other action render methods?

longfin:
> > I'm not sure if we need a hierarchy...
> 
> You mean to make `IRenderer<T>` interface to have only block/reorg render methods, and add a new interface something like `IActionRenderer<T>` which inherits `IRenderer<T>` and has other action render methods?

Right.

moreal:
[...]

PR #963, opened 2020-08-21
Unify all state change events on BlockChain<T> into IRenderer<T>

dahlia:
Closes #959.

This pull request introduces a new interface named `IRenderer<T>` and replaces rendering methods of `IAction` and `TipChanged`/`Reorged` events on `BlockChain<T>` with it.

To make `IAction` to satisfy [separation of concerns][1], the interface becomes no more aware of rendering, but only has `Execute()` and knows how to serialize and deserialize itself.  Instead, rendering mechanism is extracted into the new interface named `IRenderer<T>`, and it needs to be *injected* into `BlockChain<T>`.  In order to watch state changes via `IRenderer<T>`, `BlockChain<T>()` constructor now takes `renderers`, `IEnumerable<IRenderer<T>>`:

```csharp
public class SampleRenderer : IRenderer<SampleAction>
{
        public void RenderAction(
            IAction action,
            IActionContext context,
            IAccountStateDelta nextStates
        ) =>
            RENDER_THINGS_HERE;
        // …
}
IRenderer<SampleAction> renderer = new SampleRenderer();
var chain = new BlockChain<SampleAction>(..., renderers: new[] { renderer });
```

Besides rendering `IAction`s, `IRenderer<T>` also replaces `BlockChain<T>.TipChanged` event (introduced in #517 & #526) with `IRenderer<T>.RenderBlock()` method, and `BlockChain<T>.Reorged` event (introduced in #945) with `IRenderer<T>.RenderReorg()` method.  It means now all events regarding state changes on `BlockChain<T>` become unified into one interface `IRenderer<T>`.

(FYI I actually left `BlockChain<T>.TipChanged` event `private` for an internal use: cancelling undergoing mining when a new block is appended.  However, still it is invisible from public surface.)

This patch also removed the `render` option from the `BlockChain<T>()` constructor (introduced in #883).  As `renderers: null` is equivalent to `render: false`, it's redundant.

For the sake of convenient development (writing unit tests in particular), I added two built-in renderers as well:

- `AnonymousRenderer<T>` mimics Java's anonymous class for implementing `IRenderer<T>`.  If you need only one-use renderer, it helps you to avoid a tedious inner class:

    ```csharp
    var renderer = new AnonymousRenderer<SampleAction>
    {
        RenderAction = (action, context, nextStates) => RENDER_THINGS_HERE,
    };
    ```

[...]

PR #883, opened 2020-05-28
Add render option to BlockChain<T>() constructor

earlbread:
This adds `allowRender` option to `BlockChain<T>`.

PR #970, opened 2020-08-27
Short-circuit evaluation of stale actions on reorg

dahlia:
This addresses #967.

The previous patch #963 unified two types of events into one: one is for rendering/unrendering actions, and other one is for watching new tips & reorgs.  At the same time, it also threw away the `BlockChain<T>()` constructor's Boolean option `render` because I thought it's redundant as `render: false` seemed equivalent to `renderers: null`.

However, those two changes regressed the performance of reorg, because the optimization made by the patch #883 was thrown away together.  The disappeared optimization was a short-circuit evaluation option for stale actions to be unrendered.  As the option `render: false` was removed, stale actions became evaluated even if there is no renderers, or all renderers do nothing for `UnrenderAction()`/`UnrenderActionError()` events…

In order to fix this performance regression, I reverted (with some adjusts and new name) `ShortCircuitActionEvaluationForUnrenderWithNoActionRenderers` (was `Render`) test, I removed in the previous patch #963, and refactored the renderer interface(s) according to the conclusion made in the issue #967.  Now we have a new interface named `IActionRenderer<T>`, is basically equivalent to the previous `IRenderer<T>`, and `IRenderer<T>` became to have only block-related events.  `BlockChain<T>.Swap()` now omits evaluation of stale actions if there is no `IActionRenderer<T>` at all.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -143,6 +143,7 @@ namespace Libplanet.Blockchain
             Renderers = renderers is IEnumerable<IRenderer<T>> r
                 ? r.ToImmutableArray()
                 : ImmutableArray<IRenderer<T>>.Empty;
+            ActionRenderers = Renderers.OfType<IActionRenderer<T>>().ToImmutableArray();
             _rwlock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
             _txLock = new object();
 
@@ -199,6 +200,12 @@ namespace Libplanet.Blockchain
         /// constructor instead.
         /// </remarks>
         public IImmutableList<IRenderer<T>> Renderers { get; }
+
+        /// <summary>
+        /// A filtered list, from <see cref="Renderers"/>, which contains only <see
+        /// cref="IActionRenderer{T}"/> instances.
+        /// </summary>
+        public IImmutableList<IActionRenderer<T>> ActionRenderers { get; }
 
         public IBlockPolicy<T> Policy { get; }
 
@@ -974,7 +981,7 @@ namespace Libplanet.Blockchain
             {
                 if (evaluation.Exception is null)
                 {
-                    foreach (IRenderer<T> renderer in Renderers)
+                    foreach (IActionRenderer<T> renderer in ActionRenderers)
                     {
                         renderer.RenderAction(
                             evaluation.Action,
@@ -985,7 +992,7 @@ namespace Libplanet.Blockchain
                 }
                 else
                 {
-                    foreach (IRenderer<T> renderer in Renderers)
+                    foreach (IActionRenderer<T> renderer in ActionRenderers)
                     {
                         renderer.RenderActionError(
                             evaluation.Action,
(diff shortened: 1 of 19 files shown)


longfin (approved):

Libplanet/Blockchain/Renderers/LoggedRenderer.cs
```diff
@@ -27,11 +27,14 @@ namespace Libplanet.Blockchain.Renderers
     /// );
     /// </code>
     /// </example>
+    /// <remarks>Since <see cref="IActionRenderer{T}"/> is a subtype of <see cref="IRenderer{T}"/>,
+    /// <see cref="LoggedRenderer(IRenderer{T}, ILogger, LogEventLevel)"/> constructor can take
```
earlbread, line 31:
```suggestion
    /// <see cref="LoggedRenderer{T}(IRenderer{T}, ILogger, LogEventLevel)"/> constructor can take
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -200,6 +201,12 @@ IEnumerable<IRenderer<T>> renderers
         /// </remarks>
         public IImmutableList<IRenderer<T>> Renderers { get; }
 
+        /// <summary>
+        /// A flitered list, from <see cref="Renderers"/>, which contains only <see
```
earlbread, line 205:
```suggestion
        /// A filtered list, from <see cref="Renderers"/>, which contains only <see
```

Libplanet/Blockchain/Renderers/IActionRenderer.cs
```diff
@@ -0,0 +1,104 @@
+#nullable enable
+using System;
+using Libplanet.Action;
+
+namespace Libplanet.Blockchain.Renderers
+{
+    /// <summary>
+    /// Listens state changes of every step of actions, besides blocks, on a
+    /// <see cref="BlockChain{T}"/>.  If you need more fine-grained events than
+    /// <see cref="IRenderer{T}"/> implement this interface instead.
```
moreal, line 10: Doesn't it need `,` after *if* usage?

```suggestion
    /// <see cref="IRenderer{T}"/>, implement this interface instead.
```

Libplanet/Blockchain/Renderers/LoggedRenderer.cs
dahlia: Amended!

Libplanet/Blockchain/BlockChain.cs
dahlia: Amended!

dahlia:
@planetarium/libplanet Applied suggestions and rebased on the up-to-date main.  Please review this again!

earlbread (approved):

Libplanet/Blockchain/Renderers/IActionRenderer.cs
dahlia: Amended!

dahlia:
@earlbread @moreal Fixed a typo. Please review this again.

earlbread (approved):

moreal (approved):

limebell (approved):

riemannulus (approved):

dahlia:
@planetarium/libplanet Rebased again!

earlbread (approved):

longfin (approved):

PR #971, opened 2020-08-27
Configure purge-artifacts

longfin:
It introduces [purge-artifacts](https://github.com/kolpav/purge-artifacts-action) to cleanup old artifacts.

### DIFF
--- a/.github/workflows/delete-old-artifacts.yml
+++ b/.github/workflows/delete-old-artifacts.yml
@@ -1,0 +1,13 @@
+name: 'Delete old artifacts'
+on:
+  schedule:
+    - cron: '0 15 * * *' # every day
+
+jobs:
+  delete-artifacts:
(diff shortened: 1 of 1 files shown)


.github/workflows/delete-old-artifacts.yml
```diff
@@ -0,0 +1,13 @@
+name: 'Delete old artifacts'
+on:
+  schedule:
+    - cron: '0 15 * * *' # every day
```
dahlia, line 4: As we have nightly builds and artifacts are useful for debugging deployment scripts, how about wait longer days before delete old artifacts, like a week?

dahlia: Hmm… I misunderstood the code.  It seems to already work as I said.

dahlia (approved):

moreal (approved):

Issue #976, opened 2020-08-28
Libplanet assembly takes too long time to be built


dahlia:
From a [CI build job][1]:

<pre>2020-08-28T04:27:13.2045031Z 
2020-08-28T04:27:14.8916945Z   Restore completed in 637.47 ms for /home/vsts/work/1/s/Libplanet.Benchmarks/Libplanet.Benchmarks.csproj.
2020-08-28T04:27:14.8918392Z   Restore completed in 637.46 ms for /home/vsts/work/1/s/Libplanet.RocksDBStore.Tests/Libplanet.RocksDBStore.Tests.csproj.
2020-08-28T04:27:15.1355159Z   Restore completed in 177.04 ms for /home/vsts/work/1/s/Libplanet.RocksDBStore/Libplanet.RocksDBStore.csproj.
2020-08-28T04:27:15.1361016Z   Restore completed in 216.97 ms for /home/vsts/work/1/s/Libplanet.Stun.Tests/Libplanet.Stun.Tests.csproj.
2020-08-28T04:27:15.4326890Z   Restore completed in 87.07 ms for /home/vsts/work/1/s/Libplanet.Stun/Libplanet.Stun.csproj.
2020-08-28T04:27:15.4328468Z   Restore completed in 174.12 ms for /home/vsts/work/1/s/Libplanet.Tests/Libplanet.Tests.csproj.
2020-08-28T04:27:15.8423725Z   Restore completed in 230.12 ms for /home/vsts/work/1/s/Libplanet.Tools/Libplanet.Tools.csproj.
2020-08-28T04:27:15.9304778Z   Restore completed in 329.39 ms for /home/vsts/work/1/s/Libplanet/Libplanet.csproj.
2020-08-28T04:27:23.9474387Z   Libplanet.Stun -> /home/vsts/work/1/s/Libplanet.Stun/bin/Release/netstandard2.0/Libplanet.Stun.dll
2020-08-28T04:27:23.9579101Z   Libplanet.Stun -> /home/vsts/work/1/s/Libplanet.Stun/bin/Release/netstandard2.1/Libplanet.Stun.dll
2020-08-28T04:27:26.0882909Z   Libplanet.Stun.Tests -> /home/vsts/work/1/s/Libplanet.Stun.Tests/bin/Release/netcoreapp3.0/Libplanet.Stun.Tests.dll
[...]

PR #977, opened 2020-08-28
Speed up build time

earlbread:
This limits the target framework and updates the SonarAnalyzer version to speed up build time.

### DIFF
--- a/Libplanet.Tests/Libplanet.Tests.csproj
+++ b/Libplanet.Tests/Libplanet.Tests.csproj
@@ -38,7 +38,7 @@
     <PackageReference Include="Serilog.Enrichers.Thread" Version="3.0.0" />
     <PackageReference Include="Serilog.Sinks.TestCorrelator" Version="3.2.0" />
     <PackageReference Include="Serilog.Sinks.XUnit" Version="1.0.7" />
-    <PackageReference Include="SonarAnalyzer.CSharp" Version="8.1.0.13383">
+    <PackageReference Include="SonarAnalyzer.CSharp" Version="8.12.0.21095">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
--- a/Libplanet.Tests/Libplanet.Tests.csproj
+++ b/Libplanet.Tests/Libplanet.Tests.csproj
@@ -1,6 +1,4 @@
-<Project
-  xmlns="http://schemas.microsoft.com/developer/msbuild/2003"
-  Sdk="Microsoft.NET.Sdk">
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netcoreapp2.2</TargetFramework>
     <IsPackable>false</IsPackable>
@@ -18,13 +16,11 @@
     <AdditionalFiles Include="..\stylecop.json" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'!='Mono' And
-                             '$(MSBuildVersion)'&gt;='16.3.0'">
+  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'!='Mono' And '$(MSBuildVersion)'&gt;='16.3.0'">
     <TargetFramework>netcoreapp3.0</TargetFramework>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'=='Mono' And
-                             '$(BuildingByReSharper)'!='true'">
+  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'=='Mono' And '$(BuildingByReSharper)'!='true'">
     <TargetFramework>net47</TargetFramework>
   </PropertyGroup>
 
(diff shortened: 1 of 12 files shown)


moreal (approved):

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -1371,7 +1371,7 @@ public async Task RemoveForkedChainWhenFillBlocksAsyncFail()
             }
         }
 
-        [Fact]
+        [Fact(Timeout=Timeout)]
```
longfin, line 1374: For consistency 😅 

```suggestion
        [Fact(Timeout = Timeout)]
```

earlbread: Amended!

limebell (approved):

Libplanet.Tests/Libplanet.Tests.csproj
```diff
@@ -18,13 +16,11 @@
...
-                             '$(MSBuildVersion)'&gt;='16.3.0'">
+  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'!='Mono' And '$(MSBuildVersion)'&gt;='16.3.0'">
     <TargetFramework>netcoreapp3.0</TargetFramework>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'=='Mono' And
-                             '$(BuildingByReSharper)'!='true'">
+  <PropertyGroup Condition=" '$(MSBuildRuntimeType)'=='Mono' And '$(BuildingByReSharper)'!='true'">
     <TargetFramework>net47</TargetFramework>
   </PropertyGroup>
```
dahlia, line 25: As Rider became based on .NET Core (was Mono before), I guess we don't need this `<PropertyGroup>` anymore… 🤔

Libplanet/Libplanet.csproj
```diff
@@ -36,7 +36,7 @@ https://docs.libplanet.io/</Description>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;netcoreapp3.1</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
dahlia, line 39: IIRC these multiple target frameworks purpose to derive benefits for .NET Core programs, especially for lib9c & NineChronicles.Standalone.

dahlia (commented):
Anyway, even if we shorten its build time almost half, it still takes more than a minute, whereas other assemblies take only about 10 seconds… Would there be more radical approach to solve #976?

Libplanet.Tests/Libplanet.Tests.csproj
earlbread: Isn't the `<PropertyGroup>` for Mono testing?

Libplanet/Libplanet.csproj
earlbread: Even if only netstandard2.0 is provided, it seems that it can be used in other projects.

earlbread:
> Anyway, even if we shorten its build time almost half, it still takes more than a minute, whereas other assemblies take only about 10 seconds… Would there be more radical approach to solve #976?

If we don't use the Sonar analyzer, the build takes only a few seconds. We need to discuss whether to continue using Sonar.

riemannulus (approved):

longfin (approved):

Libplanet.Tests/Libplanet.Tests.csproj
dahlia: IIRC the second condition `'$(BuildingByReSharper)'!='true` purposed to be *not* run on Rider, because at that time under Rider `$(MSBuildRuntimeType)` was set to `'Mono'`.

Libplanet/Libplanet.csproj
dahlia: Yeah, .NET Standard 2.0 is compatible with both Mono and .NET Core, because that's why .NET Standard exists, but at the same time [.NET Standard 2.0 assemblies are optimized for neither Mono nor .NET Core][1]. For example, [*Microsoft.Bcl.AsyncInterfaces*][2], which is one of our indirect dependencies, is basically not necessary for .NET Standard 2.1 or .NET Core 3 as they have built-in `IAsyncEnumerable<T>` & others, so using `IAsyncEnumerable<T>` & others in the NuGet version of *Microsoft.Bcl.AsyncInterfaces* has performance impact in these newer runtimes AFAIK.

[1]: https://andrewlock.net/converting-a-netstandard-2-library-to-netcore-3/#upgrading-a-net-standard-2-0-library-to-net-core-3-is-it-necessary-
[2]: https://www.nuget.org/packages/System.Linq.Async

limebell:
> > Anyway, even if we shorten its build time almost half, it still takes more than a minute, whereas other assemblies take only about 10 seconds… Would there be more radical approach to solve #976?
> 
> If we don't use the Sonar analyzer, the build takes only a few seconds. We need to discuss whether to continue using Sonar.

Can we have build option to use Sonar analyzer or not? We can develop easily on local by turn off sonar analyzer, and in CI's build turn on Sonar analyzer.

dahlia:
> If we don't use the Sonar analyzer, the build takes only a few seconds. We need to discuss whether to continue using Sonar.

One of problems is that we conduct a check of Sonar Analyzer for *every concurrent build job*.  How about temporarily turning off the check in the CI build jobs with leaving only one build job, so that the check is conducted only once through a build?

earlbread:
> > If we don't use the Sonar analyzer, the build takes only a few seconds. We need to discuss whether to continue using Sonar.
> 
> One of problems is that we conduct a check of Sonar Analyzer for _every concurrent build job_. How about temporarily turning off the check in the CI build jobs with leaving only one build job, so that the check is conducted only once through a build?

To try this way I made #979. Please take a look.

PR #981, opened 2020-09-02
Find specific peer from network

limebell:
There was `Swarm<T>.FindSpecificPeerAsync()` method but it was not precise. This PR makes it to work more precisely.

### DIFF
--- a/Libplanet/Net/Protocols/KademliaProtocol.cs
+++ b/Libplanet/Net/Protocols/KademliaProtocol.cs
@@ -363,7 +363,7 @@ namespace Libplanet.Net.Protocols
                 history.Add(viaPeer);
             }
 
-            BoundPeer specificPeerFound = await ProcessFoundForSpecificAsync(
+            return await ProcessFoundForSpecificAsync(
                 history,
                 found,
                 target,
@@ -371,8 +371,6 @@ namespace Libplanet.Net.Protocols
                 timeout,
                 cancellationToken
             );
-
-            return specificPeerFound;
         }
 
         internal async Task PingAsync(
@@ -537,27 +535,12 @@ namespace Libplanet.Net.Protocols
             List<BoundPeer> neighbors = _routing.Neighbors(target, _bucketSize, false).ToList();
             var found = new List<BoundPeer>();
             int count = Math.Min(neighbors.Count, _findConcurrency);
-            var timeoutOccurred = true;
             for (var i = 0; i < count; i++)
             {
-                try
-                {
-                    var peers =
-                        await GetNeighbors(neighbors[i], target, timeout, cancellationToken);
-                    history.Add(neighbors[i]);
-                    found.AddRange(peers.Where(peer => !found.Contains(peer)));
-                    timeoutOccurred = false;
-                }
-                catch (TimeoutException)
-                {
-                }
-            }
-
-            if (count != 0 && timeoutOccurred)
-            {
-                _logger.Debug($"Timeout occurred during {nameof(QueryNeighborsAsync)}.");
-                throw new TimeoutException(
-                    $"Timeout occurred during {nameof(QueryNeighborsAsync)}.");
+                var peers =
+                    await GetNeighbors(neighbors[i], target, timeout, cancellationToken);
+                history.Add(neighbors[i]);
+                found.AddRange(peers.Where(peer => !found.Contains(peer)));
             }
 
             return found;
@@ -586,7 +569,7 @@ namespace Libplanet.Net.Protocols
             catch (TimeoutException)
             {
                 RemovePeer(addressee);
-                throw;
+                return ImmutableArray<BoundPeer>.Empty;
             }
         }
 
--- a/Libplanet/Net/Protocols/KademliaProtocol.cs
+++ b/Libplanet/Net/Protocols/KademliaProtocol.cs
@@ -100,7 +100,7 @@ namespace Libplanet.Net.Protocols
                 {
                     _logger.Error("Version is different from seed peer.");
                 }
-                catch (TimeoutException)
+                catch (PingTimeoutException)
                 {
                     _logger.Error("A timeout exception occurred connecting to seed peer.");
                     RemovePeer(peer);
@@ -246,7 +246,7 @@ namespace Libplanet.Net.Protocols
 
                         await PingAsync(replacement, _requestTimeout, cancellationToken);
                     }
-                    catch (TimeoutException)
+                    catch (PingTimeoutException)
                     {
                         _logger.Debug(
                             "Remove stale peer {Peer} from replacement cache.",
@@ -399,7 +399,9 @@ namespace Libplanet.Net.Protocols
             }
             catch (TimeoutException)
             {
-                throw new TimeoutException($"Timeout occurred during {nameof(PingAsync)}().");
+                throw new PingTimeoutException(
+                    target,
+                    $"Timeout occurred during dial to {target}.");
             }
             catch (DifferentAppProtocolVersionException)
             {
@@ -429,7 +431,7 @@ namespace Libplanet.Net.Protocols
                 await PingAsync(peer, timeout, cancellationToken);
                 _routing.Check(peer, check, DateTimeOffset.UtcNow);
             }
-            catch (TimeoutException)
+            catch (PingTimeoutException)
             {
                 _logger.Debug("Peer {Peer} is invalid, removing...", peer);
                 RemovePeer(peer);
@@ -634,14 +636,24 @@ namespace Libplanet.Net.Protocols
             {
                 await Task.WhenAll(awaitables);
             }
-            catch (Exception e)
+            catch (AggregateException ae)
             {
-                // FIXME: If timeout occurred, the peer should be removed from the peers list.
-                _logger.Error(
-                    e,
-                    "Some responses from neighbors found unexpectedly terminated: {Exception}",
-                    e
-                );
+                foreach (var e in ae.InnerExceptions)
+                {
+                    if (e is PingTimeoutException pte)
+                    {
+                        // Remove timed out peers from the list.
+                        peers.Remove(pte.Target);
+                    }
+                    else
+                    {
+                        _logger.Error(
+                            e,
+                            "Some responses from neighbors found unexpectedly terminated: {e}",
+                            e
+                        );
+                    }
+                }
             }
 
             var findPeerTasks = new List<Task>();
@@ -708,10 +720,6 @@ namespace Libplanet.Net.Protocols
             }
 
             peers = Kademlia.SortByDistance(peers, target);
-            if (peers[0].Address.Equals(target))
-            {
-                return peers[0];
-            }
 
             Task[] awaitables = peers.Select(peer =>
                 PingAsync(peer, _requestTimeout, cancellationToken)
@@ -720,14 +728,29 @@ namespace Libplanet.Net.Protocols
             {
                 await Task.WhenAll(awaitables);
             }
-            catch (Exception e)
+            catch (AggregateException ae)
             {
-                // FIXME: If timeout occurred, the peer should be removed from the peers list.
-                _logger.Error(
-                    e,
-                    "Some responses from neighbors found unexpectedly terminated: {Exception}",
-                    e
-                );
+                foreach (var e in ae.InnerExceptions)
+                {
+                    if (e is PingTimeoutException pte)
+                    {
+                        // Remove timed out peers from the list.
+                        peers.Remove(pte.Target);
+                    }
+                    else
+                    {
+                        _logger.Error(
+                            e,
+                            "Some responses from neighbors found unexpectedly terminated: {e}",
+                            e
+                        );
+                    }
+                }
+            }
+
+            if (peers[0].Address.Equals(target))
+            {
+                return peers[0];
             }
 
             var findNeighboursTasks = new List<Task<BoundPeer>>();
--- a/Libplanet/Net/Protocols/KademliaProtocol.cs
+++ b/Libplanet/Net/Protocols/KademliaProtocol.cs
@@ -320,7 +320,7 @@ namespace Libplanet.Net.Protocols
                 target,
                 depth);
 
-            if (_routing.ContainsAddress(target) is BoundPeer boundPeer)
+            if (_routing.GetPeer(target) is BoundPeer boundPeer)
             {
                 try
                 {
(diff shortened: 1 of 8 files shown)


Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -294,28 +294,25 @@ public string Trace()
...
         /// message. If null, selects 3 <see cref="Peer"/>s from <see cref="RoutingTable"/> of
         /// self.</param>
         /// <param name="depth">Target depth of recursive operation.</param>
-        /// <param name="searchTarget">The address to search for among peers.</param>
         /// <param name="timeout"><see cref="TimeSpan"/> for waiting reply of
         /// <see cref="FindNeighbors"/>.</param>
         /// <param name="cancellationToken">A cancellation token used to propagate notification
         /// that this operation should be canceled.</param>
-        /// <returns>BoundPeer containing the <paramref name="searchTarget"/>'s address.</returns>
+        /// <returns>A BoundPeer with <see cref="Address"/> of <paramref name="target"/>.</returns>
```
dahlia, line 310:
```suggestion
        /// <returns>A <see cref="BoundPeer"/> with <see cref="Address"/> of
        /// <paramref name="target"/>.</returns>
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -328,6 +325,12 @@ public string Trace()
 
             if (history is null)
             {
+                if (_routing.ContainsAddress(target) is BoundPeer boundPeer)
+                {
+                    _logger.Debug("Target peer was in routing table. {BoundPeer}", boundPeer);
```
dahlia, line 330:
```suggestion
                    _logger.Debug("{BoundPeer}, a target peer, is in the routing table.", boundPeer);
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -684,22 +690,22 @@ private void ReceivePing(Ping ping)
...
         {
-            BoundPeer peerFound = null;
             List<BoundPeer> peers = found.Where(
                 peer => !peer.Address.Equals(_address)).ToList();
 
             if (peers.Count == 0)
             {
                 _logger.Debug("No any neighbor received.");
-                return peerFound;
+                return null;
```
dahlia, line 701: Would it be difficult to have a unit test for this case?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -684,22 +690,22 @@ private void ReceivePing(Ping ping)
...
-            Address searchAddress)
+            CancellationToken cancellationToken)
         {
-            BoundPeer peerFound = null;
             List<BoundPeer> peers = found.Where(
                 peer => !peer.Address.Equals(_address)).ToList();
 
             if (peers.Count == 0)
             {
                 _logger.Debug("No any neighbor received.");
```
dahlia, line 700:
```suggestion
                _logger.Debug("Received no neighbors at all.");
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -708,53 +714,35 @@ private void ReceivePing(Ping ping)
             {
                 await Task.WhenAll(awaitables);
             }
-            catch (AggregateException e)
+            catch (Exception e)
```
dahlia, line 717: @limebell What kind of exceptions do you expect to be thrown here?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -708,53 +714,35 @@ private void ReceivePing(Ping ping)
...
             {
-                if (string.CompareOrdinal(
-                        closestNeighbors[i].Address.ToHex(),
-                        searchAddress.ToHex()
-                   ) == 0 && _routing.Contains(closestNeighbors[i]))
+                if (history.Contains(peer))
                 {
-                    peerFound = closestNeighbors[i];
-                    return peerFound;
+                    continue;
```
dahlia, line 733: Would it be difficult to have a unit test for this case?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -708,53 +714,35 @@ private void ReceivePing(Ping ping)
...
                 {
-                    findNeighboursTasks.Add(FindSpecificPeerAsync(
-                        history,
-                        target,
-                        peers[i],
-                        (depth == -1) ? depth : depth - 1,
-                        searchAddress,
-                        timeout,
-                        cancellationToken));
+                    break;
```
dahlia, line 745: Would it be difficult to have a unit test for this case?

dahlia (commented):
Instead of directly referring to the constant `Kademlia.FindConcurrency` from `KademliaProtocol`, how about making `KademliaProtocol()` constructor to take that value?

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: Usually `TimeoutException`, which occurs when the received peer is not alive actually.

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: I guess we have to make `ProcessFoundForSpecificAsync` method public (or internal) to make unit test for this case. Although, we can make test to cover this line for this case, but with no meaning 😥

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -328,6 +329,14 @@ public string Trace()
...
+                if (_routing.ContainsAddress(target) is BoundPeer boundPeer)
+                {
+                    _logger.Debug(
+                        "{BoundPeer}, a target peer, is in the routing table.",
+                        boundPeer);
+                    return boundPeer;
+                }
+
                 history = new ConcurrentBag<BoundPeer>();
             }
```
earlbread, line 341: How about this?
```suggestion
            if (history is null && _routing.ContainsAddress(target) is BoundPeer boundPeer)
            {
                _logger.Debug(
                    "{BoundPeer}, a target peer, is in the routing table.",
                    boundPeer);
                return boundPeer;
            }

            history ??= new ConcurrentBag<BoundPeer>();
```

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -184,10 +208,17 @@ public IEnumerable<BoundPeer> Neighbors(Address target, int k)
...
             var peers = new List<BoundPeer>();
-            foreach (var peer in sorted.Where(peer => !peer.Address.Equals(target)))
+
+            // Select maximum k * 2 peers excluding the target itself.
+            bool containsTarget = sorted.Any(peer => peer.Address.Equals(target));
+            int maxCount = (includeTarget && containsTarget) ? k * 2 + 1 : k * 2;
+
+            foreach (var peer in includeTarget
+                ? sorted
+                : sorted.Where(peer => !peer.Address.Equals(target)))
```
earlbread, line 218: How about changing it like this?
```
            var peers = includeTarget
                ? sorted
                : sorted.Where(peer => !peer.Address.Equals(target));

            return peers.Take(maxCount);
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -708,53 +720,35 @@ private void ReceivePing(Ping ping)
...
                     e,
                     "Some responses from neighbors found unexpectedly terminated: {Exception}",
                     e
                 );
             }
 
-            for (int i = 0; i < closestNeighbors.Count; i++)
+            var findNeighboursTasks = new List<Task<BoundPeer>>();
+            var count = 0;
+            foreach (var peer in peers)
```
earlbread, line 735: I think we can omit `continue` statement below like this:
```
            foreach (var peer in peers.Where(peer => !history.Contains(peer)))
```

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: This if statement was not appropriate for the purpose of this method... So I decided to delete this condition.

earlbread (dismissed):

Libplanet/Net/Swarm.cs
```diff
@@ -926,20 +926,32 @@ await foreach (var pair in completedBlocks)
...
+        /// will recursive until the closest <see cref="BoundPeer"/> to the
+        /// <paramref name="target"/> is found.</param>
+        /// <param name="timeout">
+        /// <see cref="TimeSpan"/> for waiting reply of <see cref="FindNeighbors"/>.
+        /// If <c>null</c> is given, <see cref="TimeoutException"/> will not be thrown.
+        /// </param>
+        /// <param name="cancellationToken">A cancellation token used to propagate notification
+        /// that this operation should be canceled.</param>
+        /// <returns>A BoundPeer with <see cref="Address"/> of <paramref name="target"/>.
+        /// Returns <c>null</c> if the peer with address does not exist.</returns>
```
dahlia, line 944:
```suggestion
        /// <returns>A <see cref="BoundPeer"/> with <see cref="Address"/> of <paramref
        /// name="target"/>. Returns <c>null</c> if the peer with address does not exist.</returns>
```

dahlia (dismissed):

limebell:
Method implemented in this PR keeps searching even the specific target is found. I will add some commits on this PR to fix the issue. Until then, leave title as WIP.

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: I made `TimeoutException` (which is replaced by `PingTimeoutException`) to remove peer from lookup candidate, which will be covered by `SwarmTest.FindSpecificPeerAsyncFail()` test.

limebell:
Used DFS beforehand, may cause stack overflow. So I changed it to BFS by using concurrent queue and stop immediately when target peer is found. However `KademliaProtocol.FindSpecificPeer()` became to search only one peer at a time, so I'll increase concurrency in next PR.

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -294,65 +297,105 @@ public string Trace()
...
+                    "{BoundPeer}, a target peer, is in the routing table.",
+                    boundPeer);
+                return boundPeer;
             }
 
-            IEnumerable<BoundPeer> found;
-            if (viaPeer is null)
+            var history = new ConcurrentBag<BoundPeer>();
+            var peersToFind = new ConcurrentQueue<BoundPeer>();
+            foreach (var peer in _routing.Neighbors(target, _findConcurrency, false))
```
longfin, line 345:
```suggestion
            foreach (BoundPeer peer in _routing.Neighbors(target, _findConcurrency, false))
```

Libplanet/Net/Protocols/PingTimeoutException.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            Target = target;
+        }
+
+        protected PingTimeoutException(
+            SerializationInfo info,
+            StreamingContext context
+        )
+            : base(info, context)
+        {
```
longfin, line 32: We need to restore `Target` from `info` here. (therefore, we also need to define `GetObjectData()` to serialize it.)

Libplanet/Net/Protocols/RoutingTable.cs
```diff
@@ -160,6 +160,12 @@ public bool Contains(BoundPeer peer)
             return BucketOf(peer).Contains(peer);
         }
 
+        public BoundPeer ContainsAddress(Address addr) =>
```
longfin, line 163: In C#, methods starting with `Contains` tend to return `bool`. how about renaming like `GetPeer()` (or `TryGetPeer()` and using out var).

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -294,65 +297,105 @@ public string Trace()
...
-            else
+
+            while (peersToFind.Any())
             {
-                found = await GetNeighbors(viaPeer, target, timeout, cancellationToken);
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    throw new TaskCanceledException(
+                        $"Task is cancelled during {nameof(FindSpecificPeerAsync)}()");
+                }
```
longfin, line 356:
```suggestion
                cancellationToken.ThrowIfCancellationRequested();
```

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -294,65 +297,105 @@ public string Trace()
...
+                        }
+                    }
+                    catch (TaskCanceledException)
+                    {
+                        throw new TaskCanceledException(
+                            $"Task is cancelled during {nameof(FindSpecificPeerAsync)}()");
+                    }
+                    catch (Exception)
+                    {
+                        continue;
```
longfin, line 393: is it okay we ignore here without any logging?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -294,65 +297,105 @@ public string Trace()
...
-                target,
-                depth,
-                timeout,
-                cancellationToken,
-                searchTarget
-            );
+                int count = 0;
+                foreach (var found in foundPeers
+                    .Where(peer => !history.Contains(peer) && !peer.Address.Equals(_address))
+                    .Take(_findConcurrency))
```
longfin, line 370: How about extracting immediate value for readability?

longfin, line 370: is it okay we take only this much?

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: I guess should ignore `PingTimeoutException` only, otherwise just throw. I'll fix it.

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: Can you please explain more about this comment? I didn't understand your intent 😥

limebell: Kademlia stores closer peers in more detail. Since `foundPeers` are sorted by the distance between `Target`, this expects to only select close peers to the `Target` to reduce network load.

But as you worried, in some situation the expectation may be incorrect. In that case, we have to increase `_findConcurrency` for more accurate result.

longfin:
```suggestion
                // idk filteredPeers is appropriate name...
                IEnumerable<BoundPeer> filteredPeers = foundPeers
                    .Where(peer => !history.Contains(peer) && !peer.Address.Equals(_address))
                    .Take(_findConcurrency);
                foreach (var found in filteredPeers)
```

limebell: And I found that we should exclude peers already in the `peersToFind` queue. I'll add it with your suggestion :)

dahlia (approved):

Libplanet/Net/Protocols/PingTimeoutException.cs
```diff
@@ -0,0 +1,47 @@
...
+            Target = (BoundPeer)info.GetValue("Target", typeof(BoundPeer));
+        }
+
+        public BoundPeer Target { get; private set; }
+
+        public override void GetObjectData(
+            SerializationInfo info, StreamingContext context)
+        {
+            if (info == null)
```
longfin, line 38:
```suggestion
            if (info is null)
```

longfin (dismissed):

longfin (approved):

earlbread (approved):

dahlia (approved):

longfin (approved):

PR #999, opened 2020-09-11
Fix swarm test

earlbread:
To improve tests:
- Use in memory `DefaultStoreFixture` by default.
- Reduce the default test timeout in `SwarmTest` to 30 seconds.
- Use helper methods instead of fixtures in `SwarmTest`.
- Parallelize tests.

### DIFF
--- a/.azure-pipelines/windows-net471.yml
+++ b/.azure-pipelines/windows-net471.yml
@@ -1,7 +1,7 @@
 parameters:
   configuration: Debug
   testArguments: -verbose
-  testTimeoutInMinutes: 12
+  testTimeoutInMinutes: 16
   copySQLitePCLRaw: false
 
 steps:
--- a/Libplanet.Tests/Net/SwarmTest.cs
+++ b/Libplanet.Tests/Net/SwarmTest.cs
@@ -1386,43 +1386,34 @@ namespace Libplanet.Tests.Net
 
             int renderCount = 0;
 
-            try
-            {
-                await StartAsync(miner1);
-                await StartAsync(miner2);
+            var privKey = new PrivateKey();
+            var addr = miner1.Address;
+            var item = "foo";
 
-                await BootstrapAsync(miner2, miner1.AsPeer);
+            miner1.BlockChain.MakeTransaction(privKey, new[] { new DumbAction(addr, item) });
+            await miner1.BlockChain.MineBlock(miner1.Address);
 
-                var privKey = new PrivateKey();
-                var addr = miner1.Address;
-                var item = "foo";
+            miner2.BlockChain.MakeTransaction(privKey, new[] { new DumbAction(addr, item) });
+            await miner2.BlockChain.MineBlock(miner2.Address);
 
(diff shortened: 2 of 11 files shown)


dahlia (commented):
Could you explain why it's okay to get rid of `-parallel none` option even if our tests are thread-unsafe? /cc @longfin

longfin:
It seems to need to rebase.

earlbread:
> Could you explain why it's okay to get rid of `-parallel none` option even if our tests are thread-unsafe? /cc @longfin

I thought it would be okay to remove `-parallel none` because the tests of the same collection (class) are not running in parallel and we usually don't use the option locally. https://xunit.github.io/docs/running-tests-in-parallel.html

limebell (approved):

riemannulus (dismissed):

longfin (approved):

longfin:
> Could you explain why it's okay to get rid of `-parallel none` option even if our tests are thread-unsafe? /cc @longfin

I missed it 😓 and I also have no idea about it.

In my memory, parallelism was abandoned, not because the tests were thread-unsafe, but because it was impossible to run too many tests at once.

limebell (approved):

.azure-pipelines/windows-net471.yml
```diff
@@ -33,7 +33,7 @@ steps:
     solution: Libplanet.sln
     msbuildVersion: "16.0"
     configuration: ${{ parameters.configuration }}
-    msbuildArguments: /restore /p:SkipSonar=true
+    msbuildArguments: /restore /p:SkipSonar=true /m
```
dahlia, line 36:
```suggestion
    msbuildArguments: /restore /p:SkipSonar=true /maxCpuCount
```

dahlia (commented):
Apparently needs to rebase again.

dahlia (approved):

limebell (dismissed):

longfin (approved):

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -1386,43 +1386,34 @@ public async Task RenderInFork()
...
-                await miner1.BlockReceived.WaitAsync();
+            await miner1.BlockReceived.WaitAsync();
 
-                Assert.Equal(miner1.BlockChain.Tip, miner2.BlockChain.Tip);
-                Assert.Equal(miner1.BlockChain.Count, miner2.BlockChain.Count);
-                Assert.Equal(2, renderCount);
-            }
-            finally
-            {
-                await StopAsync(miner1);
```
limebell, line 1505: It is okay to remove these post processes?

earlbread: If you create'Swarm' with'CreateSwarm', it will be disposed [here](https://github.com/planetarium/libplanet/blob/main/Libplanet.Tests/Net/SwarmTest.cs#L132).

riemannulus (approved):

limebell (approved):

dahlia (approved):

moreal (approved):

longfin (approved):

PR #1001, opened 2020-09-11
Discard StateRootHash check from BlockPolicy<T>

longfin:


longfin:
It seems unnecessary. I'll close it.

PR #1008, opened 2020-09-16
CI: Use -stoponfail on .NET Core tests

dahlia:
In the pull request #992, I applied `-stoponfail` option to Xunit, but it's only Mono and .NET Framework, because .NET Core jobs used `dotnet test` instead.  This replaces `dotnet test` with [xunit-cli][1] which can support all options `xunit.console.exe` provides.  Meanwhile, for coverage it still uses `dotnet test` because all tests should be run anyway even if some fails.

[1]: https://github.com/natemcmaster/xunit-cli

### DIFF
--- a/.azure-pipelines/dotnet-core.yml
+++ b/.azure-pipelines/dotnet-core.yml
@@ -3,6 +3,7 @@ parameters:
   testArguments: ""
   locale: "en_US.UTF8"
   testTimeoutInMinutes: 12
+  useDotnetTest: false
 
 steps:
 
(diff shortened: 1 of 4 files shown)


longfin (approved):

earlbread (dismissed):

longfin (approved):

earlbread (dismissed):

riemannulus (approved):

limebell (dismissed):
Isn't this may take longer for the successful tests, because it installs xunit-cli?

dahlia:
> Isn't this may take longer for the successful tests, because it installs xunit-cli?

You might be right, but it's a tiny program and takes only a short amount of time to install. Try to install it in your local shell.

dahlia:
@planetarium/libplanet Fixed errors.  Please review this again!

longfin (approved):

moreal (approved):

PR #1017, opened 2020-09-23
Omit state root hash check when evaluateActions is disabled

longfin:
This PR omits a check for `Block<T>.StateRootHash` when `evaluateActions` is `false` to avoid unexpected evaluation.

I also skip the changelog because we didn't release 0.10.0 that introduced MPT yet(!).

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -866,7 +866,7 @@ namespace Libplanet.Blockchain
             Block<T> prevTip = Tip;
             try
             {
-                InvalidBlockException e = ValidateNextBlock(block);
+                InvalidBlockException e = ValidateNextBlock(block, evaluateActions);
 
                 if (!(e is null))
                 {
@@ -1669,7 +1669,10 @@ namespace Libplanet.Blockchain
             }
         }
 
-        private InvalidBlockException ValidateNextBlock(Block<T> nextBlock)
+        private InvalidBlockException ValidateNextBlock(
+            Block<T> nextBlock,
+            bool validateStateRootHash
+        )
         {
             InvalidBlockException e = Policy.ValidateNextBlock(this, nextBlock);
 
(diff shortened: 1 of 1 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1669,7 +1669,7 @@ HashDigest<SHA256> blockHash
             }
         }
 
-        private InvalidBlockException ValidateNextBlock(Block<T> nextBlock)
+        private InvalidBlockException ValidateNextBlock(Block<T> nextBlock, bool evaluateActions)
```
moreal, line 1672: It is just a suggestion, IMO it should have another naming like `validateStateRootHash` or better to present its purpose.

earlbread (dismissed):
How about moving the state root hash validation code to [here](https://github.com/planetarium/libplanet/pull/1017/files#diff-f9c16aaa8d3a312223ca86876f791202R912) so that the actions doesn't execute twice?

Libplanet/Blockchain/BlockChain.cs
longfin: Amended https://github.com/planetarium/libplanet/pull/1017/commits/6de74c53b6275a13ee47e26a7cba86ff15e456aa

longfin:
> How about moving the state root hash validation code to [here](https://github.com/planetarium/libplanet/blob/6de74c53b6275a13ee47e26a7cba86ff15e456aa/Libplanet/Blockchain/BlockChain.cs#L912) so that the actions doesn't execute twice?

I added https://github.com/planetarium/libplanet/pull/1017/commits/2b710f0d8792c5eaafe7deb98dff74d0e5046d2b to try it. PTAL @planetarium/libplanet

earlbread (approved):

moreal (approved):

earlbread (approved):

moreal (approved):

PR #1000, opened 2020-09-11
More events on renderers

dahlia:
Added two more render methods:

- `IActionRenderer<T>.RenderBlockEnd()`
- `IRenderer<T>.RenderReorgEnd()`

See also the docs on `IRenderer<T>`:

IRenderer<T>

… and `IActionRenderer<T>`:

IActionRenderer<T>

By utilizing `RenderBlockEnd()` event, `DelayedActionRenderer<T>` became to atomically update its internal buffers (it's further than the approach made in #998).

I also refactored `DumRenderer<T>` (in the tests project) into `RecordingRenderer<T>` to reorganize `RenderRecord` into hierarchical layers.

PR #1019, opened 2020-09-24
Fix inaccurate render events order according to documented automata

dahlia:
This fixes the inaccurate order of rendering events that `BlockChain<T>` emits so that the order conforms to the [documented rendering automata][1], i.e.:

> See also the docs on `IRenderer<T>`:
> 
> IRenderer<T>
> 
> … and `IActionRenderer<T>`:
> 
> IActionRenderer<T>

In order to find the causes of these bugs, I also added `ValidatingActionRenderer<T>` which is a profiler-guided analyzer.

[1]: https://github.com/planetarium/libplanet/pull/1000

### DIFF
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -46,7 +46,7 @@ namespace Libplanet.Tests.Blockchain
                 .ForContext<BlockChainTest>();
 
             _fx = new DefaultStoreFixture(memory: true);
-            _renderer = new RecordingRenderer<DumbAction>();
+            _renderer = new ValidatingActionRenderer<DumbAction>();
             _policy = new BlockPolicy<DumbAction>(new MinerReward(1));
             _blockChain = new BlockChain<DumbAction>(
                 _policy,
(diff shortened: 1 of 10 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -967,9 +967,13 @@ public long GetNextTxNonce(Address address)
                 _rwlock.ExitUpgradeableReadLock();
             }
 
-            if (renderActions)
+            if (renderBlocks && ActionRenderers.Any())
             {
-                RenderBlock(evaluations, block, stateCompleters);
+                if (renderActions)
+                {
+                    RenderBlock(evaluations, block, stateCompleters);
```
limebell, line 974: Do not invoke `RenderBlock` when `renderActions` is false but `renderBlocks` is true?

dahlia: Although the name is confusing, `RenderBlock()` here refers to `BlockChain<T>.RenderBlock()` method which calls `IActionRenderer<T>.RenderAction()` etc, and it does not call `IRenderer<T>.RenderBlock()` method.

limebell (approved):

earlbread (approved):

dahlia:
@planetarium/libplanet Please review this! 🙇🏻‍♂️

longfin (approved):

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -1485,24 +1482,42 @@ public void SwapForSameTip(bool render)
...
+        [InlineData(false)]
+        public async Task SwapWithoutReorg(bool render)
         {
             BlockChain<DumbAction> fork = _blockChain.Fork(_blockChain.Tip.Hash);
 
-            // When direction == true:  the higher chain goes to the lower  chain  [#N -> #N-1]
-            // When direction == false: the lower  chain goes to the higher chain  [#N -> #N+1]
-            await (direction ? _blockChain : fork).MineBlock(default);
-            var prevRecords = _renderer.ReorgRecords;
+            // The lower  chain goes to the higher chain  [#N -> #N+1]
```
moreal, line 1491:
```suggestion
            // The lower chain goes to the higher chain  [#N -> #N+1]
```

It seems just useless space not affecting its functions. If it makes your task confused, please ignore this.

riemannulus (approved):

moreal (approved):

PR #1050, opened 2020-10-22
Limit the maximum bytes length of blocks

### LINKED ISSUE #201: Limit size of block (byte-wise)
Currently, Libplanet doesn't define how many transactions will be stored in a block, and the size of a transaction is also not explicitly defined. however, in order to predict figures such as storage space and network traffic, it is necessary to be able to determine how much physical space a block occupies.

The block size is not a protocol-level constraint, and game developers need to adjust the unit to synchronize the status of each game, so I suggest offering it as an configurable parameter.

dahlia:
This fixes #201.  Please read added changelogs for details.


earlbread (dismissed):

dahlia:

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -682,63 +685,120 @@ public long GetNextTxNonce(Address address)
...
+                index: index,
+                difficulty: difficulty,
+                totalDifficulty: Tip.TotalDifficulty,
+                nonce: default,
+                miner: miner,
+                previousHash: prevHash,
+                timestamp: currentTime,
+                transactions: new Transaction<T>[0]
+            ).BytesLength;
+            int maxBlockBytes = Math.Max(Policy.GetMaxBlockBytes(index), 1);
```
limebell, line 744: What is the meaning of `1`?

riemannulus (dismissed):

Libplanet/Blockchain/BlockChain.cs
dahlia: This makes the `maxBlockBytes` 1 at least even if `Policy.GetMaxBlockBytes()` is less than 1 (e.g., `Max(-1, 1)` → `1`, `Max(5, 1)` → `5`).

Libplanet/Blocks/InvalidBlockBytesLengthException.cs
```diff
@@ -0,0 +1,49 @@
...
+                $"Allowed (max): #{maxBlockBytesLength}")
+        {
+            BlockBytesLength = blockBytesLength;
+            MaxBlockBytesLength = maxBlockBytesLength;
+        }
+
+        /// <summary>
+        /// The bytes length of the actual block.
+        /// </summary>
+        public int BlockBytesLength { get; set; }
```
longfin, line 41: I guess we need to add codes about serialization for these fields to preserve them.

Libplanet/Blockchain/BlockChain.cs
limebell: I was wondering why the value is exactly 1 🤔

dahlia: This implements the below documented behavior:

<https://github.com/planetarium/libplanet/blob/e7d963bae4a757f45ac808d679e31a060a5ed780/Libplanet/Blockchain/Policies/IBlockPolicy.cs#L81-L82>

which assumes there is no block taking 0 bytes or negative length of bytes.

limebell (dismissed):

Libplanet/Blocks/InvalidBlockBytesLengthException.cs
dahlia: Added.

dahlia:
@planetarium/libplanet Addressed reviews and rebased.  Please review this again!

longfin (dismissed):

earlbread (dismissed):

limebell (dismissed):

riemannulus (dismissed):

moreal (dismissed):

dahlia:
@planetarium/libplanet I rebased.  Please review this again.

earlbread (approved):

longfin (approved):

riemannulus (approved):

moreal (approved):

dahlia:
@planetarium/libplanet As it seems to reach Azure Pipelines' rate limit, the build is being timed out again and again.  Instead retring the build again, I am going to merge this anyway.

PR #1063, opened 2020-10-30
Do not append genesis during fork

limebell:
~Instead `IStore.ForkBlockIndexes()` became to iterate from 0 (was 1).~

Instead just append its index only.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -172,7 +172,13 @@ namespace Libplanet.Blockchain
                 GetBalance,
                 trieGetter);
 
-            if (Count == 0)
+            if (inFork && !(StateStore is IBlockStatesStore))
+            {
+                // If the store is BlockStateStore, have to fork state reference too so
+                // should use Append().
+                Store.AppendIndex(Id, genesisBlock.Hash);
+            }
+            else if (Count == 0)
             {
                 Append(
                     genesisBlock,


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -174,13 +174,16 @@ IEnumerable<IRenderer<T>> renderers
...
             if (Count == 0)
             {
-                Append(
-                    genesisBlock,
-                    currentTime: genesisBlock.Timestamp,
-                    renderBlocks: !inFork,
-                    renderActions: !inFork,
-                    evaluateActions: !inFork
-                );
+                if (!inFork)
```
earlbread, line 177: How merge if statements?
```suggestion
                if (count == 0 && !inFork)
```

limebell: It will throw exception in next `elseif` statement 😢

dahlia, line 177: How about merging two conditions into one using an `&&` operator?

dahlia (approved):

Libplanet/Blockchain/BlockChain.cs
dahlia: Oh, [fair enough](https://github.com/planetarium/libplanet/pull/1063#discussion_r514946332).

limebell: It will throw exception in next else if statement 😢

limebell:
I made constructor of `BlockChain<T>` instance to not append genesis block every time during fork. Please review again! @libplanet

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -172,7 +172,11 @@ IEnumerable<IRenderer<T>> renderers
                 GetBalance,
                 trieGetter);
 
-            if (Count == 0)
+            if (inFork && !(StateStore is IBlockStatesStore))
```
longfin, line 175: It would be better if we left comments about why we query `StateStore`'s type.

limebell: 👌

longfin (approved):

earlbread (approved):

PR #1067, opened 2020-11-03
Maintenance/9c-main

limebell:
Cherrypicked from 9c-main.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -172,7 +172,13 @@ namespace Libplanet.Blockchain
                 GetBalance,
                 trieGetter);
 
-            if (Count == 0)
+            if (inFork && !(StateStore is IBlockStatesStore))
+            {
+                // If the store is BlockStateStore, have to fork state reference too so
+                // should use Append().
+                Store.AppendIndex(Id, genesisBlock.Hash);
+            }
+            else if (Count == 0)
             {
                 Append(
                     genesisBlock,
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1905,22 +1905,21 @@ namespace Libplanet.Blockchain
             if (nextBlock.Index != index)
             {
                 return new InvalidBlockIndexException(
-                    $"The expected block index is #{index}, but its index " +
-                    $"is #{nextBlock.Index}.");
+                    $"The expected index of block {nextBlock.Hash} is #{index}, " +
+                    $"but its index is #{nextBlock.Index}.");
             }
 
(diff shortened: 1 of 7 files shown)


dahlia:
I believe these patches should be ported to the 0.10-maintenance first.

limebell:
> I believe these patches should be ported to the 0.10-maintenance first.

Rebased! Please review again.

dahlia (dismissed):

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1899,22 +1905,21 @@ private InvalidBlockException ValidateNextBlock(Block<T> nextBlock)
             if (nextBlock.Index != index)
             {
                 return new InvalidBlockIndexException(
-                    $"The expected block index is #{index}, but its index " +
-                    $"is #{nextBlock.Index}.");
+                    $"The expected index of block [{nextBlock.Hash}] is #{index}, " +
```
dahlia, line 1908: How about this:

```
Expected the block #{Index}, but got the block #{NextBlock.Index} {NextBlock.Hash} instead.
```

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1899,22 +1905,21 @@ private InvalidBlockException ValidateNextBlock(Block<T> nextBlock)
...
+                    $"but its index is #{nextBlock.Index}.");
             }
 
             if (nextBlock.Difficulty < difficulty)
             {
                 return new InvalidBlockDifficultyException(
-                    $"The expected difficulty of the block #{index} " +
-                    $"is {difficulty}, but its difficulty is " +
-                    $"{nextBlock.Difficulty}.");
+                    $"The expected difficulty of the block #{index} [{nextBlock.Hash}] " +
```
dahlia, line 1915: It would be better if `InvalidBlockDifficultyException` has more properties like `BlockHash` and `BlockIndex` in the future…

```suggestion
                    $"The expected difficulty of the block #{index} {nextBlock.Hash} " +
```

dahlia (approved):

Libplanet/Blockchain/BlockChain.cs
limebell: How about make `InvalidBlockException` to have `BlockHash` and `BlockIndex`?

dahlia: > How about make `InvalidBlockException` to have `BlockHash` and `BlockIndex`?

That sounds good if it's possible. Some of its subtypes might be unable to have these properties though?

dahlia (approved):

dahlia:

moreal (approved):

PR #909, opened 2020-06-16
Reconnect to TURN when the connection is lost

earlbread:
This fixes a bug where `Swarm<T>` had terminated when disconnected from TURN. The flaky `StopGracefullyWhileStarting` test was also fixed.

PR #1074, opened 2020-11-04
Move initializing code to TurnClient

earlbread:
This PR moves TURN related features from `NetMQTransport` to `TurnClient` and enables restarting in `TurnClient` when relayed connection and TURN control session had disconnected. (see also #909 )


longfin:
@planetarium/libplanet Rebased. PTAL

dahlia (approved):
When this pull request is ported to the upstream, it should go to the main branch rather than the 0.10-maintenance branch IMHO.

PR #1075, opened 2020-11-05
Dispose _control when ProcessMessage is ended

earlbread:


longfin:
I believe that we can achieve this PR's purpose by #1074. so I'll close it, thanks.

PR #1078, opened 2020-11-06
Fix dealer

earlbread:


longfin:
I guess we can replace in dealer sockets in `_dealers` by comparing as like below

```csharp
                    string endpoint = ToNetMQAddress(peer);
                    if (!_dealers.TryGetValue(peer.Address, out DealerSocket dealer))
                    {
                        dealer = new DealerSocket(endpoint);
                        _dealers[peer.Address] = dealer;
                    }
                    else if (dealer.Options.LastEndpoint != endpoint)
                    {
                        dealer.Dispose();
                        dealer = new DealerSocket(endpoint);
                        _dealers[peer.Address] = dealer;
                    }
```

longfin:
I believe that we can achieve this PR's purpose by #1074. so I'll close it, thanks.

PR #1088, opened 2020-11-17
Ignore stale block demand

limebell:
Additionally, same block demand will not refresh the value.


dahlia:
@limebell Could we retarget this pull request to the main branch?

PR #1104, opened 2020-11-25
Fix BlockChain<T>.Append() to disallow block with too many txs

dahlia:
In the commit 1d357776a640839e34d4195299fc09238e4ce5d8, we added `IBlockPolicy<T>.MaxTransactionsPerBlock` property, but it had only functioned as a default value for `BlockChain<T>.MineBlock()` method's `maxTransactions` option.  This had misled and been unintended, because `BlockChain<T>.Append()` still had accept blocks violating its policy (`Block<T>` having more `Transactions` than the number a policy allows still can be made with `Block<T>.Mine()` method).

Although this adds `BlockExceedingTransactionsException` class, since it would be released as a patch release, it's not `public` yet.  We should make it `public` when it's ported to the upstream main (next major/minor release).

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -565,8 +565,8 @@ namespace Libplanet.Blockchain
         /// <exception cref="InvalidBlockBytesLengthException">Thrown when the block to mine is
         /// too long (according to <see cref="IBlockPolicy{T}.GetMaxBlockBytes(long)"/>) in bytes.
         /// </exception>
-        /// <exception cref="InvalidBlockException">Thrown when the given
-        /// <paramref name="block"/> is invalid, in itself or according to
+        /// <exception cref="InvalidBlockException">Thrown when the given <paramref name="block"/>
+        /// is invalid (e.g., too many transactions), in itself or according to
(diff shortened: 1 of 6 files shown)


longfin (approved):

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1042,6 +1042,16 @@ public long GetNextTxNonce(Address address)
 
             _logger.Debug("Trying to append block {blockIndex}: {block}", block?.Index, block);
 
+            if (block.Transactions.Count() is { } txCount &&
```
moreal, line 1045: It is a question for this line, `block.Transactions.Count()`'s return value is nullable? I wonder how this line works.

moreal (approved):

Libplanet/Blockchain/BlockChain.cs
dahlia: It does not have to necessarily use pattern matching here; it just functions as an inline variable declaration (`txCount`)…

moreal: I understood. Thanks for your kind description. 🙏

PR #1112, opened 2020-11-26
Break when getting blocks count is less then hash chunk size

riemannulus:
If getting blocks size in `Swarm<T>.FillBlocksAsync()` is less then `Swarm<T>.FindNextHashesChunkSize`, it means getting all of blocks in this try. So, do not processing next step for check.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@ To be released.
 ### Backward-incompatible API changes
 
 ### Backward-incompatible network protocol changes
+ -   `Swarm<T>` became no longer retry when `Swarm<T>` receives
+      less than 500 blocks.  [[#1112]]
 
 ### Backward-incompatible storage format changes
 
@@ -61,6 +63,7 @@ To be released.
 [#1101]: https://github.com/planetarium/libplanet/pull/1101
 [#1102]: https://github.com/planetarium/libplanet/pull/1102
 [#1110]: https://github.com/planetarium/libplanet/pull/1110
+[#1112]: https://github.com/planetarium/libplanet/pull/1112
 
 
 Version 0.10.2
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -1862,6 +1862,7 @@ namespace Libplanet.Net
                         cancellationToken
                     );
 
+                    var receivedBlockCountCurrentLoop = 0;
                     await foreach (Block<T> block in blocks)
                     {
                         _logger.Debug(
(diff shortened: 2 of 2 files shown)


dahlia (commented):
- Could we have a small unit test for this?  (If it's not easy we should have a `TODO`/`FIXME` comment then.)
- This behavioral change should be recorded in the changelog.

Libplanet/Net/Swarm.cs
```diff
@@ -1891,6 +1891,16 @@ await foreach (Block<T> block in blocks)
                             block.Hash
                         );
                     }
+
+                    var getBlockListCount = await blocks.CountAsync(cancellationToken);
```
longfin, line 1895: I guess we can use a separated counter manually instead of  `.CountAsync()`. (or, just save `receivedBlockCount` before `GetBlocksAsync()`)

Libplanet/Net/Swarm.cs
```diff
@@ -1786,7 +1786,7 @@ CancellationToken cancellationToken
                             "Peer [{0}] didn't return any hashes; ignored.",
                             peer.Address.ToHex()
                         );
-                        return workspace;
+                        break;
```
limebell, line 1789: What is this change for?

Libplanet/Net/Swarm.cs
```diff
@@ -1891,6 +1891,16 @@ await foreach (Block<T> block in blocks)
                             block.Hash
                         );
                     }
+
+                    var getBlockListCount = await blocks.CountAsync(cancellationToken);
+                    if (getBlockListCount < FindNextHashesChunkSize)
+                    {
+                        _logger.Debug(
+                            $"Get all blocks from Peer [{peer}]",
```
limebell, line 1899:
```suggestion
                            $"Got all blocks from Peer [{peer}]",
```
And how about log block count here?

Libplanet/Net/Swarm.cs
riemannulus: It is ok because return `workspace` after the `finally` statement. Other return `workspace` things are all of the return at the finish of this method. So I want to standardize this too.

Libplanet/Net/Swarm.cs
riemannulus: amended at https://github.com/planetarium/libplanet/pull/1112/commits/969588c050c433949763a7ec7fee88f9f25e7cfa

Libplanet/Net/Swarm.cs
riemannulus: amended at https://github.com/planetarium/libplanet/pull/1112/commits/969588c050c433949763a7ec7fee88f9f25e7cfa. I added log when adding test code.

longfin (dismissed):

longfin (approved):

CHANGES.md
```diff
@@ -9,6 +9,8 @@ To be released.
 ### Backward-incompatible API changes
 
 ### Backward-incompatible network protocol changes
+ -  If the count of blocks `Swarm<T>` have received is less than
+    the `Swarm<T>.FindNextHashesChunkSize`, `Swarm<T>` will no longer retry.  [[#1112]]
```
dahlia, line 13: As `Swarm<T>.FindNextHashesChunkSize` is `internal`, this should be like instead:

```suggestion
 -  `Swarm<T>` became no longer retry when `Swarm<T>` receives less than 500
    blocks.  [[#1112]]
```

dahlia (dismissed):

longfin (approved):

PR #1116, opened 2020-11-26
Reuse action evaluations from mining

longfin:
This PR increases the performance of `BlockChain<T>.MineBlock()` and `Block<T>.Deserialize()`.

- `BlockChain<T>.MineBlock()` became to reuse action evaluations while `.Append()`.
- `Block<T>.Deserielize()` became to accept all serialized values(e.g. `.Hash`, `.TxHash`) instead of re-calculating them.

and it also changes the behavior of `Block<T>.Validate()` to only checks its `.TxHash`, not all `.Transactions`. for it, I created `InvalidBlockTxHashException` as a new exception.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,7 +63,8 @@ To be released.
  -  `Swarm<T>` became not to receive states from trusted peers.
     [[#1061], [#1102]]
  -  `Swarm<T>` became not to retry when block downloading.  [[#1062], [#1102]]
- -  `Block<T>.Validate()` became to validate only `Block<T>.TxHash` instead of its all transactions.  [[#1116]]
+ -  `Block<T>.Evaluate()` became to validate only `Block<T>.TxHash` instead of
+    its all transactions.  [[#1116]]
  -  Improved performance of `BlockChain<T>.MineBlock()`.  [[#1116]]
  -  Improved performance of `Block<T>.Deserialize()`.  [[#1116]]
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -53,6 +53,7 @@ To be released.
     class.  [[#1119]]
  -  Added `Libplanet.Blockchain.Renderers.Debug.InvalidRenderException<T>`
     class.  [[#1119]]
+ -  Added `InvalidBlockTxHashException` class.  [[#1116]]
 
 ### Behavioral changes
 
@@ -62,6 +63,9 @@ To be released.
  -  `Swarm<T>` became not to receive states from trusted peers.
     [[#1061], [#1102]]
  -  `Swarm<T>` became not to retry when block downloading.  [[#1062], [#1102]]
+ -  `Block<T>.Validate()` became to validate only `Block<T>.TxHash` instead of its all transactions.  [[#1116]]
+ -  Improved performance of `BlockChain<T>.MineBlock()`.  [[#1116]]
+ -  Improved performance of `Block<T>.Deserialize()`.  [[#1116]]
 
 ### Bug fixes
 
@@ -82,6 +86,7 @@ To be released.
 [#1102]: https://github.com/planetarium/libplanet/pull/1102
 [#1110]: https://github.com/planetarium/libplanet/pull/1110
 [#1112]: https://github.com/planetarium/libplanet/pull/1112
+[#1116]: https://github.com/planetarium/libplanet/pull/1116
 [#1119]: https://github.com/planetarium/libplanet/pull/1119
 
 
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -687,6 +687,7 @@ namespace Libplanet.Tests.Blockchain
                     nonce: nonce,
                     privateKey: signer);
                 heavyTxs.Add(heavyTx);
+                nonce += 1;
             }
 
             var block1 = TestUtils.MineNext(
--- a/Libplanet.Tests/Blocks/BlockTest.cs
+++ b/Libplanet.Tests/Blocks/BlockTest.cs
@@ -447,76 +447,6 @@ namespace Libplanet.Tests.Blocks
         }
 
         [Fact]
-        public void EvaluateInvalidTxSignature()
-        {
-            RawTransaction rawTx = new RawTransaction(
-                0,
-                _fx.TxFixture.Address1.ByteArray,
-                _fx.Genesis.Hash.ByteArray,
-                ImmutableArray<ImmutableArray<byte>>.Empty,
-                _fx.TxFixture.PublicKey1.Format(false).ToImmutableArray(),
-                DateTimeOffset.UtcNow.ToString(
-                    "yyyy-MM-ddTHH:mm:ss.ffffffZ",
-                    CultureInfo.InvariantCulture
-                ),
-                ImmutableArray<IValue>.Empty,
-                new byte[10].ToImmutableArray()
-            );
-            var invalidTx = new Transaction<DumbAction>(rawTx);
-            Assert.Throws<InvalidTxSignatureException>(() =>
-                MineNext(
-                    MineGenesis<DumbAction>(),
-                    new List<Transaction<DumbAction>>
-                    {
-                        invalidTx,
-                    }
-                )
-            );
-        }
-
-        [Fact]
-        public void EvaluateInvalidTxPublicKey()
-        {
-            RawTransaction rawTxWithoutSig = new RawTransaction(
-                0,
-                new PrivateKey().ToAddress().ByteArray,
-                _fx.Genesis.Hash.ByteArray,
-                ImmutableArray<ImmutableArray<byte>>.Empty,
-                _fx.TxFixture.PublicKey1.Format(false).ToImmutableArray(),
-                DateTimeOffset.UtcNow.ToString(
-                    "yyyy-MM-ddTHH:mm:ss.ffffffZ",
-                    CultureInfo.InvariantCulture
-                ),
-                ImmutableArray<IValue>.Empty,
-                ImmutableArray<byte>.Empty
-            );
-            byte[] sig = _fx.TxFixture.PrivateKey1.Sign(
-                new Transaction<DumbAction>(rawTxWithoutSig).Serialize(false)
-            );
-            var invalidTx = new Transaction<DumbAction>(
-                new RawTransaction(
-                    0,
-                    rawTxWithoutSig.Signer,
-                    rawTxWithoutSig.GenesisHash,
-                    rawTxWithoutSig.UpdatedAddresses,
-                    rawTxWithoutSig.PublicKey,
-                    rawTxWithoutSig.Timestamp,
-                    rawTxWithoutSig.Actions,
-                    sig.ToImmutableArray()
-                )
-            );
-            Assert.Throws<InvalidTxPublicKeyException>(() =>
-                MineNext(
-                    MineGenesis<DumbAction>(),
-                    new List<Transaction<DumbAction>>
-                    {
-                        invalidTx,
-                    }
-                )
-            );
-        }
-
-        [Fact]
         public void EvaluateInvalidTxUpdatedAddresses()
         {
             ImmutableArray<IValue> rawActions =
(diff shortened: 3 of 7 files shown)


Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -686,6 +686,7 @@ public void AppendThrowsInvalidBlockBytesLengthException()
                     nonce: nonce,
                     privateKey: signer);
                 heavyTxs.Add(heavyTx);
+                nonce += 1;
```
riemannulus, line 689: How pass this test before? 🤔

Libplanet.Tests/Blocks/BlockTest.cs
```diff
@@ -446,76 +446,6 @@ public void EvaluateActionsPerTx()
             );
         }
 
-        [Fact]
-        public void EvaluateInvalidTxSignature()
```
riemannulus, line 450: Doesn't need this test anymore?

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -917,13 +917,18 @@ public long GetNextTxNonce(Address address)
                 try
                 {
                     SetStates(block, actionEvaluations, false);
+                    block = new Block<T>(block, trieStateStore.GetRootHash(block.Hash));
+                    SetStates(block, actionEvaluations, false);
```
riemannulus, line 921: IMO We don't need `SetStates` twice in this statement.

Libplanet.Tests/Blockchain/BlockChainTest.cs
longfin: Because we didn't check `.TxHash`, but re-calculate them. 😅

Libplanet/Blockchain/BlockChain.cs
longfin: It's needed. (it's a quite tricky part.)

In first `SetStates(block, actionEvaluations, false)`, `block.Hash` isn't a real block hash. it can't be determined since `Block<T>.Hash` needs `StateRootHash` from `SetState()`. so we need to do `SetStates` to calculate and confirm `.Hash`, and repeat it to store it.

I think it would be better there is some comment about it. I'll add it.

Libplanet.Tests/Blocks/BlockTest.cs
longfin: Yes, because we'll skip `Transaction.Validate()` during `Block<T>.Validate()`.

Libplanet/Blockchain/BlockChain.cs
longfin: I amended to leave comments. https://github.com/planetarium/libplanet/pull/1116/commits/78d50d6750a3a4f281bb119497bdd42a946ade28

but if you have any questions and/or ideas yet, feel free to left here.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -917,13 +917,22 @@ public long GetNextTxNonce(Address address)
                 try
                 {
                     SetStates(block, actionEvaluations, false);
+                    block = new Block<T>(block, trieStateStore.GetRootHash(block.Hash));
+
+                    // it's needed because `block.Hash` was updated with the state root hash.
+                    // FIXME: we need a method for calculating the state root hash without
+                    // `.SetStates()`.
+                    SetStates(block, actionEvaluations, false);
```
limebell, line 925: It seems some duplicated action is conducted in here. Can't we reuse some evaluations from first `SetStates()`?

Libplanet/Blockchain/BlockChain.cs
riemannulus: I think it is ok to leave `FIXME` comments.

riemannulus (dismissed):

Libplanet/Blocks/InvalidBlockTxHashException.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            if ((byte[])info.GetValue(nameof(CalculatedTxHash), typeof(byte[])) is { } cTxHashBytes)
+            {
+                CalculatedTxHash = new HashDigest<SHA256>(cTxHashBytes);
+            }
+        }
+
+        /// <summary>
+        /// The hashdigest from actual block.
```
limebell, line 45:
```suggestion
        /// The hash digest from actual block.
```

Libplanet/Blocks/InvalidBlockTxHashException.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        public HashDigest<SHA256>? BlockTxHash { get; }
+
+        /// <summary>
+        /// The calculated hashdigest from transactions in the block.
```
limebell, line 50:
```suggestion
        /// The calculated hash digest from transactions in the block.
```

Libplanet/Blocks/InvalidBlockTxHashException.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace Libplanet.Blocks
+{
+    [Serializable]
+    [Equals]
+    public class InvalidBlockTxHashException : InvalidBlockException
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="InvalidBlockTxHashException"/> class.
+        /// </summary>
+        /// <param name="blockTxHash">The hashdigest of <see cref="Block{T}.TxHash"/>.</param>
```
limebell, line 14:
```suggestion
        /// <param name="blockTxHash">The hash digest of <see cref="Block{T}.TxHash"/>.</param>
```

Libplanet/Blocks/InvalidBlockTxHashException.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <param name="calculatedTxHash">The calculated hashdigest from
```
limebell, line 15:
```suggestion
        /// <param name="calculatedTxHash">The calculated hash digest from
```

Libplanet/Blocks/Block.cs
```diff
@@ -188,6 +171,36 @@ private Block(RawBlock rb)
...
+            Difficulty = difficulty;
+            TotalDifficulty = totalDifficulty;
+            Nonce = nonce;
+            Miner = miner;
+            PreviousHash = previousHash;
+            Timestamp = timestamp;
+            Hash = hash;
+            PreEvaluationHash = preEvaluationHash ??
+                throw new ArgumentNullException(nameof(preEvaluationHash));
+            StateRootHash = stateRootHash;
```
limebell, line 199: No null check is needed here?

longfin: `Block<T>.StateRootHash` is `HashDigest<SHA256>?`, so it can be null.  (I don't have ideas about why.)

Libplanet/Blockchain/BlockChain.cs
longfin: `.SetState()` doesn't return any value. it only stores to `.StateStore`. 

We need the state root hash to create a new block. and, at this point, the only way to do this is to get the stored state root hash via `.SetState()`.

limebell: I mean
```
            IImmutableSet<Address> stateUpdatedAddresses = actionEvaluations
                .SelectMany(a => a.OutputStates.StateUpdatedAddresses)
                .ToImmutableHashSet();
            IImmutableSet<(Address, Currency)> updatedFungibleAssets = actionEvaluations
                .SelectMany(a => a.OutputStates.UpdatedFungibleAssets
                    .SelectMany(kv => kv.Value.Select(c => (kv.Key, c))))
                .ToImmutableHashSet();
```
reuse these values which are inside of `SetStates()`.

Libplanet/Blocks/Block.cs
limebell: I was wondering why `PreEvaluationHash` throws `ArgumentNullException` when null while `StateRootHash` isn't. And I found that there can be a block with null `StateRootHash` which is intended.

dahlia: @moreal Could you explain what does the situation `Block<T>.StateRootHash = null` mean?

CHANGES.md
```diff
@@ -48,6 +49,9 @@ To be released.
  -  `Swarm<T>` became not to receive states from trusted peers.
     [[#1061], [#1102]]
  -  `Swarm<T>` became not to retry when block downloading.  [[#1062], [#1102]]
+ -  `Block<T>.Validate()` became to validate only `Block<T>.TxHash` instead of its all transactions.  [[#1116]]
```
dahlia, line 52: As `Block<T>.Validate()` is `internal`, we mention `Block<T>.Evaluate()` rather than `Validate()` here.  Also, we probably should update the docs here too?:

https://github.com/planetarium/libplanet/blob/ae6833181e6a58c78c4e081d366f96e1ec9bc08a/Libplanet/Blocks/Block.cs#L501-L526

Libplanet/Blocks/Block.cs
moreal: @dahlia `Block<T>.StateRootHash = null` means it is using another `IStateStore` implementation not `TrieStateStore`. Or it may be a block to caculate `StateRootHash`. To calculate `StateRootHash`, it should have a block not having `StateRootHash`.

Libplanet/Blockchain/BlockChain.cs
longfin: Ah, I see.

Libplanet/Blocks/Block.cs
dahlia: @moreal 

1. Even if a `BlockChain<T>` does not use `TrieStateStore`, shouldn't a `Block<T>` in itself anyway calculate/have its `StateRootHash`?  The current way makes `Block<T>`'s behavior to depend on its `BlockChain<T>`, which depends on `Block<T>` again; [it causes circular dependency][1] which is a [code smell][2].
2. Further on the above point, I believe `Block<T>` should not expose its state that “it should calculate its `StateRootHash` but it hasn't been yet at this *moment*.”

[1]: https://en.wikipedia.org/wiki/Acyclic_dependencies_principle
[2]: https://en.wikipedia.org/wiki/Code_smell

Libplanet/Blockchain/BlockChain.cs
longfin: @limebell On second thought, if we adjust codes as above `FIXME:`, we can omit to repeat `SetStates()` calls and it would be more natural to me. so I'll waive [it](https://github.com/planetarium/libplanet/pull/1116#discussion_r535806980) at this time.

longfin:
@planetarium/libplanet I applied feedbacks and rebased. PTAL

CHANGES.md
```diff
@@ -82,7 +87,11 @@ To be released.
 [#1102]: https://github.com/planetarium/libplanet/pull/1102
 [#1110]: https://github.com/planetarium/libplanet/pull/1110
 [#1112]: https://github.com/planetarium/libplanet/pull/1112
+<<<<<<< HEAD
 [#1119]: https://github.com/planetarium/libplanet/pull/1119
+=======
+[#1116]: https://github.com/planetarium/libplanet/pull/1116
+>>>>>>> 6137f0014... Changes
```
riemannulus, line 94: It looks like you have some mistake when rebasing PR.

longfin: Amended. (https://github.com/planetarium/libplanet/pull/1116/commits/22eae2ac2c598488d38c3684d5b1e2d300149b46)

limebell (approved):

riemannulus (approved):

Libplanet/Blocks/Block.cs
```diff
@@ -178,6 +158,9 @@ private Block(RawBlock rb)
                     rb.Header.Timestamp,
                     BlockHeader.TimestampFormat,
                     CultureInfo.InvariantCulture).ToUniversalTime(),
+#pragma warning disable MEN002 // Line is too long
+                rb.Header.TxHash.Any() ? new HashDigest<SHA256>(rb.Header.TxHash) : (HashDigest<SHA256>?)null,
+#pragma warning restore MEN002 // Line is too long
```
dahlia, line 163: Can't we expand this expression into multiple lines?

```suggestion
                rb.Header.TxHash.Any()
                    ? new HashDigest<SHA256>(rb.Header.TxHash)
                    : (HashDigest<SHA256>?)null,
```

Libplanet/Blocks/Block.cs
dahlia: It would be better if we have a `FIXME` comment here.

Libplanet/Blocks/Block.cs
```diff
@@ -513,13 +526,9 @@ public byte[] Serialize()
         /// <exception cref="InvalidBlockNonceException">Thrown when
         /// the <see cref="Nonce"/> does not satisfy its
         /// <see cref="Difficulty"/> level.</exception>
-        /// <exception cref="InvalidTxSignatureException">Thrown when its
-        /// <see cref="Transaction{T}.Signature"/> is invalid or not signed by
-        /// the account who corresponds to its
-        /// <see cref="Transaction{T}.PublicKey"/>.</exception>
-        /// <exception cref="InvalidTxPublicKeyException">Thrown when its
-        /// <see cref="Transaction{T}.Signer"/> is not derived from its
-        /// <see cref="Transaction{T}.PublicKey"/>.</exception>
```
dahlia, line 522: One thing I wonder about this: Are these exception classes still used from any other places? Can't they be safely removed now?

longfin: They're still thrown when staging tx on `BlockChain<T>` or calling `Transaction<T>.Validate()` manually.

Libplanet/Blocks/Block.cs
longfin: If then, we'll face [SA1118](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1118.md). it is okay to change it to ignore SA1118, but I chose to keep the existing code style.

dahlia: Okay. It's still confusing that the following argument has multiple lines though… 🤔:

```
                rb.Transactions
                    .Select(tx => Transaction<T>.Deserialize(tx.ToArray()))
                    .ToList(),
```

longfin:
PTAL @planetarium/libplanet

limebell (approved):

Libplanet/Blocks/Block.cs
moreal: @dahlia 

I agree with your opinions:

> it causes circular dependency which is a code smell.
> I believe Block<T> should not expose its state that “it should calculate its StateRootHash but it hasn't been yet at this moment.”

Step by step:

1. To calculate and have `StateRootHash` of `Block<T>`, `Block<T>.Mine()` needs to get `IKeyValueStore` as a parameter at least. It can be smoothly resolved. But if the `BlockChain<T>`does not use `TrieStateStore`, since the time, `Block<T>` loses the duty to do that. A block's shape and the `IStateStore` have strong relations since the block has a field to represent a state of a blockchain.
1. I also agree with your opinion but it has something in common with the above descriptions. As you said, to avoid exposing the state, we must force to use some Merkle trie implementations like `TrieStateStore` to every user. I wonder libplanet's direction in the future for it. If it is the direction, I wonder how `IStateStore` can be used. cc @planetarium/libplanet

area363 (approved):

dahlia (approved):

moreal (approved):

PR #1124, opened 2020-12-10
Improve tx selecting performance

longfin:
This PR improves tx selecting performance in `BlockChain<T>.MineBlock()` by storing and reusing tx's nonce instead of calculating them per each txs.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -780,6 +780,8 @@ namespace Libplanet.Blockchain
 
             foreach (Transaction<T> tx in stagedTransactions)
             {
+                // We don't care about nonce ordering here because `.ListStagedTransactions()`
+                // returns already ordered transactions by its nonce.
                 if (!storedNonces.ContainsKey(tx.Signer))
                 {
                     storedNonces[tx.Signer] = Store.GetTxNonce(Id, tx.Signer);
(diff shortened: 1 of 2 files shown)


moreal (approved):

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -775,9 +775,25 @@ public long GetNextTxNonce(Address address)
...
+                    storedNonces[tx.Signer] = Store.GetTxNonce(Id, tx.Signer);
+                }
+
+                if (nextNonces.TryGetValue(tx.Signer, out long prevNonce))
+                {
+                    nextNonces[tx.Signer] = prevNonce + 1;
+                }
+                else
+                {
+                    nextNonces[tx.Signer] = storedNonces[tx.Signer];
```
limebell, line 794: Shouldn't we add 1 to this?

longfin: It's fine at this moment because I also [changed the condition](https://github.com/planetarium/libplanet/pull/1124/files#diff-ecd6275ea8686534dc160d341079910ede8da3d2ec020236582bdb0906cddc21R831).

but on second thought, adding 1 here is more natural to me too. I'll change it and condition.

area363 (approved):

moreal (commented):
These changes seem based on that `ListStagedTransactions` returns transactions order by tx nonce. How about leaving a comment for it to help code reading?

moreal (approved):

moreal (approved):

longfin:
@planetarium/libplanet PTAL

dahlia (approved):

moreal (approved):

PR #1131, opened 2020-12-15
Parametrize the transaction staging strategy & make it volatile by default

dahlia:
This patch parametrizes `BlockChain<T>`'s strategy to stage transactions through the new `IStagePolicy<T>` interface, and introduced `VolatileStagePolicy<T>`, which is an only built-in & default implementation.

See also the changelog.


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -624,7 +639,7 @@ public void StageTransaction(Transaction<T> transaction)
                 if (!_transactions.ContainsKey(transaction.Id))
                 {
                     _transactions[transaction.Id] = transaction;
```
longfin, line 641: Even if we manage the stage in-memory, it will still use a disk if add the `transaction` to the store here.

dahlia: Applied: a496a0edb6785417491b56088f25fe5ac652e6cd.

dahlia:
@planetarium/libplanet PTAL.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -615,21 +635,9 @@ public void StageTransaction(Transaction<T> transaction)
...
-                {
-                    _transactions[transaction.Id] = transaction;
-                    Store.StageTransactionIds(ImmutableHashSet.Create(transaction.Id));
-                }
-            }
-            finally
+            if (!StagePolicy.HasStaged(this, transaction.Id, includeUnstaged: true))
             {
-                _rwlock.ExitWriteLock();
+                StagePolicy.Stage(this, transaction);
```
longfin, line 640: Is `StagePolicy.Stage()` thread-safe?

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1192,21 +1183,25 @@ public IImmutableSet<TxId> GetStagedTransactionIds()
...
                 }
 
+                ISet<TxId> txIds = StagePolicy.Iterate(this)
+                    .Where(tx => maxNonces.TryGetValue(tx.Signer, out long nonce) &&
+                        tx.Nonce <= nonce)
+                    .Select(tx => tx.Id)
+                    .ToImmutableHashSet();
+                foreach (TxId txId in txIds)
+                {
+                    StagePolicy.Unstage(this, txId);
```
longfin, line 1199: Is `StagePolicy.Unstage()` also thread-safe?

Libplanet/Net/Swarm.MessageHandlers.cs
```diff
@@ -162,12 +163,15 @@ private void ProcessMessageHandler(object target, Message message)
...
-            IEnumerable<Transaction<T>> txs = getTxs.TxIds
-                .Where(txId => _store.ContainsTransaction(txId))
-                .Select(BlockChain.GetTransaction);
-
-            foreach (Transaction<T> tx in txs)
+            foreach (TxId txid in getTxs.TxIds)
             {
+                Transaction<T> tx = BlockChain.GetTransaction(txid);
+
+                if (tx is null)
```
longfin, line 170: Can `BlockChain<T>.GetTransaction()` return `null`?

longfin (approved):

Libplanet/Blockchain/BlockChain.cs
dahlia: It should be so (interface-wise), and I believe `VolatileStagePolicy<T>`, the only implementation, is so.  (It uses `ConcurrentDictionary<K, V>` under the hood.)

Libplanet/Net/Swarm.MessageHandlers.cs
dahlia: If `txid` is invalid there's the possibility.

Libplanet/Blockchain/BlockChain.cs
longfin: OK, if `.Unstage()` is too, it seems that we can strip `_rwlock` in `BlockChain<T>.UnstageTransaction()`.

Libplanet/Net/Swarm.MessageHandlers.cs
longfin: 🆗

Libplanet/Blockchain/BlockChain.cs
dahlia: I made every operation on `IStagePolicy<T>` (protocol-wise) & `VolatileStagePolicy<T>` (implementation-wise) sure they are thread safe: 131ed7ddd7bccc7054cabfa5f06ac3097a7a1796.

dahlia:
@planetarium/libplanet PTAL.

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
```diff
@@ -0,0 +1,152 @@
...
+                _queue.Add(transaction.Id);
+            }
+            finally
+            {
+                _lock.ExitWriteLock();
+                if (stageAgain)
+                {
+                    _lock.ExitUpgradeableReadLock();
+                }
+            }
```
longfin, line 72: How about this?

```suggestion
            try
            {
                if (!_set.TryAdd(transaction.Id, transaction))
                {
                    _lock.EnterUpgradeableReadLock();

                    if (_queue.Contains(transaction.Id))
                    {
                        return;
                    }
                }

                _lock.EnterWriteLock();
                _queue.Add(transaction.Id);
            }
            finally
            {
                _lock.ExitWriteLock();
                if (_lock.IsUpgradableReadLockHeld)
                {
                    _lock.ExitUpgradeableReadLock();
                }
            }
```

Or, I guess it can be fine if we always hold `EnterUpgradeableReadLock` like below

```csharp
            try
            {
                _lock.EnterUpgradeableReadLock();
                if (!_set.TryAdd(transaction.Id, transaction) && _queue.Contains(transaction.Id))
                {
                    return;
                }

                _lock.EnterWriteLock();
                try
                {
                    _queue.Add(transaction.Id);
                }
                finally
                {
                    _lock.ExitWriteLock();
                }
            }
            finally
            {
                _lock.ExitUpgradeableReadLock();
            }

```

dahlia: What I intended is that `_lock` is only for `_queue` as `_set` in itself is a `ConcurrentDictionary<TxId, Transaction<T>>`.  Do you think both `_queue` and `_set` should be locked?

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
```diff
@@ -0,0 +1,152 @@
...
+        private readonly List<TxId> _queue;
+
+        private readonly ReaderWriterLockSlim _lock;
+
+        /// <summary>
+        /// Creates a new <see cref="VolatileStagePolicy{T}"/> instance.
+        /// </summary>
+        public VolatileStagePolicy()
+        {
+            _set = new ConcurrentDictionary<TxId, Transaction<T>>();
```
moreal, line 30: It seems there is no code to delete `_set`'s element. Is there a default capacity?

dahlia: No, we need to implement lifetiming, but I'm going to do it on a different PR.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -441,6 +456,11 @@ public bool ContainsBlock(HashDigest<SHA256> blockHash)
         /// <see cref="Transaction{T}"/> with a given <paramref name="txId"/>.</exception>
         public Transaction<T> GetTransaction(TxId txId)
         {
+            if (StagePolicy.Get(this, txId, includeUnstaged: true) is { } tx)
+            {
+                return tx;
+            }
+
             _rwlock.EnterReadLock();
```
riemannulus, line 464: Are doesn't need `_rwlock.ExitWriteLock()` in this method?

riemannulus, line 464: Are doesn't need _rwlock.ExitWriteLock() in this method?

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
longfin: > the first suggestion might be a problem when any exception is thrown before _lock.EnterWriteLock() is reached.

Right. then...

```suggestion
            try
            {
                if (!_set.TryAdd(transaction.Id, transaction))
                {
                    _lock.EnterUpgradeableReadLock();

                    if (_queue.Contains(transaction.Id))
                    {
                        return;
                    }
                }

                _lock.EnterWriteLock();
                try
                {
                    _queue.Add(transaction.Id);
                }
                finally
                {
                   _lock.ExitWriteLock();
                }
            }
            finally
            {
                if (_lock.IsUpgradableReadLockHeld)
                {
                    _lock.ExitUpgradeableReadLock();
                }
            }
```

> The second suggestion seems more coarse-grained than before.

It's also right, but IMHO, it doesn't affect the parallelism seriously. 😕

longfin (approved):
There are [comments](https://github.com/planetarium/libplanet/pull/1131#pullrequestreview-555454598) about the style of the code left, but there seems to be no big problem with the behavior right now.

moreal (approved):
LGTM.

Libplanet/Blockchain/BlockChain.cs
dahlia: You mean, why `_rwlock` is necessary here? It's for protecting `_transactions` from other threads.

riemannulus: I misread about the lock. thanks for your answer.

riemannulus (dismissed):

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
dahlia: @longfin Applied your suggestion: 97263d76c.

dahlia:
@planetarium/libplanet I rebased on the current main.  Please review this again!

riemannulus (approved):

longfin (approved):

moreal (approved):

limebell (approved):

PR #1133, opened 2020-12-17
fix: TransferAsset vulnerability

riemannulus:
Previously, TransferAsset used the previously imported GetBalance of the recipient, so if the recipient and the sender were the same, there was a phenomenon that gold was duplicated continuously. Fix that phenomenon.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -90,6 +90,9 @@ To be released.
 
 ### Bug fixes
 
+ -  Fixed a vulnerability that `AccountStateDeltaImpl.TransferAsset()` if
+    sender and recipient are same. [#1133]
+
 ### CLI tools
 
 [#795]: https://github.com/planetarium/libplanet/issues/795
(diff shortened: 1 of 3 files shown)


longfin (changes requested):
It seems to need a changelog.

riemannulus:
> 
> 
> It seems to need a changelog.

appended at https://github.com/planetarium/libplanet/pull/1133/commits/376085c860c65a828a0448c54e476f6ad5aa3710

dahlia:
I believe this patch must be applied to 0.10-maintenance first, and then be ported to main later.

riemannulus:
> 
> 
> I believe this patch must be applied to 0.10-maintenance first, and then be ported to the main branch later.

Then I'll make 0.10-maintenance ported version PR too.

dahlia:
> Then I'll make 0.10-maintenance ported version PR too.

@riemannulus FYI GitHub allows you to change the existing PR's target branch.

riemannulus:
It merged at https://github.com/planetarium/libplanet/pull/1134. Closed this PR

PR #1134, opened 2020-12-17
fix: TransferAsset vulnerability

riemannulus:
Previously, TransferAsset used the previously imported GetBalance of the recipient, so if the recipient and the sender were the same, there was a phenomenon that gold was duplicated continuously. Fix that phenomenon.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,8 +6,12 @@ Version 0.10.3
 
 To be released.
 
- -  Fixed a vulnerability that `AccountStateDeltaImpl.TransferAsset()` if
-    sender and recipient are same. [[#1134]]
+ -  Fixed a vulnerability of the `IAccountStateDelta.TransferAsset()`'s internal
+    implementation that it had doubled the balance for transferring to themselves.
+    *Since this behavioral change breaks the protocol-level compatibility if any actions
+    in your app had used `IAccountStateDelta.TransferAsset()`, you must make the action
+    to conditionally work as it had done for the hard-coded block index before
+    this patch is applied, or you may be possible to hard-fork the network.*  [[#1134]]
 
 [#1134]: https://github.com/planetarium/libplanet/pull/1134
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@ Version 0.10.3
 
 To be released.
 
+ -  Fixed a vulnerability that `AccountStateDeltaImpl.TransferAsset()` if
+    sender and recipient are same. [[#1134]]
+
+[#1134]: https://github.com/planetarium/libplanet/pull/1134
+
 
 Version 0.10.2
 --------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,12 +6,14 @@ Version 0.10.3
 
 To be released.
 
- -  Fixed a vulnerability of the `IAccountStateDelta.TransferAsset()`'s internal
-    implementation that it had doubled the balance for transferring to themselves.
-    *Since this behavioral change breaks the protocol-level compatibility if any actions
-    in your app had used `IAccountStateDelta.TransferAsset()`, you must make the action
-    to conditionally work as it had done for the hard-coded block index before
-    this patch is applied, or you may be possible to hard-fork the network.*  [[#1134]]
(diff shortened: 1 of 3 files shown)


longfin (approved):

limebell (approved):

Libplanet/Action/AccountStateDeltaImpl.cs
```diff
@@ -180,6 +185,15 @@ public IAccountStateDelta BurnAsset(Address owner, FungibleAssetValue value)
             );
         }
 
+        [Pure]
+        private FungibleAssetValue GetBalance(
+            Address address,
+            Currency currency,
+            IImmutableDictionary<(Address, Currency), BigInteger> state) =>
```
moreal, line 192: How about renaming as others like `fungibleAssets` or better? because I think the word `state` can cause confusion with `state` used in `GetState`, `updatedStates`.

dahlia: It would be better to have a different name like `balances`.  At least, the singular `state` should be a plural `states`.

CHANGES.md
```diff
@@ -6,6 +6,11 @@ Version 0.10.3
 
 To be released.
 
+ -  Fixed a vulnerability that `AccountStateDeltaImpl.TransferAsset()` if
+    sender and recipient are same. [[#1134]]
```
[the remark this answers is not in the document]
riemannulus: https://github.com/planetarium/libplanet/pull/1134/commits/37b996fc3a7717d0853dbe077569279412e8c67d amended.

Libplanet/Action/AccountStateDeltaImpl.cs
riemannulus: https://github.com/planetarium/libplanet/pull/1134/commits/22aa91a244839f71dfd6155a734d0151f0182d7f amended.

longfin (dismissed):

dahlia (approved):

CHANGES.md
```diff
@@ -6,6 +6,15 @@ Version 0.10.3
 
 To be released.
 
+ -  Fixed a vulnerability of the `IAccountStateDelta.TransferAsset()`'s internal
+    implementation that it had doubled the balance for transferring to themselves.
+    *Since this behavioral change breaks the protocol-level compatibility if any actions
+    in your app had used `IAccountStateDelta.TransferAsset()`, you must make the action
+    to conditionally work as it had done for the hard-coded block index before
+    this patch is applied, or you may be possible to hard-fork the network.*  [[#1134]]
```
moreal, line 14: It seems longer than 80 and it makes check `check-build` failed. 😢

moreal (changes requested):

CHANGES.md
riemannulus: 74a5111866fcef81b4c94048c34ea67e69ff984b amended.

junorouse (approved):
🚢

dahlia:
> *Since this behavioral change breaks the protocol-level compatibility if any actions in your app had used `IAccountStateDelta.TransferAsset()`, you must make the action to conditionally work as it had done for the hard-coded block index before this patch is applied, or you may be possible to hard-fork the network.*

Although it suggest client code to leave the existing behavior for compatibility besides the new behavior in their side, it is even impossible to implement the existing behavior by themselves… So we need to provide backward compatibility in our side (Libplanet) first.

CHANGES.md
```diff
@@ -6,6 +6,17 @@ Version 0.10.3
...
 To be released.
 
+ -  Fixed a vulnerability of the `IAccountStateDelta.TransferAsset()`'s
+    internal implementation that it had doubled the balance for transferring
+    to themselves. *Since this behavioral change breaks the protocol-level
+    compatibility if any actions in your app had used
+    `IAccountStateDelta.TransferAsset()`, you must make the action to
+    conditionally work as it had done for the hard-coded block index before
+    this patch is applied, or you may be possible to
+    hard-fork the network.*  [[#1134]]
```
limebell, line 16:
```suggestion
    this patch is applied, or you should hard-fork the network.*
    [[#1134]]
```

dahlia:
As I have no permission to rebase or add commits to on this pull request, I opened a new pull request: #1152.

PR #1141, opened 2021-01-05
Register WatchTip at beggining of MineBlock

limebell:
Mining hadn't been cancelled if the tip of the chain is changed in specific situation. This patch register event handler at the very beginning of `MineBlock()` to cover the case where tip is changed before event registration.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -101,6 +101,9 @@ To be released.
 
 ### Bug fixes
 
+ -  Fixed a bug where `BlockChain<T>.MineBlock()` was not automatically
+    cancelled when the tip of the chain was changed occasionally.  [[#1141]]
+
 ### CLI tools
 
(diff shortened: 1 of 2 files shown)


longfin (commented):
It would be better if we have a changelog about it.

dahlia (approved):

limebell:
> It would be better if we have a changelog about it.

Added. Please review again :)

longfin (dismissed):

moreal (approved):

riemannulus (dismissed):

dahlia (approved):
I'm sorry, but I've just merged another pull request.  Could you rebase this on the current main again?

limebell:
Sure, rebased.

dahlia (approved):

riemannulus (approved):

PR #1143, opened 2021-01-05
Add property `IActionContext.IsBlockAction`

riemannulus:
Previously, there was no way in the action itself to check whether it was a block action or not. You can use the IActionContext.IsBlockAction property to know if the action itself is a block action or not.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ To be released.
  -  Removed `Swarm<T>.GetTrustedStateCompleterAsync()` method.  [[#1117]]
  -  Removed `trustedStateValidators` parameter from `Swarm<T>.PreloadAsync()`
     method.  [[#1117]]
+ -  Added `IActionContext.BlockAction` property. [[#1143]]
 
 ### Backward-incompatible network protocol changes
  -   `Swarm<T>` became no longer retry when `Swarm<T>` receives
@@ -146,6 +147,7 @@ To be released.
 [#1136]: https://github.com/planetarium/libplanet/pull/1136
 [#1137]: https://github.com/planetarium/libplanet/pull/1137
 [#1141]: https://github.com/planetarium/libplanet/pull/1141
+[#1143]: https://github.com/planetarium/libplanet/pull/1143
 
 
 Version 0.10.2
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -1854,7 +1854,7 @@ namespace Libplanet.Tests.Blockchain
             Assert.Equal(
                 (Integer)2,
                 (Integer)blockActionEvaluation.OutputStates.GetState(miner));
-            Assert.True(blockActionEvaluation.InputContext.IsBlockAction);
+            Assert.True(blockActionEvaluation.InputContext.BlockAction);
 
             _blockChain.ExecuteActions(genesis);
             _blockChain.Append(
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -1567,6 +1567,11 @@ namespace Libplanet.Tests.Blockchain
             chain.StageTransaction(tx1);
             await chain.MineBlock(_fx.Address1);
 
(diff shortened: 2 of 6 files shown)


CHANGES.md
```diff
@@ -76,6 +76,7 @@ To be released.
  -  Added `IProtocol` interface.  [[#1120]]
  -  Added `KademliaProtocol` class which implements `IProtocol`.
     [[#1120], [#1135]]
+ -  Added `IActionContext.IsBlockAction` property. [[#1143]]
```
longfin, line 79: IMHO `IActionContext.BlockAction` is more natural to me. (`IsBlockAction` seems a method because it's a statement).

longfin, line 79: And also I guess we may need to move this section to "Backward-incompatible API changes" because `IActionContext` was open to users.

dahlia (commented):
Although this might be out of scope, how about adding `TxId? IActionContext.TxId` property instead of `IsBlockAction`?  By introducing `TxId` property, an action code can query both of which transaction it belongs to (if it is not a block action) and whether it is a block action.

`IAction.Execute()` code could be written like:

```csharp
if (context.TxId is null)
{
    // When it is a block action
}
```

… instead of:

```csharp
if (context.IsBlockAction)
{
    // When it is a block action
}
```

When an action needs to behave differently depending on whether it is a block action or not and what transaction it belongs to if it's not a block action, the action code can be written like:

```csharp
if (context.TxId is { } txid)
{
    // When it is not a block action.
    // The "txid" here refers to the transaction it belongs to.
}
else
{
    // When it is a block action.
}
```

riemannulus:
>  By introducing TxId property, an action code can query both of which transaction it belongs to (if it is not a block action) and whether it is a block action.

Knowing the TxId, what's good about the action code? (Except for block action verification.)

CHANGES.md
riemannulus: amended https://github.com/planetarium/libplanet/pull/1143/commits/c6d761364eab4aa0297b920d14d3cf534dfc6841

riemannulus: amended https://github.com/planetarium/libplanet/pull/1143/commits/c6d761364eab4aa0297b920d14d3cf534dfc6841

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -1854,6 +1854,7 @@ public void EvaluateBlockAction()
             Assert.Equal(
                 (Integer)2,
                 (Integer)blockActionEvaluation.OutputStates.GetState(miner));
+            Assert.True(blockActionEvaluation.InputContext.BlockAction);
```
dahlia, line 1857: Could we have a more test case where this property turns out to be `false`?

riemannulus: https://github.com/planetarium/libplanet/pull/1143/commits/f46d693c43f9f1d76f33b1f36489408580812bb0 I add it

riemannulus:
@libplanet PTAL 🙏

limebell (approved):

Libplanet/Action/IActionContext.cs
```diff
@@ -70,6 +70,12 @@ public interface IActionContext
         /// </summary>
         HashDigest<SHA256>? PreviousStateRootHash { get; }
 
+        /// <summary>
+        /// Boolean flag indicating if this action was executed as a BlockAction.
```
dahlia, line 74: It's trivial though…

```suggestion
        /// Whether this action is executed as a block action.
        /// <c>false</c> if it belongs to a transaction.
```

dahlia (approved):

Libplanet/Action/IActionContext.cs
riemannulus: amended https://github.com/planetarium/libplanet/pull/1143/commits/99d1396f3d958304550794d358eff604327ec3e6

dahlia: @riemannulus It seems you've missed the second statement.

riemannulus: you're right, I amended 105b7ff one more time

dahlia (approved):

limebell (approved):

longfin (approved):

Issue #1142, opened 2021-01-05
Libplanet-level protocol versioning


dahlia:
Besides `AppProtocolVersion`, introduce Libplanet-level protocol versioning.

### Requirements

- Behaviors should be changed in the newer version.
- Even if a behavior is changed in a new protocol version, blocks created in the older versions and transactions/actions belonging to them should behave as it had done.
- Once a new version is introduced on the blockchain, blocks of older protocol versions should be disallowed.
- Each node does not automatically join newer version without its own explicit choice (e.g., application update).

### Specification

- If a change breaks the consensus of the protocol, this change is considered a new protocol version.
- Every block contains what protocol version it belongs to, or it's treated as the version zero.
- Every transaction/action in a block behaves according to the protocol version of the block.
- All blocks except the genesis block is valid if and only if its protocol version is equal to or newer than its previous block's version.
- A node creates only blocks of the latest protocol version it is aware of.
- A node accepts only the block of the protocol version it is aware of, and blocks of the version it is not aware of is considered invalid.
     - Therefore, networks are partitioned (hard forked) if a new protocol version is introduced.
     - Suppressing hard-forking can be achieved through application-level update policy (i.e., `AppProtocolVersion`).

PR #1147, opened 2021-01-06
Block<T>.ProtocolVersion & BlockHeader.ProtocolVersion

dahlia:
This implements #1142.


dahlia:
@planetarium/libplanet PTAL.

CHANGES.md
```diff
@@ -38,11 +50,10 @@ To be released.
  -  Added `BlockExceedingTransactionsException` class.  [[#1104], [#1110]]
  -  Added `BlockChain<T>.GetStagedTransactionIds()` method.  [[#1089]]
  -  (Libplanet.RocksDBStore) Added `maxTotalWalSize`, `keepLogFileNum` and
-    `maxLogFileSize` parameters into `RocksDBStore` constructor.
+    `maxLogFileSize` parameters into `RocksDBStore()` constructor.
     [[#1065], [#1102], [#1132]]
  -  Added `Swarm<T>.BlockDemand` property.  [[#1068], [#1102]]
  -  Added `BlockDemand` struct.  [[#1068], [#1102]]
- -  Added `Block<T>.Header` property.  [[#1070], [#1102]]
```
limebell, line 45: Was this change intended?

Libplanet.Tests/Blocks/BlockTest.cs
```diff
@@ -338,19 +338,22 @@ public void EvaluateActionsPerTx()
 
             Transaction<DumbAction>[] blockIdx2Txs =
             {
+                // Note that these timestamps in themselves does not have any meanings but are
+                // only arbitrary.  These purpose to make their evaluation order in a block
```
riemannulus, line 342: Is there any reason insert two space?

Libplanet/Blocks/Block.cs
```diff
@@ -21,6 +21,11 @@ namespace Libplanet.Blocks
     public class Block<T>
         where T : IAction, new()
     {
+        /// <summary>
+        /// The most latest protocol version.
+        /// </summary>
+        public const int CurrentProtocolVersion = 1;
```
limebell, line 27: Latest protocol version is okay with constant?

Libplanet/Blocks/Block.cs
```diff
@@ -126,9 +135,11 @@ public Block(Bencodex.Types.Dictionary dict)
         {
         }
 
+        // FIXME: Should this be
```
limebell, line 138: Something is missing I guess...

Libplanet.Tests/Blocks/BlockHeaderTest.cs
```diff
@@ -12,13 +14,124 @@ public class BlockHeaderTest : IClassFixture<BlockFixture>
...
+                    BlockHeader.TimestampFormat,
+                    CultureInfo.InvariantCulture
+                ),
+                preEvaluationHash: TestUtils.GetRandomBytes(32).ToImmutableArray(),
+                stateRootHash: ImmutableArray<byte>.Empty
+            );
+
+            Assert.Throws<InvalidBlockProtocolVersionException>(() =>
+                header.Validate(DateTimeOffset.UtcNow)
+            );
```
riemannulus, line 124: Could we have a more test case where valid block protocol version? Like https://github.com/planetarium/libplanet/pull/1147/files#diff-b70ec4dc8f1efa4e59f3584b07849b9155f1f3d448a2e026a5d80e61c3de4bf5R152

Libplanet/Blocks/BlockHeader.cs
```diff
@@ -196,6 +210,11 @@ public Bencodex.Types.Dictionary ToBencodex()
                 .Add(NonceKey, Nonce.ToArray())
                 .Add(HashKey, Hash.ToArray());
 
+            if (ProtocolVersion != 0)
+            {
+                dict = dict.Add(ProtocolVersionKey, ProtocolVersion);
```
limebell, line 215: How about remove if statement and force to record protocol version afterward?

Libplanet/Blocks/Block.cs
riemannulus, line 138: It seems does not write the word to follow.

CHANGES.md
dahlia: I just moved this line upper.

Libplanet/Blocks/Block.cs
dahlia: Yeah, this shouldn't be changed by applications, since it's for Libplanet-level protocol versioning.

Libplanet/Blocks/BlockHeader.cs
dahlia: It has to accept the field being omitted for backward compatibility (which is the point of the [spec][1]).

[1]: https://github.com/planetarium/libplanet/issues/1142

Libplanet/Blocks/Block.cs
limebell: Wording "current" is confusing to me. What this constant means?

Libplanet.Tests/Blocks/BlockTest.cs
dahlia: AFAIK it's an [old convention][1], and I believe it improves readability on plain text (two spaces eventually will be reduced to one space when it's rendered anyway).

[1]: https://en.wikipedia.org/wiki/Sentence_spacing

Libplanet/Blocks/Block.cs
dahlia: It's the latest protocol version that this revision of Libplanet is aware of.

Libplanet.Tests/Blocks/BlockHeaderTest.cs
dahlia: The link seems lost.

Libplanet/Blocks/Block.cs
limebell: Understood. Thx :)

Libplanet/Blocks/BlockHeader.cs
limebell: 👌

Libplanet/Blocks/Block.cs
dahlia: Completed the sentence.

dahlia: Completed the sentence.

dahlia:
@planetarium/libplanet Rebased.  PTAL.

Libplanet.Tests/Blocks/BlockHeaderTest.cs
riemannulus: Like this https://github.com/planetarium/libplanet/blob/main/Libplanet.Tests/Blocks/BlockHeaderTest.cs#L38-L39

junorouse (dismissed):
I understand that libplanet separately control the flow of actions through protocol version management (this patch) instead of creating a new action file (i.e., Buy {2,3,4}.cs), am I right? -- for proper patching of https://github.com/planetarium/lib9c/pull/248

If so, this PR looks OK for me. However, from an attacker's point of view, if an attacker has enough mining power, the block's previous version can continue to be mined even if the new protocol version introduces.

There are some mitigations for it:

1. Set the maximum number of blocks(=limit) allowed with the previous protocol version.
2. Like lib9c, use an authorized miner to introduce a new version.
3. Change to a consensus algorithm to that selecting the next block proposer randomly.
4. ...

Libplanet.Tests/Blocks/BlockHeaderTest.cs
dahlia: Added:

```diff
--- a/Libplanet.Tests/Blocks/BlockHeaderTest.cs
+++ b/Libplanet.Tests/Blocks/BlockHeaderTest.cs
@@ -98,6 +98,13 @@ namespace Libplanet.Tests.Blocks
             Assert.Equal(expected, new BlockHeader(expected).ToBencodex());
         }
 
+        [Fact]
+        public void ValidateValidHeader()
+        {
+            _fx.Genesis.Header.Validate(DateTimeOffset.UtcNow);
+            _fx.Next.Header.Validate(DateTimeOffset.UtcNow);
+        }
+
         [Fact]
         public void ValidateProtocolVersion()
         {
```

dahlia:
@junorouse Thanks for your review and suggestions!

> However, from an attacker's point of view, if an attacker has enough mining power, the block's previous version can continue to be mined even if the new protocol version introduces.

Even if we don't merge this PR, that attackers can outpace honest nodes is the critical problem in itself and we need to address it as soon as possible.  Therefore, IMHO we'd better deal with it in the separate issue/patches.

CHANGES.md
```diff
@@ -99,7 +119,6 @@ To be released.
  -  `Swarm<T>` became not to fill states from trusted peers, because now states
     can be validated rather than trusted due to MPT.  [[#1117]]
  -  `HashDigest<SHA256>` became serializable.  [[#795], [#1125]]
-
```
moreal, line 102: Is this change intended? Of course, it seems not to affect to the rendered result because it is markdown.

riemannulus (approved):

CHANGES.md
dahlia: No, it's removed by mistake.  I'm going to revert it.

dahlia: Removed.

moreal (approved):

longfin (approved):

PR #1152, opened 2021-01-08
Fix TransferAsset vulnerability (take 2)

dahlia:
I rebased #1134 on the current main, and made it to maintain backward compatibility with [protocol version] 0.

Please review this instead of #1134. The commit I added besides #1134 is: 506ae01d5b6d29d93acdbcca0a7a10c79065cb16.

Closes #1134.

[protocol version]: https://github.com/planetarium/libplanet/issues/1142

### DIFF
--- a/Libplanet.Tests/Action/AccountStateDeltaTest.cs
+++ b/Libplanet.Tests/Action/AccountStateDeltaTest.cs
@@ -1,0 +1,331 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Numerics;
+using Bencodex.Types;
+using Libplanet.Action;
+using Libplanet.Assets;
+using Libplanet.Blockchain;
+using Libplanet.Crypto;
+using Libplanet.Store;
+using Libplanet.Store.Trie;
+using Libplanet.Tests.Blockchain;
+using Libplanet.Tests.Common.Action;
+using Libplanet.Tx;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Libplanet.Tests.Action
+{
+    public abstract class AccountStateDeltaTest
+    {
+        protected readonly PrivateKey[] _keys;
+        protected readonly Address[] _addr;
+        protected readonly Currency[] _currencies;
+        protected readonly IImmutableDictionary<Address, IValue> _states;
+        protected readonly IImmutableDictionary<(Address, Currency), BigInteger> _assets;
+        protected readonly IAccountStateDelta _init;
+
+        protected AccountStateDeltaTest(ITestOutputHelper output)
+        {
+            _keys = new[]
+            {
+                new PrivateKey(),
+                new PrivateKey(),
+                new PrivateKey(),
+            };
+
+            _addr = _keys.Select(AddressExtensions.ToAddress).ToArray();
+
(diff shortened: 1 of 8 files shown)


junorouse (approved):
:ship:

dahlia:
@planetarium/libplanet PTAL.

Libplanet/Action/AccountStateDeltaImpl.cs
```diff
@@ -16,11 +16,13 @@ namespace Libplanet.Action
     [Pure]
     internal class AccountStateDeltaImpl : IAccountStateDelta
     {
-        private readonly AccountStateGetter _accountStateGetter;
-        private readonly AccountBalanceGetter _accountBalanceGetter;
-        private readonly Address _signer;
-        private IImmutableDictionary<Address, IValue> _updatedStates;
-        private IImmutableDictionary<(Address, Currency), BigInteger> _updatedFungibleAssets;
+#pragma warning disable SA1401
```
limebell, line 19:
```suggestion
#pragma warning disable SA1401  // FieldsMustBePrivate
```
Can you please add something like this for code legibility?

limebell (commented):
It seems like constructor of `AccountStateDeltaImpl` is also in `Transaction.cs`. It is okay to leave it just as is?

Libplanet/Action/AccountStateDeltaImpl.cs
dahlia: Okay, I'm rather going to make them properties.

dahlia:
> It seems like constructor of `AccountStateDeltaImpl` is also in `Transaction.cs`. It is okay to leave it just as is?

It's used by `Transaction<T>.Create()`.  I believe it's okay, because new transactions should behave according to the latest protocol version?

Libplanet/Action/AccountStateDeltaImpl.cs
dahlia: Fixed: 46aad246d7d43595daa9b531ed6c5fec96c7a0c2.

Libplanet.Tests/Action/AccountStateDeltaTest.cs
```diff
@@ -0,0 +1,331 @@
...
+using Libplanet.Tx;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Libplanet.Tests.Action
+{
+    public abstract class AccountStateDeltaTest
+    {
+        protected readonly PrivateKey[] _keys;
+        protected readonly Address[] _addr;
```
moreal, line 24: How about renaming this variable name as plural like `_addresses` as other variables do? Is it too long? 🤔

moreal (approved):

Libplanet.Tests/Action/AccountStateDeltaTest.cs
dahlia: First of all, these names were not named in this pull request.  The file was renamed and slightly modified (class name, etc).

I guess these names were named such short with intention so that they look like prefixes rather than their own names.  If you look into their usages, they indeed are always used with indexers…

dahlia:
@planetarium/libplanet Rebased; PTAL.

riemannulus (approved):

moreal (approved):

PR #1154, opened 2021-01-13
Test a block with index after nonexistent

### LINKED ISSUE #947: BlockChain<T>.Append() does not validate the given block's index
Where `chain is BlockChain<T>` and `nextBlock is Block<T>`, `chain.Append(nextBlock)` should accept `nextBlock` only if `nextBlock.Index == chain.Count`, but the current implementation does not throw any exception.

dahlia:
I claimed `BlockChain<T>.Append()` does not validate block indices in the issue #947, but it turns out false.  However, I added a more test case if it invalidates a block with index after nonexistent.

Close #947.


moreal (approved):
👍

limebell (approved):

PR #1163, opened 2021-01-18
Add `clearBufferInterval` parameter into `DelayedActionRenderer<T>()` constructor

riemannulus:
Add `clearBufferInterval` parameter into `DelayedActionRenderer<T>()` constructor for clear `DelayedActionRenderer<T>._bufferedActionRender` and `DelayedActionRenderer<T>._bufferedActionUnrender` each block interval.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -90,7 +90,7 @@ To be released.
  -  Added `IProtocol` interface.  [[#1120]]
  -  Added `KademliaProtocol` class which implements `IProtocol`.
     [[#1120], [#1135]]
- -  Added `clearBufferInterval` parameters into `DelayedActionRenderer<T>()`
+ -  Added `reorgResistantHeight` parameters into `DelayedActionRenderer<T>()`
     constructor. [[#1163]]
 
 ### Behavioral changes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -90,6 +90,8 @@ To be released.
  -  Added `IProtocol` interface.  [[#1120]]
  -  Added `KademliaProtocol` class which implements `IProtocol`.
     [[#1120], [#1135]]
+ -  Added `clearBufferInterval` parameters into `DelayedActionRenderer<T>()`
+    constructor. [[#1163]]
 
 ### Behavioral changes
 
@@ -178,6 +180,7 @@ To be released.
 [#1147]: https://github.com/planetarium/libplanet/pull/1147
 [#1152]: https://github.com/planetarium/libplanet/pull/1152
 [#1162]: https://github.com/planetarium/libplanet/pull/1162
+[#1163]: https://github.com/planetarium/libplanet/pull/1163
 
 
 Version 0.10.2
--- a/Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
+++ b/Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
@@ -319,6 +319,16 @@ namespace Libplanet.Blockchain.Renderers
             return Iterate().Reverse().ToImmutableArray();
         }
 
+        internal int GetBufferedActionRendererCount()
+        {
+            return _bufferedActionRenders.Count;
+        }
+
+        internal int GetBufferedActionUnRendererCount()
+        {
+            return _bufferedActionUnrenders.Count;
+        }
+
         /// <inheritdoc cref="DelayedRenderer{T}.OnTipChanged(Block{T}, Block{T}, Block{T}?)"/>
         protected override void OnTipChanged(
             Block<T> oldTip,
--- a/Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
+++ b/Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
@@ -71,7 +71,8 @@ namespace Libplanet.Blockchain.Renderers
         /// See also the <see cref="DelayedRenderer{T}.Confirmations"/> property.</param>
         /// <param name="reorgResistantHeight">Configures the height of blocks to maintain the
         /// <see cref="ActionEvaluation"/> buffer. Buffered <see cref="ActionEvaluation"/>s
-        /// that belong to blocks older than this height from the tip are gone.</param>
+        /// that belong to blocks older than this height from the tip are gone.
+        /// If zero, which is a default value, is passed the buffer is not cleared.</param>
         public DelayedActionRenderer(
             IActionRenderer<T> renderer,
             IStore store,
--- a/Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
+++ b/Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
@@ -58,7 +58,7 @@ namespace Libplanet.Blockchain.Renderers
 
         private HashDigest<SHA256>? _eventReceivingBlock;
         private Reorg? _eventReceivingReorg;
-        private long _clearBufferInterval;
+        private long _reorgResistantHeight;
 
         /// <summary>
         /// Creates a new <see cref="DelayedRenderer{T}"/> instance decorating the given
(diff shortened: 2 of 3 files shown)


Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
```diff
@@ -68,14 +69,21 @@ public class DelayedActionRenderer<T> : DelayedRenderer<T>, IActionRenderer<T>
         /// <param name="store">The same store to what <see cref="BlockChain{T}"/> uses.</param>
         /// <param name="confirmations">The required number of confirmations to recognize a block.
         /// See also the <see cref="DelayedRenderer{T}.Confirmations"/> property.</param>
-        public DelayedActionRenderer(IActionRenderer<T> renderer, IStore store, int confirmations)
+        /// <param name="clearBufferInterval">The required number of interval to remove stale
```
dahlia, line 72: It seems unclear what unit does this mean here by <q>number</q>: height of blocks, hours, or a kind of ticks?  How about this instead?

> Configures the height of blocks to maintain the `ActionEvaluation` buffer.  Buffered `ActionEvaluation`s that belong to blocks older than this height from the tip are gone.

dahlia (commented):
Instead of `clearBufferInterval`, which is imperative, how about a goal-oriented word: `reorgResistantHeight` or `bufferedBlocks`.

riemannulus:
> Instead of `clearBufferInterval`, which is imperative, how about a goal-oriented word: `reorgResistantHeight` or `bufferedBlocks`.
Agree, IMO `reorgResistantHeight` seems more meaningful.

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
riemannulus: changed on 826fbbc

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
```diff
@@ -68,14 +69,22 @@ public class DelayedActionRenderer<T> : DelayedRenderer<T>, IActionRenderer<T>
         /// <param name="store">The same store to what <see cref="BlockChain{T}"/> uses.</param>
         /// <param name="confirmations">The required number of confirmations to recognize a block.
         /// See also the <see cref="DelayedRenderer{T}.Confirmations"/> property.</param>
-        public DelayedActionRenderer(IActionRenderer<T> renderer, IStore store, int confirmations)
+        /// <param name="reorgResistantHeight">Configures the height of blocks to maintain the
+        /// <see cref="ActionEvaluation"/> buffer. Buffered <see cref="ActionEvaluation"/>s
+        /// that belong to blocks older than this height from the tip are gone.</param>
```
dahlia, line 74: We seem to miss the explanation the meaning of the default value (0).

```suggestion
        /// that belong to blocks older than this height from the tip are gone.
        /// If zero, which is a default value, is passed the buffer is not cleared.</param>
```

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
```diff
@@ -244,6 +253,27 @@ public void RenderBlockEnd(Block<T> oldTip, Block<T> newTip)
 
             _localRenderBuffer.Value = new Dictionary<HashDigest<SHA256>, List<ActionEvaluation>>();
             DiscoverBlock(oldTip, newTip);
+
+            if (_reorgResistantHeight == 0)
+            {
+                return;
+            }
+
+            var quotient = Math.DivRem(newTip.Index, _reorgResistantHeight, out _);
```
longfin, line 262: Is it okay that we clear all buffers before threshold(`newTip.Index - _reorgResistantHeight`) instead of dividing?

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
```diff
@@ -466,6 +506,25 @@ bool unrender
...
+            ConcurrentDictionary<HashDigest<SHA256>, List<ActionEvaluation>> buffer)
+        {
+            var invalidityBlock = new List<HashDigest<SHA256>>();
+            foreach (var (hash, evaluations) in buffer.Select(x => (x.Key, x.Value)))
+            {
+                if (evaluations.First().InputContext.BlockIndex < targetBlockIndex)
+                {
+                    invalidityBlock.Add(hash);
+                }
+            }
```
longfin, line 520: I guess LINQ can help us. (you may need to check it again 😅 )

```suggestion
            IEnumerable<HashDigest<SHA256>> invalidityBlockHashes = buffer
                .Select(x => (x.Key, x.Value)))
                .Where(x.Item2.First().InputContext.BlockIndex < targetBlockIndex)
                .Select(t => t.Item1);
```

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
riemannulus: Yes, that's okay. In this case, there is at least one checkpoint block in the buffer, so there is no case where the ActionEvaluation after the checkpoint block does not exist when a reorg occurs.

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
riemannulus: How about...

```suggestion
            IEnumerable<HashDigest<SHA256>> invalidityBlockHashes = buffer
                .Where(x => x.Value.First().InputContext.BlockIndex < targetBlockIndex)
                .Select(t => t.Key);
```

riemannulus: chage at a3b8adb

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
riemannulus: 1c969e6 added

dahlia (approved):

riemannulus:
@planetarium/libplanet PTAL 🙏

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
```diff
@@ -244,6 +254,27 @@ public void RenderBlockEnd(Block<T> oldTip, Block<T> newTip)
 
             _localRenderBuffer.Value = new Dictionary<HashDigest<SHA256>, List<ActionEvaluation>>();
             DiscoverBlock(oldTip, newTip);
+
+            if (_reorgResistantHeight == 0)
+            {
+                return;
+            }
+
+            var quotient = Math.DivRem(newTip.Index, _reorgResistantHeight, out _);
```
limebell, line 263: `quotient` seems not used elsewhere. How about just check `newTip.Index` < `_reorgResistantHeight` instead of `quotient` <= 0?

limebell (commented):
Buffer clearing is performed every `RenderBlockEnd()`? Why not run it every `reorgResistantHeight` blocks to reduce calculation?

Libplanet/Blockchain/Renderers/DelayedActionRenderer.cs
riemannulus: Previously I coded cleaning buffer every `reorgResistantHeight` as you say. That's why calculate `quotient`.
And yes, you're right. it working exactly same.

limebell: In my opinion, clear every interval looks better because clearing needs array iteration.

longfin: IMHO, clearing buffer per each render isn't a big matter because array iteration overhead isn't heavy.

limebell (dismissed):

riemannulus:
@planetarium/libplanet PTAL 🙏

limebell (approved):

dahlia (approved):

longfin (approved):

CHANGES.md
```diff
@@ -90,6 +90,8 @@ To be released.
  -  Added `IProtocol` interface.  [[#1120]]
  -  Added `KademliaProtocol` class which implements `IProtocol`.
     [[#1120], [#1135]]
+ -  Added `reorgResistantHeight` parameters into `DelayedActionRenderer<T>()`
```
moreal, line 93: Is the added parameter okay to be presented as plural?

```suggestion
 -  Added `reorgResistantHeight` parameter into `DelayedActionRenderer<T>()`
```
```suggestion
 -  Added the parameter `reorgResistantHeight` into `DelayedActionRenderer<T>()`
```

moreal (approved):

Issue #1155, opened 2021-01-13
Motivate miners to upgrade protocol version


dahlia:
The current design of [protocol versioning](https://github.com/planetarium/libplanet/issues/1142) does not motivate miners to produce a block with the latest protocol version.  Therefore, if a new protocol version makes them less profitable, they can reject the version and keep the older version endless.

In order to disincentive them to stick to the older protocol version, Libplanet needs to give priority to high-versioned chains (i.e., chains with the high-versioned tip) with being the canonical chain (best chain).  Among blocks with the same version, the block with the most total difficulty should be promoted, as it has been.

Questions, ideas, and comments are welcome.

dahlia:
@planetarium/libplanet @junorouse Could you take a look, please?

dahlia:
Also: we may need to consider time a node received blocks.

dahlia:
Also: we may need to consider time a node received blocks.

PR #1165, opened 2021-01-20
Customizable rule to determine the canonical chain

dahlia:
This is the first step to address the issue #1155.  This refactors duplicated logic, which had existed here and there, to determine the canonical chain (i.e., the most total difficulty = the best chain), and now the logic can be customizable through `IBlockPolicy<T>.CanonicalChainComparer`, which is an instance of `IComparer<IBlockExcerpt>`.

`IBlockExcerpt` is a new interface to deal with three different types (`Block<T>`, `BlockHeader`, & `ChainStatus`), which are involved to determine the canonical chain, in a unified way.  These three types all implement `IBlockExcerpt`.

`CanonicalChainComparer`, which complies `IComparer<IBlockExcerpt>`, implements the existing logic based on the most total difficulty.  This should be changed later according to #1155.


Libplanet/Blockchain/CanonicalChainComparer.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// The default canonical chain comparer (which purpose to be a <see
+    /// cref="IBlockPolicy{T}.CanonicalChainComparer"/>).
+    /// <para>The chain which has the most <see cref="Block{T}.TotalDifficulty"/> is considered
+    /// the greatest, i.e., canonical chain.</para>
+    /// </summary>
+    /// <seealso cref="IBlockPolicy{T}.CanonicalChainComparer"/>
+    /// <seealso cref="IBlockExcerpt"/>
+    public class CanonicalChainComparer : IComparer<IBlockExcerpt>
```
longfin, line 16: IMHO, `CanonicalChainComparer` doesn't represent its behavior... how about a name like `DifficultyBasedChainComparer`?

longfin (commented):
I guess this PR is pair enough, but added tests seem to fail. 😢

Libplanet/Blockchain/CanonicalChainComparer.cs
dahlia: The logic behind this will be changed soon.  How about `DefaultCanonicalChainComparer`, to imply it is the officially most recommended configuration?

dahlia:
@planetarium/libplanet Fixed the timed out tests.  PTAL.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -464,7 +455,13 @@ public Transaction<T> GetTransaction(TxId txId)
             _rwlock.EnterReadLock();
             try
             {
-                return _transactions[txId];
+                if (Store.GetTransaction<T>(txId) is { } transaction)
```
limebell, line 458: How about removing `GetTransaction` from `BlockChain<T>`?

Libplanet/Blockchain/CanonicalChainComparer.cs
limebell: There could be multiple implementation of comparer, so to clarify its usage `DifficultyBasedChainComparer` seems better to me. To suggest it as default, maybe write it as docstring will help I guess?

Libplanet/Blockchain/Policies/IBlockPolicy.cs
```diff
@@ -15,6 +17,14 @@ namespace Libplanet.Blockchain.Policies
     public interface IBlockPolicy<T>
         where T : IAction, new()
     {
+        /// <summary>
+        /// A comparer to determine which branch is the canonical chain (i.e., best chain).
+        /// The most greater one according to this comparer is considered to be the canon.
+        /// </summary>
+        /// <seealso cref="IBlockExcerpt"/>
+        /// <seealso cref="Blockchain.CanonicalChainComparer"/>
+        IComparer<IBlockExcerpt> CanonicalChainComparer { get; }
```
limebell, line 26: IMO chain comparer is not really a property related to `Block<T>` but `BlockChain<T>`. Maybe we need something like `IBlockChainPolicy`? Not a strong suggestion though.

Libplanet/Net/Messages/ChainStatus.cs
```diff
@@ -1,36 +1,46 @@
 using System.Collections.Generic;
 using System.Numerics;
 using System.Security.Cryptography;
+using Libplanet.Blocks;
 using NetMQ;
 
 namespace Libplanet.Net.Messages
 {
-    internal class ChainStatus : Message
+    internal class ChainStatus : Message, IBlockExcerpt
```
limebell, line 9: This looks a little bit awkward for me. Why a message inherits `IBlockExcerpt`?

Libplanet/Blockchain/CanonicalChainComparer.cs
dahlia: If we put the implementation details into the name, we should rename it when we change the details.  This can be considered as a kind of leaky abstraction in theory, and also makes difficult to maintain backward compatibility of API due to the name: we should keep the name even if it no more reflects the implementation, or rename it to break backward compatibility.  I believe this would not be good for the default implementation. (It'd be good if C# supports class aliases like `public class DefaultChainComparer = DifficultyBasedChainComparer`, but it's impossible anyway…)

dahlia: Or we can make `DefaultChainComparer` to subclass `DifficultyBasedChainComparer`.  How about it?

Libplanet/Blockchain/BlockChain.cs
dahlia: If it's unused by other projects (e.g., Lib9c, NineChronicles.Headless) that would be good.

Libplanet/Blockchain/CanonicalChainComparer.cs
longfin: IMO, if we want to represent this comparer as default, we can also consider adding static field `BlockPolicy.DefaultChainCoamparer` and setting it `new DifficultyBasedChainComparer()`.

Libplanet/Blockchain/Policies/IBlockPolicy.cs
dahlia: I believe `IBlockPolicy<T>` is already about `BlockChain<T>`, not `Block<T>`… its namespace implies that too: `Libplanet.Blockchain`.  It might be better to rename to `IBlockChainPolicy<T>` or `IConsensus<T>`? /cc @planetarium/libplanet

Libplanet/Net/Messages/ChainStatus.cs
dahlia: It's because this struct is used to determine the canonical chain:

https://github.com/planetarium/libplanet/blob/8f8bceb4589cefd6e5ca2304b0e88e9a0e377315/Libplanet/Net/Swarm.cs#L1338-L1344

Note that it was originally:

https://github.com/planetarium/libplanet/blob/b5fdad939dadde80ea4007e1dadc908669a21c0f/Libplanet/Net/Swarm.cs#L1337-L1342

Libplanet/Blockchain/BlockChain.cs
dahlia: It's still used by `Swarm<T>`:

https://github.com/planetarium/libplanet/blob/ebce8116322b5cada571aa0055dde601b48f4f77/Libplanet/Net/Swarm.MessageHandlers.cs#L168

Libplanet/Blockchain/CanonicalChainComparer.cs
dahlia: I renamed it to `TotalDifficultyComparer`: 9c5063451c1e012c2df804ebc733481017194203.

dahlia:
@planetarium/libplanet Rebased and addressed reviewed comments.  PTAL.

longfin (approved):

Libplanet/Blockchain/BlockChain.cs
limebell: IMO replace it to `IStore.GetTransaction` looks better, for unity.

dahlia: > IMO replace it to `IStore.GetTransaction` looks better, for unity.

@limebell Could you elaborate?

IMHO the methods of `IStore` are `BlockChain<T>`'s implementations details that should not be exposed on its surface.  If game apps use (= need) `IStore` directly, I'll find it [design smell][1], more exactly, *missing abstraction*.

On the other hand, if game app use (= need) neither `IStore.GetTransaction<T>()` nor `BlockChain<T>.GetTransaction()`, it means there's no demand for `BlockChain<T>.GetTransaction()` so that having this method is *deficient encapsulation*.

Anyway, I try to look into other projects that use Libplanet… and I found several demands in reality:

https://github.com/planetarium/lib9c/blob/614ab68915f14f0bcf933f06afe629ad0af1ff03/Lib9c/Miner.cs#L51
https://github.com/planetarium/lib9c/blob/614ab68915f14f0bcf933f06afe629ad0af1ff03/Lib9c/Miner.cs#L62
https://github.com/planetarium/lib9c/blob/614ab68915f14f0bcf933f06afe629ad0af1ff03/Lib9c/ReorgMiner.cs#L73
https://github.com/planetarium/lib9c/blob/614ab68915f14f0bcf933f06afe629ad0af1ff03/Lib9c/ReorgMiner.cs#L84

One more example, but it's not open sourced (yet):

https://github.com/planetarium/nekoyume-unity/blob/cc3d8a82d60f3fe04a4da0cb6f5e396f34c4e093/nekoyume/Assets/_Scripts/BlockChain/Agent.cs#L190

[1]: https://en.wikipedia.org/wiki/Design_smell

Libplanet/Blockchain/Policies/IBlockPolicy.cs
riemannulus: I think `IBlockPolicy<T>` could be confusing enough as @limebell said. `IConsensus<T>` is much better name for describe.

limebell (approved):

PR #1168, opened 2021-01-23
Add Hash validation to Block<T>.Validate()

longfin:
This PR adds hash validation to `BlockHeader.Validate()` (and `Block<T>`.Validate()).

### DIFF
--- a/Libplanet/Blocks/Block.cs
+++ b/Libplanet/Blocks/Block.cs
@@ -546,6 +546,8 @@ namespace Libplanet.Blocks
         /// </param>
         /// <returns>An <see cref="ActionEvaluation"/> for each
         /// <see cref="IAction"/>.</returns>
+        /// <exception cref="InvalidBlockHashException">Thrown when
+        /// the <see cref="Hash"/> is invalid.</exception>
         /// <exception cref="InvalidBlockTimestampException">Thrown when
         /// the <see cref="Timestamp"/> is invalid, for example, it is the far
         /// future than the given <paramref name="currentTime"/>.</exception>
--- a/Libplanet/Blocks/Block.cs
+++ b/Libplanet/Blocks/Block.cs
@@ -84,11 +84,11 @@ namespace Libplanet.Blocks
             Transactions = transactions.OrderBy(tx => tx.Id).ToArray();
             TxHash = CalcualteTxHashes(Transactions);
 
-            PreEvaluationHash = preEvaluationHash ?? Hashcash.Hash(SerializeForHash());
+            PreEvaluationHash = preEvaluationHash ?? Hashcash.Hash(Header.SerializeForHash());
             StateRootHash = stateRootHash;
 
             // FIXME: This does not need to be computed every time?
-            Hash = Hashcash.Hash(SerializeForHash(stateRootHash));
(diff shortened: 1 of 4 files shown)


Libplanet/Blocks/Block.cs
```diff
@@ -382,8 +382,8 @@ public BlockHeader Header
 
             // Poor man' way to optimize stamp...
             // FIXME: We need to rather reorganize the serialization layout.
-            byte[] emptyNonce = MakeBlock(new Nonce(new byte[0])).SerializeForHash();
-            byte[] oneByteNonce = MakeBlock(new Nonce(new byte[1])).SerializeForHash();
+            byte[] emptyNonce = MakeBlock(new Nonce(new byte[0])).Header.SerializeForHash();
+            byte[] oneByteNonce = MakeBlock(new Nonce(new byte[1])).Header.SerializeForHash();
```
moreal, line 386: It is just a suggestion, how about making the `MakeBlockHeader()` function which creates a new `BlockHeader`?
`Block<T>.SerializedForHash()` was moved to `BlockHeader` so using `Block<T>.BlockHeader` getter seems useless. I believe it can be related to performance issue a little 🤔

Libplanet/Blocks/BlockHeader.cs
```diff
@@ -337,11 +385,43 @@ internal void Validate(DateTimeOffset currentTime)
...
+                    Index,
+                    Timestamp,
+                    Difficulty,
+                    Nonce,
+                    Miner,
+                    PreviousHash,
+                    TxHash,
+                    StateRootHash
+                )
+            );
```
moreal, line 406: It is a question, Could you describe what difference between the above code and `internal byte[] SerializeForHash()` code? It seems same to me.

```suggestion
            HashDigest<SHA256> calculatedHash = Hashcash.Hash(SerializeForHash());
```

longfin: Yes, it's the same. 😅

Libplanet/Blocks/Block.cs
longfin: It can, but IMHO it isn't slightly significant because `MakeBlock()` calls appear only here during the whole mining...  😕

riemannulus: I think there will be a performance drop here, too, but I don't think there will be more performance drops than before.

riemannulus: I think there will be a performance drop here, too, but I don't think there will be more performance drops than before.

moreal: I agree.

riemannulus (approved):

limebell (approved):

dahlia (dismissed):
We may need to port this to the 0.10-maintenance branch too… 🤔

moreal (dismissed):

longfin:
Rebased. PTAL @libplanet

riemannulus (dismissed):

moreal (approved):

dahlia (dismissed):

longfin:
Rebased again. PTAL @libplanet

moreal (approved):

limebell (approved):

PR #1180, opened 2021-01-29
Ignore transactions that have lower nonce than the highest nonce among transaction in storage

### LINKED ISSUE #1173: Ignore transactions with lower nonce
Transactions which has lower nonce are now ignored in mining step only. In order to reduce the network load, transactions with low nonce ​​should be ignored during the staging phase.

limebell:
Transactions with lower nonce can not be added to chain unless huge reorg occurs. This PR prevent those transactions to be staged to prevent network load. This closes #1173 .


riemannulus (approved):

CHANGES.md
```diff
@@ -147,6 +147,10 @@ To be released.
     can be validated rather than trusted due to MPT.  [[#1117]]
  -  `HashDigest<SHA256>` became serializable.  [[#795], [#1125]]
  -  `Transaction<T>()` constructors became not to validate itself.  [[#1149]]
+ -  `Blockchain<T>` became not to stage transactions with nonce that is
+    less than or equal to the highest nonce among the transactions
+    it has in its storage for a transaction signed by a specific address.
+    [[#1173], [#1180]]
```
dahlia, line 153:
```suggestion
 -  `BlockChain<T>` became not to stage transactions having nonce
    less than or equal to the nonce of the same signer's latest
    transaction in already mined blocks.  [[#1173], [#1180]]
```

Libplanet.Tests/Menees.Analyzers.Settings.xml
```diff
@@ -2,5 +2,5 @@
 <Menees.Analyzers.Settings>
   <MaxLineColumns>100</MaxLineColumns>
   <MaxMethodLines>300</MaxMethodLines>
-  <MaxFileLines>2600</MaxFileLines>
+  <MaxFileLines>2700</MaxFileLines>
```
dahlia, line 5: Rather than increasing this, how about making a partial class of `SwarmTest` instead?

longfin (approved):

dahlia (approved):

riemannulus (approved):

Issue #1161, opened 2021-01-14
THE-9C-017


junorouse:
https://docs.google.com/document/d/1Arx6UoYonnqsGw0Mf9lmqbdlKt2Nx8qf4fp_HZmPe3c/edit#heading=h.yw19yxpjxyx7

PR #1181, opened 2021-02-02
Sign all messages

riemannulus:
SSIA

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -160,6 +160,7 @@ To be released.
  -  `Blockchain<T>` became not to stage transactions having nonce
     less than or equal to the nonce of the same signer's latest
     transaction in already mined blocks.  [[#1173], [#1180]]
+ -  Improved security of `Message.Parse()`.  [[#1161], [#1181]]
 
 ### Bug fixes
 
@@ -223,6 +224,7 @@ To be released.
 [#1152]: https://github.com/planetarium/libplanet/pull/1152
 [#1155]: https://github.com/planetarium/libplanet/issues/1155
 [#1160]: https://github.com/planetarium/libplanet/issues/1160
+[#1161]: https://github.com/planetarium/libplanet/issues/1161
 [#1162]: https://github.com/planetarium/libplanet/pull/1162
 [#1163]: https://github.com/planetarium/libplanet/pull/1163
 [#1165]: https://github.com/planetarium/libplanet/pull/1165
@@ -232,6 +234,7 @@ To be released.
 [#1172]: https://github.com/planetarium/libplanet/pull/1172
 [#1173]: https://github.com/planetarium/libplanet/issues/1173
 [#1180]: https://github.com/planetarium/libplanet/pull/1180
+[#1181]: https://github.com/planetarium/libplanet/pull/1181
 
 
 Version 0.10.3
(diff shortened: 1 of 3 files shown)


Libplanet/Net/Messages/Message.cs
```diff
@@ -341,14 +351,30 @@ private enum MessageFrame
...
+            message.Clear();
+
+            // Make message
+            foreach (NetMQFrame frame in DataFrames)
+            {
+                message.Append(frame);
+            }
+
+            // Write headers. (inverse order)
+            message.Push(signature);
```
limebell, line 377: According to [this](https://github.com/zeromq/netmq/blob/0b98a648ef0dd7933b5819257dee97e64bfb6420/src/NetMQ/NetMQMessage.cs#L48), we can make `NetMQMessage` with an `IEnumerable` instance. How about use `IList` typed message first before signing in order to inject signature easily without make message instance twice?

CHANGES.md
```diff
@@ -160,6 +160,7 @@ To be released.
  -  `Blockchain<T>` became not to stage transactions having nonce
     less than or equal to the nonce of the same signer's latest
     transaction in already mined blocks.  [[#1173], [#1180]]
+ -  Improved security of `Message.Parse()`  [[#1181]]
```
moreal, line 163:
```suggestion
 -  Improved security of `Message.Parse()`.  [[#1181]]
```

It seems also related to https://github.com/planetarium/libplanet/issues/1161, doesn't it?

Libplanet/Net/Messages/Message.cs
```diff
@@ -341,14 +351,30 @@ private enum MessageFrame
 
             var message = new NetMQMessage();
 
+            // Make signature
```
moreal, line 354: Did you intend this comment for all processes, from building message to make its signature? Or if you intended the comment only for making a signature, it seems more natural to be above on the below lines.

```
            var signature = key.Sign(message.ToByteArray());
            message.Clear();
```

But I think it can be just my personal taste, You can ignore this opinion free if you don't think so.

Libplanet/Net/Messages/Message.cs
riemannulus: According to [this](https://github.com/zeromq/netmq/blob/0b98a648ef0dd7933b5819257dee97e64bfb6420/src/NetMQ/NetMQMessage.cs#L249), there is a problem that the implementation of the part where the type conversion occurs when pushing is partly imported into `ToMQMessage`. So, what do you think we're trying to change in the following way? 

```suggestion
            byte[] signature = key.Sign(message.ToByteArray());
            List<NetMQFrame> frames = message.ToList();
            frames.Insert((int)MessageFrame.Sign, new NetMQFrame(signature));
            message = new NetMQMessage(frames);
```

CHANGES.md
riemannulus: https://github.com/planetarium/libplanet/pull/1181/commits/390689589ce98793466b78fbe444f83fdacc976a amended.

Libplanet/Net/Messages/Message.cs
riemannulus: https://github.com/planetarium/libplanet/pull/1181/commits/f3c25c05e3a1a5cbc5215ebc28b9c8adc77b76f8 I fix this comment. please check this out.

moreal (approved):

limebell (approved):

longfin (approved):

PR #1185, opened 2021-02-04
Reset all relays after reconnecting TURN

longfin:
This PR fixes a bug where `TurnClient` hadn't reset its relays after resetting.


limebell (approved):
Can we have unit tests related to TURN?

longfin:
> Can we have unit tests related to TURN?

IMO we can use `ReconnectToTurn`  but it's also a little tricky because we don't have interfaces to access `TurnClient` from outside.

As another approach, it would be better if we add unit tests for `TurnClient` directly but it's needed some additional works.

area363 (approved):

PR #1186, opened 2021-02-05
Prevent invalid transactions from being staged repeatedly

dahlia:


longfin (approved):

riemannulus (approved):

Libplanet/Blockchain/Policies/IStagePolicy.cs
```diff
@@ -35,17 +36,22 @@ public interface IStagePolicy<T>
...
-        /// <param name="includeUnstaged">Whether to include transactions that had once staged but
-        /// unstaged then.</param>
-        /// <returns><c>true</c> if a transaction is staged.  For transactions that had once staged
-        /// but unstaged then, it returns <c>true</c> iff <paramref name="includeUnstaged"/> is
-        /// turned on; otherwise it returns <c>false</c> for those cases.  For transactions
-        /// that have never staged, it returns <c>false</c>.</returns>
-        public bool HasStaged(BlockChain<T> blockChain, TxId id, bool includeUnstaged);
+        /// <returns><c>true</c> if a transaction should be ignored.
+        /// Otherwise, <c>false</c>.</returns>
+        public bool Ignores(BlockChain<T> blockChain, TxId id);
```
moreal, line 54: `Ignores` seems a verb like *'it ignores ...'*. So, to me, it looks like a method to some actions, not to check. I wonder your intention of naming 😀 IMHO How about `Ignored`?

moreal (approved):

Libplanet/Blockchain/Policies/IStagePolicy.cs
dahlia: `IsIgnored` sounds less natural to me, because sentences like “the chain stage _is ignored_ by the transaction” while we intended rather the other way: “the transaction _is ignored_ by the chain stage.” If we turn this passive voice into active, it would be like “the chain stage _ignores_ the transaction” in the end.

moreal: Yes, I understood. Thanks for the kind description 🙏

PR #1193, opened 2021-02-24
Handshake protocol

limebell:
This patch make a peer to authenticate the message received before add it to routing table.

For example, if peer A sends `Ping` to peer B, peer B used to add peer A to its routing table. But after this patch peer B will reply `Pong` as usual, and send `Ping` again to peer A to verify that the information it sent is valid.

Todo: Should ignore non-`Ping` messages (i.e. `GetBlocks`) if the remote peer of the message is not in the routing table to prevent DDoS attack from maleficent peers.

### DIFF
--- a/Libplanet.Tests/Net/Protocols/ProtocolTest.cs
+++ b/Libplanet.Tests/Net/Protocols/ProtocolTest.cs
@@ -212,9 +212,9 @@ namespace Libplanet.Tests.Net.Protocols
             await StartTestTransportAsync(transportB);
             await StartTestTransportAsync(transportC);
 
-            await transportA.AddPeersAsync(new[] { transport.AsPeer }, null);
-            await transportB.AddPeersAsync(new[] { transport.AsPeer }, null);
-            await transportC.AddPeersAsync(new[] { transport.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportA.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportB.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportC.AsPeer }, null);
 
             Assert.Single(transportA.Peers);
             Assert.Contains(transportA.AsPeer, transport.Peers);
@@ -240,10 +240,10 @@ namespace Libplanet.Tests.Net.Protocols
             await StartTestTransportAsync(transportB);
             await StartTestTransportAsync(transportC);
 
-            await transportA.AddPeersAsync(new[] { transport.AsPeer }, null);
-            await transportB.AddPeersAsync(new[] { transport.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportA.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportB.AsPeer }, null);
             await Task.Delay(100);
-            await transportC.AddPeersAsync(new[] { transport.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportC.AsPeer }, null);
 
             Assert.Single(transportA.Peers);
             Assert.Contains(transportA.AsPeer, transport.Peers);
@@ -276,8 +276,8 @@ namespace Libplanet.Tests.Net.Protocols
             await StartTestTransportAsync(transportB);
             await StartTestTransportAsync(transportC);
 
-            await transportA.AddPeersAsync(new[] { transport.AsPeer }, null);
-            await transportB.AddPeersAsync(new[] { transport.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportA.AsPeer }, null);
+            await transport.AddPeersAsync(new[] { transportB.AsPeer }, null);
 
             Assert.Single(transport.Peers);
             Assert.Contains(transportA.AsPeer, transport.Peers);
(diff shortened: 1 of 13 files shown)


Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -57,6 +60,8 @@ public class KademliaProtocol : IProtocol
                 requestTimeout ??
                 TimeSpan.FromMilliseconds(5000);
             _transport.ProcessMessageHandler += ProcessMessageHandler;
+            _cache = new ConcurrentDictionary<Address, BoundPeer>();
+            _cacheSize = int.MaxValue;
```
longfin, line 64: Is it okay we always use `int.MaxValue`?

Libplanet/Net/Swarm.cs
```diff
@@ -908,22 +908,27 @@ await foreach (var pair in demandBlockHashes)
             await kademliaProtocol.CheckAllPeersAsync(timeout, cancellationToken);
         }
 
-        internal async Task AddPeersAsync(
-            IEnumerable<Peer> peers,
-            TimeSpan? timeout,
-            CancellationToken cancellationToken = default(CancellationToken))
+        /// <summary>
+        /// This methods directly adds given peer to routing table.
+        /// It is not recommended to use this method outside of testing purpose.
```
longfin, line 913: How about extracting this method and moving to `Libplanet.Tests` project?

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: I have no idea for this value actually... Do you have suggestion?

longfin: Can we use the value related `Kademlia.TableSize` or `Kademlia.BucketSize` here?

limebell:
PTAL! @libplanet

Libplanet.Tests/Net/Protocols/ProtocolTest.cs
```diff
@@ -181,7 +155,9 @@ public async Task BootstrapAsyncTest()
                 await StartTestTransportAsync(transportC);
 
                 await transportB.BootstrapAsync(new[] { transportA.AsPeer });
+                await Task.Delay(100);
```
riemannulus, line 158: Is delay need here? if then, why doesn't need it before?

Libplanet/Net/Protocols/KademliaProtocol.cs
```diff
@@ -180,7 +185,6 @@ public class KademliaProtocol : IProtocol
             {
                 _logger.Debug(
                     $"Timeout occurred during {nameof(AddPeersAsync)}() after {timeout}.");
-                throw;
```
riemannulus, line 183: Then now, just logging and ignore `TimeoutException`?

Libplanet.Tests/Net/Protocols/ProtocolTest.cs
limebell: Previously `BootstrapAsync` method guaranteed that two peers (peer, seed) will have each other peer's information in routing table right after the function call. But new logic guarantees peer to have seed's information in its table but seed peer may not have remote peer's information in its table. So I added context changer for seed to have remote peer in its table.

Libplanet/Net/Protocols/KademliaProtocol.cs
limebell: Right.

Libplanet.Tests/Net/Protocols/ProtocolTest.cs
```diff
@@ -82,39 +86,9 @@ public async Task Ping()
...
-            }
-            finally
-            {
-                await transportA.StopAsync(TimeSpan.Zero);
-                await transportB.StopAsync(TimeSpan.Zero);
-            }
-        }
-
-        [Fact(Timeout = Timeout)]
-        public async Task PingTwice()
```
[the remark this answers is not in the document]
limebell: The test was to check each peer can receive and respond to `ping`/`pong`. Logic became to handshake before connection, it will resend another `ping` message when a `ping` message is received. Thus the `PingTwice` test became meaningless.

riemannulus (dismissed):

longfin (approved):

Libplanet.Tests/Net/Protocols/TestTransport.cs
```diff
@@ -420,6 +426,11 @@ public bool ReceivedTestMessageOfData(string data)
             return ReceivedMessages.OfType<TestMessage>().Any(msg => msg.Data == data);
         }
 
+        private string ToAddress(BoundPeer peer)
+        {
+            return $"{peer.EndPoint.Host}:{peer.EndPoint.Port}";
```
dahlia, line 431: Is this format used only for unit tests?

dahlia (commented):
@limebell Your branch seems outdated long ago.  Could you rebase this on the up-to-date *main* branch?

dahlia:
@limebell Would we merge this?  If we would like to merge this, this needs to be rebased.

PR #1198, opened 2021-03-05
Improve performance of BlockChain<T>.Fork()

longfin:


Libplanet/Store/IStore.cs
```diff
@@ -244,5 +244,16 @@ void PutBlock<T>(Block<T> block)
...
+
+        /// <summary>
+        /// Forks transaction nonces from
+        /// <paramref name="sourceChainId"/> to
+        /// <paramref name="destinationChainId"/>.
+        /// </summary>
+        /// <param name="sourceChainId">The chain ID of transaction nonces to
+        /// fork.</param>
+        /// <param name="destinationChainId">The chain ID of destination
+        /// transaction nonces.</param>
```
moreal, line 256: How about referencing words to its meaning code? I don't think my suggestion is the best way but I think it will be better to reference like my suggestion does.

```suggestion
        /// <summary>
        /// Forks <see cref="Transaction{T}"/> <see cref="Transaction{T}.Nonce"/>s from
        /// <paramref name="sourceChainId"/> to
        /// <paramref name="destinationChainId"/>.
        /// </summary>
        /// <param name="sourceChainId">The chain <see cref="BlockChain{T}.Id"/> of
        /// <see cref="Transaction{T}"/> <see cref="Transaction{T}.Nonce"/>s to fork.</param>
        /// <param name="destinationChainId">The chain <see cref="BlockChain{T}.Id"/> of destination
        /// <see cref="Transaction{T}"/> <see cref="Transaction{T}.Nonce"/>s.</param>
```

Libplanet/Store/IStore.cs
```diff
@@ -244,5 +244,16 @@ void PutBlock<T>(Block<T> block)
...
+        void ForkTxNonces(Guid sourceChainId, Guid destinationChainId);
```
moreal, line 257: Of course, I know there is a test to check it works in `BlockChainTest` but could you add unittest for this new method in `StoreTest`?

longfin: I'll add.

longfin: [Added!](https://github.com/planetarium/libplanet/pull/1198/files#diff-4682d77dc711a5c0e7ec81aaeeb20519c75c475bd24ccfbfa3d6caec762caff1)

moreal (approved):

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -29,6 +29,7 @@ public class RocksDBStore : BaseStore
         private const string TxIndexDbName = "txindex";
         private const string StagedTxDbName = "stagedtx";
         private const string ChainDbName = "chain";
+        private const int ForkWriteBatchSize = 100000;
```
limebell, line 32: How about making this value configurable?

longfin: That's possible, but this batch isn't affected by the nature of the application (only affected by internal data structure).  and there are too many configurable variables in this class 😅

limebell (approved):

riemannulus (approved):

dahlia (approved):

PR #1236, opened 2021-04-09
Fix memory leakage in relayed clients

limebell:
This patch fixes memory leakage caused by not freeing disposed TCP clients in `TurnClient`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -218,6 +218,7 @@ To be released.
     updated cached peers multiple times.  [[#1219]]
  -  Fixed memory leak due to undisposed `CancellationTokenRegistration`s.
     [[#1228]]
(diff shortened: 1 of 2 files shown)


dahlia (approved):

longfin:
It seems that we need to resolve conflicts... 😢

riemannulus (approved):

dahlia (dismissed):

longfin (dismissed):

dahlia:
The test `ProtocolTest.ReplacementCache` failed on Unity's Mono runtime.

dahlia (commented):
For `internal`/`private` methods, how about require explicit `cancellationToken`s (rather than giving it a `default` value).  IMHO it would be less bug-prone.

PR #1243, opened 2021-04-15
Utilize multiple TURN/ICE servers to balance loads

dahlia:
This patch makes the test suite to take multiple `TURN_SERVER_URLS` instead of a single `TURN_SERVER_URL` so that these loads that TURN/ICE servers receive become balanced.  I already configured `turnServerUrls` on Azure Pipelines too.

Read the changes in _CONTRIBUTING.md_ for details.


limebell (approved):

dahlia:
@longfin @riemannulus Could you take a look?

riemannulus (approved):

PR #1255, opened 2021-04-19
Testing with NetMQ upstream

longfin:


longfin:
I'll close this PR and re-open with [Planetarium.NetMQ 4.0.0.261](https://www.nuget.org/packages/Planetarium.NetMQ/4.0.0.261-planetarium) due to interface changing that hadn't been applied NetMQ upstream.

PR #1261, opened 2021-04-22
Prepare 0.11.2

longfin:
SSIA

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,25 @@
 Libplanet changelog
 ===================
 
+Version 0.11.2
+-------------
+
+To be released.
+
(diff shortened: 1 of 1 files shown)


dahlia (commented):
As patch releases have only bug fixes, we don't have such sections.  (See also other patch releases' changelogs.)

longfin:
> As patch releases have only bug fixes, we shouldn't have such sections. (See also other patch releases' changelogs.)

I added a commit to clean up. https://github.com/planetarium/libplanet/pull/1261/commits/49916ebe71863667a15785d53b2edb811b7847c3

dahlia (approved):

PR #1263, opened 2021-04-22
Fix `new Null()` to `default(Null)`

boscohyun:
There is only line to construct the `Null` struct with `new Null()` and we need change it to `default(Null)` if possible.

### DIFF
--- a/Libplanet/Assets/Currency.cs
+++ b/Libplanet/Assets/Currency.cs
@@ -210,7 +210,7 @@ namespace Libplanet.Assets
 #pragma warning disable SA1129  // See also: https://github.com/planetarium/bencodex.net/issues/20
             IValue minters = Minters is ImmutableHashSet<Address> a
                 ? new List(a.OrderBy(m => m).Select(m => (IValue)new Binary(m.ToByteArray())))
-                : (IValue)new Null();
+                : (IValue)default(Null);
 #pragma warning restore SA1129
             IValue serialized = Dictionary.Empty
                 .Add("ticker", Ticker)


dahlia (commented):
Could you sign our CLA?  (It's necessary only for the first time.)

This also needs to be rebased on the up-to-date _main_ branch.

dahlia:
Relevant issue: <https://github.com/planetarium/bencodex.net/issues/20>.

longfin (approved):

riemannulus (approved):

dahlia:
It would be better if *Bencodex* packaged is upgraded to the latest version and `Null.Value` field is used instead after <https://github.com/planetarium/bencodex.net/pull/37> is merged.

limebell (approved):

limebell:
Can you please update this branch? @boscohyun

dahlia (changes requested):
As *Bencodex* package now has `Null.Value` property (see also <https://github.com/planetarium/bencodex.net/pull/37>), we don't need to suppress `SA1129` warning anymore.  You can simply use `Null.Value` instead of `default(Null)` or `new Null()`.

@boscohyun Could you rebase this on the up-to-date *main* branch, and use `Null.Value` instead?

dahlia:
Oh, I already worked on this in the commit 2c75b0f9c1ad18e4d0fcf054bf9222fe98670516.  This pull request seems now unnecessary.  I close this.  Thanks @boscohyun!

boscohyun:
> As _Bencodex_ package now has `Null.Value` property (see also [planetarium/bencodex.net#37](https://github.com/planetarium/bencodex.net/pull/37)), we don't need to suppress `SA1129` warning anymore. You can simply use `Null.Value` instead of `default(Null)` or `new Null()`.
> 
> @boscohyun Could you rebase this on the up-to-date _main_ branch, and use `Null.Value` instead?

@dahlia That's good news! Now there is no longer any need to worry between `default(Null)` and `new Null()`. Thank you!
I'll close this PR.

Issue #1267, opened 2021-04-26
Actions in a transaction are executed all or nothing at all


dahlia:
Even though the name `Transaction` implies actions in it are atomic (i.e., evaluated all or nothing at all, not partly), but the current implementation does not.  If any action in a transaction throws an exception, every action in that transaction should be no-op both rendering-wise and state-wise.

Issue #1156, opened 2021-01-14
Leaving `Transaction<T>`s' results


longfin:
(Continued from #860)

When `IAction.execute()` throws an exception, Libplanet only calls `RenderError()`. it can be useful for UI rendering. but if we want to track exceptions on previous `IAction.execute()`s from the chain, it isn't enough.

so I suggest "Action execution result" like [Ethereum's receipt](https://github.com/ethereum/go-ethereum/blob/v1.9.25/core/types/receipt.go). it should have contents as bellowing contents 

- Tx Id
- Block Hash
- Output states
- Exception

PR #1272, opened 2021-04-27
Idempotent RocksDBStore.Dispose()

longfin:

PR #1275, opened 2021-04-29
Execute `Transaction<T>.Actions` all or nothing at all

dahlia:
This addresses fixes <https://github.com/planetarium/libplanet/issues/1267>.

Previously, `Transaction<T>.Actions` had been executed and affected states whether the actions throw an exception or not.  However, as the term “transaction” in computer science in general imply it's atomic (i.e., executed all or nothing), this behavior had been quite misleading.

This patch regards such previous behavior as a bug, and fixes it so that if any action in a `Transaction<T>` throws an exception all actions in the transaction do not commit any state changes to the current `IStore`.

However, there is still a quirk for it: `IActionRenderer<T>` receives render events caused by unsuccessful transactions too, as it has done.  In other words, transactions' atomicity still is not guaranteed for action renderers.  In order to work around this quirk, this patch introduces `AtomicActionRenderer<T>` as well.  If an `IActionRenderer<T>` is wrapped by `AtomicActionRenderer<T>` the inner action renderer will receive no `RenderActionError()`/`UnrenderActionError()` events except for block actions (since they do not belong to any transactions).

The reasons why I stick with the existing behavior for action renderers are:

- Ignoring render events from unsuccessful transactions making the current shape of `IActionRenderer<T>` no more natural.  Because `RenderActionError()`/`UnrenderActionError()` methods are no more necessary.
- However, changing the interface breaks backward compatibility.  In particular, `IActionRenderer<T>` is one of the key interfaces game app programmers must implement.  For example, such change on the interface would require many changes on [Nine Chronicles](https://nine-chronicles.com/).
- We may eventually need to refactor the interface when we implement <https://github.com/planetarium/libplanet/issues/1156>.

For details, please read the added changelogs and XML doc comments too.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -143,6 +143,7 @@ To be released.
  -  Added `NetMQTransport.QueryAppProtocolVersion()` static method.  [[#1235]]
  -  Added `BoundPeer.Parse()` static method.  [[#1240]]
  -  Added `TransportException` class.  [[#1242]]
+ -  Added `AtomicActionRenderer<T>` class.  [[#1267], [#1275]]
 
 ### Behavioral changes
 
@@ -174,6 +175,8 @@ To be released.
         action, action rendering methods in `IActionRenderer<T>`
         (`RenderAction()`, `RenderActionError()`, `UnrenderAction()`, and
         `UnrenderActionError()`) became not invoked.
+        If you want to dismiss all actions in unsuccessful transactions at all,
+        wrap your action renderer with `AtomicActionRenderer<T>`.
  -  Fixed a bug where `KademliaProtocol.BootstrapAsync()` has sent multiple
     `Ping` messages to other peers.  [[#1219]]
  -  Fixed a bug where `KademliaProtocol.CheckReplacementCacheAsync()` has
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -118,6 +118,7 @@ To be released.
      -  The type of `TrieStateStore.PruneStates()` method's `excludeBlockHashes`
         parameter became `IImmutableSet<BlockHash>`
         (was `ImmutableHashSet<HashDigest<SHA256>>`).
+ -  Added `IActionContext.TxId` property.  [[#1275]]
  -  Removed `StunMessage.Parse(Stream)` method.  [[#1228]]
  -  Moved `ITransport` and `NetMQTransport` from `Libplanet.Net` to
     `Libplanet.Net.Transports`.  [[#1235]]
@@ -195,6 +196,7 @@ To be released.
 [#1268]: https://github.com/planetarium/libplanet/pull/1268
 [#1272]: https://github.com/planetarium/libplanet/pull/1272
 [#1274]: https://github.com/planetarium/libplanet/pull/1274
+[#1275]: https://github.com/planetarium/libplanet/pull/1275
 [#1278]: https://github.com/planetarium/libplanet/pull/1278
 
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -159,6 +159,21 @@ To be released.
 
 ### Bug fixes
 
+ -  Fixed a bug where executing `Transaction<T>.Actions` had not been atomic.
+    `Actions` in a `Transaction<T>` now became executed all or nothing at all.
+    [[#1267], [#1275]]
+     -  `Transaction<T>.EvaluateActions()` method became atomic.
+     -  `Transaction<T>.EvaluateActionsGradually()` method had returned
+        the same number of `ActionEvaluation`s to `Transaction<T>.Actions`,
+        but now became to omit the evaluations after the first action throwing
+        an exception.  If no action throws any exception, it still returns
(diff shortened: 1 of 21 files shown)


CHANGES.md
```diff
@@ -151,6 +151,21 @@ To be released.
...
+    `Actions` in a `Transaction<T>` now became executed all or nothing at all.
+    [[#1267], [#1275]]
+     -  `Transaction<T>.EvaluateActions()` method became atomic.
+     -  `Transaction<T>.EvaluateActionsGradually()` method had returned
+        the same number of `ActionEvaluation`s to `Transaction<T>.Actions`,
+        but now became to omit the evaluations after the first action throwing
+        an exception.  If no action throws any exception, it still returns
+        the same number of `ActionEvaluation`s to `Transaction<T>.Actions`.
+     -  State-wise, `Transaction<T>`s having any exception-throwing action
+        now do not commit any changes at all to `IStore`.
```
moreal, line 165: What is the meaning of `State` you mean? In my understanding, it should use `IStateStore` instead of `IStore`.

Libplanet/Blockchain/Renderers/IActionRenderer.cs
```diff
@@ -30,6 +32,15 @@ namespace Libplanet.Blockchain.Renderers
     /// (one time)</description></item>
     /// </list>
     /// </summary>
+    /// <remarks>Although <see cref="Transaction{T}"/>s affect the states in the <see
+    /// cref="IStore"/> all or nothing at all (i.e., atomically), <see cref="IActionRenderer{T}"/>
```
moreal, line 36: It has the same context with the review comment in `CHANGES.md`. Does the `states` mean state in `IStateStore`? If then, it seems better to use `IStateStore` instead of `IStore`.

Libplanet/Blockchain/Renderers/AtomicActionRenderer.cs
```diff
@@ -0,0 +1,179 @@
...
+            _errored = false;
+        }
+
+        /// <summary>
+        /// The inner action renderer which has the <em>actual</em> implementations and expects to
+        /// receive no <see cref="RenderActionError"/>/<see cref="UnrenderActionError"/> events.
+        /// </summary>
+        public IActionRenderer<T> ActionRenderer { get; }
+
+        /// <inheritdoc cref="DelayedRenderer{T}.RenderBlock(Block{T}, Block{T})"/>
```
moreal, line 51: Is there a reason to inherit document of `DelayedRenderer` not `IRenderer` for only this method, like other methods do? In my sight, it seems not related with `DelayedRenderer{T}.RenderBlock` method.

longfin (commented):
It seems that need to rebase.

CHANGES.md
dahlia: I was be blurred about that.  Amended: 1f1ed497c892b56c6802a9d15cdeb4cddd1b98db.

Libplanet/Blockchain/Renderers/IActionRenderer.cs
dahlia: Amended: 1f1ed497c892b56c6802a9d15cdeb4cddd1b98db.

Libplanet/Blockchain/Renderers/AtomicActionRenderer.cs
dahlia: It was a mistake made during I copy-and-pasted things… 🙄 Anyway amended: 69b6474d7914f3929154ed9a6474c5f8828f6661.

longfin (approved):

CHANGES.md
```diff
@@ -182,9 +201,12 @@ To be released.
 [#1240]: https://github.com/planetarium/libplanet/pull/1240
 [#1242]: https://github.com/planetarium/libplanet/pull/1242
 [#1265]: https://github.com/planetarium/libplanet/pull/1265
+[#1267]: https://github.com/planetarium/libplanet/issues/1267
 [#1268]: https://github.com/planetarium/libplanet/pull/1268
 [#1272]: https://github.com/planetarium/libplanet/pull/1272
 [#1274]: https://github.com/planetarium/libplanet/pull/1274
+[#1275]: https://github.com/planetarium/libplanet/pull/1275
+
```
moreal, line 209: It seems to make the interval between changelogs for each versions, into three new lines. In my memory, it was intended to make the interval into two new lines but [other pull request](#1272) also does and it was occurred during rebasing. Other intervals seems all two lines except for between *0.11.1* and *0.11.0*... Could you remove this line?

```suggestion
```

moreal (changes requested):
Thanks to apply reviews 🙏🏻  I believe this is my last review 👍🏻

moreal (approved):
🚀

dahlia:
@planetarium/libplanet PTAL.

riemannulus (approved):

dahlia:
@planetarium/libplanet All green now.  Please take a look again!

riemannulus (approved):

longfin (approved):

PR #1276, opened 2021-04-30
[skip changelog] Nullable enabled on all non-build-breaking files

greymistcube:
Added `#nullable enable` directive to all possible `.cs` files where `dotnet build` doesn't break.

Stats:
* Number of `.cs` files without `#nullable enable` before change: 154
* Number of `.cs` files without `#nullable enable` after change: 50

Note:
* Only `.cs` files inside projects with specified language version 8.0 *and* project-wide nullable context not already enabled are considered.
* Although this is relates to issue [#458](https://github.com/planetarium/libplanet/issues/458), doesn't close the issue.

### DIFF
--- a/Libplanet.Explorer.UnitTests/Common/Action/NoOpAction.cs
+++ b/Libplanet.Explorer.UnitTests/Common/Action/NoOpAction.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using Bencodex.Types;
 using Libplanet.Action;
 
--- a/Libplanet.Explorer.UnitTests/Controllers/GenericControllerFeatureProviderTests.cs
+++ b/Libplanet.Explorer.UnitTests/Controllers/GenericControllerFeatureProviderTests.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Immutable;
 using System.Linq;
--- a/Libplanet.Explorer.UnitTests/GraphTypes/AddressTypeTest.cs
+++ b/Libplanet.Explorer.UnitTests/GraphTypes/AddressTypeTest.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using Libplanet.Explorer.GraphTypes;
 using Xunit;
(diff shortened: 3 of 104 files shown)


moreal (approved):

longfin (approved):

dahlia (changes requested):
Although this could be built without errors, every single file should be reviewed with the naked eyes.  For example, `BlockHeader.PreviousHash` should be `ImmutableArray<byte>?` (the type should end with a question mark), because it can be `null` for genesis blocks.

Anyway, this patch is too big to be properly reviewed.  I recommend you to split this into many small patches instead so that peers can deeply look into them one by one.  /cc @planetarium/libplanet

greymistcube:
@dahlia 
Thanks for the comment. I see there are further considerations for properly propagating null context. Dividing up the commits shouldn't be too hard, but to achieve what you've suggested, I presume it would require a considerable amount of additional resources to go over every single type declarations. Not sure what the best course of action would be then.

I think then pure number of affected files may be too large. Perhaps splitting this PR to smaller PR's may be desirable?

greymistcube:
@dahlia 
> For example, `BlockHeader.PreviousHash` should be `ImmutableArray<byte>?` (the type should end with a question mark), because it can be null for genesis blocks.

On a side note, I thought compilers were suppose to catch those errors. From what I see from the code, `BlockHeader.PreviousHash` gets assigned `ImmutableArray<byte>.Empty` not `null`.

dahlia:
> Dividing up the commits shouldn't be too hard, but to achieve what you've suggested, I presume it would require a considerable amount of additional resources to go over every single type declarations.

The issue #458 would be the long-term mission the team needs to work on altogether, so you don't need to check every file this time.  This work would take few months IMHO.

> I think then pure number of affected files may be too large. Perhaps splitting this PR to smaller PR's may be desirable?

I suggest you to just pick few files to read carefully, and make patches for these files first.  If it goes well, you could work on more files then, or work on another issue instead.

greymistcube:
Okay. I'll reset and parse through files and try to make more sensible changes to a smaller unit.

greymistcube:
Think this better be closed for now.

PR #1277, opened 2021-05-03
add error log to rocksdbstore

area363:
This PR adds error logs to `RocksDBStore` to better assess preloading failures.

### DIFF
--- a/Libplanet.RocksDBStore/RocksDBStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBStore.cs
@@ -22,7 +22,6 @@ namespace Libplanet.RocksDBStore
     /// <seealso cref="IStore"/>
     public class RocksDBStore : BaseStore
     {
-        #pragma warning disable MEN002 // Line is too long
         private const string BlockDbRootPathName = "block";
         private const string BlockIndexDbName = "blockindex";
         private const string BlockPerceptionDbName = "blockpercept";
@@ -190,11 +189,7 @@ namespace Libplanet.RocksDBStore
                 }
                 catch (Exception e)
                 {
-                    _logger.Error(
-                        e,
-                        $"An unexpected exception occurred on {nameof(ListChainIds)}: {{Message}}",
-                        e.Message
-                    );
+                    LogUnexpectedException(nameof(ListChainIds), e);
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -127,6 +127,7 @@ To be released.
         `IReadOnlyList<T>` (was `IEnumerable<T>`).
      -  The type of `Transactions` property of `Block<T>` became
         `IReadOnlyList<T>` (was `IEnumerable<T>`).
+ -  Added error logging to `RocksDBStore`. [[#1277]]
```
dahlia, line 130: IMHO this shouldn't be recorded in the changelog.  You could bypass the changelog check by adding `[skip changelog]` in the commit message.

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -22,6 +22,7 @@ namespace Libplanet.RocksDBStore
     /// <seealso cref="IStore"/>
     public class RocksDBStore : BaseStore
     {
+        #pragma warning disable MEN002 // Line is too long
```
dahlia, line 25: This warning shouldn't be disabled.  How about defining a simple method to log an unexpected exception?  For example:

```csharp
private string LogUnexpectedException(string methodName, Exception e)
{
    string msg = $"An unexpected exception occurred on {methodName}: {{Message}}";
    _logger.Error(e, msg, e.Message);
}
```

dahlia (dismissed):

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -187,6 +187,11 @@ public override IEnumerable<Guid> ListChainIds()
                 {
                     continue;
                 }
+                catch (Exception e)
+                {
+                    LogUnexpectedException(nameof(ListChainIds), e);
+                    continue;
+                }
```
longfin, line 194: Strictly speaking, this change changes the semantics so that unexpected exceptions are not thrown out of this block. In general, these changes don't cause much of a problem, but if you don't intend to change the behavior, it's better to use `throw`.

```suggestion
                catch (Exception e)
                {
                    LogUnexpectedException(nameof(ListChainIds), e);
                    throw;
                }
```

dahlia:
Rather than making a merge commit (02ec3cc0e2428464f4c52192ec2ff91f619e6166), could you rebase commits on the up-to-date *main* branch instead?

```
git pull --rebase git://github.com/planetarium/libplanet.git main
```

area363:
Applied your suggestions! 🙇  @longfin @dahlia

area363:
PTAL one more time! @longfin @dahlia 🙏

dahlia (approved):

longfin (approved):

PR #1292, opened 2021-05-13
Refactor/valid block enforcement

greymistcube:
This refactoring patch attempts to enforce creation of only valid `Block<T>` instances. Here, validity refers to having a proper `Block<T>` object with all the necessary fields filled in, *not* the consistency of the values for such `Block<T>`.

Other than the prevention of exposing invalid `Block<T>` representation, Internally there are additional structural changes in order to streamline the `Block<T>` creation process.

Creation of a `Block<T>` instance can be largely split into two parts:
- Filling in and calculating up to `Block<T>.PreEvaluationHash`.
- Additional processing to find and filling in `Block<T>.StateRootHash` and `Block<T>.Hash`.

The former is handled by introducing a new intermediary class called `BlockDraft<T>`.


greymistcube:
Closing stale PR. I'll open a new one once this is ready again.

Issue #1297, opened 2021-05-13
Put `TxExecution` also in `BlockChain<T>.MineBlock()`


moreal:
Since #1289, `TxExecution`, the transaction execution summary, was introduced. But it seems not to put `TxExecution` when appending through `BlockChain<T>.MineBlock()` method. It is because `BlockChain<T>.ExecuteActions()` method puts `TxExecution` and `BlockChain<T>.Append()` method calls `BlockChain<T>.ExecuteActions()` only when the `actionEvaluations` is `null`. If this is not intended, it should be fixed.

If you want to test this, write test code like below and it will fail.
```csharp
[Fact]
public void MineBlock_ShouldPut_TxExecution()
{
  // Initialize IStore store, BlockChain<T> blockChain, IAction action
  var tx = blockChain.MakeTransaction(minerPrivateKey, new [] { action });
  var block = await blockChain.MineBlock(minerPrivateKey.ToAddress());
  Assert.NotNull(store.GetTxExecution(block.Hash, tx.Id));
}
```

PR #1304, opened 2021-05-18
Let `BlockChain<T>.MineBlock()` also updates `TxExecution`s

dahlia:
Fixes the bug #1297.


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1000,6 +1000,20 @@ public long GetNextTxNonce(Address address)
                 {
                     _rwlock.ExitWriteLock();
                 }
+
+                // Prepare TxExecutions.  Note that these should be realized into an actual array
+                // before _rwlock.EnterWriteLock(), because the following function tries to access
+                // to the current states.
+                TxExecution[] txExecutions = MakeTxExecutions(block, actionEvaluations).ToArray();
```
longfin, line 1007: How about unifying `MakeTxExecution()` calls here and the inside of `Append()`? If `TxExecution` is of a type similar to `ActionEvaluation`, it might be treated similarly. (Calculate once and store only once)

Libplanet/Blockchain/BlockChain.TxExecution.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        internal void UpdateTxExecutions(IEnumerable<TxExecution> txExecutions)
+        {
+            int count = 0;
+            foreach (TxExecution txExecution in txExecutions)
+            {
+                // Note that there are two overloaded methods of the same name PutTxExecution()
+                // in IStore.  As those two have different signatures, run-time polymorphism
+                // does not work.  Instead, we need the following hard-coded branch:
+                switch (txExecution)
```
limebell, line 87: If only `TxSuccess` and `TxFailure` is allowed as last discussion, how about deal other cases with `default` and throw exception?

dahlia: I left [`Trace.Assert()`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.trace.assert?view=netstandard-2.0) instead: 8272e2c4db35bf3a308bedba5edbd993e74dfdab.

longfin (approved):

dahlia:
@planetarium/libplanet Rebased and the build succeeds now.  Please take a look again!

riemannulus (approved):

Issue #1301, opened 2021-05-17
Streamline `Block<T>` evaluation process


greymistcube:
It seems current `Block<T>` evaluation is unnecessarily complicated resulting in high dependency. When evaluating `IAction`s of a `Block<T>`, necessary information pertaining to evaluation is passed on to several different classes:

* `BlockEvaluator`: `Block<T>` instance
* `Block<T>`: `AccountStateGetter`, `AccountBalanceGetter`
* `Transaction<T>`: meta info of `Block<T>`
* `ActionEvaluation`: meta info of `Block<T>` and `Transaction<T>`

As evaluation from `Block<T>` and `Transaction<T>` classes rely on external context, I don't think these classes should have their own instantiated evaluation methods. I propose moving `Block<T>.Evaluate()` and `Transaction<T>.EvaluateActions()` to either `BlockEvaluator` or some other external helper class.

dahlia:
Historically speaking, `BlockEvaluator` was introduced along with MPT much later than others, and it purposed to eliminate circular dependency that `BlockChain<T>()` requires a `genesisBlock` and the genesis block needs to know how to evaluate its block action (for its state root hash) hash at a time.  At that time, it was (probably) considered as a workaround of that, so we've never tried to give `BlockEvaluator` the dedicated role in evaluating actions.  If we're willing to give such role to `BlockEvaluator`, IMHO we'd better rename it, like `ActionEvaluator` for example.

greymistcube:
Yes, I agree renaming `BlockEvaluator` to something else might be better, as a newly created class might have to additionally expose something like "`EvaluateTransaction()`". If there is no larger issue in regards to separating evaluation from `Block<T>` and `Transaction<T>` entities, I'll be temporarily working on this first before continuing with #1164.

dahlia:
> If there is no larger issue in regards to separating evaluation from `Block<T>` and `Transaction<T>` entities, I'll be temporarily working on this first before continuing with #1164.

If we separate two things it is probably for better testability.

dahlia:
> Historically speaking, `BlockEvaluator` was introduced along with MPT much later than others, and it purposed to eliminate circular dependency that `BlockChain<T>()` requires a `genesisBlock` and the genesis block needs to know how to evaluate its block action (for its state root hash) hash at a time. At that time, it was (probably) considered as a workaround of that, so we've never tried to give `BlockEvaluator` the dedicated role in evaluating actions.

[...]

PR #1305, opened 2021-05-18
Refactor/streamline block evaluation

greymistcube:
This PR attempts to resolve the problem mentioned in #1301.

Currently, what should be a relatively simple `BlockEvaluator.Evaluate()` call is handled in four different places. To mitigate this issue, a new `ActionEvaluator` class is introduced to keep the entire evaluation process within this new class. All methods for `ActionEvaluator` are pretty much copied from `BlockEvaluator`, `Block<T>`, `Transaction<T>`, and `ActionEvaluation`, aside from minor internal refactoring and renaming for better consistency and clarity.


CHANGES.md
```diff
@@ -260,6 +270,7 @@ To be released.
 [#1289]: https://github.com/planetarium/libplanet/pull/1289
 [#1295]: https://github.com/planetarium/libplanet/issues/1295
 [#1296]: https://github.com/planetarium/libplanet/pull/1296
+[#1301]: https://github.com/planetarium/libplanet/issues/1301
```
limebell, line 273:
```suggestion
[#1301]: https://github.com/planetarium/libplanet/issues/1301
[#1305]: https://github.com/planetarium/libplanet/pull/1305
```

limebell (changes requested):
Some tests seem to fail because of the changes of this PR. Please check [Azure DevOps](https://dev.azure.com/planetarium/libplanet/_build/results?buildId=5108&view=results) for more detail.

CHANGES.md
greymistcube: Yup, this isn't fully ready for deployment. :(

greymistcube:
https://github.com/planetarium/libplanet/blob/25f62f8561c1f9037c1033ddf951d7a612030240/Libplanet.Tests/Blockchain/BlockChainTest.cs#L1826-L1837

This is one of the failing test, but I'm not sure what the intention is here. If we are evaluating a block action for the `genesis` block, shouldn't assertion be equal to `1`?

limebell:
Not a strong suggestion though, how about squash some commits? (i.e. [`414e8fd`](https://github.com/planetarium/libplanet/pull/1305/commits/414e8fd9d6eabb499a5c1df1dc273dd1765b66ee) and [`85710af`](https://github.com/planetarium/libplanet/pull/1305/commits/85710af8285457231d14155c765d3083256368b4))

greymistcube:
Should be ready for review. Since the changes are quite substantial, I'll be writing a summary soon.

greymistcube:
# Summary

New `ActionEvaluator<T>` class is introduced to handle all evaluation.

## Migration

- `ActionEvaluator<T>.Evaluate()`: Changed from `BlockEvaluator.EvaluateActions()`.
- `ActionEvaluator<T>.EvaluateBlock()`: Changed from `Block<T>.Evaluate()`.
- `ActionEvaluator<T>.EvaluateTxs()`: Changed from `Block<T>.EvaluateActionsPerTx()`.
- `ActionEvaluator<T>.EvaluateTx()`: Changed from `Transaction<T>.EvaluateActionsGradually()`.
- `ActionEvaluator<T>.EvaluateTxResult()`: Changed from `Transaction<T>.EvaluateActions()`.
- `ActionEvaluator<T>.EvaluatePolicyBlockAction()`: Changed from `BlockEvaluator.EvaluateBlockAction()`.
- `ActionEvaluator<T>.EvaluateActions()`: Changed from `ActionEvaluation.EvaluateActionsGradually()`.
- Other corresponding test methods moved to `ActionEvaluatorTest.cs`.

## Introduced

- `ActionEvaluator<T>.GetUpdatedAddresses()`: Separated from `ActionEvaluator<T>.EvaluateTxResult()`.
  - Used by `Transaction<T>` to get updated addresses.
  - Hard coded dependency injection from `Transaction<T>` removed. This is now handled internally.

## Notable changes

- Only the main entry point for evaluation, namely `ActionEvaluator<T>.Evaluate()`, directly takes external delegates.
  - All other methods depend on the previous states of type `IAccountStateDelta` before method call as evaluation context.
- Access modifier of `ActionEvaluator<T>.Evaluate()` has changed to `public` from `internal` and that of `ActionEvaluator<T>.EvaluateBlock()` has changed to `internal` from `public`.
- Number of parameters have been greatly reduced.
  - This is mainly done by requiring `Block<T>` as an argument instead of using individual values of `Block<T>`'s fields.
- Lazy evaluation is now strictly enforced for all methods with return type `IEnumerable<T>`.
  - Entire evaluation is postponed until right before returning the whole list of `ActionEvaluation`s from `ActionEvaluator<T>.Evaluate()` which has a return type of `IReadOnlyList<T>`.
  - This was not the case where a realization of `IEnumerable<T>` was prematurely happening inside `ActionEvaluator<T>.EvaluateBlock()`.
- `ActionEvaluator<T>.EvaluateTxs()` now has a simple `IEnumerable<Transaction<T>>` return type.
  - This is done by moving "updated addresses check" down to `ActionEvaluator<T>.EvaluateTx()` from `ActionEvaluator<T>.EvaluateBlock()`.
  - This removes the necessity for keeping track of which `Transaction<T>` an `ActionEvaluation` is tied to.
- A couple of private helper methods, `ActionEvaluator<T>.GetPreviousBlockOutputStates()` and `ActionEvaluator<T>.InitializeAccountGettersPair()`, are introduced to initialize the setup.

## Future suggestions

- A call to `Block<T>.Validate()` should not be inside `ActionEvaluator<T>.Evaluate()`.
- I suggest making `_policyBlockAction` non-nullable.
  - Designwise , it is probably better to have a default `IAction` that does nothing when evaluated.
- I also suggest making `_trieGetter` non-nullable.
  - Not even sure if trie data type should be enforced. Either way, if we want a mock storage, this should be handled outside the `ActionEvaluator<T>` class.
- Probably better to have parameters of `ActionEvaluator<T>.EvaluateActions` simplified. All the necessary information can be inferred from `Block<T>` and `Transaction<T>` to evaluate.
- The usage of `StateCompleterSet<T>` together with `StateGetter<T>`, `BalanceGetter<T>`, `AccountStateGetter`, and `AccountBalanceGetter` is quite convoluted. Concurrent refactoring may be needed.
  - Although I have introduced `StateGetter<T>` and `BalanceGetter<T>` as `public` `delegate`s for convenience, these `delegate`s are lacking documentation. I do not think I fully understand the purpose of these, so I can't concisely explain what these are for.
- One additional miner qualm is that `Block<T>.Miner` is nullable at the moment. It is probably better to be able to use `Block<T>.Miner` instead of `Block<T>.Miner.GetValueOrDefault()` from the perspective of the `ActionEvaluator<T>` class.

## Remarks

- Substantial amount of documentation has been reworked.
- I did my best to fix the tests to accommodate the changes made, but due to my lack of experience in unit testing, updated test code may need extra care while reviewing.
- I am sure there are other things I forgot to mention. I'll make additional comments as things come up.

limebell:
image

Build seems to fail in mono.

greymistcube:
@limebell 
Fixed. Older language versions not supporting `Zip()` taking a single argument was the issue.

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+                signer: tx.Signer,
+                signature: tx.Signature,
+                actions: tx.Actions.Cast<IAction>().ToImmutableList(),
+                rehearsal: true,
+                previousBlockStatesTrie: null);
+
+            if (evaluations.Any())
+            {
+                return evaluations.Last().OutputStates.UpdatedAddresses;
+            }
```
longfin, line 146: Trivial
```suggestion
            if (evaluations.LastOrDefault() is { } lastEval)
            {
                return lastEval.OutputStates.UpdatedAddresses;
            }
```

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+        /// <see cref="IAccountStateDelta.UpdatedAddresses"/> of the given
+        /// <paramref name="previousStates"/> as well.</returns>
+        [Pure]
+        internal IAccountStateDelta EvaluateTxResult(
+            Block<T> block,
+            Transaction<T> tx,
+            IAccountStateDelta previousStates,
+            bool rehearsal = false)
+        {
+            var evaluations = EvaluateTx(
```
longfin, line 472:
```suggestion
            ImutableList<ActionEvaluation> evaluations = EvaluateTx(
```

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+                blockIndex: block.Index,
+                txid: null,
+                previousStates: previousStates,
+                miner: block.Miner.GetValueOrDefault(),
+                signer: block.Miner.GetValueOrDefault(),
+                signature: Array.Empty<byte>(),
+                actions: new[] { _policyBlockAction }.ToImmutableList(),
+                rehearsal: false,
+                previousBlockStatesTrie: previousBlockStatesTrie,
+                blockAction: true).First();
```
longfin, line 529:
```suggestion
                blockAction: true).Single();
```

Libplanet/Action/ActionEvaluator.cs
greymistcube: Hmm... haven't really put my thought into it when writing this, but looking back, wouldn't this be problematic if for whatever reason `default` is implemented and `AccountStateDeltaImpl` becomes a `struct`? I know it is clear from the code that `AccountStateDeltaImpl` is a `class` (and also it implements an `interface`, which also makes it "likely" to be a `class`) but conceptually, I don't think an evaluation output should necessarily be a `class`.

longfin: Fair enough.

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+        /// <see cref="IAccountStateDelta.UpdatedAddresses"/> of the given
+        /// <paramref name="previousStates"/> as well.</returns>
+        [Pure]
+        internal IAccountStateDelta EvaluateTxResult(
+            Block<T> block,
+            Transaction<T> tx,
+            IAccountStateDelta previousStates,
+            bool rehearsal = false)
+        {
+            ImutableList<ActionEvaluation> evaluations = EvaluateTx(
```
greymistcube, line 472:
```suggestion
            ImmutableList<ActionEvaluation> evaluations = EvaluateTx(
```
Typo.

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+            }
+
+            byte[] preEvaluationHashBytes = preEvaluationHash.ToByteArray();
+            int seed =
+                (preEvaluationHashBytes.Length > 0
+                    ? BitConverter.ToInt32(preEvaluationHashBytes, 0) : 0)
+                ^ (signature.Any() ? BitConverter.ToInt32(hashedSignature, 0) : 0);
+
+            IAccountStateDelta states = previousStates;
+            ILogger logger = Log.ForContext<ActionEvaluation>();
```
dahlia, line 225: Why not use `_logger` instead?

Libplanet.Tests/Action/ActionEvaluatorTest.cs
```diff
@@ -0,0 +1,997 @@
...
+                timestamp: timestamp,
+                transactions: txs);
+            var stateRootBlock = TestUtils.MineGenesis(
+                timestamp: timestamp,
+                transactions: txs).AttachStateRootHash(stateStore, null);
+            StateGetter<RandomAction> nullStateGetter =
+                (address, hashDigest, stateCompleter) => null;
+            BalanceGetter<RandomAction> nullBalanceGetter =
+                (address, currency, hashDigest, fungibleAssetStateCompleter)
+                    => new FungibleAssetValue(currency);
```
dahlia, line 84: Would such null state/balance getters be frequently used?  If so, how about changing the visibility of `ActionEvaluator<T>._nullStateGetter` & `_nullBalanceGetter` into `public static readonly`?  (Or it might be better to define them in normal static methods instead.  If a method's signature is matched to a delegate type-wise, the compile treat the method as a value of the delegate type.)

Libplanet.Tests/Action/ActionEvaluatorTest.cs
```diff
@@ -0,0 +1,997 @@
...
+                genesisHash: chain.Genesis.Hash,
+                actions: new[] { action });
+
+            chain.StageTransaction(tx);
+            await chain.MineBlock(_storeFx.Address1);
+
+            var actionEvaluation = chain.ActionEvaluator.Evaluate(
+                chain.Tip,
+                StateCompleterSet<TestEvaluateAction>.Recalculate);
+            Assert.False(actionEvaluation[0].InputContext.BlockAction);
```
dahlia, line 144: Could you test if `ActionEvaluation`'s other properties have correct values too?  Especially, it would be better if it properly has `Exception` on failed actions.

dahlia (commented):
> - I suggest making `_policyBlockAction` non-nullable.
>   - Designwise , it is probably better to have a default `IAction` that does nothing when evaluated.

On .NET, interfaces cannot have `default` values, because they are treated as reference types.  Their default values are merely `null`.

> - The usage of `StateCompleterSet<T>` together with `StateGetter<T>`, `BalanceGetter<T>`, `AccountStateGetter`, and `AccountBalanceGetter` is quite convoluted. Concurrent refactoring may be needed.

If it's possible I believe simply adding parameters to the existing `AccountStateGetter` & `AccountBalanceGetter` delegates would be easier to understand and maintain than having one more pair of `StateGetter<T>` & `BalanceGetter<T>`.  Of course, they need to have generic parameters (`AccountStateGetter<T>` & `AccountBalanceGetter<T>`) too if we'd like to do so.

---

Good job.  Thanks for your hard work.  Anyway, your branch is now outdated.  Needs to rebase again.

Libplanet/Action/ActionEvaluator.cs
greymistcube: Partly because `EvaluateActions()` is left as `static`. Not sure if it was the best design choice, but the my reasoning behind it was to leave last mile/heavy lifting as `static` since this may be considered more of as a utility method, and possibly for optimization.

Libplanet.Tests/Action/ActionEvaluatorTest.cs
greymistcube: Went back and forth with this a little bit. Thought about exposing `ActionEvaluator<T>._nullStateGetter` and such as `public`, but then realized it would most likely only be for testing (and possibly for benchmarking). Now that I'm going over `ActionEvaluator.cs` again, `ActionEvaluator<T>._nullStateGetter` and `ActionEvaluator<T>._nullBalanceGetter` don't even get used. 🙃

I'll just remove them from `ActionEvaluator<T>`.

Libplanet/Action/AccountBalanceGetter.cs
```diff
@@ -16,4 +18,11 @@ namespace Libplanet.Action
     /// The <paramref name="address"/>'s balance of the <paramref name="currency"/>.
     /// </returns>
     public delegate FungibleAssetValue AccountBalanceGetter(Address address, Currency currency);
+
+    public delegate FungibleAssetValue BalanceGetter<T>(
+        Address address,
+        Currency currency,
+        BlockHash? hashDigest,
+        FungibleAssetStateCompleter<T> fungibleAssetStateCompleter)
+            where T : IAction, new();
```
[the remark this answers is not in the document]
greymistcube: Sorry, I got bit lazy towards the end. Didn't really decide what to do with `delegates`. After a brief survey, the consensus seems to be to either put them in the related class file, which would be `ActionEvaluator.cs` in this case, or have separate files, which seems bit excessive. In any case, I'll separate them out.

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+                accountStateGetter = address => _stateGetter(
+                    address,
+                    block.PreviousHash,
+                    stateCompleterSet.StateCompleter);
+                accountBalanceGetter = (address, currency) => _balanceGetter(
+                    address,
+                    currency,
+                    block.PreviousHash,
+                    stateCompleterSet.FungibleAssetStateCompleter);
+            }
```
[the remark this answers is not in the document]
greymistcube: ~Empty braces non-null pattern matching weirds me out.~

I'll fix it.

greymistcube:
> On .NET, interfaces cannot have default values, because they are treated as reference types. Their default values are merely null.

I disagree. I was thinking something akin to `_nullStateGetter` and such. I know the language does not support `default` values for `interface`es or `class`es per se, but I don't think this means we can't have a default behavior. To drive my point, I would say something like

```csharp
int Sum(List<int>? list)
{
    if (list is { })
    {
        return list.Sum();
    }
    else
    {
        return 0;
    }
}
```

is, in general, a bad design choice. A `null` check should not be deferred to an operating entity, but handled by the caller or, better yet, the list creator. Besides, `IBlockPolicy<T>.BlockAction` is not supposed to be `null`, but it just so happens `BlockPolicy<T>.BlockAction` implementation can be, due to discrepancy of nullable contexts, which should be fixed anyway.

Also, there is already a `NullAction` class found in `StoreExtensions.cs`. 🙃

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+        private static readonly StateGetter<T> _nullStateGetter =
+            (address, hashDigest, stateCompleter) => null;
+
+        private static readonly BalanceGetter<T> _nullBalanceGetter =
+            (address, currency, hashDigest, fungibleAssetStateCompleter)
+                => new FungibleAssetValue(currency);
+
+        private static readonly AccountStateGetter _nullAccountStateGetter = address => null;
+        private static readonly AccountBalanceGetter _nullAccountBalanceGetter =
+            (address, currency) => new FungibleAssetValue(currency);
```
greymistcube, line 30:
```suggestion
        private static readonly AccountStateGetter _nullAccountStateGetter = address => null;
        private static readonly AccountBalanceGetter _nullAccountBalanceGetter =
            (address, currency) => new FungibleAssetValue(currency);
```

Remove unused code.

Libplanet/Action/AccountBalanceGetter.cs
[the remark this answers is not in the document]
greymistcube: See cc305a5fbf1c1da20980b58283c715f9efb92e6b.

Libplanet/Action/ActionEvaluator.cs
dahlia: As logger contexts are for filtering logs, to me the best practice is to make logs belong to the context where the logging code is placed file-wise.  The current context implies the logging code is placed in *ActionEvaluation.cs*, it would be rather a trap when we investigate issues in production.

Libplanet.Tests/Action/ActionEvaluatorTest.cs
dahlia: If these are not useful to Libplanet users, but still useful for test/benchmark code how about turning them `internal static readonly`?

greymistcube: Hmm... I'm not fully on board with the idea of making things `internal` for purely testing and/or benchmarking. However, from a slightly different point of view, I can get behind the idea of providing additional `internal` tools for developers using the library, including us, of course. I'll make the changes. 🙃

Libplanet/Action/ActionEvaluator.cs
greymistcube: Well, I'll just turn the method into an instance method. Shouldn't affect the performance too hard anyways.

I think a more coherent solution would be to make it `private` and handle it from outside.

Libplanet.Tests/Action/ActionEvaluatorTest.cs
greymistcube: On a second thought, I still think the answer is no. I'll make the changes regardless for now, since this can be separated to a different issue.

dahlia:
> I disagree. I was thinking something akin to `_nullStateGetter` and such. I know the language does not support `default` values for `interface`s or `class`es per se, but I don't think this means we can't have a default behavior.

Fair enough, but we still don't need to fill it with a null getter by default.  If we make the constructor parameter required callers should explicitly pass a null getter by their hands.  To me, this way is a more error-prone design.

Libplanet.Tests/Action/ActionEvaluatorTest.cs
greymistcube: Done. See 04ae033430c70f543115322e3f613da1a3d8a113.

Libplanet/Action/ActionEvaluator.cs
greymistcube: Eh... ended up making `_logger` static instead. See 87849869dbdab8c8548cc56db3e054bd1dd291d9.

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -0,0 +1,581 @@
...
+using Libplanet.Blockchain;
+using Libplanet.Blockchain.Policies;
+using Libplanet.Blocks;
+using Libplanet.Store.Trie;
+using Libplanet.Tx;
+using Serilog;
+
+namespace Libplanet.Action
+{
+    public class ActionEvaluator<T>
```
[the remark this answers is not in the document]
greymistcube: See 0b854364be609ed868e6422f385739bee6a6120a. Might be tad bit short. 😶

longfin (dismissed):
But another conflict appears 😢

longfin (dismissed):

Libplanet.Tests/Action/ActionEvaluatorTest.cs
dahlia: @greymistcube It seems this is not addressed yet.

greymistcube: Still working on it. 😢

greymistcube: Implemented with mostly copy pasted code. See 55e4ec57c71953cead85c364719d3832aa5fad56.

dahlia (approved):

longfin (approved):

limebell (approved):

PR #1324, opened 2021-06-02
Find blocks in store instead of in chain

area363:
This PR changes the `BlockCompletion` predicate to use the `ContainsBlock` method of `IStore` instead of `BlockChain`. This way, `Swarm` won't have to download blocks again if preloading is aborted during the `Block Execution` stage.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,13 @@ To be released.
 
 ### Bug fixes
 
+ -  `Swarm<T>.PreloadAsync()` now checks the existence of blocks in the storage
+    (was in the blockchain).  [[#1324]]
+
 ### CLI tools
+
+[#1324]: https://github.com/planetarium/libplanet/pull/1324
+
 
 Version 0.12.0
 --------------
--- a/Libplanet/Net/Swarm.cs
+++ b/Libplanet/Net/Swarm.cs
@@ -572,7 +572,7 @@ namespace Libplanet.Net
                 FillBlocksAsyncStarted.Set();
 
                 var blockCompletion = new BlockCompletion<BoundPeer, T>(
-                    completionPredicate: workspace.ContainsBlock,
+                    completionPredicate: wStore.ContainsBlock,
                     window: InitialBlockDownloadWindow
                 );
 


dahlia (commented):
Could you add a minimum test case to reproduce the problem this PR tries to address?  Would it be difficult to reproduce in deterministic manner?

longfin:
it would be better if we have changelog about this patch. (as behaviour change?)

limebell:
PTAL @libplanet

dahlia (approved):

kfangw (approved):

kfangw:
LGTM.
I am not sure, but if there is performance issue(db read is expensive than memory read) then, how about the following?
- look up the blockchain first. If not exists, look up the store

limebell:
> LGTM.
> I am not sure, but if there is performance issue(db read is expensive than memory read) then, how about the following?
> 
> * look up the blockchain first. If not exists, look up the store

Looking up in the blockchain looks up storage too.

kfangw:
> > LGTM.
> > I am not sure, but if there is performance issue(db read is expensive than memory read) then, how about the following?
> > 
> > * look up the blockchain first. If not exists, look up the store
> 
> Looking up in the blockchain looks up storage too.

- BlockChain.ContainsBlock return true if blockchain `and` store both contains the block with other conditions 
- Store.ContainsBlock return true if store contains the block

IMO, what I mentioned(in blockchain but not in store) is not covered by BlockChain.ContainsBlock

Btw, BlockChain.ContainsBlock already searching db too, hence there is no further problem with performance by this change.

greymistcube (dismissed):

greymistcube:
I thought the main issue was unnecessarily redownloading a block where given block is *not* in blockchain but in store? Improving storage access performance is a separate issue.

dahlia (dismissed):

longfin (dismissed):

dahlia:
Retargeted to the 0.12-maintenance branch and rebased on that.

dahlia (approved):

kfangw (approved):

longfin (approved):

Issue #1314, opened 2021-05-26
`Hashcash`'s hash algorithm needs to be configurable


dahlia:
Currently `Hashcash` is dedicated SHA-256.  This should be configurable.

dahlia:
Closed by #1352.

PR #1326, opened 2021-06-02
Patch/transaction ordering update

### LINKED ISSUE #1322: Accommodate variable length `Block<T>.PreEvaluationHash` for `Block<T>.Transactions` ordering
Current "reordering" algorithm of `Block<T>.Transactions` depend on `Block<T>.PreEvaluationHash`'s bit length to be longer than that of the target reference values, currently XOR aggregated `Transaction<T>.Id`. Pertaining to #1314, allowing a usage of custom hash function with different output length for computing `Block<T>.PreEvaluationHash` is in the works. `Block<T>.Transactions` ordering mechanism should reflect/accommodate the upcoming change.

### LINKED ISSUE #1323: Fix improper `Block<T>.Transactions` ordering
Currently, `Block<T>.Transactions` are "shuffled" using `Block<T>.PreEvaluationHash` as an XOR bit mask on XOR aggregated `Transaction<T>.Id`s for `Transaction<T>`s with the same `Transaction<T>.Signer`. The problem is, as `Block<T>.PreEvaluationHash` is also used to show proof of work, by targeting given difficulty, `Block<T>.PreEvaluationHash` starts with `0` values when represented as bits. This results in XOR bit masking via `Block<T>.PreEvaluationHash` doing nothing in almost all cases. This can be resolved by using a hashed value of `Block<T>.PreEvaluationHash` instead of directly using `Block<T>.PreEvaluationHash` itself.

greymistcube:
Closes #1322
Closes #1323

Ordering `Transaction<T>`s for evaluation has been separated from `Block<T>()` constructor as `Block<T>.OrderTxsForEvaluation()` `static` method. Additionally, in preparation for #1314, a new ordering algorithm has been introduced. New algorithm will kick in once `ProtocalVersion` gets increased to at least `2`. Finally, pre-existing algorithm has been optimized for time and memory.

Here are some scripted benchmark results:
```
Number of Signers: 10
Number of Txs per Signer: 10
Running 1000 trials...

Old shuffling time: 2452 ticks
NewV0 shuffling time: 776 ticks
NewV2 shuffling time: 379 ticks

Number of Signers: 100
Number of Txs per Signer: 100
Running 1000 trials...

Old shuffling time: 190516 ticks
NewV0 shuffling time: 103822 ticks
NewV2 shuffling time: 26516 ticks
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -254,6 +254,8 @@ To be released.
  -  `BlockChain<T>.Append()` cumulates indexes for pairs (TxId and BlockHash).
     A transaction inclusion for a block is retrievable by using this index.
     [[#1315], [#1329]]
+ -  `Block<T>.Transactions` will be ordered using a different scheme once
+    the protocol version is bumped to `2`.  [[#1322], [#1323], [#1326]]
  -  `ActionEvaluator<T>.EvaluateActions()` now throws an unmanaged exception
     if `OutOfMemoryException` is caught from `IAction.Execute()`.
     [[#1320], [#1343]]
@@ -358,7 +360,10 @@ To be released.
 [#1315]: https://github.com/planetarium/libplanet/issues/1315
 [#1316]: https://github.com/planetarium/libplanet/issues/1316
 [#1320]: https://github.com/planetarium/libplanet/issues/1320
+[#1322]: https://github.com/planetarium/libplanet/issues/1322
+[#1323]: https://github.com/planetarium/libplanet/issues/1323
 [#1325]: https://github.com/planetarium/libplanet/pull/1325
+[#1326]: https://github.com/planetarium/libplanet/pull/1326
 [#1328]: https://github.com/planetarium/libplanet/pull/1328
 [#1329]: https://github.com/planetarium/libplanet/pull/1329
 [#1334]: https://github.com/planetarium/libplanet/pull/1334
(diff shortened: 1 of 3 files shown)


limebell (commented):
Can you please add a test case for this change? It would be great the added test fails at previous protocol, and pass at modified protocol.

And it would be great to have changelog describing this patch (in behavioral changes section).

greymistcube:
@limebell 

> Can you please add a test case for this change?

See 89a0addc686cce0f54fa8a8567caa902848c613a. Hard coded spec is added to ensure that there is no behavioral change between the new optimized V0 implementation and the old implementation.

> It would be great the added test fails at previous protocol, and pass at modified protocol.

Previous protocol doesn't "fail" per se, but does not work as intended. This is described in more detail in #1323. Statistical analysis may be done, but I don't see much point in doing so, and I'm not sure such test should be a part of unit testing. 😶 

> And it would be great to have changelog describing this patch (in behavioral changes section).

See 025a55f16d012df4353966de7de8fae534cbf8b8.

limebell (dismissed):

longfin:
@greymistcube it has a conflicting file...

greymistcube:
@longfin Resolved.

limebell:
Build seems to fail 😢

limebell:
not stale

dahlia (dismissed):

limebell (approved):

limebell (approved):

Issue #1321, opened 2021-06-01
[WIP] refactor: divide and reorganize IStore methods according to determinism


kfangw:
## Background
`IStore` has a bunch of functionalities. Hence, it leads to several issues.

- Heavy loads
Some of them do not need to be stored in the persistent database.
For example, the block perception time or transaction candidates can be handled in transient memory well.

- Lacks flexibility

- Lacks Understandability

## TO-BE
Divide IStore methods into Deterministic/Non-Deterministic.
Deterministic methods should remain in the IStore.
Non-Deterministic methods can be moved to another service/database.

### Deterministic and Persistent
Participants should maintain or able to retrieve the data.

* `BLOCK` _blockIndexDb
```csharp
        Block<T> GetBlock<T>(BlockHash blockHash);
        void PutBlock<T>(Block<T> block);
        bool DeleteBlock(BlockHash blockHash);
        IEnumerable<BlockHash> IterateBlockHashes();
        BlockDigest? GetBlockDigest(BlockHash blockHash);
        bool ContainsBlock(BlockHash blockHash);
        long CountBlocks();
```
* `TX` _txIndexDb
```csharp
        Transaction<T> GetTransaction<T>(TxId txid);
        void PutTransaction<T>(Transaction<T> tx);
        bool DeleteTransaction(TxId txid);
        bool ContainsTransaction(TxId txId);
        long CountTransactions();
        IEnumerable<TxId> IterateTransactionIds();
```

* `NONCE` _chainDb ---> move to state store?
```csharp
        long GetTxNonce(Guid chainId, Address address);
        void IncreaseTxNonce(Guid chainId, Address signer, long delta = 1);
        void ForkTxNonces(Guid sourceChainId, Guid destinationChainId);
        IEnumerable<KeyValuePair<Address, long>> ListTxNonces(Guid chainId);
```

### Transient
Below data are retrievable by replaying the deterministic data

#### Transient & deterministic
The intact node retrieves the data in a deterministic manner.

* `BLOCK INDEX` _chainDb ---> move to block indexer(not yet introduced)
```csharp
        long? GetBlockIndex(BlockHash blockHash);
        BlockHash? IndexBlockHash(Guid chainId, long index);
        long AppendIndex(Guid chainId, BlockHash hash);
        void ForkBlockIndexes(Guid sourceChainId, Guid destinationChainId, BlockHash branchpoint);
[...]

PR #1328, opened 2021-06-03
feat: introduce txid->blockhash index on IStore

### LINKED ISSUE #1294: Build `TxId` → `BlockHash` index in `IStore`
The current storage does not have such index, so we're virtually impossible to figure out the belonging block by a transaction.  The necessary methods in `IStore` would look like:

```csharp
BlockHash IndexBlockHash(Guid chainId, TxId txid);
void AppendIndex(Guid chainId, BlockHash blockHash, TxId txid);
```

One downside of the below interface would be that it bloats the size of each chain ID even more.  If we can fix the issue #1232, the interface would look like instead:

```csharp
BlockHash IndexBlockHash(BlockHash tipHash, TxId txid)
```

kfangw:
closes #1294 

### APIs
Following APIs are added to `IStore`

```csharp
void PutTxIdBlockHashIndex(TxId txId, BlockHash blockHash);
BlockHash? GetFirstTxIdBlockHashIndex(TxId txId);
IEnumerable<BlockHash> IterateTxIdBlockHashIndex(TxId txId);
void DeleteTxIdBlockHashIndex(TxId txId, BlockHash blockHash);
```

### Contract
* There exists only one record for a pair <TxId, BlockHash>.
* There are multiple BlockHash paired with the same TxId (consider `reorg` situation.)

### DIFF
--- a/Libplanet.RocksDBStore/MonoRocksDBStore.cs
+++ b/Libplanet.RocksDBStore/MonoRocksDBStore.cs
@@ -544,22 +544,19 @@ namespace Libplanet.RocksDBStore
             return null;
         }
 
-        /// <inheritdoc cref="BaseStore.PutTxIdBlockHashIndex(Guid, TxId, BlockHash)"/>
-        public override void PutTxIdBlockHashIndex(Guid chainId, TxId txId, BlockHash blockHash)
+        /// <inheritdoc cref="BaseStore.PutTxIdBlockHashIndex(TxId, BlockHash)"/>
+        public override void PutTxIdBlockHashIndex(TxId txId, BlockHash blockHash)
         {
-            var cf = GetColumnFamily(_chainDb, chainId);
-            _chainDb.Put(
+            _txExecutionDb.Put(
                 TxIdIndexKey(txId),
-                blockHash.ToByteArray(),
-                cf
+                blockHash.ToByteArray()
                 );
         }
 
-        /// <inheritdoc cref="BaseStore.GetTxIdBlockHashIndex(Guid, TxId)"/>
-        public override BlockHash? GetTxIdBlockHashIndex(Guid chainId, TxId txId)
+        /// <inheritdoc cref="BaseStore.GetTxIdBlockHashIndex(TxId)"/>
+        public override BlockHash? GetTxIdBlockHashIndex(TxId txId)
         {
-            var cf = GetColumnFamily(_chainDb, chainId);
-            if (!(_chainDb.Get(TxIdIndexKey(txId), cf) is { } blockHashByte))
+            if (!(_txExecutionDb.Get(TxIdIndexKey(txId)) is { } blockHashByte))
             {
                 return null;
             }
@@ -567,13 +564,11 @@ namespace Libplanet.RocksDBStore
             return new BlockHash(blockHashByte);
         }
 
-        /// <inheritdoc cref="BaseStore.DeleteTxIdBlockHashIndex(Guid, TxId)"/>
-        public override void DeleteTxIdBlockHashIndex(Guid chainId, TxId txId)
+        /// <inheritdoc cref="BaseStore.DeleteTxIdBlockHashIndex(TxId)"/>
+        public override void DeleteTxIdBlockHashIndex(TxId txId)
         {
-            var cf = GetColumnFamily(_chainDb, chainId);
-            _chainDb.Remove(
-                TxIdIndexKey(txId),
-                cf
+            _txExecutionDb.Remove(
+                TxIdIndexKey(txId)
             );
         }
 
--- a/Libplanet.RocksDBStore/MonoRocksDBStore.cs
+++ b/Libplanet.RocksDBStore/MonoRocksDBStore.cs
@@ -36,6 +36,7 @@ namespace Libplanet.RocksDBStore
         private static readonly byte[] TxNonceKeyPrefix = { (byte)'N' };
         private static readonly byte[] StagedTxKeyPrefix = { (byte)'t' };
         private static readonly byte[] TxExecutionKeyPrefix = { (byte)'e' };
+        private static readonly byte[] TxIdBlockHashIndexPrefix = { (byte)'i' };
         private static readonly byte[] IndexCountKey = { (byte)'c' };
         private static readonly byte[] CanonicalChainIdIdKey = { (byte)'C' };
 
@@ -543,6 +544,39 @@ namespace Libplanet.RocksDBStore
             return null;
         }
 
+        /// <inheritdoc cref="BaseStore.PutTxIdBlockHashIndex(Guid, TxId, BlockHash)"/>
+        public override void PutTxIdBlockHashIndex(Guid chainId, TxId txId, BlockHash blockHash)
+        {
+            var cf = GetColumnFamily(_chainDb, chainId);
+            _chainDb.Put(
+                TxIdIndexKey(txId),
+                blockHash.ToByteArray(),
+                cf
+                );
+        }
+
+        /// <inheritdoc cref="BaseStore.GetTxIdBlockHashIndex(Guid, TxId)"/>
+        public override BlockHash? GetTxIdBlockHashIndex(Guid chainId, TxId txId)
+        {
+            var cf = GetColumnFamily(_chainDb, chainId);
+            if (!(_chainDb.Get(TxIdIndexKey(txId), cf) is { } blockHashByte))
+            {
+                return null;
+            }
+
+            return new BlockHash(blockHashByte);
+        }
+
+        /// <inheritdoc cref="BaseStore.DeleteTxIdBlockHashIndex(Guid, TxId)"/>
+        public override void DeleteTxIdBlockHashIndex(Guid chainId, TxId txId)
+        {
+            var cf = GetColumnFamily(_chainDb, chainId);
+            _chainDb.Remove(
+                TxIdIndexKey(txId),
+                cf
+            );
+        }
+
         /// <inheritdoc cref="BaseStore.SetBlockPerceivedTime(BlockHash, DateTimeOffset)"/>
         public override void SetBlockPerceivedTime(
             BlockHash blockHash,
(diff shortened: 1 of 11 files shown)


limebell (commented):
Please update this branch to main and resolve conflicts. 🙇‍♂️

Libplanet/Store/IStore.cs
```diff
@@ -231,6 +231,42 @@ void PutBlock<T>(Block<T> block)
...
+        /// <summary>
+        /// Records a index for given pair <paramref name="txId"/> and <paramref name="blockHash"/>.
+        /// If there exist a record for <paramref name="txId"/> already,
+        /// it overwrites the record silently.
+        /// </summary>
+        /// <param name="chainId">The chain ID of the index.</param>
+        /// <param name="txId">The <see cref="TxId"/> of the <see cref="Transaction{T}"/>.</param>
+        /// <param name="blockHash">The <see cref="BlockHash"/> of the <see cref="Block{T}"/>.
+        /// </param>
+        public void PutTxIdBlockHashIndex(Guid chainId, TxId txId, BlockHash blockHash);
```
dahlia, line 243: The visibility keyword `public` is unnecessary in this case.

```suggestion
        void PutTxIdBlockHashIndex(Guid chainId, TxId txId, BlockHash blockHash);
```

Libplanet/Store/IStore.cs
```diff
@@ -231,6 +231,42 @@ void PutBlock<T>(Block<T> block)
...
+
+        /// <summary>
+        /// Determines whether the <see cref="IStore"/> contains index to <see cref="BlockHash"/>
+        /// with given <paramref name="txId"/>.
+        /// </summary>
+        /// <param name="chainId">The chain ID of the index.</param>
+        /// <param name="txId">The <see cref="TxId"/> of the <see cref="Transaction{T}"/>.</param>
+        /// <returns><c>true</c> if the index with given <paramref name="txId"/> is exists.
+        /// Otherwise, <c>false</c>.</returns>
+        public bool HasTxIdBlockHashIndex(Guid chainId, TxId txId);
```
dahlia, line 253:
```suggestion
        bool HasTxIdBlockHashIndex(Guid chainId, TxId txId);
```

Libplanet/Store/IStore.cs
```diff
@@ -231,6 +231,42 @@ void PutBlock<T>(Block<T> block)
...
+
+        /// <summary>
+        /// Retrieves the <see cref="BlockHash"/> indexed by the <paramref name="txId"/>.
+        /// </summary>
+        /// <param name="chainId">The chain ID of the index.</param>
+        /// <param name="txId">The <see cref="TxId"/> of the <see cref="Transaction{T}"/>.</param>
+        /// <returns><see cref="BlockHash"/> if the index exists. Otherwise <c>null</c>.</returns>
+        public BlockHash? GetTxIdBlockHashIndex(Guid chainId, TxId txId);
```
dahlia, line 261:
```suggestion
        BlockHash? GetTxIdBlockHashIndex(Guid chainId, TxId txId);
```

Libplanet/Store/IStore.cs
```diff
@@ -231,6 +231,42 @@ void PutBlock<T>(Block<T> block)
...
+
+        /// <summary>
+        /// Delete the index for the <paramref name="txId"/>.
```
dahlia, line 264: [We tend to use third person singular verbs for XML comments.][1]

```suggestion
        /// Deletes the index for the <paramref name="txId"/>.
```

[1]: https://weblogs.asp.net/dixin/csharp-coding-guidelines-6-documentation

Libplanet/Store/IStore.cs
```diff
@@ -231,6 +231,42 @@ void PutBlock<T>(Block<T> block)
...
+        /// </summary>
+        /// <param name="chainId">The chain ID of the index.</param>
+        /// <param name="txId">The <see cref="TxId"/> of the <see cref="Transaction{T}"/>.</param>
+        public void DeleteTxIdBlockHashIndex(Guid chainId, TxId txId);
```
dahlia, line 268:
```suggestion
        void DeleteTxIdBlockHashIndex(Guid chainId, TxId txId);
```

Libplanet/Store/BaseStore.cs
```diff
@@ -138,6 +138,22 @@ public Block<T> GetBlock<T>(BlockHash blockHash)
         /// <inheritdoc cref="IStore.GetTxExecution(BlockHash, TxId)"/>
         public abstract TxExecution GetTxExecution(BlockHash blockHash, TxId txid);
 
+        /// <inheritdoc cref="IStore.PutTxIdBlockHashIndex(Guid, TxId, BlockHash)"/>
+        public abstract void PutTxIdBlockHashIndex(Guid chainId, TxId txId, BlockHash blockHash);
+
+        /// <inheritdoc cref="IStore.HasTxIdBlockHashIndex(Guid, TxId)"/>
+        public bool HasTxIdBlockHashIndex(Guid chainId, TxId txId)
+        {
+            return GetTxIdBlockHashIndex(chainId, txId) != null;
```
dahlia, line 147: Although it's a trivial stylistic choice, in modern C# it tends to be like:

```suggestion
            return GetTxIdBlockHashIndex(chainId, txId) is { };
```

limebell (dismissed):

dahlia (approved):

Libplanet.RocksDBStore/MonoRocksDBStore.cs
```diff
@@ -543,6 +544,39 @@ public override TxExecution GetTxExecution(BlockHash blockHash, TxId txid)
             return null;
         }
 
+        /// <inheritdoc cref="BaseStore.PutTxIdBlockHashIndex(Guid, TxId, BlockHash)"/>
+        public override void PutTxIdBlockHashIndex(Guid chainId, TxId txId, BlockHash blockHash)
+        {
+            var cf = GetColumnFamily(_chainDb, chainId);
+            _chainDb.Put(
```
longfin, line 551: if we decide to store these indexes on `chainDb`, we also need to add a new method for forking because we can delete the previous `chainDb` after forking... 

Or, how about using the separated DB for storing?

kfangw:
The chain id may not be needed for this change. Please see the patch https://github.com/planetarium/libplanet/pull/1328/commits/5f43072a382023b1c2822e9ceccfd8a03e93f998

```csharp
public void PutTxIdBlockHashIndex(TxId txId, BlockHash blockHash);
public bool HasTxIdBlockHashIndex(TxId txId);
public BlockHash? GetTxIdBlockHashIndex(TxId txId);
public void DeleteTxIdBlockHashIndex(TxId txId);
```

IMHO, TxId->BlockHash index can be handled like TxExecution.
What do you think about this? @longfin @dahlia

Libplanet.RocksDBStore/MonoRocksDBStore.cs
kfangw: I was thinking that It(`forking`) is handled by other codebases. 
And I realized that the chain id may not be needed. So the `forking` also not needed.
please see the patch 5f43072

longfin: Yes, if we don't put it with `ColumnFamily`, there is no matter. 😄

kfangw:
The APIs were changed according to your advice @longfin

`chain id` is removed from the argument, and let the multiple pair <TxId, BlockHash> which share the same TxId.
Hence, `IterateTxIdBlockHashIndex` method is introduced to traverse the list of pairs by using the prefix(TxId).
And `GetTxIdBlockHashIndex` is changed to `GetFirstTxIdBlockHashIndex` method, which returns the first element or null if the list is empty.

```csharp
void PutTxIdBlockHashIndex(TxId txId, BlockHash blockHash);
BlockHash? GetFirstTxIdBlockHashIndex(TxId txId);
IEnumerable<BlockHash> IterateTxIdBlockHashIndex(TxId txId);
void DeleteTxIdBlockHashIndex(TxId txId, BlockHash blockHash);
```

Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -759,6 +760,33 @@ public override bool ContainsBlock(BlockHash blockHash)
             return false;
         }
 
+        /// <inheritdoc cref="BaseStore.PutTxIdBlockHashIndex(TxId, BlockHash)"/>
+        public override void PutTxIdBlockHashIndex(TxId txId, BlockHash blockHash)
+        {
+            _txExecutionDb.Put(
```
kfangw, line 766: Is it OK for you to share the same db with TxExecution? @dahlia 
IMHO, txExecutionDB consists of transient and deterministic data. So they can share the same db.  #1321

kfangw:
PTAL

Libplanet.RocksDBStore/RocksDBStore.cs
kfangw: cdf2e8d

kfangw:
PTAL. @longfin @dahlia @greymistcube @limebell @riemannulus 
Can I merge it?

CI passed except for Codecov test coverage.
Codecov coverage degradation is caused by the exclusion of test projects.

longfin:
Can you tell me when `GetFirstTxIdBlockHashIndex()` is useful? IMO, `IterateTxIdBlockHashIndex()` is sufficient in the general case.

longfin (approved):
Everything else looks fine.

limebell (approved):

kfangw:
> Can you tell me when `GetFirstTxIdBlockHashIndex()` is useful? IMO, `IterateTxIdBlockHashIndex()` is sufficient in the general case.

The API responds to the caller that a transaction is included in at least one block or not. It has a similar use case like `hasXXX` API.
As you mentioned, the caller can discover the fact thru the iterator API. 
However, IMHO, this API may useful who want to know the fact without iterating them.

longfin (approved):

PR #1329, opened 2021-06-03
feat: indexing TxId->BlockHash when appending block

### LINKED ISSUE #1315: Build `index`(txid -> blockhash) when a block is committed
depends on #1294 

It would be better if this operation can be configurable by the individual nodes.

kfangw:
closes #1315 

`BlockChain<T>.Append()` cumulates indexes for pairs (TxId and BlockHash).
A transaction inclusion for a block is retrievable by using this index.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1245,6 +1245,11 @@ namespace Libplanet.Blockchain
 
                     Store.AppendIndex(Id, block.Hash);
 
+                    foreach (var tx in block.Transactions)
+                    {
+                        Store.PutTxIdBlockHashIndex(Id, tx.Id, block.Hash);
+                    }
+
                     const string unstageStartMsg =
                         "Unstaging {Txs} transaction(s) which belong to the block " +
                         "#{BlockIndex} {BlockHash}...";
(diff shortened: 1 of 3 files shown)


limebell (dismissed):

longfin:
@kfangw it has conflict. 😢

kfangw:
PTAL.

longfin (dismissed):

limebell (dismissed):

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1245,6 +1245,11 @@ public IImmutableSet<TxId> GetStagedTransactionIds()
 
                     Store.AppendIndex(Id, block.Hash);
 
+                    foreach (var tx in block.Transactions)
```
longfin, line 1248:
```suggestion
                    foreach (Transaction<T> tx in block.Transactions)
```

longfin (approved):

Libplanet/Blockchain/BlockChain.cs
kfangw: Isn't it straightforward for you to guess the Type of the variable `tx` for this statement?

When you want to cast the right value into the left value, you'd better declare the Type explicitly and let the compiler do implicit casting. Or If it is very hard for a code reviewer to guess the Type, you'd better do that.

However
IMHO,  in many cases, for static type language, It is recommended to use `var` instead of declaring Type explicitly. 
If there is something that I missed, please teach me. And please put that rule into the coding rule.

PR #1338, opened 2021-06-07
Retry flaky tests using xRetry

dahlia:
We have some flaky tests, and it seems these won't be fixed in the near future.  Instead of manually rerunning the whole CI/CD job to retry one or two failed tests among about 700 succeeded tests.  And these tend to fail even after several retries.

As the team @planetarium/libplanet became more members, there are more pending pull requests, and rebasing barely succeeded pull requests becomes more painful today.

Instead of bearing such pain, this pull request introduces [xRetry] which provides `[RetryFact]` and `[RetryTheory]` attributes.  I attributed some flaky tests as `[RetryFact]`/`[RetryTheory]` in this patch.  FYI I found the list of flaky tests from [Azure Pipelines' test failure report][1].

[xRetry]: https://github.com/JoshKeegan/xRetry
[1]: https://dev.azure.com/planetarium/libplanet/_test/analytics?definitionId=3&contextType=build

### DIFF
--- a/Libplanet.Tests/Libplanet.Tests.csproj
+++ b/Libplanet.Tests/Libplanet.Tests.csproj
@@ -37,6 +37,7 @@
       </IncludeAssets>
     </PackageReference>
     <PackageReference Include="System.Collections.Immutable" Version="1.7.*" />
+    <PackageReference Include="xRetry" Version="1.5.0" />
     <PackageReference Include="xunit" Version="2.4.1" />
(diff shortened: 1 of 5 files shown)


limebell:
How about adding `SwarmTest.ReconnectToTurn`? It seems flaky too.

longfin (approved):

limebell (approved):

PR #1340, opened 2021-06-08
CLI for `TxId` → `BlockHash` index

### LINKED ISSUE #1316: Add `planet store index` command
depends on #1294 

The index `TxId -> BlockHash` is built when the block is committed. (by #1315)
Hence, the pair which is committed before the indexing introduced is not indexed yet.
This command helps the operator of the node to index the pair.

kfangw:
closes #1316 

add commands for indexing TxId->BlockHash
* planet store build-index-tx-block [STORE] [OFFSET] [LIMIT]
* planet store blocks-by-tx-id [STORE] [TX-ID]
* planet store block-hashes-by-tx-id [STORE] [TX-ID]

### DIFF
--- a/Libplanet.Extensions.Cocona.Tests/Commands/StoreCommandTest.cs
+++ b/Libplanet.Extensions.Cocona.Tests/Commands/StoreCommandTest.cs
@@ -1,10 +1,17 @@
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
 using Cocona;
+using Libplanet.Blockchain.Renderers.Debug;
+using Libplanet.Blocks;
+using Libplanet.Crypto;
 using Libplanet.Extensions.Cocona.Commands;
 using Libplanet.RocksDBStore.Tests;
+using Libplanet.Tests;
+using Libplanet.Tests.Common.Action;
 using Libplanet.Tests.Store;
+using Libplanet.Tx;
 using Xunit;
 
 namespace Libplanet.Extensions.Cocona.Tests.Commands
@@ -12,13 +19,20 @@ namespace Libplanet.Extensions.Cocona.Tests.Commands
     public class StoreCommandTest : IDisposable
     {
         private readonly ImmutableArray<StoreFixture> _storeFixtures;
-        private readonly StoreCommand _command;
         private readonly TextWriter _originalWriter;
+        private readonly Block<Utils.DummyAction> _genesisBlock;
+        private readonly Block<Utils.DummyAction> _block1;
+        private readonly Block<Utils.DummyAction> _block2;
+        private readonly Block<Utils.DummyAction> _block3;
+        private readonly Block<Utils.DummyAction> _block4;
+        private readonly Transaction<Utils.DummyAction> _transaction1;
+        private readonly Transaction<Utils.DummyAction> _transaction2;
+        private readonly Transaction<Utils.DummyAction> _transaction3;
 
         public StoreCommandTest()
         {
-            _command = new StoreCommand();
             _originalWriter = Console.Out;
+
             try
             {
                 _storeFixtures = ImmutableArray.Create<StoreFixture>(
@@ -32,13 +46,37 @@ namespace Libplanet.Extensions.Cocona.Tests.Commands
                 throw new SkipException("RocksDB is not available.");
             }
 
+            {
+                _genesisBlock = TestUtils.MineGenesis<Utils.DummyAction>();
+                _transaction1 = DummyTransaction();
+                _transaction2 = DummyTransaction();
+                _transaction3 = DummyTransaction();
+
+                _block1 = TestUtils.MineNext(_genesisBlock, new[] {_transaction1});
+                _block2 = TestUtils.MineNext(_block1, new[] {_transaction2});
+                _block3 = TestUtils.MineNext(_block2, new[] {_transaction3});
+                _block4 = TestUtils.MineNext(_block3);
+            }
+
+            var guid = Guid.NewGuid();
             foreach (var v in _storeFixtures)
             {
-                var guid = Guid.NewGuid();
                 v.Store.SetCanonicalChainId(guid);
-                v.Store.PutBlock(v.Block1);
-                v.Store.AppendIndex(guid, v.Block1.Hash);
-                v.Store.PutTransaction(v.Transaction1);
+                v.Store.PutBlock(_genesisBlock);
+                v.Store.AppendIndex(guid, _genesisBlock.Hash);
+
+                v.Store.PutBlock(_block1);
+                v.Store.AppendIndex(guid, _block1.Hash);
+                v.Store.PutTransaction(_transaction1);
+
+                v.Store.PutBlock(_block2);
+                v.Store.AppendIndex(guid, _block2.Hash);
+                v.Store.PutTransaction(_transaction2);
+
+                v.Store.PutBlock(_block3);
+                v.Store.AppendIndex(guid, _block3.Hash);
+                v.Store.PutTransaction(_transaction3);
+
                 (v.Store as IDisposable)?.Dispose();
                 (v.StateStore as IDisposable)?.Dispose();
             }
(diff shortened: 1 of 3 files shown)


Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
```diff
@@ -95,6 +135,21 @@ private static BlockHash GetBlockHash(IStore store, int blockIndex)
...
+        private static BlockHash GetBlockHash(IStore store, TxId txId)
+        {
+            if (!(store.GetFirstTxIdBlockHashIndex(txId) is { } blockHash))
+            {
+                throw Utils.Error($"cannot find the block with the TxId[{txId.ToString()}]");
+            }
+
+            if (store.IterateTxIdBlockHashIndex(txId).Count() > 1)
+            {
+                throw Utils.Error($"multiple BlockHashes found for TxId[{txId.ToString()}]");
```
limebell, line 147: Instead of throw error, how about return all `BlockHash`es found? This may occur pretty often if reorg happens.

kfangw: I Introduced a new command `BlockHashesByTxId` instead of changing an existing one for the purpose.
Because `BlockByTxId` command prints the whole of the block itself, it would be messy when the command list up blocks it found. @limebell

greymistcube: It could be useful to have a query for a block with the implemented behavior above. I'm not sure if it would be the least expected surprise though. Perhaps updating the command description may help. I think it might also be useful to query all the blocks with given `TxId`. How about having `GetBlocks()` to mirror the behaviour of `GetBlock()`?

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
```diff
@@ -26,7 +27,58 @@ public class StoreCommand
...
+        public void BuildIndexTxBlock(
+            [Argument("STORE", Description = StoreArgumentDescription)]
+            string home,
+            [Argument("OFFSET", Description = "block index")]
+            int offset,
+            [Argument("LIMIT", Description = "block index")]
+            int limit
+        )
+        {
+            var store = LoadStoreFromUri(home);
```
longfin, line 40: We prefer specifying type explicitly if code-reviewers can't guess its type from a statement. 
```suggestion
            IStore store = LoadStoreFromUri(home);
```

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
```diff
@@ -26,7 +27,58 @@ public class StoreCommand
...
+            var prev = DateTime.UtcNow;
```
longfin, line 41: `DateTimeOffset` is better than `DateTime` because it has a timezone offset too. (like aware datetime in python)
```suggestion
            DateTimeOffset prev = DateTimeOffset.UtcNow;
```

Or, we can use [`Stopwatch`](https://docs.microsoft.com/en-US/dotnet/api/system.diagnostics.stopwatch?view=net-5.0) for this purpose.

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
```diff
@@ -26,7 +27,58 @@ public class StoreCommand
...
+            foreach (var index in BuildTxIdBlockHashIndex(store, offset, limit))
+            {
+                Console.WriteLine($"processing {index}/{offset + limit}...");
+            }
+
+            Console.WriteLine($"It taken {DateTime.UtcNow - prev}");
```
longfin, line 47:
```suggestion
            Console.WriteLine($"It taken {DateTimeOffset.UtcNow - prev}");
```

limebell:
Please resolve conflict in changelog.

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
kfangw: I make a compromise with you guys about the command.
It changed to `GetBlocks` and return a list of blocks found.
And `BlockHashesByTxId` still there to return a list of block hashes

@limebell @greymistcube

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
kfangw: I prefer to use `var` instead of declaring type explicitly for static type language. It looks more readable for me.
But as you mentioned, for some statements like function calls, it is not straightforward for readers to guess the type of variable.

So I conclude that some statements like function calls would be better to be declared with Type.

Thanks for your advice.

kfangw:
PTAL.

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
```diff
@@ -95,6 +157,16 @@ private static BlockHash GetBlockHash(IStore store, int blockIndex)
             return blockHash;
         }
 
+        private static IEnumerable<Block<T>> IterateBlocks<T>(IStore store, TxId txId)
+            where T : IAction, new()
+        {
+            foreach (var blockHash in store.IterateTxIdBlockHashIndex(txId))
+            {
+                var block = GetBlock<T>(store, blockHash);
+                yield return block;
```
longfin, line 166:
```suggestion
                yield return GetBlock<T>(store, blockHash);
```

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
```diff
@@ -106,6 +178,31 @@ private static Transaction<T> GetTransaction<T>(IStore store, TxId txId)
...
+            {
+                yield return index++;
+                if (!(store.GetBlockDigest(blockHash) is { } blockDigest))
+                {
+                    throw Utils.Error(
+                        $"Block is missing for BlockHash: {blockHash} index: {index}.");
+                }
+
+                foreach (var txId in blockDigest.TxIds.Select(bytes => new TxId(bytes.ToArray()))
+                    .ToImmutableArray())
```
longfin, line 199:
```suggestion
                foreach (TxId txId in blockDigest.TxIds.Select(bytes => new TxId(bytes.ToArray())))
```

Libplanet.Extensions.Cocona/Commands/StoreCommand.cs
```diff
@@ -106,6 +178,31 @@ private static Transaction<T> GetTransaction<T>(IStore store, TxId txId)
...
 
+        private static IEnumerable<int> BuildTxIdBlockHashIndex(IStore store, int offset, int limit)
+        {
+            if (!(store.GetCanonicalChainId() is { } chainId))
+            {
+                throw Utils.Error("Cannot find the main branch of the blockchain.");
+            }
+
+            var index = offset;
+            foreach (var blockHash in store.IterateIndexes(chainId, offset, limit))
```
longfin, line 189:
```suggestion
            foreach (BlockHash blockHash in store.IterateIndexes(chainId, offset, limit))
```

Libplanet.Extensions.Cocona.Tests/Commands/StoreCommandTest.cs
```diff
@@ -2,23 +2,36 @@
...
+        private readonly Block<Utils.DummyAction> _genesisBlock;
+        private readonly Block<Utils.DummyAction> _block1;
+        private readonly Block<Utils.DummyAction> _block2;
+        private readonly Block<Utils.DummyAction> _block3;
+        private readonly Block<Utils.DummyAction> _block4;
+        private readonly Block<Utils.DummyAction> _block5;
+        private readonly Transaction<Utils.DummyAction> _transaction1;
+        private readonly Transaction<Utils.DummyAction> _transaction2;
+        private readonly Transaction<Utils.DummyAction> _transaction3;
+        private readonly Transaction<Utils.DummyAction> _transaction4;
```
limebell, line 29: How about making an array?

kfangw: I am not sure about the benefit of doing that.
Is there any weak point declaring 4 individual variables in this test case? @limebell

limebell: Just because it's clean? Not a strong suggestion though. Please check

kfangw: IMHO, those variables don't share any semantics, such as living in the same block or signed by the same signer, in the following testcases. So I think that there is no need to aggregate them. And the beauty or cleanliness of code is up to the individuals unless it does not break the coding rule(lint)

Hence If you don't mind, may I keep them as a separate variable?

longfin (approved):

longfin (dismissed):

limebell (approved):

limebell (approved):

Libplanet.Extensions.Cocona.Tests/Commands/StoreCommandTest.cs
limebell: 🆗

PR #1341, opened 2021-06-10
Miner is no longer nullable

greymistcube:
Unless we want to support blockchain architecture where blocks without miners are allowed as a feature, there doesn't seem to be a real necessity for having `null` miner `Address` for `Block<T>`. On a side note, having a dummy `Address` or a special `Address` reserved for unforeseen future purposes might come in handy for the convenience of development. Just spitballing, but something like `/dev/null` `Address` "eating up" `Transaction<T>`s and/or `IAction`s might be useful. Perhaps we should reserve `0` and/or `1`?


CHANGES.md
```diff
@@ -282,7 +284,7 @@ To be released.
 [#1301]: https://github.com/planetarium/libplanet/issues/1301
 [#1305]: https://github.com/planetarium/libplanet/pull/1305
 [#1325]: https://github.com/planetarium/libplanet/pull/1325
-
+[#1341]: https://github.com/planetarium/libplanet/pull/1341
```
limebell, line 287:
```suggestion
[#1341]: https://github.com/planetarium/libplanet/pull/1341

```
Just for consistency

limebell (dismissed):

limebell:
What happens if deserializing pre-existing blocks with null miner?

greymistcube:
Hmm... `Block<T>.Miner` holds `Address` and `BlockHeader.Miner` holds `ImmutableArray<byte>`.
If serialized data is missing the miner key/value pair, deserialization would fail on:

https://github.com/greymistcube/libplanet/blob/9e0e63d533f7de46b9ec10d8012e33d57249faaa/Libplanet/Blocks/BlockHeader.cs#L119

If a key/value pair is present, but the value is empty, then again, deserialization would fail on: 

https://github.com/greymistcube/libplanet/blob/9e0e63d533f7de46b9ec10d8012e33d57249faaa/Libplanet/Blocks/Block.cs#L168

longfin (dismissed):

Libplanet/Blocks/BlockHeader.cs
```diff
@@ -1,3 +1,4 @@
+#nullable enable
```
kfangw, line 1: is it necessary?

kfangw (approved):
LGTM

Libplanet/Blocks/BlockHeader.cs
greymistcube: https://github.com/planetarium/libplanet/issues/458

Kinda. We are trying to migrate (albeit slowly) the existing codebase to have `null` types checked. 😅

greymistcube: If you mean that it isn't strictly necessary since the `BlockHeader` class is strictly `struct` with all its properties also being a value type, then no, but I'd say it's still better to have it for management.

limebell (approved):

CHANGES.md
```diff
@@ -188,6 +188,8 @@ To be released.
  -  Added `IStore.DeleteTxIdBlockHashIndex(TxId, BlockHash)` method.
     [[#1294], [#1328]]
  -  Added `IStore.IterateTxIdBlockHashIndex(TxId)` method.  [[#1294], [#1328]]
+ -  Parameter `miner` for `Block<T>()` and `Block<T>.Mine()` can no longer be
+    `null`.  [[1341]]
```
limebell, line 192:
```suggestion
    `null`.  [[#1341]]
```

PR #1343, opened 2021-06-14
Patch: Throw an unmanaged exception for critical failure during evaluation

### LINKED ISSUE #1320: Throw non-deterministic exceptions from `IAction.Execute()`
In .NET, some exceptions(like [`OutOfMemoryException`](https://docs.microsoft.com/en-us/dotnet/api/system.outofmemoryexception?view=net-5.0)) are raised non-deterministically depending on runtime states. it may occur non-deterministic action results and disturb state root hash validation.

So, I think it is safer to throw non-deterministic errors such as OOM in `IAction.Execute()` right away so that the entire mining or verification operation fails.

greymistcube:
Closes #1320.

Pretty lousy implementation at the moment. 😶 
There are two main concerns.

I thought about repackaging a caught exception to something like `CriticallyTerminatedActionException`, but for a critical exception such as `OutOfMemoryException`, decided it would be better not to try and instantiate any new exception.

Testing is flimsy. It would be much better to see if thrown exception gets caught on a higher level, such as `BlockChain<T>.Mine()` or `BlockChain<T>.Append()`, but this is made rather hard to test as actions are run with `rehearsal` set to `true` for creation of transactions and run with `rehearsal` set to `false` for creation of blocks through these methods. The best I could do was to observe a thrown exception from `ActionEvaluator<T>.Evaluate()`.

### DIFF
--- a/Libplanet/Action/ActionEvaluator.cs
+++ b/Libplanet/Action/ActionEvaluator.cs
@@ -265,9 +265,20 @@ namespace Libplanet.Action
                 }
                 catch (Exception e)
                 {
+                    if (e is OutOfMemoryException)
+                    {
+                        var message =
+                            $"The action {action} (block #{blockIndex}, pre-evaluation hash " +
+                            $"{preEvaluationHash}, tx {txid}) threw an exception " +
+                            "during execution:\n" +
+                            $"{e}";
+                        _logger.Error(e, message);
+                        throw;
+                    }
+
                     if (rehearsal)
                     {
-                        var msg =
+                        var message =
                             $"The action {action} threw an exception during its " +
                             "rehearsal.  It is probably because the logic of the " +
                             $"action {action} is not enough generic so that it " +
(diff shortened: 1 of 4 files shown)


limebell (approved):

dahlia (approved):

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -265,9 +265,20 @@ ActionContext CreateActionContext(IAccountStateDelta prevStates, int randomSeed)
...
+                    if (e is OutOfMemoryException)
+                    {
+                        var message =
+                            $"The action {action} (block #{blockIndex}, pre-evaluation hash " +
+                            $"{preEvaluationHash}, tx {txid}) threw an exception " +
+                            "during execution:\n" +
+                            $"{e}";
+                        _logger.Error(e, message);
+                        throw;
+                    }
```
longfin, line 277: I don't know if it will pass the lint(😅), but how about...

```suggestion
                catch (OutOfMemoryException)
                {
                    // Because OutOfMemory is thrown non-deterministically depending on the state of the node, 
                    // we should throw without further handling. 
                    var message = 
                        $"The action {action} (block #{blockIndex}, pre-evaluation hash " +
                        $"{preEvaluationHash}, tx {txid}) threw an exception " +
                        "during execution:\n" +
                        $"{e}";
                    _logger.Error(e, message);
                    throw;
                }
                catch (Exception e)
                {
```

limebell (approved):

longfin (approved):

Issue #1349, opened 2021-06-23
Circumstances where the chain flips regardless of `TotalDifficultyCompare`.


riemannulus:
A slightly strange log was detected.

```
[15:52:38 DBG] 815992201/1598638844: Block #1745949 e394a727e9a78f42f4c3e81c1e68e5ed2de38c42befd22bb6facda8849d5d6dd was appended.
[15:52:38 DBG] Downloaded 1 block(s) from 0x4d92cd06Fc559520d038BDA3bBc63a9e51b76e72.Unspecified/3.22.224.238:50586.42.189.229.92.
[15:52:38 DBG] 815992201/1598638844: Got 1 blocks from Peer 4d92cd06Fc559520d038BDA3bBc63a9e51b76e72 (tip: #1745949 e394a727e9a78f42f4c3e81c1e68e5ed2de38c42befd22bb6facda8849d5d6dd)
[15:52:38 DBG] 815992201: FillBlocksAsync() completed (chain ID: 85b185bc-70e1-461e-aeb7-66a554b760a0, tip: #1745949 e394a727e9a78f42f4c3e81c1e68e5ed2de38c42befd22bb6facda8849d5d6dd).
[15:52:38 DBG] 815992201: Finished FillBlocksAsync().
[15:52:38 DBG] 815992201: Swap the chain 8780651b-241a-47f5-8af0-4015cfc10f1c for the chain 85b185bc-70e1-461e-aeb7-66a554b760a0...
[15:52:38 DBG] The blockchain was reorged from 8780651b-241a-47f5-8af0-4015cfc10f1c (#1745950 314ed0ee12c64df8aaa77946e61caf360ee9c56fdcf5d362fc5b3b9db39b7829) to 85b185bc-70e1-461e-aeb7-66a554b760a0 (#1745949 e394a727e9a78f42f4c3e81c1e68e5ed2de38c42befd22bb6facda8849d5d6dd).
```

Note the chain is swapped to the wrong chain.
![image](https://user-images.githubusercontent.com/16367497/123054204-98695480-d43f-11eb-9f97-e605ef766583.png)

And I found some strange code on `Swarm.BlockSync.SyncPreviousBlockSync()`.
https://github.com/planetarium/libplanet/blob/2eaab5a8b1/Libplanet/Net/Swarm.BlockSync.cs#L175

May this code cause wrong logic if the chain tip is null.

PS. I attach some code snippets and chain data for the debug.
[...]

PR #1350, opened 2021-06-23
Fail-fast when block sync

longfin:
This PR removes null check before chain swapping when block synchronization. it's to fail quickly instead of malfunctioning when a problem situation such as #1349

### DIFF
--- a/Libplanet/Net/Swarm.BlockSync.cs
+++ b/Libplanet/Net/Swarm.BlockSync.cs
@@ -172,14 +172,9 @@ namespace Libplanet.Net
                 var canonComparer = BlockChain.Policy.CanonicalChainComparer;
                 if (synced is { } syncedB
                     && !syncedB.Id.Equals(blockChain?.Id)
-                    && (!(blockChain.Tip is { } tip && syncedB.Tip is { } syncedTip)
-                        || canonComparer.Compare(
-                            blockChain.PerceiveBlock(tip),
-                            blockChain.PerceiveBlock(
-                                syncedTip,
-                                syncedB.PerceiveBlock(syncedTip).PerceivedTime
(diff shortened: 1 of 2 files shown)


Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -172,14 +172,9 @@ CancellationToken cancellationToken
...
-                        || canonComparer.Compare(
-                            blockChain.PerceiveBlock(tip),
-                            blockChain.PerceiveBlock(
-                                syncedTip,
-                                syncedB.PerceiveBlock(syncedTip).PerceivedTime
-                            )
-                        ) < 0
+                    && (canonComparer.Compare(
+                        blockChain.PerceiveBlock(blockChain.Tip),
+                        blockChain.PerceiveBlock(syncedB.Tip)) < 0
```
limebell, line 177: Is it okay without `PerceivedTime`?

longfin: I think it's okay because [`BlockChain<T>.PerceiveBlock()` uses `perceivedTime` only when the given perception is not found on its own storage.](https://github.com/planetarium/libplanet/blob/2eaab5a8b12a6ead49f8f822abe00396d18a55ba/Libplanet/Blockchain/BlockChain.cs#L732-L738) cc @dahlia

limebell: 🆗

limebell (approved):

limebell (dismissed):

dahlia (dismissed):

limebell (approved):

dahlia (approved):

PR #1357, opened 2021-06-28
Replace ColumnFamily to prefix

longfin:


kfangw:
go away stale bot :(

kfangw:
handled by #1392

PR #1360, opened 2021-06-28
Fix replying delay

longfin:
This PR remove `while` loop in `ProcessMessage()` to release `NetMQPoller` for `_router`. it's [common technique for poller performance](https://netmq.readthedocs.io/en/latest/poller/#performance), but in our case, it also can block other callbacks (i.e. `DoReply()`) that use poller.

### DIFF
--- a/Libplanet/Net/Transports/NetMQTransport.cs
+++ b/Libplanet/Net/Transports/NetMQTransport.cs
@@ -533,73 +533,71 @@ namespace Libplanet.Net.Transports
 
         private void ReceiveMessage(object sender, NetMQSocketEventArgs e)
         {
-            NetMQMessage raw = new NetMQMessage();
-            while (e.Socket.TryReceiveMultipartMessage(ref raw))
+            try
             {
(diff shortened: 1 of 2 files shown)


Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -664,6 +662,8 @@ private void DoReply(object sender, NetMQQueueEventArgs<NetMQMessage> e)
             NetMQMessage msg = e.Queue.Dequeue();
             string identityHex = ByteUtil.Hex(msg[0].Buffer);
 
+            _logger.Debug("Dequeued message. ({identity})", identityHex);
```
limebell, line 665: How about lower this log's level to verbose?

longfin:
```suggestion
            _logger.Verbose("Dequeued message. ({identity})", identityHex);
```

limebell (dismissed):

limebell (approved):

dahlia (approved):

Issue #1266, opened 2021-04-26
Drop .NET Standard 2.0 support


dahlia:
We've supported both .NET Standard 2.0 & 2.1, since [Unity](https://unity.com/) hadn't implemented .NET Standard 2.1, but only 2.0.  [However, Unity recently announced their fork of Mono now became to have .NET Standard 2.1 compatibility with full C# 8.][1]  If Unity became compliant with .NET Standard 2.1, we can drop .NET Standard 2.0 support with pleasure, and leave only .NET Standard 2.1 support.

[1]: https://forum.unity.com/threads/unity-future-net-development-status.1092205/#post-7032367

dahlia:
If we can drop .NET Standard 2.0 support, we don't need to stick with C# 7.1 for unit tests anymore.

TarogStar:
It seems that this cannot currently be implemented until the newer version of Unity is used, so this will need to wait.

dahlia:
[Unity supports .NET Standard 2.1 since 2021.2.0b6](https://unity3d.com/unity/beta/2021.2.0b6):

> Scripting: Allow user code to build against .NET Standard 2.1 and .NET Framework 4.8

dahlia:
@planetarium/nine-chronicles-engineers Do you have any plan to upgrade Unity version of Nine Chronicles?

dahlia:
New Unity LTS which [supports .NET Standard 2.1](https://docs.unity.cn/Manual/WhatsNew2021LTS.html) is finally available now: [Unity 2021 LTS](https://docs.unity.cn/Manual/WhatsNew2021LTS.html).

PR #1393, opened 2021-07-22
Remove .netstandard 2.0 support

TarogStar:
Fix for issue #1266 
Removed references to .net standard 2.0 and updated all language versions to at least 8.0.  Removed support for .NET Framework as that is not supported any longer.  Left references to NET as that may be used by Mono.  If that is not needed additional pragmas could be removed.


TarogStar:
It looks like the pipelines are requiring .NET Framework support, which is specifically NOT supported by .net standard 2.1, so the pipelines may need to be edited to support this particular change.

dahlia (commented):
Thanks for your first contribution, but I believe we cannot merge this patch at this moment right now, because [Nine Chronicles](https://github.com/planetarium/NineChronicles), the game created by our studio, still depends on Unity 2020.3.4f1, whereas [Unity has supported .NET Standard 2.1 since 2021.2.0a18.][1]

/cc @planetarium/9c-dev 

[1]: https://forum.unity.com/threads/unity-future-net-development-status.1092205/#post-7032367

PR #1415, opened 2021-07-29
Refactor/broadcast before rendering

### LINKED ISSUE #1385: Broadcast synced block before rendering
Currently, `Swarm<T>` broadcast new blocks that synced from remote after its rendering. 

Since action rendering is not related to block validation, it seems that it can speed up the block propagation of the entire network by speeding it up.

greymistcube:
Closes #1385.

There is no public facing API change. Other than general refactoring, the only notable change is `BlockChain<T>.Swap()` now returning packaged rendering method as a `System.Action`. As a result, `BlockChain<T>.Swap()` does not perform any rendering and returned `System.Action` must be called explicitly, regardless of whether passed `render` parameter is `true` or `false`. The `render` parameter only determines the behavior of the returned delegate.

Other internal behavioral changes include:
 - All rendering postponed until `Swap()` is complete: Necessary for implementing #1385. Before, unrendering/rendering calls were interweaved with tip rewind and fast-forward.
 - Rendering call loop hierarchy change: This is done for consistency and intuitiveness. Before, a renderer selection was done in the outermost loop, i.e. select a renderer then make a render call for each evaluation in a block, whereas now, a renderer selection is done in the innermost loop, i.e. for an evaluation in a block make a render call for each renderer.
 - `BroadcastBlock()` call is done before rendering inside `Swarm<T>.ProcessFillBlocks()`: As a result, `BroadcastBlock()` also comes before `BlockReceived.Set()` and `BlockAppend.Set()`. Logically, `BroadcastBlock()` should actually come after `BlockReceived.Set()` and `BlockAppendSet()`, but doing so by having rendering after `BlockReceived.Set()` and `BlockAppendSet()` seem to cause too many problems due to asynchronicity. 😶


Libplanet/Blockchain/BlockChain.Swap.cs
```diff
@@ -0,0 +1,462 @@
...
+namespace Libplanet.Blockchain
+{
+    public partial class BlockChain<T>
+    {
+        // FIXME it's very dangerous because replacing Id means
+        // ALL blocks (referenced by MineBlock(), etc.) will be changed.
+        // we need to add a synchronization mechanism to handle this correctly.
+        internal System.Action Swap(
+            BlockChain<T> other,
+            bool render,
```
longfin, line 20: IMO, If the rendering is done directly by calling the returned function, this flag seems to be unnecessary now...

Libplanet/Blockchain/BlockChain.Swap.cs
```diff
@@ -0,0 +1,462 @@
...
+            finally
+            {
+                _rwlock.ExitUpgradeableReadLock();
+            }
+
+            return renderSwap;
+        }
+
+        internal void RenderSwap(
+            bool render,
```
longfin, line 135: How about removing this parameter too?

greymistcube: There are unnecessary parameters still left in the code here and there just for the sake of consistency, but `render` is not one of them. If it is to be removed, it should be from `Swap()` not `RenderSwap()`. More logical implementation would be to return `Action<bool>` instead of `Action`, but either way, I'd say `render` is necessary in `RenderSwap()`.

Libplanet/Blockchain/BlockChain.Swap.cs
greymistcube: I wasn't sure what the inline comment was referring to, i.e. for which case a synchronization mechanism is needed.

longfin: Rendering may be okay from the synchronization issue. (When I first wrote that, I was worried because of the consistency of data between chains.)

Libplanet/Blockchain/BlockChain.Swap.cs
greymistcube: Although I agree the design isn't optimal at this point, if you don't mind too much, I think I'd like to deal with this issue separately. See #1417. Another minor reason `render` is still left there is for consistency, without too much thought into how to streamline the rendering processes.

Libplanet/Blockchain/BlockChain.Swap.cs
longfin: I meant, `render` flag doesn't seem to be needed anymore. (not a comment 😅)
At least as far as I understand, code that calls Swap() can now decide whether to render directly without going through the render flag. That is, in the past

```csharp
bool render = IsRenderNeeded();
chain.Swap(render: render);
```

but we can use after this PR like...

```csharp
Action renderSwap = chain.Swap();
if (IsRenderNeeded()) {
    renderSwap();
}
```

Please let me know if I'm misunderstanding anything.

greymistcube: Ah... sorry about that. I misunderstood the original comment then. However, I agree with the rationale, but as in the comment below, I'd like to deal with it in another PR if possible.

longfin: 🆗  fair enough.

longfin (dismissed):

longfin (approved):

Libplanet/Blockchain/BlockChain.Swap.cs
```diff
@@ -0,0 +1,461 @@
...
+                    rewindPath: rewindPath,
+                    fastForwardPath: fastForwardPath,
+                    stateCompleters: completers);
+            }
+            finally
+            {
+                _rwlock.ExitUpgradeableReadLock();
+            }
+
+            return renderSwap;
```
dahlia, line 130: The `Swap()` method is seemingly used in a form like `Swarm(...)()`, which means `renderSwap`, the return value, is always immediately invoked once and only once.  Then, how about invoke it here before returning the function, and make `Swap()`'s return type `void`?  The current way has more opportunities to make a mistake because:

- The returned action could be never invoked by mistake.
- The returned action could be invoked twice or more. (E.g., `var swap = Swap(...); swap(); swap();`)

```suggestion
            renderSwap();
```

dahlia (changes requested):

Libplanet/Blockchain/BlockChain.Swap.cs
longfin: IMO, separating `renderSwap()` call from `.Swap()` is the key of this PR because we want to broadcast block before 
 `renderSwap()`.

dahlia: I misread the code.  I overlooked the usage in the `Swarm<T>.ProcessFillBlocks()` method.  Fair enough.

dahlia (approved):

Libplanet/Blockchain/BlockChain.Swap.cs
dahlia: It would be better if we test whether the returned action does the rendering, e.g., check if nothing is rendered when the returned action is not invoked.

Issue #674, opened 2019-11-13
'Connect' feature on TurnClient


longfin:
`TurnClient` needs other peer's public IP addresses to grant connect permission by `CreatePermission`. And to accomplish this, `Swarm<T>` must resolve it's public IP address on `StartAsync()`. it's okay if peer's host IP equals resolved public IP. but if it didn't, `TurnClient` can't receive connection via the TURN server because they can't be granted.

If we can use '[Connect](https://tools.ietf.org/html/rfc6062#section-4.3)' feature on TURN spec, we also can omit public IP address resolving and replace it with TURN server IP.

PR #1423, opened 2021-08-03
Remove CreatePermission in NetMQTransport

longfin:
This PR strips `TurnClient.CreatePermission()` calls in `NetMQTransport` due to below reasons.

-  TurnClient can't operate with TURN server that requires the permission of peers even with CreatePermission calls in certain environments because it requires public IP of nodes can be determined by TURN relay server.
    - TURN servers in 9c mainnet are running with [`--server-relay` option](https://github.com/coturn/coturn/blob/553ce0216823904d4e52a9629c5bcbed4713f3d2/examples/etc/turnserver.conf#L746) to bypass this problem.
    - In order to completely solve this, Connect must also go through TURN to be used.(#674)
- `CreatePermission` calls also occur when a specific node connects to another node, which has the side effect of losing the entire peer table if the connection to the TURN server is lost.
- Because `CreatePermission` is not being cached thoughtfully, this leads to excessive connections to the TURN server and leads to poor performance.
---

- Lib9c: https://github.com/planetarium/lib9c/pull/606
- NineChronicles.Headless: TBD
- NineChronicles: TBD

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -122,7 +122,9 @@ To be released.
  -  Similarly, `BlockChain<T>.MineBlock()` now internally uses
     `BlockChain<T>.Policy.GetMaxBlockBytes()` literally.  [[#1449], [#1463]]
  -  `NetMQTransport` became to no more send CreatePermission to TURN client and
-    requires permission-less TURN server.  [[#1423]]
+    require permission-less TURN server.  See [coturn's relevant configuration](
+    https://github.com/coturn/coturn/blob/dc8f405f8543a83ad8c059ba6b9f930e1e5a1349/man/man1/turnserver.1#L402-L410)
+    as well.  [[#1423]]
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


dahlia:
Please change the target branch to 0.14-maintenance.

longfin:
I rebased onto `main`. (Given the release schedule, it's probably fine. IMO)

limebell:
Can you please rebase?

longfin:
PTAL @planetarium/libplanet

dahlia (approved):

CHANGES.md
```diff
@@ -121,13 +121,16 @@ To be released.
         value between `1` and `BlockChain<T>.Policy.MaxTransactionsPerBlock`.
  -  Similarly, `BlockChain<T>.MineBlock()` now internally uses
     `BlockChain<T>.Policy.GetMaxBlockBytes()` literally.  [[#1449], [#1463]]
+ -  `NetMQTransport` became to no more send CreatePermission to TURN client and
+    requires permission-less TURN server.  [[#1423]]
```
dahlia, line 125:
```suggestion
    require permission-less TURN server.  See [coturn's relevant configuration](
    https://github.com/coturn/coturn/blob/dc8f405f8543a83ad8c059ba6b9f930e1e5a1349/man/man1/turnserver.1#L402-L410)
    as well.  [[#1423]]
```

dahlia (dismissed):

PR #1425, opened 2021-08-04
Sync recent blocks from multiple peers simultaneously

limebell:
This PR allows `Swarm<T>` to sync blocks from multiple peers by re-using preloading logic (`BlockCompletion`). And fixed error that `Swarm<T>.PreloadAsync()` fails to preload blocks from the peer that has chain with higher total block difficulty but lower tip index.

Removed `Swarm<T>.BlockDemand`, I removed tests related to it too which should be restored in further PR. New unit tests testing the newly added logic also need to be added too.


Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -16,436 +20,513 @@ public partial class Swarm<T>
...
-        public BlockDemand? BlockDemand { get; private set; }
+        public ConcurrentDictionary<BoundPeer, BlockDemand> BlockDemands { get; private set; }
 
         internal AsyncAutoResetEvent FillBlocksAsyncStarted { get; } = new AsyncAutoResetEvent();
 
         internal AsyncAutoResetEvent FillBlocksAsyncFailed { get; } = new AsyncAutoResetEvent();
 
         internal AsyncAutoResetEvent ProcessFillBlocksFinished { get; } = new AsyncAutoResetEvent();
 
+        #pragma warning disable S125
```
greymistcube, line 31: Seems unnecessary?

Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -16,436 +20,513 @@ public partial class Swarm<T>
...
-                    logSessionId,
-                    workspace?.Id,
-                    workspace?.Tip?.Index,
-                    workspace?.Tip?.Hash
+                throw new ArgumentException(
+                    "The workspace chain must not have any renderers.",
+                    nameof(workspace)
                 );
-                foreach (var id in scope.Where(guid => guid != workspace?.Id))
+            }
```
greymistcube, line 488: Can you clarify as to why `ExecuteActions()` care whether `workspace` has any renderer or not?

limebell: There were condition in preloading logic, I just brought it from the previous code :)

CHANGES.md
```diff
@@ -8,6 +8,7 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  Removed `Swarm<T>.BlockDemand`.  [[#1419], [#1425]]
```
dahlia, line 11:
```suggestion
 -  Removed `Swarm<T>.BlockDemand` property.  [[#1419], [#1425]]
```

CHANGES.md
```diff
@@ -16,12 +17,26 @@ To be released.
 
 ### Added APIs
 
+ -  Added `Swarm<T>.BlockDemands`.  [[#1419], [#1425]]
```
dahlia, line 20:
```suggestion
 -  Added `Swarm<T>.BlockDemands` property.  [[#1419], [#1425]]
```

Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -16,436 +20,556 @@ public partial class Swarm<T>
         /// It is null when the <see cref="Swarm{T}"/> does not have any block to demand.
         /// <seealso cref="BlockDemand"/>
         /// </summary>
-        public BlockDemand? BlockDemand { get; private set; }
+        public ConcurrentDictionary<BoundPeer, BlockDemand> BlockDemands { get; private set; }
```
dahlia, line 23: From outside, wouldn't it better be `IDictionary<BoundBeer, BlockDemand>` so that its concrete implementation is not exposed?

```suggestion
        public IDictionary<BoundPeer, BlockDemand> BlockDemands { get; private set; }
```

greymistcube: I generally agree with the statement, but ultimately, I think it would be preferable to have its own internal logic as a class, be it `BlockDemands` and/or `BlockDemand`. 🙃

Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -16,436 +20,556 @@ public partial class Swarm<T>
...
+        /// </summary>
+        /// <param name="timeout">
+        /// The timeout value for the request to get the tip of the block.
+        /// </param>
+        /// <param name="maximumPollPeers">The maximum targets to send request to.</param>
+        /// <param name="cancellationToken">
+        /// A cancellation token used to propagate notification that this
+        /// operation should be canceled.</param>
+        /// <returns>An awaitable task without value.</returns>
+        public async Task ProcessFillBlocksAsync(
```
greymistcube, line 43: Why has this changed to `public`? From the looks of it, I think it makes more sense to have `ProcessFillBlocksAsync()` as private and `FillBlocksAsync()` as `public` instead.

limebell: `FillBlocksAsync()` is a task polls blocks by calling `ProcessFillBlocksAsync()` periodically. Do you think we should make polling task to public instead of actual function?

greymistcube: I don't think it's a cut and dried issue. Multiple asynchronous calls to `ProcessFillBlocksAsync()` and/or `FillBlocksAsync()` don't seem to make sense anyway. So if it is all the same, if we were to expose a syncing method, I think more automated higher level function is the way to go. 🙃

Libplanet/Net/Swarm.BlockSync.cs
limebell: I've added `BlockDemandTable` class. PTAL :)

Libplanet/Net/Swarm.BlockSync.cs
limebell: I lowered its access level to `internal`.

longfin (approved):

greymistcube (approved):

PR #1429, opened 2021-08-09
Add cache for action evaluations

longfin:
This PR adds a cache for action evaluations to prevent action re-executing frequently.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,9 +6,6 @@ Version 0.14.1
 
 To be released.
 
-### Behavioral changes
-
-
  -  Improved performance of `ActionEvaluator.Evaluate()` due to cache.
     [[#1429]]
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,14 @@ Version 0.14.1
 
 To be released.
 
+### Behavioral changes
+
+
+ -  Improved performance of `ActionEvaluator.Evaluate()` due to cache.
+    [[#1429]]
+
+[#1429]: https://github.com/planetarium/libplanet/pull/1429
+
 
 Version 0.14.0
 --------------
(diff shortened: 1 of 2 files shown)


kfangw:
It looks good to me. However, could you consider the nature feature of MPT in the following enhancement?
If the statestore is loadable from a specific block height, and the dangling states are kept when reorg happened,  how about use that feature?

longfin:
> It looks good to me. However, could you consider the nature feature of MPT in the following enhancement?
> If the statestore is loadable from a specific block height, and the dangling states are kept when reorg happened, how about use that feature?
> 
> I am not sure whether the dangling states are kept or not.

If we need only the output state (`IAccountStateDelta`), we can get the output state right now without recalculation without a separate cache. however, other parts of Libplanet require `ActionEvaluation` that contains various execution information in addition to the output state, which is not stored by IStateStore, so it was necessary to save it and retrieve it.

CHANGES.md
```diff
@@ -6,6 +6,14 @@ Version 0.14.1
 
 To be released.
 
+### Behavioral changes
+
+
```
dahlia, line 11: Patch changelogs don't have to be sectioned.

```suggestion
```

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -389,10 +393,23 @@ ActionContext CreateActionContext(IAccountStateDelta prevStates, int randomSeed)
...
 
-            return EvaluateTxs(
+            if (_evCache.TryGetValue(block.Hash, out IEnumerable<ActionEvaluation> cached))
+            {
+                _logger.Verbose(
+                    "_evCache hit. block.Hash: {blockHash}, # of cached: {count}",
+                    block.Hash,
+                    cached.Count()
+                );
+                return cached;
```
dahlia, line 403: According to coverage report, this code path is not tested.  It would be better if we test whether this method returns the same result for the same block hash.

dahlia (dismissed):

greymistcube (approved):

longfin:
As a result of applying it to Nine Chronicles first, caching all ActionEvaluations used excessive memory. This is basically a matter of the application, but it means that maybe you need to tweak the implementation idea, or at least make the cache size a configurable. 

So I will close the PR and reopen a separate PR once the issue is resolved.

PR #1430, opened 2021-08-10
`BlockChain<T>.Tip` made non-nullable

### LINKED ISSUE #1409: `BlockChain<T>.Tip` should not be null
Under the current implementation of `BlockChain<T>` constructors, there should be no case where its `Tip` is `null`. Catching an exception from `IStore` and turning it into `null` then further coalescing the `null` value for a (wrongful) usage seems quite dangerous.

As can be read from its class description, `BlockChain<T>` always has at least its genesis block and is never empty. Not being able to retrieve its `Tip` for whatever reason (most likely due to some kind of desync from its storage) should result in fast failure.

I suggest removing `null` value being possible for `BlockChain<T>.Tip`.

greymistcube:
Closes #1409.

----

- <https://github.com/planetarium/libplanet/pull/1430>
- <https://github.com/planetarium/lib9c/pull/575>
- <https://github.com/planetarium/NineChronicles.Headless/pull/644>

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -246,22 +246,8 @@ namespace Libplanet.Blockchain
 
         /// <summary>
         /// The topmost <see cref="Block{T}"/> of the current blockchain.
-        /// Can be <c>null</c> if the blockchain is empty.
         /// </summary>
-        public Block<T> Tip
-        {
-            get
-            {
-                try
-                {
-                    return this[-1];
-                }
-                catch (ArgumentOutOfRangeException)
-                {
-                    return null;
-                }
-            }
-        }
+        public Block<T> Tip => this[-1];
 
         /// <summary>
         /// The first <see cref="Block{T}"/> in the <see cref="BlockChain{T}"/>.
@@ -1266,7 +1252,7 @@ namespace Libplanet.Blockchain
             }
 
             _rwlock.EnterUpgradeableReadLock();
-            Block<T> prevTip = Tip;
+            Block<T> prevTip = Count > 0 ? Tip : null;
             try
             {
                 InvalidBlockException e = ValidateNextBlock(block);
@@ -1734,6 +1720,9 @@ namespace Libplanet.Blockchain
         {
             int actualProtocolVersion = nextBlock.ProtocolVersion;
             const int currentProtocolVersion = Block<T>.CurrentProtocolVersion;
+
+            // FIXME: Crude way of checking protocol version for non-genesis block.
+            // Ideally, whether this is called during instantiation should be made more explicit.
             if (actualProtocolVersion > currentProtocolVersion)
             {
                 string message =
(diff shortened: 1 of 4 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1745,11 +1731,11 @@ private InvalidBlockException ValidateNextBlock(Block<T> nextBlock)
                     message
                 );
             }
-            else if (Tip is { } tip && actualProtocolVersion < tip.ProtocolVersion)
+            else if (Count > 0 && actualProtocolVersion < Tip.ProtocolVersion)
```
dahlia, line 1734: Isn't the added condition (`Count > 0`) always `true`?

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1832,7 +1818,7 @@ private InvalidBlockException ValidateNextBlock(Block<T> nextBlock)
             _rwlock.EnterUpgradeableReadLock();
             try
             {
-                if (offset is null && Tip is null)
+                if (offset is null && Count == 0)
```
dahlia, line 1821: How about removing this `if` block at all?

dahlia (approved):
:+1:

Libplanet/Blockchain/BlockChain.cs
greymistcube: Actually no. When creating `BlockChain<T>`, given genesis block is "appended" instead of being set via `Append()` where `ValidateNextBlock()` is called, in which case the blockchain can be empty. Removing this results in an error since there is no `Tip` to retrieve.

Libplanet/Blockchain/BlockChain.cs
greymistcube: I agree the logic looks little weird. It seems to be more of a safety measure than a feature. The only possible failure is when trying to access the states before `BlockChain<T>` construction is complete. I'll try removing it and see if it breaks anything. 🙃

Libplanet/Blockchain/BlockChain.cs
dahlia: Fair enough.  I suggest to adding a private property like…

```csharp
private BeingInstantiated => Count < 1;
```

and replacing such predicates with `BeingInstantiated` so that other contributors easily can be aware of their intention.

greymistcube: Eh, still feels a bit icky.

I agree using something like `BeingInstantiated` as a private property makes the code clearer locally. Half way through implementing your suggestion, however, I realized the condition of being instantiated is not the same as `Count` being zero. Implementing with a private field `_beingInstantiated` and actually setting the flag before exiting a constructor can do the job instead, but this has its own set of problems of making the code more prone to mistakes. I can't quite think of a better yet simple solution. I think method overloading or some kind of redesign of `Append()` might be preferable, in which case, the scope of this PR gets too big. I'll try to get back on this if I get the chance.

Libplanet/Blockchain/BlockChain.cs
greymistcube: Done. See c99cde40e6c645466bf15cf78f4d7109388c865e.

Libplanet/Blockchain/BlockChain.cs
dahlia: Okay, then please leave comments on such predicates to provide some context.

Libplanet/Blockchain/BlockChain.cs
greymistcube: Arg... this has been reverted back. Since `genesis` is appended just like any other block via `Append()`, state look-up is performed while `Tip` is not accessible. Due to this, many tests break in Lib9c. 😞

limebell (approved):

longfin (approved):

PR #1445, opened 2021-08-25
Put in place an option for MinTransactionsPerBlock

TarogStar:
Put in place an option for MinTransactionsPerBlock to keep energy and cpu time to a minimum if a block policy (such as implemented by 9c) requires that blocks contain at least 1 or more transactions.  This code will stop the miner from trying to get the hash if there are less than the MinTrasactionsPerBlock as defined in the policy.

Currently the miner will continue mining an invalid block, and 9c has implemented a custom setting to force at least 1 transaction.  This will allow that to be configurable in the future.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.MineBlock.cs
+++ b/Libplanet/Blockchain/BlockChain.MineBlock.cs
@@ -363,6 +363,20 @@ namespace Libplanet.Blockchain
                 }
             }
 
+            if (transactionsToMine.Count < Policy.MinTransactionsPerBlock)
+            {
+                if (stagedTransactions.Count > transactionsToMine.Count)
+                {
+                    throw new OperationCanceledException(
+                        "stagedTransactions higher than transactionsToMine.");
+                }
+
+                throw new InvalidBlockMinTxException(
+                    transactionsToMine.Count,
+                    Policy.MinTransactionsPerBlock,
+                    "Below Minimum Transactions");
+            }
+
             _logger.Information(
                 "Gathered total of {TransactionsToMineCount} transactions to mine for " +
                 "block #{Index} from {StagedTransactionsCount} staged transactions.",
--- a/Libplanet/Blockchain/BlockChain.MineBlock.cs
+++ b/Libplanet/Blockchain/BlockChain.MineBlock.cs
@@ -365,13 +365,7 @@ namespace Libplanet.Blockchain
 
             if (transactionsToMine.Count < Policy.MinTransactionsPerBlock)
             {
-                if (stagedTransactions.Count > transactionsToMine.Count)
-                {
-                    throw new OperationCanceledException(
-                        "stagedTransactions higher than transactionsToMine.");
-                }
-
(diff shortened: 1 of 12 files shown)


longfin (approved):
Thanks for your contribution. I think it would be also great if we have a unittest about `minTransactionsPerBlock` too.

greymistcube:
I apologize in advance for overwriting numerous commits due to rebasing.

TarogStar:
I apologize for creating more work for you.  I was just about to work on fixing this myself, it looks like you cleaned up all the merges in a better way with rebase though, so thank you.

greymistcube:
It's mainly due to us taking way long to decide whether to include this feature or not and there were too many changes since the original request. 🙃

Libplanet/Blocks/InvalidBlockMinTxException.cs
```diff
@@ -0,0 +1,64 @@
...
+using Libplanet.Blockchain.Policies;
+
+namespace Libplanet.Blocks
+{
+    /// <summary>
+    /// The exception that is thrown when a <see cref="Block{T}"/>'s
+    /// <see cref="Block{T}.Transactions"/> Count is too small.
+    /// </summary>
+    [Serializable]
+    public sealed class InvalidBlockMinTxException : InvalidBlockException, ISerializable
```
dahlia, line 13: Would there be more descriptive name than this?  The exception thrown when there are too many transactions is `BlockExceedingTransactionsException`.  I suggest `BlockInsufficientTransactionsException`.

Also, this needs to be added to the changelog.

Libplanet/Blockchain/BlockChain.MineBlock.cs
```diff
@@ -363,6 +363,20 @@ public partial class BlockChain<T>
                 }
             }
 
+            if (transactionsToMine.Count < Policy.MinTransactionsPerBlock)
+            {
+                if (stagedTransactions.Count > transactionsToMine.Count)
+                {
+                    throw new OperationCanceledException(
+                        "stagedTransactions higher than transactionsToMine.");
+                }
```
dahlia, line 372: Would we need to deal with this case separately?  IMHO, we don't need this special casing.

Libplanet/Blocks/InvalidBlockMinTxException.cs
greymistcube: Done. See the latest commit.

greymistcube: Um... on the other hand, now that I think of it, the new descriptive name doesn't exactly match its use case since the exception is thrown while "gathering" transactions, not while appending. 🙃

Libplanet/Blockchain/BlockChain.MineBlock.cs
greymistcube: Done. See the latest commit.

dahlia (approved):

greymistcube (approved):

PR #1459, opened 2021-08-31
Restore block push logic

limebell:
Block sync using `BlockDemand` became not to use `BlockCompletion`.

### DIFF
--- a/Libplanet/Net/Swarm.BlockSync.cs
+++ b/Libplanet/Net/Swarm.BlockSync.cs
@@ -44,6 +44,11 @@ namespace Libplanet.Net
             int maximumPollPeers,
             CancellationToken cancellationToken)
         {
+            if (maximumPollPeers <= 0)
+            {
+                return;
+            }
+
             List<(BoundPeer, IBlockExcerpt)> peersWithBlockExcerpt =
                 await GetPeersWithExcerpts(
                     BlockChain.Tip, timeout, maximumPollPeers, cancellationToken);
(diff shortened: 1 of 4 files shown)


limebell:
Can you please take a look? @longfin

Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -112,15 +117,18 @@ CancellationToken cancellationToken
             {
                 if (BlockDemandTable.Any())
                 {
-                    List<(BoundPeer, IBlockExcerpt)> peersWithExcerpt = BlockDemandTable.Demands
-                        .Select(pair => (pair.Key, (IBlockExcerpt)pair.Value)).ToList();
-                    await ProcessFillBlocksAsync(
-                        peersWithExcerpt,
+                    var largest =
```
longfin, line 120: Can we specify type directly?

Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -112,15 +117,18 @@ CancellationToken cancellationToken
...
+                        BlockDemandTable.Demands.Values.OrderBy(demand => demand.TotalDifficulty)
+                            .Last();
```
longfin, line 122: IMO, we don't need full ordering.

```suggestion
                        BlockDemandTable.Demands.Values
                        .Aggregate(acc, next => next.TotalDifficulty > acc.TotalDifficulty ? next : acc)
```

dahlia (approved):

longfin (approved):

PR #1471, opened 2021-09-03
Docfx version bump

greymistcube:
Still haven't figured out as to why documentation isn't updating properly on local builds, but this seems to work. 🙃


greymistcube:
Aaaaaand, certificate fail. 😢

dahlia (approved):

PR #1474, opened 2021-09-06
`IStore` is now `IDisposable`

### LINKED ISSUE #1448: Make `IStore` inherit `IDisposable`
Design-wise, unless there is an explicit reason not to, it is probably better to make `IStore` itself be `IDisposable` instead of implementing multiple interfaces for store classes.

greymistcube:
Closes #1448.

I don't think it is too far fetched to think `IStore`s (together with `IStateStore`s and `IKeyValueStore`s) in general might have unmanaged resources to be cleaned up from a user's perspective. I think it would be better to forgo type checking specific implementation whether a store is disposable or not and simply disposing an `IStore` as a general rule is better.

P.S. On a side note, I thought about also implementing finalizers, but there seem to be a general consensus to *avoid* using finalizers if possible. Any thoughts?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -115,6 +115,7 @@ To be released.
 [#1440]: https://github.com/planetarium/libplanet/pull/1440
 [#1442]: https://github.com/planetarium/libplanet/pull/1442
 [#1443]: https://github.com/planetarium/libplanet/pull/1443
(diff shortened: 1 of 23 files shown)


greymistcube:
Woops, this isn't review ready yet. I'll add a changelog.

limebell (approved):

PR #1475, opened 2021-09-07
`PrivateKey.FromString()` method

dahlia:
`new PrivateKey(ByteUtil.FromHex(hex))` now can be shortened as `PrivateKey.FromString(hex)`.

### DIFF
--- a/Libplanet.Tests/Crypto/PrivateKeyTest.cs
+++ b/Libplanet.Tests/Crypto/PrivateKeyTest.cs
@@ -4,11 +4,32 @@ using System.Linq;
 using System.Text;
 using Libplanet.Crypto;
 using Xunit;
+using static Libplanet.Tests.TestUtils;
 
 namespace Libplanet.Tests.Crypto
 {
     public class PrivateKeyTest
     {
+        [Fact]
+        public void FromString()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => PrivateKey.FromString(string.Empty));
+            Assert.Throws<ArgumentOutOfRangeException>(() => PrivateKey.FromString("a"));
+            Assert.Throws<FormatException>(() => PrivateKey.FromString("zz"));
+            PrivateKey actual = PrivateKey.FromString(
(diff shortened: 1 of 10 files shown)


Libplanet/Blocks/PreEvaluationBlock.cs
```diff
@@ -0,0 +1,68 @@
+using System;
```
kfangw, line 1: is this file added with intention?

Libplanet.Tests/Crypto/PrivateKeyTest.cs
```diff
@@ -4,11 +4,32 @@
...
+            Assert.Throws<ArgumentOutOfRangeException>(() => PrivateKey.FromString(string.Empty));
+            Assert.Throws<ArgumentOutOfRangeException>(() => PrivateKey.FromString("a"));
+            Assert.Throws<FormatException>(() => PrivateKey.FromString("zz"));
+            PrivateKey actual = PrivateKey.FromString(
+                "e07107ca4b0d19147fa1152a0f2c7884705d59cbb6318e2f901bd28dd9ff78e3"
+            );
+            AssertBytesEqual(
+                new byte[]
+                {
+                    0xe0, 0x71, 0x07, 0xca, 0x4b, 0x0d, 0x19, 0x14, 0x7f, 0xa1, 0x15,
```
kfangw, line 25: just wonder. how could you pass the lint test for these multiple lines?

kfangw (dismissed):
nit comments. 

and LGTM.

Libplanet/Blocks/PreEvaluationBlock.cs
dahlia: It was mistake; I'm going to remove it.

Libplanet.Tests/Crypto/PrivateKeyTest.cs
dahlia: It can't pass the lint without suppressing `SA1118`.  It was disabled from the commit I missed. I'm going to add the commit as well.

kfangw (approved):

PR #1477, opened 2021-09-08
`txPriority` option for `BlockChain<T>.MineBlock()` method

dahlia:
This patch adds a new option `txPriority` to `BlockChain<T>.MineBlock()` method (and its internal parts, `GatherTransactionsToMine()` & `ListStagedTransactions()`).  The parameter takes an [`IComparer<Transaction<T>>`][1] to determine the priority of staged transactions.  Here's a very simple example:

```csharp
// Prioritize a certain signer's transactions over others':
Address prioritySigner = new Address("…");
IComparer<Transaction<T>> txPriority = Comparer<Transaction<T>>.Create(
    (a, b) => a.Signer.Equal(prioritySigner) ? -1 : (b.Signer.Equal(prioritySigner) ? 1 : 0);
);
```

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icomparer-1?view=netstandard-2.0

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1231,9 +1231,25 @@ namespace Libplanet.Blockchain
             }
         }
 
-        internal ImmutableList<Transaction<T>> ListStagedTransactions()
+        /// <summary>
+        /// Lists the all staged transactions, with properly ordered nonces.
+        /// </summary>
+        /// <param name="txPriority">An optional comparer for give certain transactions to
+        /// priority to belong to the block.  No certain priority by default.</param>
+        /// <returns>A list of staged transactions.  This guarantees that for transactions signed
+        /// by the same address, those with greater nonce never comes before those with
+        /// lesser nonce.</returns>
+        internal ImmutableList<Transaction<T>> ListStagedTransactions(
+            IComparer<Transaction<T>> txPriority = null
+        )
         {
-            Transaction<T>[] txs = StagePolicy.Iterate(this).ToArray();
+            IEnumerable<Transaction<T>> unorderedTxs = StagePolicy.Iterate(this);
+            if (txPriority is { } comparer)
+            {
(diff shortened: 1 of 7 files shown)


dahlia:
Rebased on the up-to-date main.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1231,9 +1231,25 @@ internal BlockLocator GetBlockLocator(int threshold = 10)
...
+        /// <summary>
+        /// Lists the all staged transactions, with properly ordered nonces.
+        /// </summary>
+        /// <param name="txPriority">An optional comparer for give certain transactions to
+        /// priority to belong to the block.  No certain priority by default.</param>
+        /// <returns>A list of staged transactions.  This guarantees that for transactions signed
+        /// by the same address, those with greater nonce never comes before those with
+        /// lesser nonce.</returns>
+        internal ImmutableList<Transaction<T>> ListStagedTransactions(
+            IComparer<Transaction<T>> txPriority = null
```
greymistcube, line 1243: Should `txPriority` be propagated all the way down to `ListStagedTransactions`? I think it is more natural to handle this as transaction selection strategy in `GatherTransactionsToMine()`.

dahlia: I just want to prevent `txPriority` accidentally reordering tx nonces per signer, which is `ListStagedTransactions()`'s key guarantee.  Applying `txPriority` before `ListStagedTransactions()` ordering txs was my immediate idea to achieve it, but I'm going to consider it if you suggest another idea.

greymistcube: Fair enough. I still have some qualms about "in place" nonce sorting's fairness in the first place, but that's a separate issue and I can't come up with a better solution at the moment. 🙃

greymistcube (approved):

PR #1478, opened 2021-09-08
Fill transactions from multiple peers

### LINKED ISSUE #1420: Parallelize tx synchronization
like #1419

limebell:
Closes #1420


longfin:
It has a conflict... 😢

Libplanet.Tests/Net/SwarmTest.cs
```diff
@@ -1551,7 +1551,9 @@ public async Task FillWhenGetAllBlocksFromSender()
             }
         }
 
-        [Fact(Timeout = Timeout)]
+        [Fact(
+            Timeout = Timeout,
+            Skip = "Swarm became to sync transaction from multiple peers simultaneously.")]
```
longfin, line 1556: Is there remained case for this test? if not, how about removing this testcase?

Libplanet/Net/Swarm.MessageHandlers.cs
```diff
@@ -212,25 +210,28 @@ private void ProcessTxIds(TxIds message)
                 message.Ids.Select(txid => txid.ToString())
             );
 
-            IStagePolicy<T> stagePolicy = BlockChain.StagePolicy;
-            ImmutableHashSet<TxId> newTxIds = message.Ids
-                .Where(id => !_demandTxIds.ContainsKey(id))
-                .Where(id => !stagePolicy.Ignores(BlockChain, id))
-                .ToImmutableHashSet();
+            HashSet<TxId> required = GetRequiredTxIds(message.Ids);
```
longfin, line 213: And, it seems that we need a new test case for basic features. ( Codecov is complaining... 😅 )

Libplanet/Net/Swarm.cs
```diff
@@ -83,6 +82,8 @@ public partial class Swarm<T> : IDisposable
             _privateKey = privateKey ?? throw new ArgumentNullException(nameof(privateKey));
             LastSeenTimestamps =
                 new ConcurrentDictionary<Peer, DateTimeOffset>();
+            _txSyncTaskQueue = new ConcurrentDictionary<BoundPeer, HashSet<TxId>>();
+            _txIgnoreQueue = new FixedSizedQueue<TxId>(100);
```
longfin, line 86: Is there any reason for separated `_txIgnoreQueue`? IMO `stagePolicy.Ignore()` does similar things.

Libplanet/Net/Swarm.MessageHandlers.cs
```diff
@@ -212,25 +210,28 @@ private void ProcessTxIds(TxIds message)
...
             _logger.Debug(
                 "Unaware transactions to receive: {@TxIds}.",
-                newTxIds.Select(txid => txid.ToString())
+                required.Select(txid => txid.ToString())
             );
-            foreach (TxId txid in newTxIds)
+
+            if (_txSyncTaskQueue.ContainsKey(peer))
+            {
+                _txSyncTaskQueue[peer].UnionWith(required);
```
longfin, line 228: AFIK, this code can be access from multiple task(thread)s and `HashSet<T>` isn't thread-safe. so I think we may need to allow concurrent download for the same tx id from multiple threads and coordinate them.

Libplanet/Net/TxCompletion.cs
```diff
@@ -0,0 +1,191 @@
...
+                    else
+                    {
+                        _logger.Information(
+                            "Failed to get transactions to stage: {@TxIds}",
+                            _requiredTxIds[peer].Select(txId => txId.ToString()));
+                    }
+                }
+                finally
+                {
+                }
```
longfin, line 185: IMO, an empty `finally` isn't needed usually.

Libplanet/Net/TxCompletion.cs
```diff
@@ -0,0 +1,191 @@
...
+using System.Threading.Tasks;
+using Libplanet.Action;
+using Libplanet.Blockchain;
+using Libplanet.Tx;
+using Nito.AsyncEx;
+using Serilog;
+
+namespace Libplanet.Net
+{
+    public class TxCompletion<TPeer, TAction> : IDisposable
```
longfin, line 16: Why do we need `TPeer` type parameter for `TxCompletion`? At look glance, there is only `BoundPeer` for that.

Libplanet/Net/TxCompletion.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        internal AsyncAutoResetEvent TxReceived { get; }
+
+        public async void Dispose()
+        {
+            if (!_disposed)
+            {
+                _cancellationTokenSource.Cancel();
+                var tasks = _txSyncTasks.Values.ToArray();
+                await Task.WhenAll(tasks);
```
longfin, line 62: If we don't have [`IAsyncDisposable`](https://docs.microsoft.com/ko-kr/dotnet/api/system.iasyncdisposable?view=net-5.0) due to .netstandard 2.0 support, it may be better than explicit blocking instead of `async void`.

```suggestion
                Task.WaitAll(tasks);
```

Libplanet/Net/TxCompletion.cs
limebell: Without finally, the analyzer throws error.

Libplanet/Net/TxCompletion.cs
limebell: This refers to `BlockCompletion`... I'm not really don't know why it used like this though.

longfin: 🤔

longfin (approved):

Issue #1476, opened 2021-09-08
Implement pluggable policy.


greymistcube:
As `IBlockPolicy` is getting more and more "fragmented", mostly due to `BlockChain<T>`'s height dependency, a new interface for handling policies may be helpful before things get out of hand.

Todos:
- Provide a default implementation of computing block difficulty function that can target block interval and do away with difficulty divisor, minimum difficulty, etc.
- Turn constant byte limits to functions.
- Maybe genesis block can be included in the policy?

PR #1485, opened 2021-09-10
Refactor/iblockpolicy

greymistcube:
This PR partially addresses #1476.

As more block chain logic gets delegated to `IBlockPolicy<T>`, need for `IBlockPolicy<T>` interface to be polished is growing. This is just a more polished version of #1479 in order to do sensible cleanup on [lib9c](https://github.com/planetarium/lib9c) and [headless](https://github.com/planetarium/NineChronicles.Headless).

Summary:
- ~No logical change in implementation.~ See the addendum below.
- `ValidateTxForNextBlock()` now returns a `TxPolicyViolationException` and `ValidateNextBlock()` now returns `BlockPolicyViolationException`.
  - Seems like either both should throw an exception or, better yet, not return any exception but log inside the implementation.
- `BlockPolicy()` constructor taking `int blockIntervalMilliseconds` has been removed.
  - The overhead of maintaining the code seemed to seriously outweigh the convenience of having an overloaded constructor. 

Note:
- As it is getting harder to divide what should be the domain of the policy or not, it might be better to move the entire validation logic to `IBlockPolicy<T>`.
- As libplanet cannot know in advance what kind of policy violation may happen in the future, having a separate abstract exception class such as `PolicyViolationException` may be useful for outside usage.
- There should be some mechanism to enforce that `BlockChain<T>` instance does not get passed as an argument to a policy before it is fully instantiated.
- To be truly "pluggable", we might also want `IBlockPolicy<T>.BlockAction`, `IBlockPolicy<T>.BlockInterval` to be assigned as a function.

----

Addendum:
- While waiting for a review, there was an additional update.
  - Parameters `int maxBlockBytes` and `int maxGenesisBytes` are removed from `BlockPolicy<T>()` constructor and is replaced by a single `Func<long, int>? getMaxBlockBytes` parameter.
  - Default implementation of `GetMaxBytes()` is now a constant function of `100`KiB.

----

- <https://github.com/planetarium/libplanet/pull/1485>
- <https://github.com/planetarium/lib9c/pull/632>


CHANGES.md
```diff
@@ -71,6 +71,17 @@ To be released.
         HashDigest<SHA256>?)` constructor instead.
  -  `IStore`, `IStateStore`, and `IKeyValueStore` interfaces now inherit
     `IDisposable`.  [[#1448], [#1474]]
+ -  Multiple changes for `IBlockPolicy<T>` interface.  [[#1485]]
+     -  `bool DoesTransactionFollowsPolicy(BlockChain<T>, Transaction<T>)`
+        changed to `TxPolicyViolationException? ValidateNextBlockTx(
```
limebell, line 76: Instead of return exception, how about just throw exception and make return type to `void`? Also for `ValidateNextBlock`.

greymistcube: I'm still on the fence about this decision. You are right in the sense that both methods should either throw designated exceptions or just return bool values. However, there are too many unknown factors at play, such as the precise responsibility of `IBlockPolicy` in particular, and the hierarchy of exceptions used by `IBlockPolicy` is not really finalized. I thought it wasn't worth the hassle and the risk to change them to throw exceptions at this moment. 😶

Libplanet/Tx/TxViolatingBlockPolicyException.cs
```diff
@@ -1,26 +0,0 @@
...
-using System;
-
-namespace Libplanet.Tx
-{
-    /// <summary>
-    /// The exception that is thrown when a <see cref="Transaction{T}"/> does not follow the
-    /// block policy.
-    /// </summary>
-    [Serializable]
-    public class TxViolatingBlockPolicyException : InvalidTxException
```
dahlia, line 11: This removal has to be recorded in the changelog.

greymistcube: Done. See 81532875ddca91a6d122800aba0704ef63fb03d3.

greymistcube:
Sorry, I removed `maxBlockBytes` and `maxGenesisBytes` and replaced with a single getter method.

longfin (approved):

Issue #1345, opened 2021-06-15
`Block<T>.Mine()` - `BlockChain<T>.Append()` doesn't work


longfin:
```csharp
                    // assume _chain is BlockChain<PolymorphicAction<ActionBase>>
                    Transaction<PolymorphicAction<ActionBase>> authProof = StageProofTransaction();
                    Block<PolymorphicAction<ActionBase>> tip = _chain.Tip;
                    Block<PolymorphicAction<ActionBase>> block = Block<PolymorphicAction<ActionBase>>.Mine(
                        tip.Index + 1,
                        _chain.Policy.GetNextBlockDifficulty(_chain),
                        tip.TotalDifficulty,
                        Address,
                        tip.Hash,
                        DateTimeOffset.UtcNow,
                        new[] { authProof },
                        Block<PolymorphicAction<ActionBase>>.CurrentProtocolVersion,
                        cancellationToken
                    );
                    _chain.Append(block);
```

The above code throws `InvalidBlockStateRootHashException` due to `block` doesn't have a state root hash yet and `.Append()` doesn't care about this case.

Currently, users can mine block only through `BlockChain<T>.MineBlock()` API.

greymistcube:
Do we want to allow mining through `Block<T>.Mine()`? Mining is inherently dependent on the state of the chain it is trying to work on. I'd say querying all the necessary information from the chain and manually passing them to `Block<T>.Mine()` needn't be publicly exposed in the first place. But if we really want it to be the case, a *proper* implementation would require a design change. We can of course pass the previous states of the block it is evaluating from, the evaluator, and the policy action to the `Block<T>.Mine()` method, but at that point, I think would be more sensible to just use the hash function to find the nonce and create a block using a constructor. 😶

Issue #1164, opened 2021-01-18
Make illegal states of `Block<T>` & `Transaction<T>` unrepresentable


dahlia:
Related: #1146.

To be elaborated.

greymistcube:
# Preamble

Well, this is really getting out of hand.

Indeed, `Block<T>` is being used in multiple stages without regards to its validity. There are historical reasons why this has come to be the case, but it might be preferable to overhaul the whole class at this point.

After a careful inspection, there seems to be two major issues:
* The content of `Block<T>` is largely divided into two groups.
  * Those not dependent on `PreEvaluationHash` and those dependent on `PreEvaluationHash`.
* Evaluation process is rather convoluted.

To summarize the issues even further, in essence, the design of `Block<T>` seems to have evolved far too much for its surrounding support classes/methods, which are way too generic, to be useful. I suggest we decide asap whether to systematically support "two-stage", or even "multi-stage", `Block<T>`.

There are also policy level complications arising from having `StateRootHash` and `Hash` derived from `PreEvaluationHash`, due to both not being secured by PoW for the newest block. Although this is not entirely a separate issue, these should be discussed more in depth if necessary.

For everyone's sanity, and especially mine, some overview of the creation processes of a `Block<T>` might be in order.

# Creation of `Block<T>`

There are mainly two ways for a `Block<T>` instance to be created: mining and deserialization.

## Mining

This is where a `Block<T>` gets created through mining. In summary, `Block<T>()` constructor is called with appropriate `Nonce` value. However, the actual process is much more complicated.

* For the creation of a new `Block<T>` through mining, `Block<T>.Mine()` is called from `BlockChain<T>.MineBlock()`.
* `Block<T>.Mine()` returns a new `Block<T>` instance only containing minimal amount of information without either `StateRootHash` or `Hash`.
  * `Transaction<T>`'s are ordered by `Id` *before* `Block<T>()` is called.
  * Inside `Block<T>()`, `Transaction<T>` is ordered by `Id` once again in preparation for calculating `PreEvaluationHash` on the fly.
    * There is an error where `SerializeForHash()` is called to *include* `StateRootHash` when calculating `PreEvaluationHash`. Since `StateRootHash` has not yet been assigned, `Block<T>.StateRootHash` of `null` gets converted to `BlockHeader.StateRootHash` of `Empty`, and then to non-existance in its dictionary representation, resulting in accidental proper evaluation of `PreEvaluationHash`.
[...]

Issue #1146, opened 2021-01-06
Do not expose the state of not-yet-filled `Block<T>.StateRootHash` as public API


dahlia:
Read the discussion: https://github.com/planetarium/libplanet/pull/1116#discussion_r535808615

Issue #1128, opened 2020-12-11
Make Libplanet support another `IStateStore` implementations universally


moreal:
This issue actually suggests making planetarium/libplanet support universally another `IStateStore` implementation, not only `TrieStateStore`. Currently, there are many *type testing* logics and fields run on only `TrieStateStore`. That makes the existence of `IStateStore` suspicious: whether to allow using another custom `IStateStore` implementation. Also, it doesn't seem like a good model to make fields of `Block<T>` nullable. So I want to suggest to update and to discuss for below:

## Rename `Block<T>.StateRootHash` more universal.

The name, `StateRootHash`, was from other blockchain's naming using [MPT] (e.g. [Ethereum]) and means the root hash of the tree. It is not suitable for another `IStateStore` implementation not using MPT. So we should rename it into a more universal name to mean "prove for the state of the block".

If you have some ideas about the naming, please leave them as a comment. cc @planetarium/libplanet

[Ethereum]: https://ethereum.org/en/
[MPT]: https://eth.wiki/en/fundamentals/patricia-tree

## Separate the logic to validate its state from `BlockChain<T>`

Currently, `BlockChain<T>` validates the state of the block when using `TrieStateStore`. And the logic is in the `BlockChain<T>` implementation and it should be separated as another interface `IStateValidator` or something with better naming.

IMHO `BlockChain<T>` should be just a replacement of its tip, which doesn't do anything like mining.

-----

Except for the above things, it should need one more interface to calculate the proof of the state.

PR #1492, opened 2021-09-15
Represent each stage of block producing as distinct types

dahlia:
### For reviewers

- @riemannulus, this patch tries to [separate](#new-types) the process of producing blocks into two explicit stages: the *evaluation* stage (which determines a block's state root hash) and the *proof-of-work* stage (which mines a block's nonce and pre-evaluation hash).  As a result, `Nonce` and `PreEvaluationHash` now belong to [`PreEvaluationBlock<T>`] (& [`PreEvaluationBlockHeader`]), while `StateRootHash` and `Hash` still remain in [`Block<T>`] (& [`BlockHeader`]).  Please take a look into this, as the concept of two types of hashes was originally designed by you (<https://github.com/planetarium/libplanet/pull/935>).
- @moreal, this patch tries to [let `Block<T>.StateRootHash` mandatory](#mandatory-state-root-hash).  This means state root hashes are no more optional but required for blocks.  Please take a look into this, because you decided to let some blocks still lack their state root hashes (<https://github.com/planetarium/libplanet/pull/939> & <https://github.com/planetarium/libplanet/pull/986>).
- @moreal, this patch made a [renovation on the existing `IStateStore` interface](#state-store).  Two key changes are: a) [`IStateStore`] became required to be a trie, and b) [`IStateStore`] no more stores associations between block hashes and state root hashes, as now every block must have its own state root hash.  Also, this introduced [`StateStoreExtensions`] too.  Please look into this too, as you introduced `IStateStore` (<https://github.com/planetarium/libplanet/pull/950>).
- @limebell, in this patch, `RawBlock` was gone, and [`BlockMetadata`]/[`BlockContent<T>`] types were introduced instead.  Also, [common interfaces](#block-interfaces) like [`IBlockMetadata`]/[`IBlockContent<T>`]/[`IBlockHeader`]/etc were introduced to group them type-wise.  Also, [some serialization methods were also reorganized.](#serialization) Please look into this [new type hierarchy](#new-types), as you designed the existing organization of block metadata types (<https://github.com/planetarium/libplanet/pull/757> & <https://github.com/planetarium/libplanet/pull/1070>).
- @longfin, @greymistcube, & @kfangw, please take a look into the overall design of [new block types](#new-types).
- There are a lot of removed lines in the test code.  As their diffs are noisy, I recommend you to read the added implementations first, and check the test cases when you're curious about corner cases.
- [`Block<T>`] and [`BlockHeader`] classes are mostly rewritten.  Reading these files through diffs wouldn't be effective.  Read the latest revision instead.
- [Diff Monster] could be helpful.

[`PreEvaluationBlockHeader`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/PreEvaluationBlockHeader.cs
[`PreEvaluationBlock<T>`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/PreEvaluationBlock.cs
[`Block<T>`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/Block.cs
[`BlockHeader`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/BlockHeader.cs
[`IStateStore`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Store/IStateStore.cs
[`StateStoreExtensions`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Store/StateStoreExtensions.cs
[`BlockMetadata`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/BlockMetadata.cs
[`BlockContent<T>`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/BlockContent.cs
[`IBlockMetadata`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/IBlockMetadata.cs
[`IBlockContent<T>`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/IBlockContent.cs
[`IBlockHeader`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/IBlockHeader.cs
[Diff Monster]: https://diff.sapzil.org/#/planetarium/libplanet/pull/1492

### <span id="new-types">Make illegal states of `Block<T>`/`BlockHeader` unrepresentable</span>

*Addresses <https://github.com/planetarium/libplanet/issues/1164>.  It does not deal with `Transaction<T>`; that should be fixed in the future.*

In this patch, several new distinct types are introduced to represent each stage of producing blocks.  These types and existing ones can be placed on the table like below:

| Stages                                                                  | Only having `TxHash`         | Having complete `Transactions` |
|-------------------------------------------------------------------------|------------------------------|--------------------------------|
| Basic validations on fields<br>(e.g., `Index` cannot be less than zero) | [`BlockMetadata`]            | [`BlockContent<T>`]            |
| Proof-of-work<br>(i.e., having `Nonce` & `PreEvaluationHash`            | [`PreEvaluationBlockHeader`] | [`PreEvaluationBlock<T>`]      |
| Evaluating actions<br>(i.e., having `StateRootHash`)                    | [`BlockHeader`]              | [`Block<T>`]                   |

In the above table, upper types are more primitive, and lower types and more verified.  Upper types can be shifted to lower types through verifying their contents. For [`BlockContent<T>`]s, they can be shifted to [`PreEvaluationBlock<T>`]s through proof-of-work *mining*:

```csharp
BlockChain<T> chain = ...;
IBlockPolicy<T> policy = chain.Policy;
HashAlgorithmType hashAlgorithm = policy.GetHashAlgorithm(content.Index);
Transaction<T>[] txs = ...;

BlockContent<T> content = new BlockContent<T>
{
    Index = chain.Tip.Index + 1L,
    Miner = minerKey.ToAddress(),
    Difficulty = policy.GetNextBlockDifficulty(chain),
    TotalDifficulty =
        chain.Tip.TotalDifficulty + policy.GetNextBlockDifficulty(chain),
    PreviousHash = chain.Tip.Hash,
    Transactions = txs,
};
PreEvaluationBlock<T> preEvalBlock = content.Mine(hashAlgorithm);
```

(As you can see, I've made habitual naming of `preEvalBlock` or more shortly `preEval` to refer `PreEvaluationBlock<T>`s here and there in this patch.)

Once again, [`PreEvaluationBlock<T>`]s also can be shifted to [`Block<T>`]s through *evaluating* actions (with an optional block action):

```csharp
Block<T> block = preEvalBlock.Evaluate(chain);
```

However, for producing genesis blocks, this method is not easy to utilize, because `BlockChain<T>` itself can be instantiated after its genesis block is produced.  Therefore, there's a more overload taking a block action (if any) and an `IStateStore` together:

```csharp
IStateStore stateStore = chain.StateStore;
IAction blockAction = policy.BlockAction;

PreEvaluationBlock<T> preEvalGenesis = ...;
Block<T> genesis = preEvalGenesis.Evaluate(blockAction, stateStore);
```

### Early validation

These types validate their integrity at the earliest time.  For example, validating a protocol version number is conducted when a [`BlockMetadata`] instance is created, so you don't have to double-check it when you deal with richer types than `BlockMetadata`, like [`PreEvaluationBlockHeader`] or [`Block<T>`].

For this reason, many test cases had existed in `BlockTest` and `BlockHeaderTest` became unnecessary.

### <span id="block-interfaces">Common block interfaces</span>

Besides these concrete types, several new hierarchical interfaces were also introduced:

- [`IBlockMetadata`]
- [`IBlockContent<T>`]
- [`IPreEvaluationBlockHeader`]
- [`IPreEvaluationBlock<T>`]
- [`IBlockHeader`]

These interfaces are helpful for *narrowing* types so that they fit for the minimum required information.  For example, methods on `ActionEvaluator<T>` now take `IPreEvaluationBlock<T>` instead of the concrete `Block<T>`.  You can pass either `Block<T>` or `PreEvaluationBlock<T>`, or even a mock type implementing the interface, to the methods.

A bonus of the new organization of types is that writing tests become easier, because we don't need complete `Block<T>` fixtures every time.  If you don't need to deal with action evaluation or PoW, `BlockContent<T>`/`BlockMetadata` is enough.

[`IPreEvaluationBlockHeader`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/IPreEvaluationBlockHeader.cs
[`IPreEvaluationBlock<T>`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/IPreEvaluationBlock.cs

### <span id="serialization">Two types of serialization</span>

Libplanet has two types of serialization, and they have their own purpose:

1. input payloads of cryptographic digests, i.e., hashing & signing
2. data transmission on the network and persistence on the disk

As block models heavily depend on the former one to check their sanity (e.g., `IPreEvaluationBlockHeader.PreEvaluationHash`, `IBlockHeader.Hash`), methods for this remain in these model classes.

However, because the latter one does not affect validation checks, I got rid of methods for this use, and added [`BlockMarshaler`] static class instead.  The common marshaling still can be easily done through extension methods:

```csharp
// These extension methods are defined in BlockMarshaler, not Block<T> nor BlockHeader:
IValue marshalledBlock = block.MarshalBlock();
IValue marshalledHeader = header.MarshalBlockHeader();
```

Since these extensions methods take abstract interfaces rather than concrete classes as their `this` parameter, narrower marshaling is easy:

```csharp
// Serializes only its header, not transactions:
IValue marshalledHeader = block.MarshalBlockHeader();
```

As a side note, we now have <span id="assert-bencodex-equal">`TestUtils.AssertBencodexEqual()`</span> method.  This visualizes the diff between two Bencodex trees if two are not equal.  I believe it's a better way to test serialization code:

```
--- Expected
+++ Actual

 Bencodex.Types.Dictionary {
   b"\x48": {
     b"\x00": 1,
     b"\x54": 2,
     b"\x64": 1,
-    b"\x69": 3,
+    b"\x69": 2,
     b"\x6e": b"\x5c\x77\x74\xc2\x39\x69\x37\x51\x87\xa5",
-    b"\x74": "0001-01-01T00:00:00.000000Z"
+    b"\x74": "2018-11-29T00:00:30.000000Z"
   }
 }
```

[`BlockMarshaler`]: https://github.com/dahlia/libplanet/blob/pre-eval-block/Libplanet/Blocks/BlockMarshaler.cs

### <span id="mandatory-state-root-hash">Now every `Block<T>` instance has its `StateRootHash`</span>

*This fixes <https://github.com/planetarium/libplanet/issues/1146> and <https://github.com/planetarium/libplanet/issues/1345>.*

At the time [we first introduced Ethereum-style <abbr title="Merkle Patricia Trie">MPT</abbr>][1], we weren't confident how it would work well on the mainnet, so we had to leave a legacy mode so that we could turn it off and roll back to the existing way if needed.

However, this dual system had confused us, as `Block<T>.StateRootHash` could be either filled or `null`.  Here and there, we'd written conditionals to ensure whether a block had `StateRootHash`.  In the test code, `TestUtils.AttachStateRootHash<T>()` method had played its role, it had made the test code complicated at a time.  (I had a habit to instantly try adding `.AttachStateRootHash()` to block fixtures when the test failed while I didn't understand why.)

Fortunately, MPT has been battle-tested on the mainnet, and is now proven that it works well.  As we don't consider turning off it anymore, I decided to remove the nullable notation from the type of `Block<T>.StateRootHash`.  Instead, block contents at pre-evaluation stage became represented as a separated new type: `PreEvaluationBlock<T>`.  `TestUtils.AttachStateRootHash<T>()` method is also gone, as it's perfectly replaced by `PreEvaluationBlock<T>.Evaluate()` method.

[1]: https://github.com/planetarium/libplanet/pull/939

### <span id="state-store">[`IStateStore`] assuming it's trie</span>

*This closes <https://github.com/planetarium/libplanet/issues/1128>.*

The [`IStateStore`] interface is also revamped, as we don't need to mind if it isn't a trie.  Now `IStateStore` requires to be a trie.

Type conditionals like `if (stateStore is TrieStateStore trieStateStore)`, which are oen of the famous anti-patterns, were also completely removed.

Previously, `IStateStore` has maintained the associations between `Block<T>.Hash`–`Block<T>.StateRootHash` and `Block<T>.PreEvaluationHash`–`Block<T>.StateRootHash`, but now they are not saved, because these associations remain the block headers.

### Catch-up pull requests for Nine Chronicles

- <https://github.com/planetarium/lib9c/pull/657>
- <https://github.com/planetarium/NineChronicles.Headless/pull/760>


Libplanet/Store/BaseStore.cs
```diff
@@ -68,32 +68,16 @@ ISet<TxId> txids
         /// <inheritdoc cref="IStore.IterateBlockHashes()"/>
         public abstract IEnumerable<BlockHash> IterateBlockHashes();
 
-        /// <inheritdoc cref="IStore.GetBlock{T}(BlockHash)"/>
-        public Block<T> GetBlock<T>(BlockHash blockHash)
+        /// <inheritdoc cref="IStore.GetBlock{T}"/>
+        public Block<T> GetBlock<T>(HashAlgorithmGetter hashAlgorithmGetter, BlockHash blockHash)
```
longfin, line 72: Do we need `hashAlgorithmGetter` here? (there is no usage in this method...)

longfin (commented):
Sorry for the late review, I'll continue within today.

Libplanet/Blocks/BlockMetadata.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        /// <inheritdoc cref="IBlockMetadata.TxHash"/>
+        [SuppressMessage(
+            "SonarQube",
+            "S2292",
+            Justification = "The backing field purposes to prevent intercepting from subclasses.")]
+        public virtual HashDigest<SHA256>? TxHash
+        {
+            get => _txHash;
+            set => _txHash = value;
```
greymistcube, line 198:
```suggestion
            protected set => _txHash = value;
```

Do we need `set` to be publicly exposed? The same goes for `BlockContent<T>`.

Libplanet/Blocks/IBlockContent.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// parameter is aligned with <see cref="Transaction{T}"/>'s type parameter.</typeparam>
+    public interface IBlockContent<T> : IBlockMetadata
+        where T : IAction, new()
+    {
+        /// <summary>
+        /// Transactions belonging to the block.
+        /// </summary>
+        /// <remarks>This is always ordered by <see cref="Transaction{T}.Id"/>.</remarks>
+        // FIXME: Need to change the type to IImmutableSet<Transaction<T>>.
+        IReadOnlyList<Transaction<T>> Transactions { get; }
```
greymistcube, line 22:
```suggestion
        IImmutableList<Transaction<T>> Transactions { get; }
```

I think `IImmutableList<T>` might be more appropriate considering its underlying class. Depending on type compatibility, we can ignore this for now.

kfangw:
I've confused about the `pre-evaluation` sometimes, but now, it looks clear with this explicit type.
If my understanding is correct, the semantic, the `pre-evaluation` based, is **N+1 finality**.

* Basically, the state root hash living in the Nth Block does not proof anything. Because it is not included in the hashed block data and not went thru PoW process.
* But, the state root hash in the Nth Block is confirmed(pass consensus somehow, PoW,  PoS, or whatever) in the N+1 th Block. Because N+1 Block consist of hash of Nth Block and the Nth Block contains the state root hash. As a result state root hash of Nth Block is confirmed when the N+1 th Block is mined.

For that reason, I suggest more tangible Type.
###  Make N+1 block contains the state root hash which is generated from executing the transactions in the Nth Block not N+1 itself.

I have no strong opinion on it. But wanna discuss about this. 
What do you think about this? If I am wrong, please teach me without hesitation.

greymistcube:
Regarding #1164, I kind of miss usage of constructors. ☹️ 

```
BlockContent<T> content = new BlockContent<T>
{
    Index = chain.Tip.Index + 1L,
    Miner = minerKey.ToAddress(),
    Difficulty = policy.GetNextBlockDifficulty(chain),
    TotalDifficulty =
        chain.Tip.TotalDifficulty + policy.GetNextBlockDifficulty(chain),
    PreviousHash = chain.Tip.Hash,
    Transactions = txs,
};
```

Maybe I'm missing something, but can you elaborate on a rationale to move to using property setters instead of constructors? I know we had general problems with parameters getting out of control, but I don't see why this is better.

Case in point:
https://github.com/dahlia/libplanet/blob/56a8308326a80bc203b8976a44a5c7071610cb42/Libplanet/Blocks/BlockMetadata.cs#L135-L141
```
        /// <remarks>This cannot not be negative.
        /// <para>When <see cref="Difficulty"/> is updated, <see cref="TotalDifficulty"/> is also
        /// updated together.  For example, when <see cref="Difficulty"/> = 10 and
        /// <see cref="TotalDifficulty"/> = 50, if <see cref="Difficulty"/> is updated to
        /// 20 (= 10 + 10) <see cref="TotalDifficulty"/> is also updated to 60 (= 50 + 10).</para>
        /// </remarks>
        public long Difficulty
```

- There seem to be too many unforeseen dependencies between properties. I don't think setting a property to update another property should be an expected behavior.
- I think one or the other should apply, not both.
  - Instead of trying to handle possible non-sensical default values, forcing a value as a requirement in a constructor may reduce implicit logic.
  - Make property checking logic as minimal as possible and have a created object validated somehow somewhere.

Instantiation through property setting may be convenient, but it'd be even harder to keep track of all edge cases that may leak.

Libplanet.Extensions.Cocona/Commands/StatsCommand.cs
```diff
@@ -79,15 +79,17 @@ public class StatsCommand
 
             foreach (var hash in hashes)
             {
-                var block = store.GetBlock<Utils.DummyAction>(hash);
+                BlockDigest blockDigest = store.GetBlockDigest(hash) ??
+                    throw Utils.Error($"Failed to load the block {hash}.");
+                BlockHeader block = blockDigest.Header;
```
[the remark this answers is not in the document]
dahlia: Changed: 8455f601fea9ac362463fab8e6a1bdbba9ed1e2a.

Libplanet/Hashcash.cs
```diff
@@ -68,7 +71,8 @@ public static class Hashcash
                 random.NextBytes(nonceBytes);
                 var nonce = new Nonce(nonceBytes);
 
-                var digest = hashAlgorithmType.Digest(stamp(nonce));
+                IEnumerable<byte[]> chunks = stamp(nonce);
+                var digest = hashAlgorithmType.Digest(chunks);
```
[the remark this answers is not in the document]
dahlia: Amended in c5f5f97c84fdda2001a65a0b1352a94fce01b7e2.

Libplanet/Store/BaseStore.cs
dahlia: It's actually needed, but I omitted passing it to the `BlockDigest`.  It's now fixed: 8455f601fea9ac362463fab8e6a1bdbba9ed1e2a.

Libplanet/Blockchain/Renderers/Debug/ValidatingActionRenderer.cs
```diff
@@ -140,28 +141,50 @@ Block<T> branchpoint
...
+
                 expectedUnrenderedActions.AddRange(
-                    block.Transactions.SelectMany(t => t.Actions).Cast<IAction>().Reverse());
-                block = store.GetBlock<T>(block.PreviousHash ??
-                    throw Error(Records, "Reorg occurred from the chain with different genesis."));
+                    transactions.SelectMany(t => t.Actions).Cast<IAction>().Reverse());
+
+                block = store.GetBlockDigest(
+                    block.Header.PreviousHash ?? throw heterogeneousGenesisError
+                ) ?? throw Error(Records, $"Failed to load block {block.Header.PreviousHash}.");
```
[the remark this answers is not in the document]
dahlia: It's difficult to extract the exception object to a shared variable, because `block.Header.PreviousHash` is not a loop invariant.

Libplanet/Blocks/IPreEvaluationBlock.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// A common interface for blocks that have their proof-of-work
+    /// <see cref="IPreEvaluationBlockHeader.Nonce"/>s and
+    /// <see cref="IBlockContent{T}.Transactions"/>, but are not evaluated yet (i.e., their state
+    /// root hashes are not yet determined).
+    /// </summary>
+    /// <typeparam name="T">A class implementing <see cref="IAction"/> to include.  This type
+    /// parameter is aligned with <see cref="Transaction{T}"/>'s type parameter.</typeparam>
+    public interface IPreEvaluationBlock<T> : IBlockContent<T>, IPreEvaluationBlockHeader
```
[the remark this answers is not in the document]
dahlia: The interface can accept `Block<T>` as well as `PreEvaluationBlock<T>`.  It's also [easier to make test fixtures without mocking.][1]  For the rationale, please read [this](https://github.com/planetarium/libplanet/pull/1492#block-interfaces).

[1]: https://softwareengineering.stackexchange.com/q/387436

Libplanet/Blocks/IBlockContent.cs
dahlia: As I left a `FIXME` comment, IMHO ideally it should be `IImmutableSet<Transaction<T>>` so that client code can't depend on its implicit ordering.  When a certain ordering is needed, it should explicitly sort them in client side.  However, changing the type seems out of the scope of this patch.  It would be good if we have a separated issue for discussing it.

dahlia:
> Make N+1 block contains the state root hash which is generated from executing the transactions in the Nth Block not N+1 itself.

@kfangw As your suggestion is quite radical, I guess it would require a lot of changes in multifaceted ways.  Anyway, it would be good if we have a separated issue for discussing it.  I personally am open to the idea.

dahlia:
@greymistcube  It makes writing tests a lot easier, especially, creating fixture objects.  For example, with object initializers it's trivial to create another fixture from an existing fixture object that differs by just one field: `new BlockMetadata(fixture) { Timestamp = DateTimeOffset.UtcNow }`.

My initial intention was introducing [builders](https://en.wikipedia.org/wiki/Builder_pattern) for blocks.  In other words, I thought it would be easier to use if we separate building blocks from representing blocks.  While representation objects should guarantee its sanity, builder objects haven't to validate its states until the complete representation is made.  On the other hand, representation objects should be immutable because it's difficult to validate the mutating states.  However, builder objects are more useful when they are open to change.

dahlia:
@riemannulus @moreal @limebell @longfin @greymistcube @kfangw I rebased on the current main, and also applied your suggestions (or left some replies).  Please take a look again!

Libplanet/Store/StateStoreExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        /// <param name="stateStore">The <see cref="IStateStore"/> to records
+        /// the <paramref name="rawStatesDelta"/>.</param>
+        /// <param name="previousStateRootHash">The state root hash on which
+        /// the <paramref name="rawStatesDelta"/> is based.</param>
+        /// <param name="rawStatesDelta">The raw states delta to be recorded.</param>
+        /// <param name="rehearsal">If turned on, the <paramref name="rawStatesDelta"/> is not
+        /// recorded to the <paramref name="stateStore"/>, but only shadow trie root is returned.
+        /// Turned off by default.</param>
+        /// <returns>The new state root hash.</returns>
```
moreal, line 33: I'm not that sure it can treat the trie as `state root hash` (feel like `Hash`) 🤔

Libplanet/Store/TrieStateStore.cs
```diff
@@ -30,76 +24,30 @@ public class TrieStateStore : IStateStore
...
-
-        public void PruneStates(IImmutableSet<BlockHash> excludeBlockHashes)
+        /// <inheritdoc cref="IStateStore.PruneStates(IImmutableSet{HashDigest{SHA256}})"/>
+        public void PruneStates(IImmutableSet<HashDigest<SHA256>> stateRootHashes)
         {
             var stopwatch = new Stopwatch();
             _logger.Verbose($"Started {nameof(PruneStates)}()");
             var excludeNodes = new HashSet<HashDigest<SHA256>>();
-            foreach (var blockHash in excludeBlockHashes)
+            foreach (HashDigest<SHA256> stateRootHash in stateRootHashes)
```
moreal, line 44: The argument's name was renamed but the logic seems not changed. 🤔 Is it okay? Originally the name was `excludeBlockHashes` and `PruneStates` removed all states except others included in the given `excludeBlockHashes`.

According to the `IStateStore.PruneStates`' document, it should prune the all states included in the given `stateRootHashes` if I understood correctly, without effect to other state roots 🤔

dahlia: It's apparently my mistake.  The word *exclude* here seemed ambiguous anyway, because it can mean both opposite ways:

1. to *exclude* (= delete) them from the state store,
2. or to *exclude* them from removal (= remain them).

During changing the code, I misunderstood its intention — I thought it meant the latter.  However, it's turned out that it meant the former, right?

As it's more misleading with the current name (`stateRootHashes`) than before, I'm going to rename it anyway, but because its original name was also misleading, I suggest a new name: `survivalStateRootHashes`.  Although it's quite long, seems more descriptive to me.  Would this name satisfy you too?

moreal: I thought a few minutes but I don't have better naming 😞 

It seems okay 👌🏻 👍🏻

Libplanet.Tests/Blocks/BlockMarshalerTest.cs
```diff
@@ -0,0 +1,128 @@
...
+        [Fact]
+        public void MarshalBlock()
+        {
+            AssertBencodexEqual(_marshaledGenesis, _fx.Genesis.MarshalBlock());
+            AssertBencodexEqual(_marshaledNext, _fx.Next.MarshalBlock());
+            AssertBencodexEqual(_marshaledHasTx, _fx.HasTx.MarshalBlock());
+        }
+
+        [Fact]
+        public void UnmarsahlBlock()
```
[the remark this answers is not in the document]
dahlia: Amended: 1d519afb5bceeb085faa03bb27521507dcd2c569.

Libplanet/Store/StateStoreExtensions.cs
dahlia: Amended: c3548d94ba1190d03ac697092fa56ed1ffa3038a.

Libplanet/Store/TrieStateStore.cs
dahlia: Applied the new name (`survivalStateRootHashes`): d65fe700f93f09d47f5d06497e459335ad8b8beb.  Feel free to suggest any better name even later.

greymistcube:
I agree that it may be a good idea to implement a separate builder to handle all these different "layers", but I think it's a slightly separate issue. I understand the convenience and the current lack of resource to consider all different dependencies, but I am still concerned about possible logical leakage at this level. Aside from the design issue, I deem part of the problem was that persisted was due to lack of control over arguments, by giving too much freedom so that we can arbitrarily plug in values *for the sake of easy testing*, where non-sensical combination of values were allowed. In my opinion, implementing a factory/builder class, to wrap the instantiation process, *and* revamping fixture helper class/methods is the right way to go. I don't think the change is realistic at this point, but I'm just voicing my concern. 🙃

greymistcube:
On a different note, I think `PreEvaluationBlock<T>.Evaluate()` and `PreEvaluationBlock<T>.DetermineStateRootHash()` and such should be offloaded somewhere else. For one, the name itself is misleading as `PreEvaluationBlock<T>.Evaluate(BlockChain<T>)` would usually read some `PreEvaluationBlock<T>` instance evaluating a `BlockChain<T>` instance.

I understand the upside of having such methods, but this strengthens coupling in too many different places, which is why I tried to remove all evaluation related logic to `ActionEvaluator<T>` in #1305.

I suggest:
- Rename methods: `Evaluate` -> `EvaluateAgainst`, `DetermineStateRootHash` -> `DetermineStateRootHashFrom`, etc.
  - Transitive verbs should be used with prepositions to clarify its meaning, as without, they mean the opposite. 😃 
- As suggested above, if possible, redirect calls to existing `BlockChain<T>.ActionEvaluator`. For instance, calling said methods `EvaluateAgainst` and `DetermineStateRootHashFrom` give a stronger vibe(?) that `BlockChain<T>` may/can be in control instead of `PreEvaluationBlock<T>`. I admit `ActionEvaluator<T>` is in pretty lousy state, which is all my fault 🙃 , but if we get to the point where builder/factory gets implemented, I think it'd be easier to port on a later date.

dahlia:
@greymistcube I'm still skeptic about whether `Evaluate()` and `DetermineStateRootHash()` methods should move to elsewhere, because they seem already a thin wrapper around `ActionEvaluator<T>`.  One and only thing they do without `ActionEvaluator<T>`'s help is dealing with state roots, but this part is merely a thin wrapper for `IStateStore` (and `StateStoreExtensions`) too.

In terms of naming, I generally agree with you that it's quite confusing as these English verbs are transitive.  It would be better if you suggest shorter alternatives though.  (`DetermineStateRootHashFrom` looks too long to me.)

longfin (approved):

Libplanet/Store/BaseStore.cs
greymistcube: Also took me a while *why* this is needed. 😕

dahlia: It's because `IStore`'s some methods require `Block<T>`, which is too rich for this low level layer.  This should be fixed anyway in the future.

Libplanet/Store/IStateStore.cs
```diff
@@ -14,56 +12,19 @@ namespace Libplanet.Store
...
-        /// <typeparam name="T">An <see cref="IAction"/> type.  It should match to
-        /// <paramref name="branchpoint"/>'s type parameter.</typeparam>
-        void ForkStates<T>(
-            Guid sourceChainId,
-            Guid destinationChainId,
-            Block<T> branchpoint)
-            where T : IAction, new();
+        /// <param name="survivalStateRootHashes">The state root hashes <em>not</em> to prune.
+        /// These state root hashes are guaranteed to survive after pruning.</param>
+        void PruneStates(IImmutableSet<HashDigest<SHA256>> survivalStateRootHashes);
```
greymistcube, line 28: I think `survivalStateRootHashes` sounds too exotic. 🙃 
Maybe `survivingStateRootHashes`, `stateRootHashesToSurvive`, `stateRootHashesNotToPrune`?
Considering the context, maybe even `keysToExclude`?

Libplanet/Store/BaseStore.cs
greymistcube: Yeah, I understand this is necessary under the current design/implementation. It was half rhetorical, and only half serious to leave some paper trail, so that we have a shared elevated concern for the issue. 😛

CHANGES.md
```diff
@@ -8,18 +8,195 @@ To be released.
...
+ -  Added `BlockMarshaler` static class.  [[#1164], [#1492]]
+ -  Added `BlockDigest.Index` property.  [[#1492]]
+ -  Added `BlockDigest.Hash` property.  [[#1492]]
+ -  Added `BlockDigest.StateRootHash` property.  [[#1492]]
+ -  Added `BlockDigest.GetHeader()` method.  [[#1492]]
+ -  Added `StateStoreExtensions` static class.  [[#1128], [#1146], [#1492]]
+ -  Added `StoreExtensions.GetStateRootHash()` extension method.
+    [[#1128], [#1146], [#1492]]
+ -  Added `DelayedRenderer<T>.HashAlgorithmGetter` property.  [[#1492]]
+ -  `BlockDigest` became to implements `IBlockExcerpt`.
```
greymistcube, line 181:
```suggestion
 -  `BlockDigest` became to implement `IBlockExcerpt`.
```

Libplanet/Store/IStateStore.cs
dahlia: I wanted to avoid the term *exclude* here… [(context)](https://github.com/planetarium/libplanet/pull/1492#discussion_r718167988).  I'm going to rename it to `survivingStateRootHashes` in the PR #1507.

CHANGES.md
dahlia: I'm going to fix the typo in the PR #1507.

Libplanet/Tx/TxId.cs
```diff
@@ -119,6 +119,7 @@ public ImmutableArray<byte> ByteArray
         [Pure]
         public override string ToString() => ToHex();
 
+        /// <inheritdoc cref="IComparable{T}.CompareTo(T)"/>
         public int CompareTo(TxId other)
```
greymistcube, line 123: Huh, didn't know `TxId`'s implemented `IComparable<T>`. Is this still needed?

dahlia: It's used for ordering transactions by their ID … when blocks are serialized?

greymistcube: Sorry, my bad. Found it. 😓

Libplanet/Blocks/BlockContentExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// Note that this validates only a block's intrinsic integrity between its transactions,
+        /// but does not guarantee integrity between blocks.  Such validation needs to be conducted
+        /// by <see cref="Blockchain.BlockChain{T}"/>.</exception>
+        /// <exception cref="InvalidTxGenesisHashException">Thrown when transactions to set have
+        /// inconsistent genesis hashes.</exception>
+        /// <exception cref="InvalidBlockTxHashException">Thrown when the given block
+        /// <paramref name="content"/>'s <see cref="IBlockMetadata.TxHash"/> is not consistent with
+        /// its <see cref="IBlockContent{T}.Transactions"/>.</exception>
+        /// <seealso cref="BlockContent{T}(IBlockContent{T})"/>
+        public static BlockContent<T> Copy<T>(this IBlockContent<T> content)
```
greymistcube, line 50: Why is this needed? I can't seem to find its use. 🙁

dahlia: It's used in test code, and I think it's also useful for applications.

    <Libplanet.Tests>  (28 usages found)
        Blocks  (28 usages found)
            BlockContentExtensionsTest.cs  (3 usages found)
                (13: 55)  AssertBlockContentsEqual(Genesis, Genesis.Copy());
                (14: 53)  AssertBlockContentsEqual(Block1, Block1.Copy());
                (15: 57)  AssertBlockContentsEqual(BlockPv0, BlockPv0.Copy());
            BlockContentTest.cs  (3 usages found)
                (119: 32)  var block = Block1.Copy();
                (145: 32)  var block = Block1.Copy();
                (177: 53)  BlockContent<Arithmetic> block = Block1.Copy();
            BlockMetadataTest.cs  (1 usage found)
                (108: 38)  BlockMetadata b = Block1.Copy();
            PreEvaluationBlockHeaderTest.cs  (2 usages found)
                (187: 56)  BlockMetadata metadata = _contents.Genesis.Copy();
                (255: 56)  BlockMetadata metadata = _contents.Genesis.Copy();
            PreEvaluationBlockTest.cs  (19 usages found)
                (28: 66)  BlockContent<Arithmetic> content = _contents.Genesis.Copy();
                (36: 40)  content = _contents.Block1.Copy();
                (47: 40)  content = _contents.Block1.Copy();
                (53: 41)  content = _contents.Genesis.Copy();
                (59: 40)  content = _contents.Block1.Copy();
                (65: 41)  content = _contents.Genesis.Copy();
                (71: 41)  content = _contents.Genesis.Copy();
                (81: 66)  BlockContent<Arithmetic> content = _contents.Genesis.Copy();
                (93: 40)  content = _contents.Block1.Copy();
                (109: 40)  content = _contents.Block1.Copy();
                (131: 66)  BlockContent<Arithmetic> content = _contents.Genesis.Copy();
                (142: 40)  content = _contents.Block1.Copy();
                (165: 40)  content = _contents.Block1.Copy();
                (171: 41)  content = _contents.Genesis.Copy();
                (177: 40)  content = _contents.Block1.Copy();
                (183: 41)  content = _contents.Genesis.Copy();
                (189: 41)  content = _contents.Genesis.Copy();
                (202: 68)  BlockContent<Arithmetic> contentPv0 = _contents.Block1.Copy();
                (221: 68)  BlockContent<Arithmetic> contentPv1 = _contents.Block1.Copy();

greymistcube: Yes, I had found it in the test code. I agree, it *may* be useful, but I think this can *also* be dangerous. I'm okay with copying `BlockMetadata<T>`, my anxiety might be unfounded, but I think being able to copy `BlockContent<T>` is a slightly different matter and may be able to cause a problem down the line. 🙁

dahlia: Is it a problem with this `.Copy()` extension method, or the `new BlockContent<T>(IBlockContent<T>)` constructor?

greymistcube: Well, I think the question can be answered more easily with another question:

> Should we be able to copy `Block<T>` instances?

Ultimately, they are just reference copies anyway, it is hard to see any harm coming from it, and I'm probably just being paranoid, but who knows what might happen in the future? 😶

dahlia: `BlockContent<T>` in itself is not a block, and it does not have any proofs like nonce, pre-evaluation hash, or state-root hash.  It's similar to `BlockMetadata`, except it has `Transactions` too (besides `TxHash`).

greymistcube: Unless I'm missing something, my thought process was `BlockMetadata` on it's own can't be consumed to create a valid `Block<T>` but `BlockContent<T>` can, hence the question above, and which is why if `Block<T>` shouldn't be copyable, neither should `PreEvaluationBlock<T>` and `BlockContent<T>`.

dahlia: If `Block<T>` shouldn't be copyable, it's because it's frozen with several proofs so that modifying few part of it is non-sense.  On the other hand, it's okay that `BlockContent<T>` & `BlockMetadata` can be copied (and changing few part of them), and it's also useful, because it's merely a draft of a block to produce.

In other perspective, copying immutable objects are not useful, because they are just duplicates.  However, copying mutable objects are useful, because in general, mutable objects change themselves in-place. In order to revise mutable objects with maintaining original contents at a time, they should be copyable.

greymistcube: Ah, I get your point. Welp, I'm approaching from a more abstracted standpoint. Just being mutable doesn't mean copying and mutating won't have any side-effect to business-logic somewhere down the line. Anyway, I'm just talking fluffy nonsense. 😛

greymistcube (approved):
Approving, but I must apologize beforehand for my utterly insufficient, ineffective, and inefficient proofreading. 🙃 

It was fun peering into your overall design philosophy tho. 😛

Libplanet/Store/IStateStore.cs
dahlia: Fixed in 2c5bab598f4cd256bae9760e345bc845c6508459.

CHANGES.md
dahlia: Fixed in bcdd46e464acba2a4eaaceccf753b2e9390e0d5e.

PR #1495, opened 2021-09-16
Difficulty adjustment algorithm separated

greymistcube:
Just some preparatory work to make difficulty adjustment algorithm more pluggable.

Other than separating out the chunk of default implementation inside `BlockPolicy<T>`, minor changes can be found in the changelog.

Making `minimumDifficulty` and `difficultyStability` parameters nullable is partly due to my OCD, but with an additional reason that these shouldn't really be easily configurable unless the user *really* knows what they are doing. I suspect the current default values used are rather inappropriate. 😶

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -86,6 +86,8 @@ To be released.
     `BlockPolicy<T>()` constructor with `Func<long, int>? getMaxBlockBytes`.
     [[#1485]]
  -  Removed `TxViolatingBlockPolicyException` class.  [[#1485]]
+ -  Optional parameter name `difficultyBoundDivisor` for `BlockPolicy<T>()`
+    constructor changed to `difficultyStability`.  [[#1495]]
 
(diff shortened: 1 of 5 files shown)


limebell (dismissed):

limebell (commented):
How about adding new tests testing newly added condition throwing exceptions?

greymistcube:
Hmm... technically they are still technically covered by https://github.com/planetarium/libplanet/blob/f4bdf0bf922fddfe860ac4bd7924b9ee0b5ce8d3/Libplanet.Tests/Blockchain/Policies/BlockPolicyTest.cs#L51

I suppose since API's are added for external use, I need to separate it.

greymistcube:
Tests written, among other bug fixes. See 7544942e4d7a788e54f01d4849a4d3966959d043.

limebell (approved):

PR #1498, opened 2021-09-20
Slight optimization of `ByteUtil.Satisfies()`

greymistcube:
Removed unnecessary array allocation and uses bit shifting instead.

Note:
- As for `digestArray`, `CopyTo` or `CopyTo<T>` is slightly faster but `IReadOnlyList<T>` does not support direct copying.
  - Although from the method's perspective, changing the parameter type to `ImmutableArray<T>` may seem appropriate, it does not work since `HashAlgorithmType.Digest(byte[])` results in `byte[]` and there is no implicit conversion from `byte[]` to `ImmutableArray<byte>`. Furthermore, conversion to `ImmutableArray<byte>` from `byte[]` is rather slow, so this defeats the whole purpose. ☹️ 
  - `ReadOnlySpan<T>` is another possible option, but in this case, converting `ImmutableArray<T>` to `ReadOnlySpan<T>` causes an issue. Aside from whether this is the right choice, significant portion of the existing code needs to be updated if we are to go this route. ☹️

### DIFF
--- a/Libplanet/ByteUtil.cs
+++ b/Libplanet/ByteUtil.cs
@@ -162,23 +162,20 @@ namespace Libplanet
                 return false;
             }
 
-            var maxTargetBytes = new byte[hashDigest.Count + 1];
-            maxTargetBytes[hashDigest.Count] = 0x01;
-            var maxTarget = new BigInteger(maxTargetBytes);
+            BigInteger maxTarget = BigInteger.One << 8 * hashDigest.Count;
             BigInteger target = maxTarget / difficulty;
 
-            var digestArray = new byte[hashDigest.Count + 1];
+            // Ensures the last byte to be zero to prevent negative result when converted
+            // to BigInteger.  Note that BigInteger(byte[]) assumes the input bytes are in
+            // little-endian order.
+            byte[] digestArray = new byte[hashDigest.Count + 1];
             int i = 0;
             foreach (byte b in hashDigest)
             {
                 digestArray[i++] = b;
             }
 
-            // Append zero to convert unsigned BigInteger.  Note that BigInteger(byte[]) assumes
-            // the input bytes are in little-endian order.
-            digestArray[i] = 0;
-
-            var result = new BigInteger(digestArray);
+            BigInteger result = new BigInteger(digestArray);
             return result < target;
         }
     }


limebell (approved):

longfin (approved):

dahlia (commented):
I'm skeptical if this patch really optimizes performance. (If your intention was to optimize something other than performance, please let me know where you focused and optimized.) I ran a simple benchmark using the [BenchmarkDotNet](https://benchmarkdotnet.org/) framework. There doesn't seem to be any significant performance difference between the code before and after the patch. (Looking at the numbers, it was rather slightly slower.) 

### The existing code

|    Method | Difficulty |     Mean |   Error |  StdDev |
|---------- |----------- |---------:|--------:|--------:|
| Satisfies |     500000 | 515.0 ms | 2.51 ms | 2.10 ms |
| Satisfies |    1000000 | 515.2 ms | 1.66 ms | 1.30 ms |
| Satisfies |    5000000 | 516.1 ms | 2.50 ms | 2.22 ms |

### The patched code

|    Method | Difficulty |     Mean |    Error |   StdDev |
|---------- |----------- |---------:|---------:|---------:|
| Satisfies |     500000 | 519.9 ms |  9.35 ms |  9.61 ms |
| Satisfies |    1000000 | 523.0 ms | 10.21 ms | 12.54 ms |
| Satisfies |    5000000 | 516.9 ms |  4.73 ms |  3.95 ms |

I may have missed something while writing the benchmark code. Would you like to check it out?  (You can run it by adding the below file in the *Libplanet.Benchmarks/* directory of the source tree and executing the `dotnet run -p Libplanet.Benchmarks -c Release -- -f Libplanet.Benchmarks.ByteUtilBenchmark.Satisfies` command.)

```csharp
using System;
using BenchmarkDotNet.Attributes;

namespace Libplanet.Benchmarks
{
    public class ByteUtilBenchmark
    {
        [Params(500_000L, 1_000_000L, 5_000_000L)]
        public long Difficulty { get; set; }
        private byte[][] _preEvaluationHashes = new byte[1_048_576][];
        private Random _random = new Random();

        [IterationSetup(Target = nameof(Satisfies))]
        public void SetPreEvaluationHashAndDifficulty()
        {
            for (int i = 0; i < _preEvaluationHashes.Length; i++)
            {
                _preEvaluationHashes[i] = new byte[32];
                _random.NextBytes(_preEvaluationHashes[i]);
            }
        }

        [Benchmark]
        public int Satisfies()
        {
            int satisfied = 0;
            foreach (byte[] hash in _preEvaluationHashes)
            {
                // Without the assignment, the compiler will get rid of the call, because
                // the method's return value is not used:
                satisfied += ByteUtil.Satisfies(hash, Difficulty) ? 1 : 0;
            }

            return satisfied;
        }
    }
}
```

greymistcube:
Gotta say, this is just what's left over from a failed attempt to meaningfully optimizing `Satisfies()`. 🙃

greymistcube:
Welp, it's really not considered a proper use case under current implementation, but the discrepancy grows as the length grows. 😶 

```
# old Satisfies() with IReadonlyList<byte> of length 128
|    Method | Difficulty |     Mean |    Error |   StdDev |
|---------- |----------- |---------:|---------:|---------:|
| Satisfies |     500000 | 925.1 ms | 18.14 ms | 18.63 ms |
| Satisfies |    1000000 | 923.3 ms | 17.42 ms | 16.29 ms |

# new Satisfies() with IReadonlyList<byte> of length 128
|    Method | Difficulty |     Mean |    Error |   StdDev |
|---------- |----------- |---------:|---------:|---------:|
| Satisfies |     500000 | 817.2 ms | 14.77 ms | 13.82 ms |
| Satisfies |    1000000 | 818.2 ms | 13.73 ms | 12.84 ms |
```

PR #1516, opened 2021-10-08
Hotfix hash algorithm type json converter

greymistcube:
Due to incomplete implementation, converting `Block<T>`s results in malformed json strings breaking [9c-launcher]. This is a quick fix to address the issue.

[9c-launcher]: https://github.com/planetarium/9c-launcher


area363:
Branch needs to be updated 🙏

PR #1518, opened 2021-10-11
OrderTxsForEvaluationV3 added

### LINKED ISSUE #1322: Accommodate variable length `Block<T>.PreEvaluationHash` for `Block<T>.Transactions` ordering

### LINKED ISSUE #1323: Fix improper `Block<T>.Transactions` ordering

greymistcube:
Related to #1326.
Closes #1322.
Closes #1323.

This tries to readdress the issues mentioned in #1322 and #1323. While working on this PR, initial attempt was made to implement [Fisher-Yates] shuffling, but it turns out this is no trivial task.

First, some context on why [Fisher-Yates] is not used in this PR. The goal is to use `PreEvaluationHash` of a `Block<T>` to deterministically pseudo-randomly shuffle transactions by signer addresses. What [Fisher-Yates] algorithm provides is an unbiased sampling, but this requires higher entropy than that of `PreEvaluationHash` can provide by itself.<sup>[1](#footnote_01)</sup> Proper implementation would require increasing entropy by extending `PreEvaluationHash` with each transaction's signature, *not* signer address, on each step of selection required for [Fisher-Yates]. As my own implementation got more and more elaborate, I became concerned about possible performance degradation, so decided to take another route.

Implementation in this PR tries to achieve at least the following.

- Given `n` signers, the probability of any signer to be selected as first is `1/n`.
- Given any pair of signers `a` and `b`, the probability of `a` being selected before `b` is `1/2`.

Keep in mind, these conditions are *not* sufficient enough to give us unbiased permutation samplings of `n` signers, which is a potential security hole, however minor it may be. As two "consecutive" signers in the original ordering is highly likely to be executed consecutively, although not necessarily in order. This *can* be somewhat mitigated easily with some additional computation, but cannot be completely normalized while using *only* `PreEvaluationHash` as the random seed.

----

<a name="footenote_01">1</a>: For unbiased sampling of permutations of `n` elements, minimum required entropy is `log2(n!)`, meaning approximately `524` randomly generated `bit`s are required for shuffling `100` elements.

[Fisher-Yates]: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,9 @@ To be released.
 
 ### Backward-incompatible network protocol changes
 
+ -  The `Block<T>.CurrentProtocolVersion` is bumped from 2 to 3.
+    [[#1322], [#1323], [#1518]]
+
 ### Backward-incompatible storage format changes
 
 ### Added APIs
@@ -43,6 +46,8 @@ To be released.
 
 ### Behavioral changes
 
+ -  `Block<T>.Transactions` is ordered using a different scheme for
+    evaluation due to protocol version bump.  [[#1322], [#1323], [#1518]]
  -  (Libplanet.Net) `NetMQTransport.SendMessageWithReplyAsync()` should now
     complete its process within given timeframe `timeout` argument instead
     of possibly taking longer on some edge cases when waiting for
@@ -67,6 +72,9 @@ To be released.
     in one line as JSON whenever a different key is found,
     than it outputs all of the different nodes at once.  [[#1729]]
 
+[#1322]: https://github.com/planetarium/libplanet/issues/1322
+[#1323]: https://github.com/planetarium/libplanet/issues/1323
+[#1518]: https://github.com/planetarium/libplanet/pull/1518
 [#1593]: https://github.com/planetarium/libplanet/pull/1593
 [#1729]: https://github.com/planetarium/libplanet/pull/1729
 [#1734]: https://github.com/planetarium/libplanet/issues/1734
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2146,8 +2146,6 @@ Released on July 23, 2021.
 [#1315]: https://github.com/planetarium/libplanet/issues/1315
 [#1316]: https://github.com/planetarium/libplanet/issues/1316
 [#1320]: https://github.com/planetarium/libplanet/issues/1320
-[#1322]: https://github.com/planetarium/libplanet/issues/1322
-[#1323]: https://github.com/planetarium/libplanet/issues/1323
 [#1325]: https://github.com/planetarium/libplanet/pull/1325
 [#1326]: https://github.com/planetarium/libplanet/pull/1326
 [#1328]: https://github.com/planetarium/libplanet/pull/1328
--- a/Libplanet.Tests/Action/ActionEvaluatorTest.cs
+++ b/Libplanet.Tests/Action/ActionEvaluatorTest.cs
@@ -318,8 +318,8 @@ namespace Libplanet.Tests.Action
                     actions: new DumbAction[0],
                     timestamp: DateTimeOffset.MinValue.AddSeconds(7)),
             };
-            int i = 0;
-            foreach (Transaction<DumbAction> tx in block1Txs)
+            foreach ((var tx, var i) in block1Txs.Zip(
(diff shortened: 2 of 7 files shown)


Libplanet/Action/ActionEvaluator.cs
```diff
@@ -590,6 +592,43 @@ ActionContext CreateActionContext(IAccountStateDelta prevStates, int randomSeed)
                 .SelectMany(group => group.OrderBy(tx => tx.Nonce));
         }
 
+        [Pure]
+        private static IEnumerable<Transaction<T>> OrderTxsForEvaluationV3(
```
riemannulus, line 596: We don't have `OrderTxsForEvaluationV3` test here. Could you add some tests for this method?

greymistcube: Actually, `OrderTxsForEvaluation()` in `ActionEvaluatorTest.cs` covers all cases. 🙃

greymistcube:
Bumped `Block<T>.CurrentProtocolVersion`.

Libplanet.Tests/Action/ActionEvaluatorTest.cs
```diff
@@ -1016,21 +997,66 @@ public void OrderTxsForEvaluation()
...
             Assert.True(originalAddresses.SequenceEqual(
                 signers.Select(signer => signer.ToAddress().ToString())));
-            Assert.True(orderedAddresses.SequenceEqual(
-                orderedTxs
-                    .Where((tx, i) => i % numTxsPerSigner == 0)
-                    .Select(tx => tx.Signer.ToString())));
+
+            // Make sure to cover every case *and* the currently running protocol does not break.
+            int[] protocolVersions = { 0, 3, BlockMetadata.CurrentProtocolVersion };
+            foreach (int protocolVersion in protocolVersions)
```
dahlia, line 1006: Rather than placing a loop in a test case, it'd be better to [parametrize the test case using `[Theory]`][1] attribute.

[1]: https://andrewlock.net/creating-parameterised-tests-in-xunit-with-inlinedata-classdata-and-memberdata/

CHANGES.md
```diff
@@ -32,6 +37,8 @@ To be released.
 
 ### Behavioral changes
 
+ -  `Block<T>.Transactions` will be ordered using a different scheme once
+    the protocol version is bumped to `3`.  [[#1322], [#1323], [#1518]]
```
dahlia, line 41: As far as I know, `ActionEvaluator<T>.OrderTxsForEvaluation()` does not affect `Block<T>.Transactions` proprerty as data.  (Actually, `Block<T>.Transactions` are ordered by `TxId`.)  It only affect the evaluation order when states are determined.

greymistcube: As per suggestion, rephrased the changelog a bit. 😗

Libplanet.Tests/Action/ActionEvaluatorTest.cs
greymistcube: Yeah... this was a long time ago before I knew about the `[Theory]` attribute. 🙃
See https://github.com/planetarium/libplanet/pull/1518/commits/fa8fc9627adc746785660d6be4e942503a6960df.

dahlia (approved):

dahlia (approved):

PR #1528, opened 2021-10-15
Introduce: `ConsensusReactor` and `ConsensusBroadcaster` for PBFT

riemannulus:
This PR introduces the consensus reactor architecture for achieving consensus, which is the basis of the PBFT implementation. 

@dahlia I've to make the architecture you've mentioned but I re-design my own way. Please review carefully between `ConsensusReactor` and `ConsensusBroadcaster`

@kfangw Try to implement exactly what you're working stuff. Please review `ConsensusReactor`'s logic and test code. If don't mind. could you propose more test code like I working on this  PR?


riemannulus:
We don't implement PBFT now.

Issue #1504, opened 2021-09-29
Reorganize policy dependent validation


greymistcube:
Possible max block bytes for a block is checked in `BlockChain<T>.Append()` but number of transactions
is checked in `BlockChain<T>.Append()` for minimum number of transactions and `BlockPolicy<T>.ValidateNextBlock()` for maximum number of transactions. Checking of minimum number of transactions was probably introduced to `BlockPolicy<T>` in [lib9c] for operational reasons (trying to implement a feature without touching [libplanet]). This should be fixed.

[libplanet]: https://github.com/planetarium/libplanet
[lib9c]: https://github.com/planetarium/lib9c
[headless]: https://github.com/planetarium/NineChronicles.Headless

PR #1600, opened 2021-11-11
Offload policy validation to `IBlockPolicy<T>`

greymistcube:
Partially addresses #1504.

Offloaded much of state agnostic, i.e. does not depend on the current state of `BlockChain<T>` object that `Block<T>` is trying to get appended to, block validation to `IBlockPolicy<T>.ValidateNextBlock()`.

Issues still remain that concerns are still not fully separated.
- `IBlockPolicy<T>` should validate `Block<T>.PreEvaluationHash` not `Block<T>.HashAlgorithm`. In my opinion, we should do away with `Block<T>.HashAlgorithm` property if possible.
- `IBlockPolicy<T>` should also validate `Block<T>.Difficulty`. This is different from other simple validations as a difficulty of a block does not solely depend on its index, but also on the chain up to that point.

There are also other unresolved icky design issues.
- By allowing uses of customizable policy through `IBlockPolicy<T>` interface, we delegate responsibility to whoever implements `IBlockPolicy<T>`. But then we *must* trust that implementers would properly validate offloaded logic and possibly other constraints on top of said constraints, which is hard to do given the complexity of the task. Better API would be to have something like `IBlockPolicy<T>.CustomValidation()` and allow users to *add* ad hoc constraints and strongly suggest to subclass `BlockPolicy<T>` instead of straight up implementing `IBlockPolicy<T>`.
- As it stands, exception granularity concerning policy violation doesn't really make much sense since all `IBlockPolicy<T>` can guarantee is to return a `BlockPolicyViolationException` if certain portion of the policy is violated. Again, we can document all different scenarios where each subclass exceptions should be returned while running `IBlockPolicy<T>.ValidateNextBlock()`, I do not think this is a realistic expectation to have.


Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -114,6 +114,76 @@ public class BlockPolicy<T> : IBlockPolicy<T>
...
+                        {
+                            var offendingGroup = groups.First();
+                            return new InvalidBlockTxCountPerSignerException(
+                                $"Block #{block.Index} {block.Hash} includes too many " +
+                                $"transactions from signer {offendingGroup.Key} where " +
+                                $"the maximum number of transactions allowed by a single signer " +
+                                $"per block is {maxTransactionsPerSignerPerBlock}: " +
+                                $"{offendingGroup.Count()}",
+                                offendingGroup.Key,
+                                offendingGroup.Count());
```
longfin, line 180:
```suggestion
                        if (groups.FirstOrDefault() is { } offendingGroup)
                        {
                            int offendingGroupCount = offendingGroup.Count();
                            return new InvalidBlockTxCountPerSignerException(
                                $"Block #{block.Index} {block.Hash} includes too many " +
                                $"transactions from signer {offendingGroup.Key} where " +
                                $"the maximum number of transactions allowed by a single signer " +
                                $"per block is {maxTransactionsPerSignerPerBlock}: " +
                                $"{offendingGroupCount()}",
                                offendingGroup.Key,
                                offendingGroupCount());
```

longfin (dismissed):

longfin (approved):

PR #1610, opened 2021-11-24
Refactor/housekeeping

greymistcube:
Just some non-essential minor changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,8 @@ To be released.
     `messageLifespan` parameter added to constructors of `NetMQMessageCodec`
     and `TcpMessageCodec`.  [[#1610]]
  -  Removed unused `PeerStates` property from `Swarm<T>`.  [[#1610]]
+ -  Method name `BoundPeer.QueryAppProtocolVersion()` changed to
+    `BoundPeer.QueryAppProtocolVersionNetMQ()`.  [[#1610]]
 
 ### Backward-incompatible network protocol changes
 
@@ -54,7 +56,7 @@ To be released.
 [#1606]: https://github.com/planetarium/libplanet/pull/1606
 [#1609]: https://github.com/planetarium/libplanet/pull/1609
 [Bencodex 0.4.0-dev.20211123080042]: https://www.nuget.org/packages/Bencodex/0.4.0-dev.20211123080042
-[#1610]: https://github.com/planetarium/libplanet/pull/1609
+[#1610]: https://github.com/planetarium/libplanet/pull/1610
 
 
 Version 0.21.1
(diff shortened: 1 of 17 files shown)


CHANGES.md
```diff
@@ -47,6 +54,7 @@ To be released.
 [#1606]: https://github.com/planetarium/libplanet/pull/1606
 [#1609]: https://github.com/planetarium/libplanet/pull/1609
 [Bencodex 0.4.0-dev.20211123080042]: https://www.nuget.org/packages/Bencodex/0.4.0-dev.20211123080042
+[#1610]: https://github.com/planetarium/libplanet/pull/1609
```
riemannulus, line 57:
```suggestion
[#1610]: https://github.com/planetarium/libplanet/pull/1610
```

Libplanet/Net/Messages/NetMQMessageCodec.cs
```diff
@@ -104,15 +109,15 @@ public NetMQMessageCodec()
             var timestamp = new DateTimeOffset(ticks, TimeSpan.Zero);
 
             var currentTime = DateTimeOffset.UtcNow;
-            if (!(lifetime is null) &&
-                (currentTime < timestamp || timestamp + lifetime < currentTime))
+            if (_messageLifespan is TimeSpan lifespan &&
```
riemannulus, line 112: 👍

Libplanet/Net/Transports/BoundPeerExtensions.cs
```diff
@@ -96,7 +96,7 @@ public static class BoundPeerExtensions
             {
                 Identity = Guid.NewGuid().ToByteArray(),
             };
-            var messageCodec = new MessageCodec();
+            var messageCodec = new TcpMessageCodec();
```
riemannulus, line 99: Is a `BoundPeerExtensions` using `NetMQTransport`? If then, I guess we need to separate this class to `TCPBoundPeerExtensions`

greymistcube: Eh, didn't think too much about it since it's only used for a console command. I agree, that the whole extension is rather problematic as `BoundPeerExtensions` is only for `TcpTransport` whereas `BoundPeer` class is agnostic of `ITransport` type. We can change the name to `TcpBoundPeerExtensions`, but either way, it is still problematic. ☹️

riemannulus: Yeah, what I wanted to say more is, we need one pair of extensions if those extensions are dependent on some transport layer. In this case, For example, use `IMessageCodec` instead `TcpMessageCodec`.

greymistcube: Huh... looking at the surrounding code again, there is already `QueryAppProtocolVersion()` for NetMQ and `QueryAppProtocolVersionTcp()` for TCP. I think #1612 should be resolved first for better interfacing this whole extension class. I'll change `QueryAppProtocolVersion()` to `QueryAppProtocolVersionNetMQ()` for now to at least somewhat mitigate the confusion.

riemannulus (approved):

PR #1613, opened 2021-11-24
Chore/error serialization

greymistcube:
Just some minor housekeeping stuff.

### DIFF
--- a/Libplanet/Net/InvalidTimestampException.cs
+++ b/Libplanet/Net/InvalidTimestampException.cs
@@ -1,6 +1,7 @@
 #nullable enable
 using System;
 using System.Runtime.Serialization;
+using Libplanet.Serialization;
 
 namespace Libplanet.Net
 {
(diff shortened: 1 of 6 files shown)


Libplanet/Net/InvalidTimestampException.cs
```diff
@@ -10,7 +11,7 @@ public class InvalidTimestampException : Exception
         internal InvalidTimestampException(
             string message,
             DateTimeOffset createdOffset,
-            TimeSpan lifespan,
+            TimeSpan? lifespan,
```
riemannulus, line 14: Is it necessary changes? I'm not sure in which case `lifespan` gets the `TimeSpan` as nullable...

greymistcube: There is some logic issue in implementations of `IMessageCodec`, which I haven't got my hands on yet. In any case, `lifespan` should be nullable since `InvalidTimestampException` should be throwable even when `lifespan` is `null`.

riemannulus (approved):

dahlia (commented):
Could you write a short description about this fix in the changelog?

greymistcube:
@dahlia Done.

dahlia (approved):

PR #1614, opened 2021-11-25
Bad tip expiration for pulling blocks fix

greymistcube:
Inequality inside `Swarm<T>.PollBlocksAsync()` determining whether tip has become stale was set up to act oppositely to what was intended. This would've resulted in not polling properly once tip has actually gone stale unless tip is changed by a push from another node.

Also, changed the polling interval to `60` seconds.

### DIFF
--- a/Libplanet/Net/Swarm.BlockSync.cs
+++ b/Libplanet/Net/Swarm.BlockSync.cs
@@ -6,7 +6,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Libplanet.Blockchain;
 using Libplanet.Blocks;
-using Libplanet.Store;
 using Libplanet.Tx;
 using Nito.AsyncEx;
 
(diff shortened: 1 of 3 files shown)


Libplanet/Net/SwarmOptions.cs
```diff
@@ -128,7 +128,7 @@ public enum TransportType : byte
         /// for the configured lifespan, <see cref="Swarm{T}"/> pulls new blocks from neighbor
         /// peers.
         /// </summary>
-        public TimeSpan TipLifespan { get; set; } = TimeSpan.FromSeconds(30);
+        public TimeSpan TipLifespan { get; set; } = TimeSpan.FromSeconds(60);
```
dahlia, line 131: This change of the default value should be recorded in the changelog.

Libplanet/Net/Swarm.BlockSync.cs
```diff
@@ -155,11 +155,11 @@ CancellationToken cancellationToken
                     lastUpdated = DateTimeOffset.UtcNow;
                     lastTip = BlockChain.Tip;
                 }
-                else if (lastUpdated + tipLifespan > DateTimeOffset.UtcNow)
+                else if (lastUpdated + tipLifespan < DateTimeOffset.UtcNow)
```
dahlia, line 158: Would it be difficult to add a regression test for this bug?

dahlia (commented):
IMO it needs to target to a maintenance branch instead of the main branch.

Libplanet/Net/SwarmOptions.cs
greymistcube: Done

Libplanet/Net/Swarm.BlockSync.cs
greymistcube: Well, it's probably possible to do a unit test by elevating the access level to `internal` and directly calling the method, but I'm not sure about a regression test. 😗

dahlia (approved):

PR #1618, opened 2021-11-25
Fix apv query command

ipdae:
- use netmq query instead of tcp


greymistcube:
I think it might be better to make the command to take an additional argument for transport type.

greymistcube (approved):

PR #1636, opened 2021-12-02
Fragmented states on Merkle tree

dahlia:
This patches tries further fragmentation on states.  It addresses the following problems:

- States has been already fragmented on Merkle tree, but their atomic unit was an account.  Therefore, if an action changed only a tiny leaf in a tree, a new copy of the entire tree needed to be added.
- If an action needs only a tiny leaf in an account state tree, the entire tree had to be loaded including unused leafs.

By utilizing Bencodex offloading,[^1] states are now further fragmented into multiple subtrees, which can be offloaded and loaded when needed.  Such fragmentation is recursively applied to nested lists/dictionaries.  However, light values, encoded to 256 bytes[^2] or shorter, are embedded to their parent tree as offloading them is rather overkill.

To optimize shallow but broad dictionaries (for instance, a dictionary with 20+ keys with short strings), dictionary values are chunked into several groups so that each group can be loaded independently.  Each group contains up to 5 values.[^3]

To optimize similar-looking dictionaries (i.e., semi-schemed dictionaries), keys are independently stored from their associated values.  Therefore, `foo`/`bar`/`baz` in `{ foo: 1, bar: 2, baz: 3 }` and `{ foo: 4, bar: 5, baz: 6 }` are stored once.

[^1]: Related pull requests:
      - <https://github.com/planetarium/bencodex.net/pull/50>
      - <https://github.com/planetarium/bencodex.net/pull/52>
      - <https://github.com/planetarium/bencodex.net/pull/53>
      - <https://github.com/planetarium/bencodex.net/pull/55>
      - <https://github.com/planetarium/bencodex.net/pull/56>
[^2]: Hard-coded in `MerkleTrie.OffloadThresholdBytes`.  This can be adjusted in the future without breaking backward compatibility.
[^3]: Hard-coded in `MerkleTrie.OffloadOptions.Offload()`.  This also can be adjusted in the future without breaking backward compatibility.

### DIFF
--- a/Libplanet/Misc/ArrayEqualityComparer.cs
+++ b/Libplanet/Misc/ArrayEqualityComparer.cs
@@ -1,0 +1,99 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+
+namespace Libplanet.Misc
+{
+    /// <summary>
+    /// An <see cref="IEqualityComparer{T}"/> implementation to compare two arrays of the same
+    /// element type.  This compares the elements in the order of the array.
+    /// <para>The way to compare each element can be customized by specifying
+    /// the <see cref="ElementComparer"/>.</para>
+    /// </summary>
+    /// <typeparam name="T">The element type of the array.</typeparam>
+    public class ArrayEqualityComparer<T> : IEqualityComparer<T[]>
+        where T : IEquatable<T>
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="ArrayEqualityComparer{T}"/>.
+        /// </summary>
+        /// <param name="elementComparer">Optionally customize the way to compare each element.
+        /// </param>
+        public ArrayEqualityComparer(IEqualityComparer<T>? elementComparer = null)
+        {
+            ElementComparer = elementComparer;
+        }
+
+        /// <summary>
+        /// Optionally customizes the way to compare each element.
+        /// </summary>
+        public IEqualityComparer<T>? ElementComparer { get; }
+
+        /// <inheritdoc cref="IEqualityComparer{T}.Equals(T, T)"/>
+        public bool Equals(T[]? x, T[]? y)
+        {
+            if (x is null && y is null)
+            {
+                return true;
+            }
+            else if (x is null || y is null)
+            {
+                return false;
+            }
+            else if (x.Length != y.Length)
+            {
+                return false;
+            }
+
+            if (ElementComparer is { } comparer)
+            {
+                for (long i = 0L; i < x.LongLength; i++)
+                {
+                    if (!comparer.Equals(x[i], y[i]))
+                    {
+                        return false;
+                    }
+                }
+            }
+            else
+            {
+                for (long i = 0L; i < x.LongLength; i++)
+                {
+                    if (!x[i].Equals(y[i]))
+                    {
+                        return false;
+                    }
+                }
+            }
+
+            return true;
+        }
+
+        /// <inheritdoc cref="IEqualityComparer{T}.GetHashCode(T)"/>
+        public int GetHashCode(T[]? obj)
+        {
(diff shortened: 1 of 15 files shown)


Libplanet/Store/Trie/Nodes/BaseNode.cs
```diff
@@ -21,8 +13,6 @@ protected BaseNode(INode? value)
         // It will not support embedded node.
         public INode? Value { get; }
 
-        public byte[] Serialize() => _codec.Encode(ToBencodex());
```
riemannulus, line 24: IMO this change needs to describe on CHANGES.md

Libplanet/Misc/ArrayEqualityComparer.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace Libplanet.Misc
+{
+    /// <summary>
+    /// An <see cref="IEqualityComparer{T}"/> implementation to compare two arrays of the same
+    /// element type.  This compares the elements in the order of the array.
+    /// <para>The way to compare each element can be customized by specifying
+    /// the <see cref="ElementComparer"/>.</para>
+    /// </summary>
+    /// <typeparam name="T">The element type of the array.</typeparam>
+    public class ArrayEqualityComparer<T> : IEqualityComparer<T[]>
```
riemannulus, line 14: Why did you build comparer instead of using [`Enumerable.SequenceEqual()`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sequenceequal?view=net-6.0)?

dahlia: Because `ConcurrentDictionary<K, V>` requires an `IEqualityComparer<T>` implementation.  It is used for accurate matching of `byte[]` keys.  (.NET arrays do not implement `IEquatable<T>`.)

Libplanet/Store/Trie/Nodes/BaseNode.cs
dahlia: I omitted it because `BaseNode` is an `internal class`.

Libplanet/Misc/ArrayEqualityComparer.cs
greymistcube: This looks a bit too generic and possibly dangerous. 😶 

Arrays and lists are base enough classes where it is expected that `IEquatable<T>` isn't implemented, i.e. `new int[] { 0 } != new int[] { 0 }`. Such high visibility, access level `public`, and scope, namespace suggesting global utility `class`, makes this prone to misuses.

Case in point, `Commit(INode node, IDictionary<byte[], byte[]> values)` has no idea whether a given `IDictionary` will be using the said comparer. Looking at the code locally, a reader might be easily confused. I don't know the best way mitigate such possible issue, but I can make a few suggestions.

- Lowering visibility and/or access level.
- Wrap `byte[]` in a new class implementing `IEquatable<T>`.
- Wrap `Dictionary<TKey, TValue>` in a new class with default comparer as `ArrayEqualityComparer<T>`.

Or you can just put this on a backlog since it isn't that critical. 🙃

longfin (approved):

Libplanet/Misc/ArrayEqualityComparer.cs
dahlia: I believe this (and any other `IEqualityComparer<T[]>`) is not that prone to misuse, because the interface in itself is not widely used and only used in a narrow scope — mostly for dictionaries. 🤔 

In general, using arrays as dictionary keys are not a good idea.  It was already designed in that way before I kicked this off, and I left it as is due to in the interest of time… 😭 Anyway, I am going to define a distinct type satisfying `IEquatable<T>` and replace arrayed keys with it in the future.

PR #1637, opened 2021-12-02
Remove `int` casting and clutter removal

greymistcube:
Just a small chore.

Unless I'm missing something, I don't think tx shuffling for evaluation is necessary here. 😶


longfin:
Tx shuffling was introduced from https://github.com/planetarium/libplanet/commit/875f484942376cd7b56d8ca1e768677afe7b4b8f.

@dahlia Do you have any idea?

dahlia (approved):

longfin (approved):

PR #1639, opened 2021-12-05
Remove message history & release 0.22.1

longfin:
Same purpose of #1638.


riemannulus:
@limebell comes to next Monday. Should we wait for his review?

riemannulus (approved):

longfin:
> @limebell comes next Monday. Should we wait for his review?

I think... nope. if there is any missing, I'll fix it on 0.22.2.

Issue #1643, opened 2021-12-05
Refactor `IStagePolicy`


greymistcube:
Related to #1583.

On its surface, `IStagePolicy` have two distinct uses:

- For `BlockChain<T>` to retrieve unconfirmed `Transaction<T>`s to mine.
- For `ITransport` to determine whether a `Transaction<T>` is needed to be downloaded by another peer.

These two use cases *do not* play nice with each other. Without any explicit constraint on what the state properties of implemented `IStagePolicy` should be, `IStagePolicy` can hold on to pretty much any `Transaction<T>`. This is too weak of a constraint and is not really useful from `BlockChain<T>`'s or `ITransport`'s perspective. The problem then arises from two different client `class`es trying to use `IStagePolicy` with offloaded implicit logic of their own where the desired properties of `IStagePolicy` for `BlockChain<T>` and `ITransport` are incompatible.

The matter is further complicated when we try to optimize for network traffic while chain reorgs are introduced. For an optimized implementation for the second use case, `IStagePolicy` either requires access to non-volatile storage or have such `Transaction<T>` copied to `IStagePolicy` at an appropriate time, which makes the name `VolatileStagePolicy` even more confusing. 🙃

Issue #1583, opened 2021-11-05
Implement a more defensive stage policy to prevent zombie transactions


greymistcube:
Currently, there is no mechanism in place to retire bad transactions from a staged state for non-mining nodes other than waiting for said transactions to "time out".

This results in unnecessary transaction propagation where non-mining nodes endlessly broadcast bad transactions to each other.

PR #1648, opened 2021-12-07
Default `IStagePolicy` reimplementation

### LINKED ISSUE #1645: Implement `BlockChain<T>.GetTransaction()` with smaller scope
Currently, `BlockChain<T>.GetTransaction()` method is only able to retrieve a `Transaction<T>` with maximal scope, i.e. it searches both in `IStore` and `IStagePolicy` and returns one if found. Although technically possible, there is no (computationally) fast and easy way to check if a given `Transaction<T>` is *actually* inside a `BlockChain<T>` as a part of a `Block<T>`. Aside from a seeming lack of basic functionality, this makes it hard to offload or design a business logic that is state dependent of a `BlockChain<T>`.

greymistcube:
Partially addresses issues #1583 and #1643.

----

This PR reimplements the `VolatileStagePolicy<T>` with simple constraints:

- `VolatileStagePolicy<T>` holds all known unconfirmed `Transaction<T>`s except:
  - `Transaction<T>` is expired due to staleness.
  - `Transaction<T>` is ignored due to policy violation.

In particular, this means for a given `Address`, `Transaction<T>`s with lower nonces than that of the highest nonce included in the `BlockChain<T>` can still be staged. This can happen if an `Address` is responsible for creating `Transaction<T>`s with duplicated nonces.

With such implementation, the downside is that the size of `Messages.TxIds` would generally get longer when broadcasting the entire staged `Transaction<T>`s. However, I would say the upsides outweighs the downside as this makes the handling of `IStagePolicy` vastly simpler and it is a lot easier to guarantee that `Transaction<T>`s do not get lost due to leaky logic.

Trying to optimize/minimize network traffic by suppressing transfers of seemingly unneeded `Transaction<T>` is nice to have, but I suggest we add new constraints when the necessity *actually* arises. I've tried to mathematically model desired properties of an `IStagePolicy` for several days, but to no avail. It looks like optimizing network traffic *whilst* maintaining network consistency *together with* guaranteeing no `Transaction<T>` loss is *a lot harder* than it seems. Either that, or its due to my vast incompetence. 🙃 

----

The interface method signatures of `IStagePolicy<T>` are left intact since I wasn't sure if there would be enough time to track down all uses all the way up to [headless], but all parameters of `BlockChain<T>` should be scrubbed from `IStagePolicy<T>` to send a clear signal that `IStagePolicy<T>` is incapable of using `BlockChain<T>` as its running context.

In my opinion, it'd be better for an implementation of `IStagePolicy<T>` to maintain its state, but the minimum requirement for this to happen would be to resolve #1645 first.

----

**Update**

Since there were some compatibility issues for supporting the [lib9c]'s implementation of `IStagePolicy<T>`, additional API change is implemented and `VolatileStagePolicy<T>`, the default policy, has been re-reimplemented. 😶 

The basic premise of `VolatileStagePolicy<T>` of holding all unconfirmed `Transaction<T>`s still holds as mentioned above with the definition of a staged `Transaction<T>` is simply being any known `Transaction<T>` that is not included in the current `BlockChain<T>`. However, the default behavior of `Iterate()` has been "reverted" to match what was "intended" as before.

[lib9c]: https://github.com/planetarium/lib9c
[headless]: https://github.com/planetarium/NineChronicles.Headless


dahlia (commented):
Could you rebase it on the latest main?

dahlia:
Also it would be better if you rename the title so that we can easily recognize what this pull request is for in the future.

Libplanet.Tests/Blockchain/BlockChainTest.Append.cs
```diff
@@ -537,7 +539,7 @@ public void UnstageAfterAppendComplete()
                 blockInterval: TimeSpan.FromSeconds(10)
             ).Evaluate(privateKey, _blockChain);
             _blockChain.Append(block3);
-            Assert.Empty(_blockChain.GetStagedTransactionIds());
+            Assert.Equal(1, _blockChain.GetStagedTransactionIds().Count);
```
riemannulus, line 542: It seems different logic... is it a bug before?

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
```diff
@@ -36,184 +36,203 @@ public VolatileStagePolicy()
...
-                                TimestampFormat, CultureInfo.InvariantCulture),
-                            DateTimeOffset.UtcNow.ToString(
-                                TimestampFormat, CultureInfo.InvariantCulture));
-                }
-                finally
-                {
-                    _lock.ExitWriteLock();
+                    try
+                    {
+                        _lock.EnterWriteLock();
```
riemannulus, line 78: According to [MSDN ReadWriteLockSlim](https://docs.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim?redirectedfrom=MSDN&view=net-6.0) example, `ReadWriteLockSlim.EnterWriteLock()` and all kind of lock should be call on outside of `try` block. If you worried about throw error from `ReadWriteLockSlim.EnterWriteLock()`, it is ok. ([ref](https://stackoverflow.com/a/24533469))

Libplanet.Tests/Blockchain/BlockChainTest.Append.cs
greymistcube: This is due to changed behavior of `VolatileStagePolicy`.

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
greymistcube: Thanks, now that I have given more thought to it and going over your reference, it *should* be outside `try`. I'll fix it right away.

Libplanet.Tests/Blockchain/BlockChainTest.Append.cs
riemannulus: Is it ok? isn't it affected nine chronicles mainnet?

greymistcube: From a miner's perspective, it should be fine as the changed implementation holds a superset of `Transaction<T>`s held by the previous implementation. `BlockChain<T>.GatherTransactionsToMine()` actually does the heavy lifting of filtering what can be mined.

As from a network's perspective, we don't have no known measured metric to compare whether this would increase or decrease the network's traffic. My gut feeling is that it would decrease the overall traffic.

I do need to make sure this plays nice with [lib9c]`s `StagePolicy` though and I'm currently working on it.

[lib9c]: https://github.com/planetarium/lib9c

greymistcube:
Updated with an overhauled API to support https://github.com/planetarium/lib9c/pull/685.

dahlia (approved):

PR #1650, opened 2021-12-08
Build Libplanet.Tools executables for linux-arm64 & win-arm64

dahlia:
Actually, I started to work on this to build `planet` executable for osx-arm64 (M1 series), but I found that .NET 6 does not support cross-compilation targeting osx-arm64 from other architectures yet.  So I added supports only for linux-arm64 and win-arm64…


riemannulus (dismissed):

dahlia:
Apparently there is an issue on build… I don't merge it right now.

dahlia:
I rebased this and fixed the build. Please take a look again!

longfin (approved):

riemannulus (approved):

longfin (changes requested):
It seems to have conflicts.

Issue #1632, opened 2021-12-01
Reduce log clutter by getting rid of duplicate `Exception.Message`s


greymistcube:
Default output format via `Serilog` is to have `{exception}` at the end of log messages. This is to automatically output a message from a thrown `Exepction` when using `Error()`. However, currently, many, if not most, of `Serilog`'s `Error()` in the code base explicitly include `Exception.Message` in `message` parameter. The result is unnecessary cluttering of the log output with duplicated `Exception.Message`.

dahlia:
Does Serilog's default behavior maintain stack traces as well?  If so, we'd better not to pass exceptions by hand.

greymistcube:
As far as I know, it does. In any case,

```
Log.Error(e, "Some exception happened");
```

Automatically outputs the exception via the default setting of `outputTemplate` in `appsettings.json` as seen in the format

```
[{Timestamp:HH:mm:ss} {Level:u3}] {Message:lj}{NewLine}{Exception}
```

where exception's `e.Message` is printed.

What

```
Log.Error(e, "Some exception happened {E}", e);
```

does is that not only does it duplicate the printing of `e.Message` for `{E}` and `{Exception}` on console, but also duplicates the property in `json` with `E` property and `Exception` property.

greymistcube:
I haven't tested for unit tests since it might not be using the provided template, since `appsettings.json` is in [headless], but in that case, in my opinion, the right way to resolve the issue is configure the logger separately for unit testing instead of abusing the library by using an unnecessary extra property.

[headless]: https://github.com/planetarium/NineChronicles.Headless

greymistcube:
**Update/Errata**: It's actually `Exception.ToString()` which contains the stack trace that gets printed not `Exception.Message`.

PR #1661, opened 2021-12-13
Prevent broadcasting invalid transactions

limebell:
Previously `TxCompletion` used to re-broadcast received all transactions even it is invalid so is not staged. It had lead to some unnecessary messages as follows:
```
           TxIds            GetTxs
         ┌-------> SwarmB ---------┐
SwarmA ----------> SwarmC ------------> SwarmA (no request since the invalid transaction does not exists)
         └-------> ...    ---------┘
```

Adding a regression test is skipped because there are no event handler such as `TxReceived` for the `TxId`s. But I've checked when receiving invalid transaction it broadcasts `TxId` and work as the diagram above.

### DIFF
--- a/Libplanet/Net/TxCompletion.cs
+++ b/Libplanet/Net/TxCompletion.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


Libplanet/Net/TxCompletion.cs
```diff
@@ -149,31 +150,42 @@ private HashSet<TxId> GetRequiredTxIds(IEnumerable<TxId> ids)
...
                         }
                         catch (InvalidTxException ite)
                         {
                             _logger.Error(
                                 ite,
-                                "Transaction {TxId} will not be staged since it is invalid.",
-                                tx.Id);
+                                "Transaction {TxId} will not be staged since it is invalid. {E}",
+                                tx.Id,
+                                ite);
```
dahlia, line 171: It might be better not to pass `ite` here.  See also <https://github.com/planetarium/libplanet/issues/1632>.

greymistcube:
This needs a rebase. 😶

longfin (approved):

PR #1672, opened 2021-12-16
Prevent premature unstaging

greymistcube:
When working with a forked chain, `Append()` on a forked chain prematurely unstages transactions in the block being appended. This results in invalid nonce calculation when trying to create a new transaction while a forked chain is being worked on. Note unstaging and restaging of relevant transactions is dealt with in `Swap()` when syncing the canonical chain to a forked chain anyway.

This PR prevents premature unstaging of transactions by checking whether its calling instance is the canonical chain or not.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,17 @@ Version 0.24.1
 
 To be released.
 
+### Added APIs
+
+ -  Added `BlockChain<T>.IsCanonical` property.  [[#1672]]
+
+### Bug fixes
+
+ -  Fixed a bug where `Transaction<T>`s were unstaged when they were not
+    supposed to.  [[#1672]]
+
+[#1672]: https://github.com/planetarium/libplanet/pull/1672
+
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -6,6 +6,17 @@ Version 0.24.1
 
 To be released.
 
+### Added APIs
+
+ -  Added `BlockChain<T>.IsCanonical` property.  [[#1672]]
```
dahlia, line 11: Patch releases cannot change APIs.  (That's why patch release notes do not have sections.)  Could you make it `internal` in this patch?

greymistcube: Hmm... wasn't sure since it wasn't breaking. I suppose I can make it internal for now.

greymistcube: Done.

dahlia (approved):

longfin (approved):

PR #1676, opened 2021-12-16
Improve performance of `BlockChain<T>.FindNextHashes()`

longfin:
This PR improves performance of `BlockChain<T>.FindNextHashes()` through two bellow adjustments.

1. Removed `_rwlock` from `BlockChain<T>.FindNextHashes()` because it has already protected by another lock on `BlockChain<T>.FindBranchpoint()`.
2. Added a memory cache for `RocksDBStore.IterateIndexes()` to speed up iterative lookups.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1040,66 +1040,60 @@ namespace Libplanet.Blockchain
             BlockHash? stop = null,
             int count = 500)
         {
-            try
+            DateTimeOffset startTime = DateTimeOffset.Now;
+
+            // FIXME Theatrically, we don't accept empty chain. so `tip` can't be null on this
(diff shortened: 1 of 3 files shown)


riemannulus (approved):

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1040,66 +1040,60 @@ public IImmutableSet<TxId> GetStagedTransactionIds()
             BlockHash? stop = null,
             int count = 500)
         {
-            try
+            DateTimeOffset startTime = DateTimeOffset.Now;
+
+            // FIXME Theatrically, we don't accept empty chain. so `tip` can't be null on this
```
greymistcube, line 1045:
```suggestion
            // FIXME Theoretically, we don't accept empty chain. so `tip` can't be null on this
```

greymistcube (approved):

limebell (approved):

PR #1678, opened 2021-12-17
Add multiget operation to `IKeyValueStore`

dahlia:
For later optimizations, this patch adds a new operation *multiget* to `IKeyValueStore`.  Further optimizations would be landed as separated patches later.

Plus, I addressed https://github.com/planetarium/libplanet/pull/1636#discussion_r764478088 that @greymistcube made in the previous pull request: now `IKeyValueStore` represents keys as `KeyBytes`, a thin wrapper around `ImmutableArray<byte>`, instead of `byte[]`.  `KeyBytes` is a readonly struct, and implements `IEquatable<KeyBytes>` & `GetHashCode()` method.  Human-readable `.ToString()` representation is a bonus.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ To be released.
         `Exists(in KeyBytes)` method.
      -  `IKeyValueStore.ListKeys()` method's return type became
         `IEnumerable<KeyBytes>` (was `IEnumerable<byte[]>`).
+ -  Added `IKeyValueStore.Get(IEnumerable<KeyBytes>)` method.  [[#1678]]
 
 ### Backward-incompatible network protocol changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@ To be released.
 
 ### Added APIs
 
+ -  Added `KeyBytes` readonly struct.  [[#1678]]
+
 ### Behavioral changes
 
   -  Improved performance of `Swarm<T>`'s block propagation.  [[#1676]]
@@ -27,6 +29,8 @@ To be released.
 ### Dependencies
 
 ### CLI tools
+
+[#1678]: https://github.com/planetarium/libplanet/pull/1678
 
 
 [#1676]: https://github.com/planetarium/libplanet/pull/1676
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,7 @@ To be released.
      -  `IKeyValueStore.ListKeys()` method's return type became
         `IEnumerable<KeyBytes>` (was `IEnumerable<byte[]>`).
  -  Added `IKeyValueStore.Get(IEnumerable<KeyBytes>)` method.  [[#1678]]
+ -  Added `IKeyValueStore.Delete(IEnumerable<KeyBytes>)` method.  [[#1678]]
 
 ### Backward-incompatible network protocol changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,22 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  Replaced `HashDigest<T>(ImmutableArray<byte>)` constructor with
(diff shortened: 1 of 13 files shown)


Libplanet/Store/Trie/KeyBytes.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <exception cref="FormatException">Thrown when the given <paramref name="hex"/> string is
+        /// not a valid hexadecimal string.</exception>
+        public static KeyBytes FromHex(string hex) =>
+            new KeyBytes(ByteUtil.ParseHex(hex));
+
+        /// <summary>
+        /// Converts to a mutable byte array.
+        /// </summary>
+        /// <returns>A new copy of mutable byte array.</returns>
+        public byte[] ToByteArray() => ByteArray.IsDefault
```
greymistcube, line 90: I think `ToByteArray()` together with `ByteArray` property above might be a bit too confusing. 😶

dahlia: It might be confusing, but is also consistent with existing byte-wrapping value types: `Address`, `TxId`, `HashDigest<T>`, `BlockHash`, etc.  I agree such naming is misleading (it might be better to rename it to `ToArray()`), but I don't want to expand the scope of this patch.  Could you make a separate issue?

CHANGES.md
```diff
@@ -10,12 +10,32 @@ To be released.
...
 
+ -  Replaced `HashDigest<T>(ImmutableArray<byte>)` constructor with
+    `HashDigest<T>(in ImmutableArray<byte>)` constructor.  [[#1678]]
+ -  `IKeyValueStore`'s key type became `KeyBytes` (was `byte[]`).  [[#1678]]
+     -  Replaced `IKeyValueStore.Get(byte[])` method with `Get(in KeyBytes)`
+        method.
+     -  Replaced `IKeyValueStore.Set(byte[], byte[])` method with
+        `Set(in KeyBytes, byte[])` method.
+     -  Replaced `IKeyValueStore.Set(IDictionary<byte[], byte[]>)` method with
+        `Set(IDictionary<BytesKey, byte[]>)` method.
```
greymistcube, line 21:
```suggestion
        `Set(IDictionary<KeyBytes, byte[]>)` method.
```
Seems like a typo.

dahlia: Amended!

Libplanet/Store/Trie/KeyBytes.cs
greymistcube: Ah, I see. The scope of the issue is larger than I thought. Fair enough.

Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
```diff
@@ -27,34 +29,52 @@ public RocksDBKeyValueStore(string path)
...
-                writeBatch.Put(kv.Key, kv.Value);
+                writeBatch.Put(kv.Key.ToByteArray(), kv.Value);
             }
 
             _keyValueDb.Write(writeBatch);
         }
 
         /// <inheritdoc/>
-        public void Delete(byte[] key)
+        public void Delete(in KeyBytes key)
```
greymistcube, line 72: Just a thought. Might be nice to have multi-key-delete just like `Get` and `Set`? 😶

dahlia: Good idea.  I'm going to give it a try!

dahlia: Added `IKeyValueStore.Delete(IEnumerable<KeyBytes>)` method: 43526ca3b72fe72933aaf660e628b5377e3fc211.

dahlia:
Rebased.  Please take a look again.

riemannulus (approved):

PR #1679, opened 2021-12-20
Nullable context for `Peer` and `BoundPeer`

greymistcube:
Minor QoL improvement.

### DIFF
--- a/Libplanet.Tests/Net/PeerTest.cs
+++ b/Libplanet.Tests/Net/PeerTest.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.IO;
+using System.Net;
 using System.Runtime.Serialization.Formatters.Binary;
 using Libplanet.Crypto;
 using Libplanet.Net;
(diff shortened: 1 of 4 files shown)


dahlia:
@limebell or @longfin Could you take a look on this?

Libplanet.Tests/Net/PeerTest.cs
```diff
@@ -9,30 +10,12 @@ namespace Libplanet.Tests.Net
...
-                        0x04, 0xb5, 0xa2, 0x4a, 0xa2, 0x11, 0x27, 0x20, 0x42, 0x3b,
-                        0xad, 0x39, 0xa0, 0x20, 0x51, 0x82, 0x37, 0x9d, 0x6f, 0x2b,
-                        0x33, 0xe3, 0x48, 0x7c, 0x9a, 0xb6, 0xcc, 0x8f, 0xc4, 0x96,
-                        0xf8, 0xa5, 0x48, 0x34, 0x40, 0xef, 0xbb, 0xef, 0x06, 0x57,
-                        0xac, 0x2e, 0xf6, 0xc6, 0xee, 0x05, 0xdb, 0x06, 0xa9, 0x45,
-                        0x32, 0xfd, 0xa7, 0xdd, 0xc4, 0x4a, 0x16, 0x95, 0xe5, 0xce,
-                        0x1a, 0x3d, 0x3c, 0x76, 0xdb,
-                    })),
+                new object[] { new Peer(new PrivateKey().PublicKey) },
+                new object[] { new Peer(new PrivateKey().PublicKey, IPAddress.Parse("1.2.3.4")) },
```
limebell, line 17: The public key is hard-coded because some portion of `RoutingTable`'s portion works depends on `Peer.Address`. Even though tests depends on addresses passes now, it may cause some problem later. So I suggest to leave it as it was.

greymistcube: I understand the concern, but can you elaborate a bit more on how this may be problematic down the road? As I understand it, yes, when testing a specific behavior of `RoutingTable`, a usage of hard coded `byte`s is granted, but for this specific test I don't think it's that critical. In my opinion, the general pros and cons for [monkey testing] applies here and the issue isn't that clear cut as to whether hard coded data would be better or not.

[monkey testing]: https://en.wikipedia.org/wiki/Monkey_testing

limebell: Oh I see. It seems it is not very important for the tests which uses the member.

limebell (approved):

PR #1680, opened 2021-12-21
Remove redundant memcpy/malloc from `HashDigest<T>.DeriveFrom()`

dahlia:
This patch adds `HashDigest<T>.DeriveFrom()` overloads to remove redundant memcpy/malloc.

In order to move the pointer to the buffer (which is never changed) to `ImmutableArray<byte>` without memcpy, I used [`Unsafe.As<TFrom, TTo>()` method][1][^1] several times.[^2]  I would be able to implement it better if [there were `ImmutableArray.UnsafeFreeze<T>(T[])` method.][4]

[^1]: It parallels C++'s `std::reinterpret_cast<T>()`.
[^2]: Note that [`T[]` and `ImmutableArray<T>` share the same memory layout.][2]  See also [this article][3].

[1]: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe.as?view=net-6.0#System_Runtime_CompilerServices_Unsafe_As__2___0__
[2]: https://github.com/dotnet/runtime/blob/v6.0.1/src/libraries/System.Threading.Tasks.Dataflow/src/Internal/ImmutableArray.cs#L32
[3]: https://devblogs.microsoft.com/dotnet/please-welcome-immutablearrayt/
[4]: https://github.com/dotnet/runtime/issues/25461

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@ To be released.
 
  -  Replaced `HashDigest<T>(ImmutableArray<byte>)` constructor with
     `HashDigest<T>(in ImmutableArray<byte>)` constructor.  [[#1678]]
+ -  Replaced `HashDigest<T>.DeriveFrom(ReadOnlySpan<byte>)` overloaded static
+    method with ohter new overloads.  [[#1680]]
  -  `IKeyValueStore`'s key type became `KeyBytes` (was `byte[]`).  [[#1678]]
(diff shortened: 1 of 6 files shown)


dahlia:
@planetarium/libplanet Please take a look!

riemannulus (approved):

Issue #1653, opened 2021-12-10
Fix `TrieStateStore.PruneStates()` to work together with fragmented states


dahlia:
Since fragmented states were introduced:

- <https://github.com/planetarium/libplanet/pull/1636>
   - <https://github.com/planetarium/bencodex.net/pull/50>
   - <https://github.com/planetarium/bencodex.net/pull/52>
   - <https://github.com/planetarium/bencodex.net/pull/53>
   - <https://github.com/planetarium/bencodex.net/pull/55>
   - <https://github.com/planetarium/bencodex.net/pull/56>

`TrieStateStore.PruneStates()` method has not worked well.  Note that there's even a comment addressing this:

<https://github.com/planetarium/libplanet/blob/bf2a321af65cd70dda9243e6a8e888a217223928/Libplanet/Store/TrieStateStore.cs#L41-L43>

There are two problems:

- [x] It throws `ArgumentOutOfRangeException` because now `IStateStore` has keys other than SHA-256 digest.
   - <https://github.com/planetarium/NineChronicles.Snapshot/issues/78>
- [ ] Reference tracking works only at MPT-level.  References between fragmented states also should be tracked.

The former one should be published as a hotfix.  The latter one would accompany API changes, so it should be published in the next minor release.

PR #1689, opened 2022-01-04
Represent `ITrie` keys using `KeyBytes`

dahlia:
A small refactoring as a preparation for fixing #1653.  @moreal You should work on your job after this patch is merged.


riemannulus (approved):
Because it looks like chore things, I checked if typos exist only.

moreal (approved):
Good fixup 👍🏻 🔨

PR #1690, opened 2022-01-04
Upgrade explorer GraphQL from 3.0.0 to 4.7.1

area363:
## Changes
* Bump GQL to 4.7.1
* Fix test to fit the new version


Libplanet.Explorer.UnitTests/GraphTypes/BlockTypeTest.cs
```diff
@@ -53,28 +54,32 @@ public async void Query()
...
-            Assert.Equal(
-                ByteUtil.Hex(block.Nonce.ToByteArray()),
-                resultData["nonce"]);
-            Assert.Equal(
-                new DateTimeOffsetGraphType().Serialize(block.Timestamp),
-                resultData["timestamp"]);
-            Assert.Equal(
-                ByteUtil.Hex(block.StateRootHash.ToByteArray()),
-                resultData["stateRootHash"]);
+            if (resultData != null)
```
dahlia, line 60: In what cases `resultData` is not present?

Libplanet.Explorer.UnitTests/GraphTypes/ByteStringTypeTest.cs
```diff
@@ -30,7 +30,9 @@ public void ParseValue(object value, object parsed)
         [InlineData("beef", new byte[] { 0xbe, 0xef })]
         public void ParseLiteral(string stringValue, object parsed)
         {
-            Assert.Equal(parsed, _type.ParseLiteral(new StringValue(stringValue)));
+            var actual =
+                stringValue != null ? _type.ParseLiteral(new StringValue(stringValue)) : null;
```
dahlia, line 34:
```suggestion
                stringValue is { } v ? _type.ParseLiteral(new StringValue(v)) : null;
```

Libplanet.Explorer.UnitTests/GraphTypes/BlockTypeTest.cs
area363: I checked again just now and it appears that there isn't a case where the `resultData` is null. I might have included this code before fixing another API. I'll remove this part along with https://github.com/planetarium/libplanet/pull/1690/files#diff-aeb0e1c689f74d1eefa3a1ca8914edc2c4c379114561a8011f9b52e610d5b5a1R45-R46. Thanks for checking. 🙏

dahlia: Note that there is another similar pattern in `TransactionTypeTest.Query()` method as well.

dahlia (approved):

dahlia (changes requested):
Wait, the build apparently failed.  I'm going to take a look.

dahlia (approved):
It was due to no luck.  Rerun fixed the build.

PR #1274, opened 2021-04-29
Fix/transactions immutable

greymistcube:
Changed property of `Transactions` of `Block` to `IReadOnlyList<T>` from `IEnumerable<T>` to partially mitigate [the multiple enumeration problem](https://www.jetbrains.com/help/resharper/PossibleMultipleEnumeration.html).

Constructor now also receives `IReadOnlyList<T>` to further enforce this point.

PR #1702, opened 2022-01-10
Funnel sending `Message`s to `SendMessageAsync()` when broadcasting for `NetMQTransport`

greymistcube:
Internal behavior for broadcasting while using `NetMQTransport` changed to use `SendMessageAsync()`.

### DIFF
--- a/Libplanet/Net/Transports/NetMQTransport.cs
+++ b/Libplanet/Net/Transports/NetMQTransport.cs
@@ -689,19 +689,25 @@ namespace Libplanet.Net.Transports
                     .Select(peer =>
                         SendMessageAsync(peer, message, _runtimeCancellationTokenSource.Token))
                     .ToList();
-                Task aggregateTask = Task.WhenAll(tasks);
-                try
-                {
-                    aggregateTask.GetAwaiter().GetResult();
-                }
-                catch (Exception)
-                {
-                    AggregateException aggregateException = aggregateTask.Exception!;
-                    _logger.Warning(
-                        aggregateException,
-                        "Failed to send {MessageType} to some peers while broadcasting.",
-                        message.Type);
-                }
+                Task aggregateTask = Task
+                    .WhenAll(tasks)
+                    .ContinueWith(task =>
+                        {
+                            if (task.IsFaulted)
+                            {
+                                AggregateException aggregateException = task.Exception;
+                                _logger.Warning(
+                                    aggregateException,
+                                    "Failed to send {Message} to some peers while broadcasting.",
+                                    message);
+                            }
+                            else if (task.IsCanceled)
+                            {
+                                _logger.Debug(
+                                    "{FName}() was canceled during its operation.",
+                                    nameof(DoBroadcast));
+                            }
+                        });
             }
             catch (Exception exc)
             {
@@ -709,7 +715,6 @@ namespace Libplanet.Net.Transports
                     exc,
                     "Unexpected error occurred during {FName}().",
                     nameof(DoBroadcast));
-                throw;
             }
         }
 
--- a/Libplanet/Net/Transports/NetMQTransport.cs
+++ b/Libplanet/Net/Transports/NetMQTransport.cs
@@ -34,7 +34,7 @@ namespace Libplanet.Net.Transports
         private readonly int _minimumBroadcastTarget;
         private readonly NetMQMessageCodec _messageCodec;
 
-        private NetMQQueue<NetMQMessage> _replyQueue;
+        private NetMQQueue<Message> _replyQueue;
         private NetMQQueue<(Address?, Message)> _broadcastQueue;
 
         private RouterSocket _router;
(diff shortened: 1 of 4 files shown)


Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -513,7 +501,7 @@ public async Task ReplyMessageAsync(Message message, CancellationToken cancellat
             }
 
             string identityHex = ByteUtil.Hex(message.Identity);
-            var tcs = new TaskCompletionSource<object>();
+            TaskCompletionSource<object> tcs = new TaskCompletionSource<object>();
```
limebell, line 504:
```suggestion
            var<object> tcs = new TaskCompletionSource<object>();
```

Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -639,7 +627,7 @@ private void ReceiveMessage(object sender, NetMQSocketEventArgs e)
             }
             catch (DifferentAppProtocolVersionException dapve)
             {
-                var differentVersion = new DifferentVersion()
+                DifferentVersion differentVersion = new DifferentVersion()
```
limebell, line 630:
```suggestion
                var differentVersion = new DifferentVersion()
```

Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -158,7 +157,7 @@ static NetMQTransport()
                     try
                     {
                         using var runtime = new NetMQRuntime();
-                        var workerTasks = new Task[workers];
+                        Task[] workerTasks = new Task[workers];
```
limebell, line 160:
```suggestion
                        var workerTasks = new Task[workers];
```

Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -813,17 +776,17 @@ await foreach (MessageRequest req in _requests.Reader.ReadAllAsync(cancellationT
...
-                "A request {RequestId} is ready to be processed in {TimeSpan}: {@Message}.",
+                "Request {RequestId} is ready to be processed in {TimeSpan}: {@Message}.",
                 req.Id,
                 DateTimeOffset.UtcNow - req.RequestedTime,
                 req.Message
             );
             DateTimeOffset startedTime = DateTimeOffset.UtcNow;
 
-            using var dealer = new DealerSocket(req.Peer.ToNetMQAddress());
+            using DealerSocket dealer = new DealerSocket(req.Peer.ToNetMQAddress());
```
limebell, line 786:
```suggestion
            using var dealer = new DealerSocket(req.Peer.ToNetMQAddress());
```

Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -834,7 +797,7 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
                 AsPeer,
                 DateTimeOffset.UtcNow,
                 _appProtocolVersion);
-            var result = new List<Message>();
+            List<Message> result = new List<Message>();
```
limebell, line 800:
```suggestion
            var result = new List<Message>();
```

limebell (commented):
Used to reuse dealer sockets in broadcast because so many dealer sockets were created and disposed if we just use logic of `SendMessageAsync` itself. It caused problems in real-product which have many nodes to broadcast, so we need to check whether it is okay or not.

Libplanet/Net/Transports/NetMQTransport.cs
greymistcube: Went with using `Linq` instead. 🙃

Libplanet/Net/Transports/NetMQTransport.cs
greymistcube: Removed while refactoring. 🙃

greymistcube:
**Update**

As per suggestion by @limebell, reusing of existing `DealerSocket`s is re-introduced with even more aggressive strategy. 😝 
Reusing dealer sockets should be bit more thread safe now with better usage of `ConcurrentDictionary`.

limebell (commented):
Looks good to me, but IMO scale test is required before merging.

Libplanet/Net/Transports/NetMQTransport.cs
greymistcube: Done.

Libplanet/Net/Transports/NetMQTransport.cs
greymistcube: Done.

greymistcube:
As per #1274, had to roll back reusing `DealerSocket`s. 😕
However, this seems to "almost" work fine on the mainnet, as I haven't encountered a problem with having too many `DealerSocket`s specifically.

greymistcube:
Trying to break up this PR.

PR #1704, opened 2022-01-11
Change the way how `TxCompletion<TPeer, TAction>` fetch transaction from other peers

limebell:
Previously `TxCompletion<TPeer, TAction>.RequestTxsFromPeerAsync()` is maintained maximum 1 task for each peer to fetch transactions from. It caused a bug where fetching does not working even received a new `TxId`s from other peer.

This PR force to spawn `TxCompletion<TPeer, TAction>.RequestTxsFromPeerAsync()` task every time new `TxId`s are received.

### DIFF
--- a/Libplanet/Net/TxCompletion.cs
+++ b/Libplanet/Net/TxCompletion.cs
@@ -1,8 +1,8 @@
 #nullable enable
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -18,8 +18,6 @@ namespace Libplanet.Net
     where TPeer : notnull
     where TAction : IAction, new()
     {
-        private readonly ConcurrentDictionary<TPeer, Task> _txSyncTasks;
-        private readonly ConcurrentDictionary<TPeer, ConcurrentBag<TxId>> _requiredTxIds;
         private readonly CancellationTokenSource _cancellationTokenSource;
         private readonly BlockChain<TAction> _blockChain;
         private readonly TxFetcher _txFetcher;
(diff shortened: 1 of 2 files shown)


Libplanet/Net/TxCompletion.cs
```diff
@@ -111,95 +99,123 @@ private HashSet<TxId> GetRequiredTxIds(IEnumerable<TxId> ids)
...
+                                "{TxId} will not be staged " +
+                                "since it does not follow policy.",
+                                peer,
                                 tx.Id);
+                            _blockChain.StagePolicy.Ignore(_blockChain, tx.Id);
+                            return false;
+                        }));
+
+                var validTxs = new List<Transaction<TAction>>();
+                IImmutableSet<TxId> stagedTxIds = _blockChain.GetStagedTransactionIds();
```
greymistcube, line 159: `BlockChain<T>.StageTransaction()` should do the job of filtering so no additional filtering is necessary at this point. See #1648.

Libplanet/Net/TxCompletion.cs
```diff
@@ -111,95 +99,123 @@ private HashSet<TxId> GetRequiredTxIds(IEnumerable<TxId> ids)
...
             TPeer peer,
+            HashSet<TxId> txIds,
             CancellationToken cancellationToken)
         {
-            while (_requiredTxIds[peer].Any())
+            try
             {
+                const string log =
+                    "Starting loop of RequestTxsFromPeerAsync from {Peer}. " +
+                    "(_requiredTxIds count: {Count})";
```
greymistcube, line 109: This isn't a loop anymore. 🙃

Libplanet/Net/TxCompletion.cs
greymistcube: Eh, then again, we can't count `validTxs` if we don't do the filtering here. I'll create an issue for this.

limebell: So you mean L158 to L176 is not required?

greymistcube: I'll deal with this in #1705.

greymistcube: `IStagePolicy` is designed not to throw any `Exception` so yes, I think it should be fine.

greymistcube: But yeah, since broadcasting is tied to `validTxs`, I think it should be left as it is for now.

Libplanet/Net/TxCompletion.cs
```diff
@@ -78,26 +74,18 @@ public void Demand(TPeer peer, IEnumerable<TxId> txIds)
...
-                _logger.Debug("No unaware transactions to receive.");
+                _logger.Debug(
+                    "No unaware transactions to receive.");
                 return;
             }
 
             _logger.Debug(
                 "Unaware transactions to receive: {TxIdCount}.",
-                required.Count
+                peer
```
greymistcube, line 84:
```suggestion
                required.Count
```

longfin (approved):

greymistcube (changes requested):
I think suggested logging related changes should suffice for now.

greymistcube (approved):

PR #1710, opened 2022-01-11
`BlockChain<T>.StageTransaction()` returns `bool`

### LINKED ISSUE #1705: Change return type from `void` to `bool` for staging methods.
Logic for staging `Transaction<T>`s should be contained within `BlockChain<T>.StageTransaction()` and `IStagePolicy<T>.Stage()`. Currently for logging how many `Transaction<T>`s get staged, staging logic is leaked outside of its scope. Interface should be updated to give easier access to whether staging attempt was successful or not.

greymistcube:
Closes #1705.


limebell (approved):
👍

PR #1712, opened 2022-01-12
Nullable context for IceServer

greymistcube:
🧹

### DIFF
--- a/Libplanet/Net/IceServer.cs
+++ b/Libplanet/Net/IceServer.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


Libplanet/Net/IceServer.cs
```diff
@@ -42,30 +43,27 @@ public class IceServer
...
                     {
-                        throw new ArgumentException($"{url} isn't valid TURN url.");
+                        throw new ArgumentException($"{url} is not a valid TURN url.");
                     }
 
                     int port = url.IsDefaultPort
                         ? TurnClient.TurnDefaultPort
                         : url.Port;
-                    var turnClient = new TurnClient(
+                    TurnClient turnClient = new TurnClient(
```
dahlia, line 52: It's better to use `var` here as the type is apparent.  See also: <https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#implicitly-typed-local-variables>.

greymistcube: Eh, I'm still on the fence about it, but the argument is fair enough. Fixed; see e7d5dfdab0e30cb9d4ea9e06edd972e1641ad705.

dahlia (approved):

PR #1719, opened 2022-01-14
Fix `RocksDBStore`'s inaccurate lock handling

ipdae:
I found this when testing for https://github.com/planetarium/NineChronicles.Headless/issues/1004.
I found similar code in `PutBlock, DeleteBlock` and these methods were locking `_blockDbCache`, so i think `GetBlockDigest` is holding the wrong lock.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -114,8 +114,8 @@ Released on January 5, 2022.
 ### Behavioral changes
 
   -  Improved performance of `Swarm<T>`'s block propagation.  [[#1676]]
-  -  Improved performance of `RocksDBStore<T>.IterateIndexes()` method.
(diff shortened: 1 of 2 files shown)


longfin:
@ipdae Could you rebase onto `main` instead of `0.25-maintenance`? (because we're going to release 0.26 tomorrow)

dahlia:
> Could you rebase onto `main` instead of `0.25-maintenance`? (because we're going to release 0.26 tomorrow)

@longfin I reverted its target branch to 0.25-maintenance again, because there are already unreleased changed in 0.25-maintenance.  I'm going to make a new patch release (0.25.4) first, and port those changes to main (0.26.x).

PR #1720, opened 2022-01-15
Update Discord links

dahlia:
We recently moved our Discord server which is separated from the existing one, as it's virtually dedicated to Nine Chronicles nowadays.


area363 (approved):
👍

moreal (approved):

PR #1743, opened 2022-01-21
Parallelize MPT path finding

dahlia:
This allows multiple states to be loaded at a time through `IKeyValueStore.Get(IReadOnlyList<KeyBytes>)` method.  In order to parallelize path finding, I replaced recursion-based path finder with explicit loops.  See also `MerkleTrie.ResolvePath()` method and `MerkleTrie.Get()` method.

### DIFF
--- a/Libplanet/Store/Trie/MerkleTrie.Path.cs
+++ b/Libplanet/Store/Trie/MerkleTrie.Path.cs
@@ -160,8 +160,8 @@ namespace Libplanet.Store.Trie
 
             // FIXME: We should declare a custom value type to represent nibbles...
             [Pure]
-            private static byte GetNibble(byte word, bool high) =>
-                high ? (byte)(word >> 4) : (byte)(word & 0b00001111);
+            private static byte GetNibble(byte octet, bool high) =>
+                high ? (byte)(octet >> 4) : (byte)(octet & 0b00001111);
         }
 
         private readonly struct PathResolution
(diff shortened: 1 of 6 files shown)


Libplanet/Store/Trie/MerkleTrie.Path.cs
```diff
@@ -0,0 +1,195 @@
...
+                        return i;
+                    }
+                }
+
+                return nibbles.Length;
+            }
+
+            // FIXME: We should declare a custom value type to represent nibbles...
+            [Pure]
+            private static byte GetNibble(byte word, bool high) =>
```
moreal, line 163: This may be not important so I'll approve this pull request, but I want to comment about the `word` word.
In my thought, It looks good about the `nibble` word usage.
Then, in my memory, the `word` unit can be different among multiple architectures. So how about using `octet` or `@byte`?
While writing this comment, I also thought that the `word` unit may mean 8-bit in .NET runtime. 🤔 

Originally I leave this comment to suggest but it seems like a question. 😅 How do you think about it?

moreal (approved):

Libplanet/Store/Trie/MerkleTrie.Path.cs
dahlia: As you said, either *word* or *byte* can be shorter or longer than 8 bits in theory,  there virtually no more exist modern architectures using 7-bit byte or 16-bit word though.  Why I choose the term *word* here was just for avoiding the language keyword `byte`, but *octet` would be indeed the most appropriate here.  I'm going to rename it!

dahlia: @moreal Changed in 30cea86db8044731337caec3046179b65bb1040a.  Please take a look again!

moreal (approved):
LGTM 🌟

dahlia:
@planetarium/libplanet Please take a look!

longfin (approved):

PR #1744, opened 2022-01-21
Remove `RoutingTable` dependency from `ITransport`

### LINKED ISSUE #1741: Remove `RoutingTable` dependency from `ITransport`
`RoutingTable` is mainly used for determining candidates of peers to broadcast message. It will be better to take a list of peers to broadcast.

limebell:
Inner behavior for disposing unused dealer socket has been changed. This closes #1741 .

### DIFF
--- a/Libplanet/Net/Transports/NetMQTransport.cs
+++ b/Libplanet/Net/Transports/NetMQTransport.cs
@@ -10,7 +10,6 @@ using System.Threading.Tasks;
 using AsyncIO;
 using Libplanet.Crypto;
 using Libplanet.Net.Messages;
-using Libplanet.Net.Protocols;
 using Libplanet.Stun;
 using NetMQ;
 using NetMQ.Sockets;
@@ -29,11 +28,11 @@ namespace Libplanet.Net.Transports
         private readonly string _host;
         private readonly IList<IceServer> _iceServers;
         private readonly ILogger _logger;
-        private readonly int _minimumBroadcastTarget;
         private readonly NetMQMessageCodec _messageCodec;
+        private readonly TimeSpan _dealerSocketLifetime;
 
         private NetMQQueue<Message> _replyQueue;
-        private NetMQQueue<(Address?, Message)> _broadcastQueue;
+        private NetMQQueue<(IEnumerable<BoundPeer>, Message)> _broadcastQueue;
 
         private RouterSocket _router;
         private NetMQPoller _routerPoller;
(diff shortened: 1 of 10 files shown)


Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -914,27 +893,31 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
...
-                        if (!peerAddresses.Contains(address) &&
-                            _dealers.TryRemove(address, out DealerSocket removed))
+                        if (DateTimeOffset.UtcNow - _dealers[address].Item2 > _dealerSocketLifetime
+                            && _dealers.TryRemove(
+                                address,
+                                out (DealerSocket DealerSocket, DateTimeOffset) pair))
                         {
-                            removed.Dispose();
+                            pair.DealerSocket.Dispose();
                         }
```
greymistcube, line 904: Although chances are slim, checking an item *and* removing an item are two separate operations, hence prone to a race condition, i.e. not thread safe. For instance, `_dealers[address]` can be determined to be stale and right before `_dealers.TryRemove()` is run, another thread can try to use `_dealers[address]` socket which will be disposed immediately by the thread controlled by `DisposeUnusedDealerSockets()`. The easiest solution would be to lock `_dealers`.

limebell: Added in 7ba2aab189a551166a66601ad16516348c947055. PTAL :)

Libplanet/Net/Transports/NetMQTransport.cs
```diff
@@ -661,52 +658,43 @@ private void DoBroadcast(object sender, NetMQQueueEventArgs<(Address?, Message)>
                     DateTimeOffset.UtcNow,
                     _appProtocolVersion);
 
-                peers.AsParallel().ForAll(peer =>
+                lock (_dealerLock)
```
greymistcube, line 661: I think this `lock` scope might be too wide?

limebell: In order to define a lock in inner scope, we should make a lock for each dealer socket.

greymistcube: Doesn't `AsParallel.Forall()` block all other threads until finished?

limebell: It works like `Task.WhenAll()` as I know, each task work asynchronously. https://docs.microsoft.com/ko-kr/dotnet/standard/parallel-programming/how-to-write-a-simple-parallel-foreach-loop

greymistcube: If I'm not mistaken, I think `Task.WhenAll()` is an asynchronous call and `AsParallel().ForAll()` is a synchronous call *from* the calling thread. 🤨

greymistcube (approved):

Libplanet/Net/Transports/NetMQTransport.cs
greymistcube: My bad, seems like a synchronous call is okay here since broadcasting is done in a single thread. 😗

PR #1752, opened 2022-01-24
Retries removed from `NetMQTransport`

### LINKED ISSUE #1751: Remove retries from `NetMQTransport`
As `NetMQ` is supposed to handle sending and receiving of messages asynchronously, there doesn't seem to be a good reason to retry request-response operation when failed. Parsing through logs in [CloudWatch], retry attempts happen very rarely. This could be either that a responding node is somehow corrupted and sending invalid replies, in which case no amount of retry attempts would help, or normally timed out receive attempt in which case a retry attempt is still an overkill as the retry code path leads to a new attempt of a request-response from a clean slate.

[CloudWatch]: https://aws.amazon.com/cloudwatch/

greymistcube:
Closes #1751. For the rationale, see #1751.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,6 +38,8 @@ To be released.
     became not to be disposed right after corresponding peer is removed
     from the routing table. Instead, it will be removed after
     a certain amount of time.  [[#1741], [#1744]]
+ -  `NetMQTransport` no longer attempts to retry failed communications.
+    [[#1751] [#1752]]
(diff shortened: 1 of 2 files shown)


limebell (approved):
Rationale makes sense. Please update the branch before merging.

greymistcube:
@limebell Rebased.

limebell (approved):

Issue #1754, opened 2022-01-25
Make messages public


limebell:
Each message class was `internal` because it had `NetMQ` dependency. But in #1503 the dependency removed.

longfin:
@limebell Could you introduce the motivation(?) about exposing `Message`s to public? 🙏

PR #1756, opened 2022-01-25
Make some message classes public

limebell:
This implements #1754, but only protocol related messages were included because some of non-protocol messages use internal class (`BlockLocator`) as its constructor arguments. It would be great to make other messages as public in future.


dahlia (dismissed):
It would be better if these public classes have their short summaries in XML comments.

dahlia (approved):

dahlia (approved):

PR #1759, opened 2022-01-25
Use `BlockCandidateTable<T>` on `PollBlocksAsync`

### LINKED ISSUE #1731: Follow-up of #1699
Since [Introduce Swarm<T>.BlockCandidateConsumer](https://github.com/planetarium/libplanet/pull/1699), the thread of downloading blocks from 'FillBlocksAsync' and the thread of attaching blocks have been separated.

However, in 'PollBlocksAsync', blocks are still attached in series, which needs to be changed to use the 'BlockCandidateTable' used in 'FillBlocksAsync'.

riemannulus:
close #1731 

Now use `BlockCandidateTable<T>` on polling blocks too.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,10 @@ To be released.
     a certain amount of time.  [[#1741], [#1744]]
  -  (Libplanet.Net) `NetMQTransport` no longer attempts to retry failed
     communications.  [[#1751], [#1752]]
+ -  (Libplanet.Net) Fixed races between `Swarm<T>`'s dual systems for block
+    synchronization (polling and event-driven in parallel) on updating chain by
+    serializing the post-download process which is common for both
+    synchronization mechanisms.  [[#1731], [#1759]]
  -  New log output tagged with `Metric` added to measure evaluation time for
     individual `Transaction<T>`s.  [[#1755], [#1758]]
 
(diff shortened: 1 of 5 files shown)


limebell (dismissed):

longfin (approved):

CHANGES.md
```diff
@@ -50,6 +50,8 @@ To be released.
     a certain amount of time.  [[#1741], [#1744]]
  -  (Libplanet.Net) `NetMQTransport` no longer attempts to retry failed
     communications.  [[#1751], [#1752]]
+ -  (Libplanet.Net) `Swarm<T>.PollBlocksAsync` now sync block like
+    `Swarm<T>.FillBlocksAsync`.  [[#1731], [#1759]]
```
dahlia, line 54: Since both `PollBlocksAsync()` & `FillBlocksAsync()` are internal methods, IMO it is not a good idea to mention them in the changelog.  Readers even don't get what it means.

How about adjusting it like below?

```suggestion
  - (Libplanet.Net) Fixed races between `Swarm<T>`'s dual systems for block
    synchronization (polling and event-driven in parallel) on updating chain by
    serializing the post-download process which is common for both
    synchronization mechanisms.  [[#1731], [#1759]]
```

riemannulus: https://github.com/planetarium/libplanet/pull/1759/commits/8ae1b673d8cf52bc934b9cff406a42430b698985 amend it!

dahlia (approved):

Issue #1421, opened 2021-08-02
Extract Libplanet.Net as a distinct assembly & NuGet package


dahlia:
I expect splitting *Libplanet.Net* as a distinct assembly/project/package to have several benefits:

- Build after small code change will be faster.
- We can apply different policies for projects such as target frameworks and testing runtimes.  (In other words, *Libplanet.Net* can be free from Unity/Mono.)
- CI builds also can be faster.

PR #1760, opened 2022-01-25
Extract Libplanet.Net as a distinct assembly & NuGet package

dahlia:
The rationale is written in #1421, this patch addresses:

> I expect splitting *Libplanet.Net* as a distinct assembly/project/package to have several benefits:
>
> - Build after small code change will be faster.
> - We can apply different policies for projects such as target frameworks and testing runtimes.  (In other words, *Libplanet.Net* can be free from Unity/Mono.)
> - CI builds also can be faster.

I'm going to let CircleCI builds skip *Libplanet.Net.Tests* on Mono/Unity jobs, but it should be done as a separate work.


Libplanet/AssemblyInfo.cs
```diff
@@ -1,6 +1,7 @@
 #nullable disable
 using System.Runtime.CompilerServices;
 
-[assembly: InternalsVisibleTo("Libplanet.Tests")]
 [assembly: InternalsVisibleTo("Libplanet.Extensions.Cocona.Tests")]
-[assembly: InternalsVisibleTo("Libplanet.Benchmarks")]
+[assembly: InternalsVisibleTo("Libplanet.Net")]
```
limebell, line 5: It will be great if we can remove this and make `Libplanet` internals used in `Libplanet.Net` public.

limebell (approved):

limebell (approved):

Libplanet/AssemblyInfo.cs
dahlia: I filed a new issue to address this: <https://github.com/planetarium/libplanet/issues/1762>.  Thanks!

PR #1775, opened 2022-02-04
Preload multiple times to catch up tip close enough & release 0.26.3

dahlia:


limebell (commented):
- It would be great to have a test for following up new tips during preloading.
- Can you leave 'FIXME's and 'TODO's as issue after this PR is merged?

Libplanet/EnumerableExtensions.cs
```diff
@@ -1,13 +1,60 @@
...
+
+                first = false;
+            }
+
+            if (first)
+            {
+                throw new InvalidOperationException("Sequence contains no elements.");
+            }
+
+            return greatestElement;
```
riemannulus, line 53: Is it possible to `null` when `if (first || value.CompareTo(greatestValue) > 0)` is all false?

dahlia:
@limebell @riemannulus I rebased commits on the latest 0.26-maintenance.  Could you take a look again?

Libplanet/EnumerableExtensions.cs
dahlia: The `greatestElement` must not be `null`, except for empty sequence, and that's why `InvalidOperationException` is thrown if `first` is still `true` after the below loop ends.

riemannulus (approved):

Issue #1734, opened 2022-01-19
Fix improper timeout cancellation for `ProcessRequest()` in `NetMQTransport`


greymistcube:
Given a `TimeSpan` for timeout when processing a `MessageRequest`, timeout is handled improperly. Instead of tying the given `TimeSpan` to the process, the given `TimeSpan` is tied to each network communication. For instance, if a timeout of `5` seconds is provided for sending some message and expecting `10` replies, instead of timing the combined process of sending and receiving all replies, the current implementation can wait up to `5` seconds for sending the message and `5` seconds each for each reply.

Issue #1722, opened 2022-01-17
Funnel sending `Message`s through `SendMessageWithReplyAsync()` when broadcasting via `NetMQTransport`


greymistcube:
Unlike newly implemented `TcpTransport`, `NetMQTransport`'s broadcasting mechanism implements its own way of sending `Message`s rather than reusing `SendMessageWithReplyAsync()` interface method.

There are several problems with this:

- There is no good reason to have a separate implementation.
- Although the functionality looks the same for sending a `Message` to a `BoundPeer` from the outside, two completely different implementations need to be maintained.
  - As an example, any additional functionality we'd want to attach to the act of sending a `Message`, this needs to be done twice, also in different manner to accommodate each implementation.

I suggest we reroute internally `DoBroadcast()` to `SendMessageWithReplyAsync()` for `NetMQTransport`.

greymistcube:
On further inspection, funneling `DoBroadcast()` to `SendMessageWithReplyAsync()` doesn't seem as simple as it seems. 😶

However, I still believe the act of "sending a message" should ultimately follow a single "controllable" code path, as there is no good reason other than some orchestrated handling of `DealerSocket`s necessary for reusing cached `DealerSocket`s.

At the moment, refactoring our custom `NetMQSocketExtensions` seem to be the easiest way to go about resolving this issue. 😗

greymistcube:
Yet, on another further inspection, current usage of `DealerSocket` seems somewhat improper. I didn't know we were using a [customized NetMQ](PlanetariumNetMQ) when sending messages to "force" asynchronous sending. After some research on [NetMQ] and [ZeroMQ], the custom implementation seems to be at odds with how these library are "supposed" to be used.

[PlanetariumNetMQ]: https://github.com/planetarium/netmq
[NetMQ]: https://github.com/zeromq/netmq
[ZeroMQ]: https://zeromq.org

greymistcube:
Eh, thought this was going to be a quick fix but is getting out of hand. I suppose some explanation is in order. Hopefully this doesn't turn into a long rant. 😶

[...]

PR #1778, opened 2022-02-07
Use `TrySendMultipartMessage` instead of `SendMultipartMessageAsync` extension method

### LINKED ISSUE #1753: Use synchronous `NetMQ`'s send instead.
See the [comment] from #1722.

As we expect individual `DealerSocket` to be neither long running nor of particularly of high throughput, using `OutgoingSocketExtensions.TrySendMultipartMessage()` with default `TimeSpan.Zero` as timeout might be enough.

From what I understand, if synchronous call to either `SendMultipartMessage()` or `TrySendMultipartMessage()` becomes problematic, we are likely using `NetMQ` not as intended. 🙃

[comment]: https://github.com/planetarium/libplanet/issues/1722#issuecomment-1019807641

greymistcube:
Closes #1753.
Partially addresses #1722 and #1734.

As `SendMultipartMessageAsync` seemed like an anti-pattern for using `NetMQ`, this PR attempts to mitigate the issue by using `TrySendMultipartMessage`. There are several reasons for this:

- There is no clear indication `SendMultipartMessageAsync` is actually working as intended, i.e. recovering or waiting when a `Message` can't be send immediately.
- Faster send and receive cycle by failing fast resulting in lower total `DealerSocket` usage.
- Reduces possible total wait time for `SendMessageWithReplyAsync`.

In my opinion, if we actually want an `async` functionality, it should be implemented at a higher level.

Although apocryphal, tested with this [docker image] on `9c-main` for more than a day without much issue, seemingly with better syncing compared to other nodes. Newly added failed to send log message was seen very rarely, once every several hours, even with immediate failure.

[docker image]: https://hub.docker.com/layers/planetariumhq/ninechronicles-headless/no-retry-sync-send/images/sha256-41c025d5ad9ae44c069966db4d479c54a9ff8d4d3ddcce4442b5f2d0272f3cdc?context=repo

### DIFF
--- a/Libplanet.Net/Transports/NetMQTransport.cs
+++ b/Libplanet.Net/Transports/NetMQTransport.cs
@@ -55,8 +55,8 @@ namespace Libplanet.Net.Transports
         private ConcurrentDictionary<string, TaskCompletionSource<object>> _replyCompletionSources;
 
         /// <summary>
-        /// The <see cref="EventHandler" /> triggered when the different version of
-        /// <see cref="Peer" /> is discovered.
+        /// The <see cref="EventHandler"/> triggered when the different version of
+        /// <see cref="Peer"/> is discovered.
         /// </summary>
         private DifferentAppProtocolVersionEncountered _differentAppProtocolVersionEncountered;
 
@@ -471,7 +471,7 @@ namespace Libplanet.Net.Transports
             }
         }
 
-        /// <inheritdoc />
+        /// <inheritdoc/>
         public void BroadcastMessage(IEnumerable<BoundPeer> peers, Message message)
         {
             if (_disposed)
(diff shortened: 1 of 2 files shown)


Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -806,18 +807,27 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
...
-                _logger.Debug(
-                    "Request {Message} {RequestId} sent to {Peer} with timeout {Timeout}.",
-                    req.Message,
-                    req.Id,
-                    req.Peer,
-                    req.Timeout);
+                if (dealer.TrySendMultipartMessage(message))
+                {
+                    _logger.Debug(
+                        "Request {Message} {RequestId} sent to {Peer} with timeout {Timeout}.",
```
limebell, line 813: Seems like timeout is not needed anymore?

greymistcube: It was to track overall timeout tied to a `RequestId`. I suppose the log message itself is confusing. Fixed in https://github.com/planetarium/libplanet/commit/debefd6a29941a2d771954f2444dd94c6443b77f.

limebell (dismissed):

greymistcube:
After changing target branch to `0.27-maintenance`, `MessageSendFailedException` has been removed and uses a generic `TimeoutException` instead. This change needs to be patched once ported to `main` branch. See #1781.

limebell (approved):

longfin (approved):

PR #1789, opened 2022-02-10
Single cancellation timer during `NetMQTransport.SendMessageWithReplyAsync()`

### LINKED ISSUE #1734: Fix improper timeout cancellation for `ProcessRequest()` in `NetMQTransport`

greymistcube:
Closes #1734. Preparatory work for #1722.

Also removes no longer used `NetMQSocketExtensions`.


riemannulus:
Seems too many tests are broken, is it ok?

greymistcube:
Tested on 9c-main node with no particular adverse effect.

longfin (approved):

PR #1792, opened 2022-02-11
Revert #1759

riemannulus:
Because #1759 has preload problem, we revert #1759

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,14 @@ Version 0.27.3
 
 To be released.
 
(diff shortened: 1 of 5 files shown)


longfin:
I guess we should add changelog instead of removing.

riemannulus:
> I guess we should add changelog instead of removing.

https://github.com/planetarium/libplanet/pull/1792/commits/1e0c80a85276847f6dee153eef10695481f01b00 done

longfin (approved):

PR #1798, opened 2022-02-15
Rollback #1759

riemannulus:
Since #1759, we had a problem with preload. So I was reverted.

If you look at the following code, there is a part where you made a mistake while deleting the parameter. 

https://github.com/planetarium/libplanet/pull/1792/files#diff-215bb45126357b0974dea14401f0213b497011890a129ae3a7df3e3468ea4911L524

If preload is always true, this `if` statement is meanless but I was left this code blocks.

In this PR. I'm re-implement #1759 and fixing those bugs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,14 @@ To be released.
 
  -  Fixed *@planetarium/cli* npm package's bug where installation had not
     worked since 0.27.0.  [[#1790], [#1799]]
+ -  (Libplanet.Net) Return the fixed races between `Swarm<T>`'s dual systems
+    for block synchronization (polling and event-driven in parallel)
+    on updating chain by serializing the post-download process
+    which is common for both synchronization mechanisms.  [[#1731], [#1798]]
+
 
 [#1790]: https://github.com/planetarium/libplanet/issues/1790
+[#1798]: https://github.com/planetarium/libplanet/pull/1798
 [#1799]: https://github.com/planetarium/libplanet/pull/1799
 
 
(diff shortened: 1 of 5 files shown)


dahlia:
@riemannulus Could you elaborate some context/background behind this patch in the pull request description?

CHANGES.md
```diff
@@ -32,6 +32,11 @@ To be released.
 
 ### Behavioral changes
 
+  - (Libplanet.Net) Return the fixed races between `Swarm<T>`'s dual systems
```
longfin, line 35: It seems the link for #1789 was missing.

Libplanet.Net/Swarm.BlockSync.cs
```diff
@@ -74,28 +74,123 @@ public partial class Swarm<T>
                 return;
             }
 
+            var totalBlocksToDownload = 0L;
+            Block<T> tempTip = BlockChain.Tip;
+            List<Block<T>> blocks = new List<Block<T>>();
```
longfin, line 79: Trivial

```suggestion
            var blocks = new List<Block<T>>();
```

longfin (commented):
Does it resolve the root cause of #1759 and #1792? if so, could you leave the comments about what was the matter?

Libplanet.Net/Swarm.BlockSync.cs
riemannulus: https://github.com/planetarium/libplanet/pull/1798/commits/b7a1d200156030175ad3ea75231b09977f74cc5a amended

CHANGES.md
riemannulus: https://github.com/planetarium/libplanet/pull/1798/commits/f8775fc1d15efe1a0453adfb97561d185f466098

dahlia (approved):

Issue #1790, opened 2022-02-10
`npm i @planetarium/cli` fails since 0.27.0


dahlia:
```
error /home/runner/work/9c-launcher/9c-launcher/node_modules/@planetarium/cli: Command failed.
Exit code: 1
Command: node install.js
Arguments: 
Directory: /home/runner/work/9c-launcher/9c-launcher/node_modules/@planetarium/cli
Output:
internal/modules/cjs/loader.js:1102
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/runner/work/9c-launcher/9c-launcher/node_modules/@planetarium/cli/node_modules/node-fetch/src/index.js
require() of ES modules is not supported.
require() of /home/runner/work/9c-launcher/9c-launcher/node_modules/@planetarium/cli/node_modules/node-fetch/src/index.js from /home/runner/work/9c-launcher/9c-launcher/node_modules/@planetarium/cli/download.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/runner/work/9c-launcher/9c-launcher/node_modules/@planetarium/cli/node_modules/node-fetch/package.json.

    at new NodeError (internal/errors.js:322:7)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1102:13)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/home/runner/work/9c-launcher/9c-launcher/node_modules/@planetarium/cli/download.js:14:15)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32) {
  code: 'ERR_REQUIRE_ESM'
}
```

This seems probably due to <https://github.com/planetarium/libplanet/commit/ed8ef7d9011eb9eceaa98ce23685db5ac10716f3>.

PR #1799, opened 2022-02-15
Fix @planetarium/cli npm package

dahlia:
This fixes <https://github.com/planetarium/libplanet/issues/1790>.  The cause of the error was that recent *node-fetch* versions[^1] has switched to [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) while *@planetarium/cli* had still remained in [CommonJS](https://en.wikipedia.org/wiki/CommonJS).  This patch switches it to ESM (see also `"type": "module"` in the *package.json* file).

In order to prevent such regressions, I added small automated tests as well.

[^1]: We upgraded it since 0.27.0 to avoid vulnerability.  See also <https://github.com/planetarium/libplanet/pull/1747>.


riemannulus (dismissed):

Libplanet.Tools/package.json
```diff
@@ -8,24 +8,23 @@
...
-  ],
+  "cpu": ["x64"],
   "files": [
     "README.md",
-    "bin/planet",
-    "download.js",
-    "install.js"
+    "bin/planet.mjs",
+    "download.mjs",
+    "install.mjs"
```
BasixKOR, line 16: Changing the extensions to `.mjs` is a nice touch, but I just wanted to point out that this isn't required. When `"type": "module"` is specified, Node.js should treat all content as ES Modules unless it has a `.cjs` extension.

Libplanet.Tools/download.mjs
```diff
@@ -0,0 +1,145 @@
+// The planet binary downloader for npm.
+// Inspired by Elm's npm packaging: <https://www.npmjs.com/package/elm>.
+"use strict";
+const nodeMajorVersion = process.versions.node.split('.')[0];
+if (nodeMajorVersion < 12) {
+  throw new Error("Requires Node.js v12 or Higher Version.");
+}
+
+import child_process from "child_process";
+import fs from "fs";
```
BasixKOR, line 10: It seems like the code is using the Promise variant exclusively. Directly importing from `fs/promises` might be preferable.

```suggestion
import { readFile, mktemp, copyFile, chmod } from "fs/promises"; // Don't forget to replace fs.promises in the same file!
```

Libplanet.Tools/download.mjs
```diff
@@ -0,0 +1,145 @@
+// The planet binary downloader for npm.
+// Inspired by Elm's npm packaging: <https://www.npmjs.com/package/elm>.
+"use strict";
+const nodeMajorVersion = process.versions.node.split('.')[0];
+if (nodeMajorVersion < 12) {
+  throw new Error("Requires Node.js v12 or Higher Version.");
+}
```
BasixKOR, line 7: You may use [`engines`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines) field for specifying the required Node.js version range. This will warn users when installing on an unsupported Node.js version.

limebell (dismissed):

BasixKOR (commented):
LGTM! I've left some thoughts I had while reading the code.

Libplanet.Tools/package.json
dahlia: That's good to know!  However, as like me, most Libplanet maintainers are not that familiar to npm, IMHO it would be better to leave them explicit.

Libplanet.Tools/download.mjs
dahlia: Applied: 656979cbc2ec6f3c9a7789bcebe5380ed4139c14.

Libplanet.Tools/download.mjs
dahlia: Applied: 923770e7793b962fda167e215c3b216c8f420e3f.

dahlia:
@planetarium/libplanet Applied @BasixKOR's review suggestions.  Please take a look again!

BasixKOR (approved):

PR #1802, opened 2022-02-16
Additional network related metric logs

greymistcube:
Adds logs for counting sockets in use and request process duration.

PR #1796, opened 2022-02-14
Fix some flaky unit tests in `Libplanet.Net.Tests`

limebell:
This PR fixes some flaky tests which are:
- `PreloadRetryWithNextPeers`
- `CreateNewChainWhenBranchPointNotExist`
- `BroadcastTxWhileMining`

PR #1809, opened 2022-02-17
Port 0.27.6 to main (0.28.x)

riemannulus:
🧹 

to @greymistcube, make sure your https://github.com/planetarium/libplanet/pull/1802 PR is correctly included. 

I try my own best but `InvalidTimestampExcption` is renamed to `InvalidMessageTimestampException` on the main branch so may I have a mistake about your PR. 🙏 

to @dahlia, 0.26.4 is ported to 0.27-maintenance. right? So this PR has a bunch of commits. Could you check is it ok?


.idea/.idea.Libplanet/.idea/aws.xml
```diff
@@ -0,0 +1,17 @@
...
+        <option value="profile:default" />
+      </list>
+    </option>
+    <option name="recentlyUsedRegions">
+      <list>
+        <option value="us-east-2" />
+      </list>
+    </option>
+  </component>
+</project>
```
dahlia, line 17: I believe this file was added by mistake.

riemannulus: https://github.com/planetarium/libplanet/pull/1809/commits/f094c9a65b9402e2da546818e5eb076eb004b3ab amended

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -871,33 +906,41 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
...
-            }
-            catch (InvalidMessageTimestampException imte)
-            {
-                tcs.TrySetException(imte);
-            }
-            catch (InvalidMessageSignatureException imse)
+            catch (Exception e) when (
+                e is DifferentAppProtocolVersionException ||
+                e is InvalidMessageTimestampException ||
+                e is TimeoutException)
```
greymistcube, line 924: This should also catch `MessageSendFailedException` and `InvalidMessageSignatureException`.

riemannulus:
```suggestion
            catch (Exception e) when (
                e is DifferentAppProtocolVersionException ||
                e is InvalidMessageTimestampException ||
                e is InvalidMessageSignatureException ||
                e is MessageSendFailedException ||
                e is TimeoutException)
```

is it right?

dahlia:
The build failed:

```
Error: Swarm.cs(632,21): error CS1739: The best overload for 'CompleteBlocksAsync' does not have a parameter named 'preload' [/home/runner/work/libplanet/libplanet/Libplanet.Net/Libplanet.Net.csproj]
```

dahlia:
The build is _consistently_ failing over reruns and jobs (Linux/macOS/Windows & .NET Core/Mono) due to the following tests:

- `SwarmTest.PullBlocks`
- `SwarmTest.PullBlocksByDifficulty`

It must be worth a look.

Libplanet.Net.Tests/SwarmTest.cs
```diff
@@ -1277,7 +1277,7 @@ public async Task CreateNewChainWhenBranchPointNotExist()
                 await receiverSwarm.BlockAppended.WaitAsync();
                 await AssertThatEventually(
                     () => receiverChain.Tip.Equals(minerChainB.Tip),
-                    15_000,
+                    30_000,
```
dahlia, line 1280: If we wait upto 30 seconds twice in a single test case it would exceeds timeout (60 seconds) at the worst case.  It also tends to be much slower in CI runners.  Is increasing these timeouts from 15 seconds to 30 seconds necessary?

riemannulus: This is a mistake. This test is broken on my machine so I want to test its success if much more waiting for this method. After that, I do not revert this change. I'll remove these changes.

dahlia (dismissed):

dahlia:
To make it sure that those failed tests are merely flaky, how about merging 0.27.6 as well?  Note that 0.27.6 includes @limebell's patch #1796 that fixed several flaky tests.

dahlia (approved):

Issue #1772, opened 2022-01-26
`NonBlockRenderer<T>` throws `ThreadStateException`


longfin:
In 0.26.0, `NonBlockRenderer<T>` throws `ThreadStateException` when it tries to invoke `.RenderBlock()` in `BlockChain<T>.Append()`.

```
[20:56:08 ERR] An exception occurred during appending blocks.
System.Threading.ThreadStateException: Thread is running or terminated; it cannot restart.
   at System.Threading.Thread.StartupSetApartmentStateInternal()
   at System.Threading.Thread.Start()
   at Libplanet.Blockchain.Renderers.NonblockRenderer`1.Queue(Action action) in C:\Users\kidon\Desktop\repo_planetarium\NineChronicles.DataProvider\NineChronicles.Headless\Lib9c\.Libplanet\Libplanet\Blockchain\Renderers\NonblockRenderer.cs:line 229
   at Libplanet.Blockchain.Renderers.NonblockRenderer`1.RenderBlock(Block`1 oldTip, Block`1 newTip) in C:\Users\kidon\Desktop\repo_planetarium\NineChronicles.DataProvider\NineChronicles.Headless\Lib9c\.Libplanet\Libplanet\Blockchain\Renderers\NonblockRend
erer.cs:line 188
   at Libplanet.Blockchain.BlockChain`1.Append(Block`1 block, Boolean evaluateActions, Boolean renderBlocks, Boolean renderActions, IReadOnlyList`1 actionEvaluations, Nullable`1 stateCompleters) in C:\Users\kidon\Desktop\repo_planetarium\NineChronicles.Da
taProvider\NineChronicles.Headless\Lib9c\.Libplanet\Libplanet\Blockchain\BlockChain.cs:line 1029
   at Libplanet.Net.Swarm`1.CompleteBlocksAsync(IList`1 peersWithExcerpt, BlockChain`1 workspace, IProgress`1 progress, Boolean preload, Boolean render, CancellationToken cancellationToken) in C:\Users\kidon\Desktop\repo_planetarium\NineChronicles.DataPro
vider\NineChronicles.Headless\Lib9c\.Libplanet\Libplanet\Net\Swarm.BlockSync.cs:line 447
```

PR #1810, opened 2022-02-17
Fix `ThreadStateException` thrown by `NonblockRenderer<T>`

dahlia:
This addresses <https://github.com/planetarium/libplanet/issues/1772>.

There were multiple tries to prevent `NonblockRenderer<T>.Queue()` from calling `Thread.Start()`, and utilizing `Interlock` was indeed the most simple but robust way to achieve that.  A side story: `Thread.IsAlive` property can be more complex than you thought.  (Or it was to me at least.)

I tried to write a substantial regression test for this bug, but I couldn't find any way to make it clear without decreasing its reproducibility.  I concluded it's better to reproduce the situation well through less clear test logic, rather than clear one that does not reproduce it well.

Apart from the automated regression test, I tested _Nine Chronicles_[^1] with this patch for days, and I believe this is now quite stable.

[^1]: FYI, `NineChronicles.Executable` already has `--nonblock-renderer` option.  Give it a try by adding it to `"HeadlessArgs"` list in your 9c-launcher configuration.

### DIFF
--- a/Libplanet.Tests/Blockchain/Renderers/NonblockRendererTest.cs
+++ b/Libplanet.Tests/Blockchain/Renderers/NonblockRendererTest.cs
@@ -97,5 +97,57 @@ namespace Libplanet.Tests.Blockchain.Renderers
             // Double disposing does nothing.
             renderer.Dispose();
         }
+
+        [Theory]
+        [InlineData(4)]
+        [InlineData(8)]
+        [InlineData(16)]
+        [InlineData(32)]
+        public void StatingWorkerThreadOnlyOnce(int parallel)
+        {
+            // This is a regression test for the follwing bug:
+            // https://github.com/planetarium/libplanet/issues/1772
+            var renderer = new NonblockRenderer<DumbAction>(
+                new AnonymousRenderer<DumbAction>(),
+                25,
+                NonblockRenderer<DumbAction>.FullMode.DropNewest
+            );
+            Thread[] threads = new Thread[parallel];
+            Exception[] errors = new Exception[threads.Length];
+            using (renderer)
+            {
+                for (int i = 0; i < threads.Length; i++)
+                {
+                    // Declare a new variable for each iteration to avoid being captured by closure:
+                    int threadNo = i;
+                    threads[threadNo] = new Thread(() =>
+                    {
+                        try
+                        {
+                            renderer.RenderBlock(_blockA, _blockB);
+                        }
+                        catch (Exception e)
+                        {
+                            errors[threadNo] = e;
+                        }
+                    });
+                }
+
+                for (int i = 0; i < threads.Length; i++)
+                {
+                    threads[i].Start();
+                }
+
+                for (int i = 0; i < threads.Length; i++)
+                {
+                    threads[i].Join();
+                }
+            }
+
+            for (int i = 0; i < threads.Length; i++)
+            {
+                Assert.Null(errors[i]);
+            }
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


Libplanet.Tests/Blockchain/Renderers/NonblockRendererTest.cs
```diff
@@ -97,5 +97,57 @@ Block<DumbAction> branchpoint
...
+        public void StatingWorkerThreadOnlyOnce(int parallel)
+        {
+            // This is a regression test for the follwing bug:
+            // https://github.com/planetarium/libplanet/issues/1772
+            var renderer = new NonblockRenderer<DumbAction>(
+                new AnonymousRenderer<DumbAction>(),
+                25,
+                NonblockRenderer<DumbAction>.FullMode.DropNewest
+            );
+            Thread[] threads = new Thread[parallel];
```
riemannulus, line 115: Why do we use `Thread` instead of `Task`? To my knowledge, we don't need to call low-level API `Thread` these days. IMO if you use `Task`, this test is way easier to read.

Plus, AFAIK `NineChronicles.Headless` is using `Renderer` with `Task`. So IMO we need to test like the user.

dahlia: > Why do we use `Thread` instead of `Task`? To my knowledge, we don't need to call low-level API `Thread` these days. IMO if you use `Task`, this test is way easier to read.

Because `Task`s are basically pooled.  In other words, they don't guarantee to start immediately, in particular when you spawn many tasks in a short time.  `Task`s can be spawned in a custom pool, but if we do that it's no more simpler than bare `Thread`… 🤔

> Plus, AFAIK `NineChronicles.Headless` is using `Renderer` with `Task`. So IMO we need to test like the user.

It does not intend to simulate *Nine Chronicles*' way to deal with renderers (by the way, does it really use renderers with tasks, and for what?), and it's even difficult to reproduce the race condition addressed by #1772 using tasks (due to the reason I wrote above).

Libplanet/Blockchain/Renderers/NonblockRenderer.cs
```diff
@@ -218,16 +225,28 @@ protected void Queue(System.Action action)
...
-                        _worker.Start();
-                    }
+                    _worker.Start();
+                }
+                catch (ThreadStateException e)
+                {
+                    _logger.Error(
+                        "The rendering thread #{WorkerThreadId} failed to start.",
+                        _worker.ManagedThreadId,
+                        e
```
greymistcube, line 248:
```suggestion
                        e,
                        "The rendering thread #{WorkerThreadId} failed to start.",
                        _worker.ManagedThreadId
```

Libplanet.Tests/Blockchain/Renderers/NonblockRendererTest.cs
riemannulus: > Because Tasks are basically pooled. In other words, they don't guarantee to start immediately,

Ah, I miss that. 👍

Libplanet/Net/Swarm.BlockCandidate.cs
```diff
@@ -78,8 +78,8 @@ public partial class Swarm<T>
             catch (Exception e)
             {
                 _logger.Error(
-                    "Cannot append blocks at {MethodName}. " +
-                    "InnerException: {InnerException}",
+                    e,
+                    "Cannot append blocks at {MethodName}",
                     nameof(BlockCandidateTable),
                     e.Message);
```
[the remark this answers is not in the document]
dahlia: Amended!

Libplanet/Blockchain/Renderers/NonblockRenderer.cs
dahlia: Amended!

dahlia:
@greymistcube @riemannulus  I addressed all review comments.  Please take a look again!

riemannulus (approved):

greymistcube (approved):

PR #1823, opened 2022-02-21
Allow `planet` command to take passphrase from stdin 

dahlia:
This patch is the second take of @moreal's patch <https://github.com/planetarium/libplanet/pull/1593>.  In particular, this tried to address a [review suggestion on that made by myself][1]:

> As @riemannulus said, avoiding to get a passphrase through stdin was basically intended.
>
> IMHO the method to get a passphrase shouldn't be silently switched.  Instead, it'd better have an explicit option to turn on the such mode, e.g., `--passphrase-stdin`.  Of course, such option and the existing `-p`/`--passphrase` option should be mutually exclusive.
>
> On the other hand, however, exposing an uncaught exception to end-users is also a bug.  For the case `Console.IsInputRedirected` (i.e., `!isatty()`) without `--passphrase-stdin`, it should print an explicit error message instead of uncaught exception, for example:
>
> > Error: The standard input is not associated to any terminal device (tty), which means it probably is piped.  If you need to pass the passphrase through pipe, turn on `--passphrase-stdin` option.

From user's perspective, there are two changes:

- The `planet` command no more unexpectedly terminates with an uncaught `InvalidOperationException` when the standard input is piped (to put in passphrase), which is the issue that PR <https://github.com/planetarium/libplanet/pull/1593> originally tried to address.  Instead, it suggests `--passphrase-file=-` option.

- The `planet` subcommands taking passphrase now have a new option named `--passphrase-file`.  As the name implies, it basically reads passphrase from the specified file, and `--passphrase-file=-` reads passphrase from the standard input, just like many other Unix/Linux programs.

API-wise, `PassphraseParameters` type was introduced to group `-p`/`--passphrase` & `--passphrase-file` options.  As it utilizes `ICommandParameterSet`, *Cocona.Lite* was upgraded from 1.5 to 1.6.

[1]: https://github.com/planetarium/libplanet/pull/1593#pullrequestreview-800822286

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -21,7 +21,8 @@
     "formulahendry.dotnet-test-explorer",
     "ms-vscode.powershell",
     "ms-vsliveshare.vsliveshare",
-    "streetsidesoftware.code-spell-checker"
+    "streetsidesoftware.code-spell-checker",
+    "GitHub.vscode-pull-request-github"
   ],
 
   // Use 'postCreateCommand' to run commands after the container is created.
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
     "ms-dotnettools.csharp",
     "formulahendry.dotnet-test-explorer",
     "ms-vscode.powershell",
-    "streetsidesoftware.code-spell-checker"
+    "streetsidesoftware.code-spell-checker",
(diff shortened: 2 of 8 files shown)


dahlia:
@planetarium/libplanet @moreal Take a look, please.

riemannulus (approved):

moreal (approved):
It has very kind guides when errors! 🚀

PR #1824, opened 2022-02-22
Funnel broadcast to send

### LINKED ISSUE #1722: Funnel sending `Message`s through `SendMessageWithReplyAsync()` when broadcasting via `NetMQTransport`

### LINKED ISSUE #1819: Fix possible socket leak in `NetMQTransport.DoBroadcast()`
Possibly introduced in #1744.

See https://github.com/planetarium/libplanet/pull/1744/files#diff-77c98fc4737fd05e38d93768876b809dd2f4e62abdca55449f72a896bd7618a7L690-R697.

Seems like failure to send a message during broadcast would possibly result in an unreferenced undisposed socket. 🙄

greymistcube:
Closes #1722.

This also effectively resolves #1819 since by eliminating an internal management of a collection of `DealerSocket`s inside `NetMQTransport`.

Unless the overhead for additional creations and teardowns of `DealerSocket` is substantial, which I don't think there is any reason to believe this is the case, this shouldn't cause an issue. As a result, overall socket usage should be further decreased by this patch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,9 @@ To be released.
     multiple replies.  [[#1734], [#1789]]
  -  `ActionEvaluator<T>.Evaluate()` method became to log message directly
     instead of other methods that use it.  [[#1773]]
+ -  (Libplanet.Net) `NetMQTransport.DoBroadcast` now uses
+    `NetMQTransport.SendMessageAsync()` internally instead of
+    using `NetMQ` directly.  [[#1722], [#1824]]
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


greymistcube:
Currently testing on 9c-main.

riemannulus:
> Currently testing on 9c-main/remote-headless-99 node.

So then, putting it back in the draft state is better I guess.

greymistcube:
Please approve this PR after test playing on remote-headless-99 node. 😗

riemannulus (approved):

Issue #1836, opened 2022-02-28
Implement more delayed disposal for sockets


greymistcube:
See #1778, #1789, and #1824.

By using `DealerSocket.TrySendMultipartMessage()`, when not waiting for a reply, `DealerSocket.Dispose()` is called immediately inside `NetMQTransport.ProcessRequest()`. This should've resulted in messages not being sent, but when testing locally without any latency involved, tests ran mostly fine, so there was an oversight. Arbitrary `Task.Delay(1)` was introduced not due to a failure of not being able to send a message without a reply, but due to internal code inside `NetMQ` during cleanup causing problems under a Windows environment. 🙄

Even when extensively tested on multiple nodes on the mainnet, this worked fine, possibly due to low network latency between k8s pods, although I am a bit surprised `Task.Delay(1)` was enough. 😶

A major issue emerged when testing with a local client for v100120 with libplanet 0.28.0. After testing with various delay amounts, `1` second delay seemed to work fine. The previous value before the above mentioned pull requests was `3` seconds.

To account for latency, `DealerSocket` should linger for a while longer when sending a message without a reply.

greymistcube:
Closed by #1839.

PR #1839, opened 2022-03-02
DealerSocket linger added

greymistcube:
Partially addresses #1836.

Few things to note:
- This further relies on the reliability of `NetMQ`.
  - There is no explicit check on whether a `Message` was actually sent.
  - ~`DealerSocket` count is only an estimate as it can linger before actually getting disposed by `NetMQ` behind the scene.~

Tested with a local chain on 9c-internal [v100120 rc1] based client.

[v100120 rc1]: https://github.com/planetarium/NineChronicles/commit/cd871445c9fde4dcd0200a28516728f63f469032

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ To be released
  -  Fixed an evaluation log to output `IPreEvaluationBlock<T>.PreEvaluationHash`
     as a hex formatted string.  [[#1835], [#1837]]
  -  Fixed a bug where some messages could not be sent using `NetMQTransport`
-    due to premature `DealerSocket` disposal.
+    due to premature `DealerSocket` disposal.  [[#1836], [#1839]]
 
 [#1835]: https://github.com/planetarium/libplanet/issues/1835
 [#1836]: https://github.com/planetarium/libplanet/issues/1836
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,9 +8,13 @@ To be released
 
  -  Fixed an evaluation log to output `IPreEvaluationBlock<T>.PreEvaluationHash`
     as a hex formatted string.  [[#1835], [#1837]]
+ -  Fixed a bug where some messages could not be sent using `NetMQTransport`
+    due to premature `DealerSocket` disposal.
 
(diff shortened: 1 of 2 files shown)


greymistcube:
Temporary conversion back into a draft so this doesn't get accidentally get merged. 😐

greymistcube:
Had some reliability issue with `Linger` option. On further inspection, seemed like a possible anti-pattern, and in case of network traffic spike, there is a buffer overflow at an even lower level than `NetMQ` where the whole thing crashes. 😑

This is a temporary hotfix until we properly address #1836.

Even before the slew of patches for `NetMQTransport`, usage of `SendMultipartMessageAsync()` using a `DealerSocket` without expecting a response was an anti-pattern for using the `NetMQ` library. There are two ways to go about this:

- Easier solution: Change the messaging protocol so that every outgoing `Message` requires a reply `Message`.
- Harder solution: Use a different `NetMQSocket` type for sending out `Message`s that does not expect a reply.

Both have problems of their own, but I personally recommend the former option.

- It is much easier to implement and deal with and reduces the complexity of an `ITransport` implementation.
- It is easier to have a tighter control over and monitor whether a `Message` was sent or not.

Although, this is more prone to run into a scalability issue earlier, I think such timing would *not* be much earlier than when we mix in a [pub/sub] model.

However, with the next arena season for [9c] looming ahead, if possible, easier solution should be implemented ASAP.

[pub/sub]: https://netmq.readthedocs.io/en/latest/pub-sub/
[9c] https://github.com/planetarium/NineChronicles

CHANGES.md
```diff
@@ -8,9 +8,13 @@ To be released
 
  -  Fixed an evaluation log to output `IPreEvaluationBlock<T>.PreEvaluationHash`
     as a hex formatted string.  [[#1835], [#1837]]
+ -  Fixed a bug where some messages could not be sent using `NetMQTransport`
+    due to premature `DealerSocket` disposal.
```
limebell, line 12:
```suggestion
    due to premature `DealerSocket` disposal.  [[#1836], [#1839]]
```

limebell (approved):

PR #1845, opened 2022-03-04
Reply `Pong` to `TxIds` and `BlockHeaderMessage`

greymistcube:
Another follow up hotfix for #1839.

I honestly wanted to do a minor tweak on `ITransport` interface on `main` branch, but figured it'd be better to break up the work and do iterative interface update on `main` branch and have this dirty fix pushed to `0.28-maintenance` first. See #1846.

Although the behavior of `Swarm<T>` has changed, this is fully compatible with `0.28.1`.


greymistcube:
Changed the base to `main`.

longfin (approved):

PR #1849, opened 2022-03-14
Deprecate `ITransport.SendMessageAsync()`

greymistcube:
See #1839 for the rationale.

To elaborate,

- With a proper pub/sub model, the responsibility of acknowledging a transportation of a `Message` that requires no reply is shifted from the sender to a receiver. In this model, the sender's application layer should not have a control to send a `Message` to a specific peer.
- With a req/rep model, 
  - It is unintuitive to implicitly expect a reply for `SendMessageAsync()`.
  - Every initiating `Message`s is a request, hence expects a reply.

In either case, there is no need for `SendMessageAsync()`.

### DIFF
--- a/Libplanet.Net/Transports/TcpTransport.cs
+++ b/Libplanet.Net/Transports/TcpTransport.cs
@@ -193,26 +193,15 @@ namespace Libplanet.Net.Transports
         /// <inheritdoc cref="ITransport.WaitForRunningAsync"/>
         public Task WaitForRunningAsync() => _runningEvent.Task;
 
-        public Task SendMessageAsync(
-            BoundPeer peer,
-            Message message,
-            CancellationToken cancellationToken)
-            => SendMessageWithReplyAsync(
-                peer,
-                message,
-                TimeSpan.FromSeconds(3),
-                0,
-                false,
-                cancellationToken);
-
-        public async Task<Message> SendMessageWithReplyAsync(
+        /// <inheritdoc/>
+        public async Task<Message> SendMessageAsync(
             BoundPeer peer,
             Message message,
             TimeSpan? timeout,
             CancellationToken cancellationToken)
         {
             IEnumerable<Message> replies =
-                await SendMessageWithReplyAsync(
+                await SendMessageAsync(
                     peer,
                     message,
                     timeout,
@@ -224,7 +213,8 @@ namespace Libplanet.Net.Transports
             return reply;
         }
 
-        public async Task<IEnumerable<Message>> SendMessageWithReplyAsync(
+        /// <inheritdoc/>
+        public async Task<IEnumerable<Message>> SendMessageAsync(
             BoundPeer peer,
             Message message,
             TimeSpan? timeout,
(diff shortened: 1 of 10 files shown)


longfin:
I guess `ITransport.SendAsync()` is still needed because sending can be a blocking operation in general manner. (It is a netmq implementation level decision not to consider blocking on transport)

Of course, this may feel invalid as the only valid implementation of `ITransport` currently is `NetMQTrnasport`, and it can't actually work as async. but, considering the contagiousness of `async`, I think that guaranteeing it at the interface level will make future refactoring even a little easier.

greymistcube:
Hmm... I don't quite understand your comment. Is it of your opinion that we need `SendMessageAsync()` as a shorthand fire-and-forget method?

longfin:
Ah, I've misleaded for it that replace `.SendMessageAsync()` to `.SendMessage()`. 

Well, Replacing all `.SendMessageAsync()` to `.SendMessageWithReplyAsync()` seems good. but if so, how about keeping `.SendMessageAsync()` name and changing its behaviour as `.SendMessageWithReplyAsync()`? (because `.SendMessageWithReplyAsync()` name was too verbose IMHO)

greymistcube:
Good point. If a reply is always expected, I guess we can drop the `WithReply` part. 😗

Libplanet.Net/Transports/TcpTransport.cs
```diff
@@ -412,10 +402,12 @@ public void BroadcastMessage(IEnumerable<BoundPeer> peers, Message message)
...
-            foreach (var peer in peers)
-            {
-                _ = SendMessageAsync(peer, message, _runtimeCancellationTokenSource.Token);
-            }
+            peers.AsParallel().ForAll(
+                peer => Task.Run(() => SendMessageAsync(
+                    peer,
+                    message,
+                    TimeSpan.FromSeconds(1),
+                    _runtimeCancellationTokenSource.Token)));
```
longfin, line 410: Do we need to wrap by `Task.Run()` here?

```suggestion
            peers.AsParallel().ForAll(
                peer => SendMessageAsync(
                    peer,
                    message,
                    TimeSpan.FromSeconds(1),
                    _runtimeCancellationTokenSource.Token));
```

greymistcube: I couldn't figure out a better way to do it. Without `Task.Run()`, we get a [CS4014] compiler warning. If we use `async` and `await` inside `ForAll()`, we get an error. 🙁

[CS4014]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014

longfin: Well, got it. `.SendMessageAsync()` was marked as `async` keyword so compiler seemed complain. (Sometimes I get confused between API that just returns a `Task` (e.g. `Task.Delay()`) and marked as `async` 😕).

I think it seems knitpick, but we can write as below.

```suggestion
            peers.AsParallel().ForAll(
                peer => async () => await SendMessageAsync(
                    peer,
                    message,
                    TimeSpan.FromSeconds(1),
                    _runtimeCancellationTokenSource.Token));
```

longfin: 🆗

longfin (approved):

PR #1854, opened 2022-03-15
Port 0.28.2 to main (0.29.x)

dahlia:
Note that I merged 66a1b9f41e7386e50e5976807f3bb3e2fc1afe48 instead of tag 0.28.2, because it does not refer to 66a1b9f41e7386e50e5976807f3bb3e2fc1afe48 the merge commit, but its parent commit instead.  It was my mistake, and it indeed didn't work with the release script on GitHub Actions.  (So I had to manually upload files to NuGet and GitHub releases).

In order to prevent such mistakes from now on, I added a small check to the main workflow.  It ensures if a new tag refers to a proper merge commit or a commit signed by GitHub @web-flow.[^1]

[^1]: As pull requests can be squashed when it's merged, sometimes tags may refer to a non-merge commit.  In such case, the commit is still signed by GitHub @web-flow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,25 @@ jobs:
       if: github.event_name == 'pull_request'
       with:
         ref: ${{ github.pull_request.head.sha }}
+    - if: github.event_name != 'schedule' && github.ref_type == 'tag'
+      run: |
+        set -evx
+        pushd /tmp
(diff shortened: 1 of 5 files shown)


dahlia:
I rebased and added a new check for tags.  See also the description (I wrote it down there).

greymistcube (dismissed):

PR #1838, opened 2022-03-02
Replace ColumnFamilies to key prefix for index db

longfin:
- [x] Implement prefix-based chain ID
- [x] Add migration method and command
- [ ] Add support to https://github.com/planetarium/ninechronicles.snapshot

PR #1862, opened 2022-03-17
Remove column family from index db and add support for migration

limebell:
This is reupload of https://github.com/planetarium/libplanet/pull/1838.


dahlia:
Could you summarize key differences of it from #1838, the previous attempt?

dahlia (approved):

PR #1876, opened 2022-03-30
Increase `Message` receive throughput

### LINKED ISSUE #1864: Local nodes not receiving sufficient `Message`s when using `TURN`

greymistcube:
Temporarily resolves #1864. `16` is just some random number I chose. 😶
This might eventually run into a scaling issue again eventually.

### DIFF
--- a/Libplanet.Stun/Stun/TurnClient.cs
+++ b/Libplanet.Stun/Stun/TurnClient.cs
@@ -18,6 +18,7 @@ namespace Libplanet.Stun
     {
         public const int TurnDefaultPort = 3478;
         private const int AllocateRetry = 5;
+        private const int ProxyCount = 16;
 
         // TURN Permission lifetime was defined in RFC 5766
         // see also https://tools.ietf.org/html/rfc5766#section-8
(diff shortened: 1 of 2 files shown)


Libplanet.Stun/Stun/TurnClient.cs
```diff
@@ -99,7 +99,7 @@ public async Task ReconnectTurn(int listenPort, CancellationToken cancellationTo
                 {
                     if (BehindNAT)
                     {
-                        _turnTasks = BindMultipleProxies(listenPort, 3, _turnTaskCts.Token);
+                        _turnTasks = BindMultipleProxies(listenPort, 16, _turnTaskCts.Token);
```
limebell, line 102: How about making `16` configurable or at least declare as a constant?

greymistcube: Wasn't sure how far I should've taken this.

- I felt icky about exposing more configurable options on top of already crowded options.
  - This is a very low level concern, but as it stands, this option will be passed through `Swarm<T>` to `ITransport` to `TurnClient` which will be *very* confusing at `SwarmOptions` level.

I agree, at least changing it to a constant would be better. 😀

limebell (approved):

PR #1878, opened 2022-03-30
Add `IStore.PruneOutdatedChains()`

### LINKED ISSUE #1874: Add pruning for RocksDbStore
Add pruning to remove all non-canonical chains.

limebell:
Adds `IStore.PruneOutdatedChains()` which deletes all non-canonical chains from the storage.

Closes #1874.

### DIFF
--- a/Libplanet/Store/IStore.cs
+++ b/Libplanet/Store/IStore.cs
@@ -332,5 +332,22 @@ namespace Libplanet.Store
         /// <exception cref="ChainIdNotFoundException">Thrown when the given
         /// <paramref name="sourceChainId"/> does not exist.</exception>
         void ForkTxNonces(Guid sourceChainId, Guid destinationChainId);
+
+        /// <summary>
+        /// Delete all non-canonical chains.
+        /// </summary>
+        /// <param name="noopWithoutCanon">
+        /// Flag to determine whether the function throws exception
+        /// when the canonical chain is not assigned.  <c>false</c> by default.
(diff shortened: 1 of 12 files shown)


Libplanet/Store/IStore.cs
```diff
@@ -332,5 +332,10 @@ void PutBlock<T>(Block<T> block)
         /// <exception cref="ChainIdNotFoundException">Thrown when the given
         /// <paramref name="sourceChainId"/> does not exist.</exception>
         void ForkTxNonces(Guid sourceChainId, Guid destinationChainId);
+
+        /// <summary>
+        /// Delete all non-canonical chains. If canonical chain id is not assigned, does nothing.
```
dahlia, line 337: What does the latter behavior purpose to?  Isn't it okay if it just removes every chain IDs when there is no canonical chain?

limebell: It used to throw `Exception` when canonical chain does not exist and I wrapped it so it does not throw anymore. I'm not sure it is okay to delete all chain IDs when there is no canonical chain because it's almost same with just removing the chain directory (To restore, must append all blocks from genesis block to tip block again).

dahlia: Fair enough.  Still those who haven't read the docs would not expect such behavior though.  How about adding an option like `bool noopWithoutCanon = false`?  It would throw an exception by default when there is no canonical chain, but such exception can be suppressed by turning on the option.

dahlia (approved):

Issue #1885, opened 2022-03-31
Unclear relationship between node behavior, `AppProtocolVersion`, `TrustedAppProtocolSigners`, `DifferentAppProtocolVersionEncountered`.


greymistcube:
Welp, here goes another diagram. The following depicts the current implementation of `Swarm<T>` in regards to `APV` handling:

```mermaid
flowchart TD
    start --> |remote message received with an APV| Q1[is the remote APV the same as the local APV?]
    Q1 --> |yes| A[accept]
    Q1 --> |no| Q2[is the remote APV signed by a trusted signer?]
    Q2 --> |yes| APV[trigger different-APV-encountered callback]
    APV --> Q3[is different APV allowed?]
    Q2 --> |no| Q4[is the received message a request?]
    Q4 --> |yes| R[reply with a different-version-message]
    Q4 --> |no| I[ignore]
    Q3 --> |yes| A
    Q3 --> |no| Q4
    R --> I
    style Q1 fill:yellow
    style Q2 fill:yellow
    style Q3 fill:yellow
    style Q4 fill:yellow
```

I'd like to raise several issues:
- If I'm not mistaken, we are using `DifferentAppProtocolVersionEncountered` to trigger an application level event that initiates an update, this should be made explicit that `DifferentAppProtocolVersionEncountered` is *only triggered* when it is from a trusted source. The current description suggests that `DifferentAppProtocolVersionEncountered` will be triggered regardless of the trustiness of a different `APV`.
- Whether to trigger `DifferentAppProtocolVersionEncountered` and what to do with `Message`s with a different `APV` are two separate concerns.
  - Personally, I don't think there should be an option to decide whether `Message`s with a different `APV` is allowed. I am highly skeptical whether network would behave properly if `Message`s with a different `APV` are allowed. I understand the underlying `BlockChain<T>` is `APV` agnostic and ideally we want as much freedom as possible for individual nodes to possibly have different `APV`s, but as an ecosystem, I fail to see a scenario where this would *actually* be useful.
  - An `APV` can be thought of as a versioning of the state transition machine owned by a node. This means, the same `APV` is supposed to guarantee the evaluation hash of a `Block<T>`. If different `APV`s are allowed on the same network, this may result in hard forks between different `APV`s, in which case, further communication becomes moot anyways.

greymistcube:
I suggest the following scheme:

```mermaid
flowchart TD
    start --> |remote message received with an APV| Q1[is the remote APV the same as the local APV?]
[...]

PR #1889, opened 2022-04-04
Decoding related exceptions refactored

greymistcube:
This is mostly to reduce noise on a larger refactoring task as part of #1885 that I am working on. 😶

### DIFF
--- a/Libplanet.Net/DifferentAppProtocolVersionException.cs
+++ b/Libplanet.Net/DifferentAppProtocolVersionException.cs
@@ -22,25 +22,25 @@ namespace Libplanet.Net
         /// received.</param>
         /// <param name="identity">The <see cref="Message.Identity"/> of the <see cref="Message"/>
         /// received.</param>
-        /// <param name="appProtocolVersion">The protocol version of
+        /// <param name="expectedAppProtocolVersion">The protocol version of
         /// the local <see cref="Swarm{T}"/>.</param>
-        /// <param name="peerAppProtocolVersion">The protocol version of the <see cref="Peer"/> that
-        /// the local <see cref="Swarm{T}"/> is trying to connect to.</param>
-        /// <param name="trusted">Whether <paramref name="peerAppProtocolVersion"/>
+        /// <param name="actualAppProtocolVersion">The protocol version of the <see cref="Peer"/>
+        /// that the local <see cref="Swarm{T}"/> is trying to connect to.</param>
+        /// <param name="trusted">Whether <paramref name="actualAppProtocolVersion"/>
         /// is signed by a trusted signer.</param>
         public DifferentAppProtocolVersionException(
             string message,
             Peer peer,
             byte[] identity,
-            AppProtocolVersion appProtocolVersion,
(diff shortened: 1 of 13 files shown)


riemannulus (dismissed):
LGTM

Libplanet.Net/DifferentAppProtocolVersionException.cs
```diff
@@ -47,13 +68,28 @@ internal sealed class DifferentAppProtocolVersionException : Exception
         /// <summary>
         /// The protocol version of the current <see cref="Swarm{T}"/>.
         /// </summary>
-        public AppProtocolVersion ExpectedVersion { get; }
+        public AppProtocolVersion Apv { get; }
```
longfin, line 71: `ExpectedVersion` / `ActualVersion` pair is more natual for me than `Apv` / `PeerApv`. 🤔 

@planetarium/libplanet Do you have idea about it?

Libplanet.Net/Protocols/KademliaProtocol.cs
```diff
@@ -150,22 +150,10 @@ public class KademliaProtocol : IProtocol
...
-                    ByteUtil.Hex(expected.Signature),
-                    expected.Signer.ToString(),
-                    actual.Version,
-                    actual.Extra,
-                    ByteUtil.Hex(actual.Signature),
-                    actual.Signer
-                );
+                    e,
+                    "Different version encountered during {FName}().",
+                    nameof(AddPeersAsync));
```
limebell, line 156: Don't we need detailed log of version information? In some environment `e` is not logged (ex. testing)

Libplanet.Net/DifferentAppProtocolVersionException.cs
riemannulus: Then what about `ExpectedApv` / `ActualApv`?

greymistcube: I agree, as an `Exception`, yes, `Expected` / `Actual` is probably more appropriate. On the other hand, this particular `Exception` carries lot more info than it really should just in order to be used as "context" in certain cases. 😐

Libplanet.Net/Protocols/KademliaProtocol.cs
greymistcube: I haven't really looked into this, but is there a reason why `e` can't be logged? If possible, I think the right solution would be to print `e` instead of nearly duplicating `e.Message`. 🤨

Libplanet.Net/DifferentAppProtocolVersionException.cs
greymistcube: Eh, sure why not. On second thought, I guess `Expected` / `Actual` might be better.

greymistcube: See da61eb5d748fdcebbbe710ee7d8b46dc2d8cf2b5. Even after changing, either way seems a bit weird tho. 🙄

Issue #1867, opened 2022-03-22
Move `Identity` handling from `Swarm<T>` to `ITransport`


greymistcube:
`Swarm<T>` should be agnostic about the underlying reply handling mechanism.

The `Message.Identity` property is used to identify for `ITransport` to figure out where to send a `Message`. Currently, setting the correct value to `Message.Identity` is handled by `Swarm<T>` manually when `Swarm<T>.ProcessMessageHandlerAsync()` is invoked. This requires `Swarm<T>` to extensively know how `ITransport` works where it really shouldn't. That is, `Message.Identity` should be generally hidden from `Swarm<T>` and handled internally by `ITransport`.

greymistcube:
This is probably long overdue, but for the sake of sanity, although strictly not a proper UML sequence diagram, here is the current scheme for `NetMQTransport`.

For everyone's convenience:
- Solid arrowhead: synchronous call
- Normal arrowhead: asynchronous call
- Dashed line: method return 

## Receiving Messages

Note: Below, queueing a reply message is done by calling `NetMQTransport.ReplyMessageAsync()` asynchronously, as each message handling is done in its own thread, there is no need for sending a reply to be asynchronous, which is why the diagram is depicted with a synchronous arrow. In my opinion, `async` is overused at the moment.

```mermaid
%%{init: {'theme': 'neutral'} }%%
sequenceDiagram
  participant WEB as Web
  participant RECS as Receive Socket<br/>with internal queue
  participant RECP as Receive Poller
  participant T1 as Thread 1
  participant T2 as Thread 2
  participant REPQ as Reply Queue<br/>as NetMQQueue
  Note over WEB, RECS: NetMQ
  Note over T1, T2: Thread Pool

  WEB -) RECS: receive and enqueue request message
  RECP ->> RECS: ask for queued request message<br/>NetMQSocket.Receive()
  activate RECP
  RECS -->> RECP: dequeue request message
  RECP -) T1: spawn new thread to process message
  deactivate RECP
  activate T1
  WEB -) RECS: receive and enqueue request message
  RECP->> RECS: ask for queued request message<br/>NetMQSocket.Receive()
  activate RECP
  WEB -) RECS: receive and enqueue request message
  RECS -->> RECP: dequeue request message
  RECP-) T2: spawn new thread to process message
  deactivate RECP
  activate T2
  T2 ->> REPQ: enqueue reply message
  deactivate T2
  T1 ->> REPQ: enqueue reply message
[...]

PR #1890, opened 2022-04-05
Separate `MessageValidator` for `IMessageCodec`s

### LINKED ISSUE #1885: Unclear relationship between node behavior, `AppProtocolVersion`, `TrustedAppProtocolSigners`, `DifferentAppProtocolVersionEncountered`.

greymistcube:
Closes #1885 for now.

Aside from generally dealing with #1885, `IMessageCodec` instances now have a fixed context for encoding and decoding in regards to `APV`s. We can also probably do away with `privatekey` and `timestamp` considering its lifecycle.

### DIFF
--- a/Libplanet.Net.Tests/SwarmTest.AppProtocolVersion.cs
+++ b/Libplanet.Net.Tests/SwarmTest.AppProtocolVersion.cs
@@ -1,4 +1,3 @@
-#nullable disable
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
--- a/Libplanet.Net.Tests/SwarmTest.AppProtocolVersion.cs
+++ b/Libplanet.Net.Tests/SwarmTest.AppProtocolVersion.cs
@@ -1,4 +1,5 @@
 #nullable disable
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
@@ -150,8 +151,21 @@ namespace Libplanet.Net.Tests
 
                 foreach (var peer in peers)
                 {
-                    await a.AddPeersAsync(new[] { peer }, null);
-                    await b.AddPeersAsync(new[] { peer }, null);
+                    try
+                    {
+                        await a.AddPeersAsync(new[] { peer }, TimeSpan.FromSeconds(1));
+                    }
+                    catch (PingTimeoutException)
+                    {
+                    }
+
+                    try
+                    {
+                        await b.AddPeersAsync(new[] { peer }, TimeSpan.FromSeconds(1));
+                    }
+                    catch (PingTimeoutException)
+                    {
+                    }
                 }
 
                 Assert.Equal(new[] { c.AsPeer }, a.Peers.ToArray());
@@ -173,11 +187,15 @@ namespace Libplanet.Net.Tests
             {
                 await StopAsync(c);
                 await StopAsync(d);
+                await StopAsync(e);
+                await StopAsync(f);
 
                 a.Dispose();
                 b.Dispose();
                 c.Dispose();
                 d.Dispose();
+                e.Dispose();
+                f.Dispose();
             }
         }
     }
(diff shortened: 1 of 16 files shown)


greymistcube:
Eh, on second thought, making `IMessageCodec` context agnostic might be a better decision. 😗
In any case, this should make it easier to move things around in preparation for #1867.

Libplanet.Net.Tests/SwarmTest.AppProtocolVersion.cs
```diff
@@ -152,8 +150,21 @@ AppProtocolVersion localVersion
...
                 {
-                    await a.AddPeersAsync(new[] { peer }, null);
-                    await b.AddPeersAsync(new[] { peer }, null);
+                    try
+                    {
+                        await a.AddPeersAsync(new[] { peer }, TimeSpan.FromSeconds(1));
+                    }
+                    catch (PingTimeoutException)
+                    {
+                    }
```
riemannulus, line 159: eh, is it ok to ignore `PingTimeoutException`?

limebell (approved):

Libplanet.Net.Tests/SwarmTest.AppProtocolVersion.cs
greymistcube: The reason for this is that the behaviors of contacted nodes have changed.

- Before: Nodes replied to *any* message regardless of `APV` signer was trustworthy or not.
- After: Nodes *ignore* messages with an `APV` that is not signed by a trusted source.

limebell: Then how about use `Assert.Throws<T>()` instead if the exception is always thrown?

greymistcube: See 9be99f8d8966fe6438a41348a3bdaa6bec54e3be.

riemannulus (approved):
LGTM

Issue #1907, opened 2022-04-11
Make `ITransport.SendMessageAsync()` fail more gracefully.


greymistcube:
Currently, the method can throw too many different types of `Exception`s, and `Exception` type is used as elaborate flow control, which is considered an [anti-pattern](https://docs.microsoft.com/en-us/visualstudio/profiling/da0007-avoid-using-exceptions-for-control-flow?view=vs-2017) in general. Unless there is a critical failure, callers do not really need to know how `ITransport.SendMessageAsync()` failed, but are only interested in its return value. It would be better to suppress `Exception`s inside `ITransport.SendMessageAsync()` and return whatever `Message`s it managed to get as replies. 😶

greymistcube:
Also, at the moment, all those `Exception` types used for flow control isn't even documented in `ITransport`. 🙄

PR #1911, opened 2022-04-14
Plug `ITransport` `Exception`s leak

greymistcube:
Follow up to #1906 and preparatory work for #1907.

Tries to plug as many holes as possible for `ITransport` logic leaking through `Exception`s into `Swarm<T>` and `IProtocol`. 😗


limebell (approved):
LGTM

PR #1912, opened 2022-04-18
`TxPolicyViolationException` should contain `InnerException`

### LINKED ISSUE #1830: `TxPolicyViolationException` should contain `InnerException`
`BlockChain<T>.Append()` method currently throws `TxPolicyViolationException` without filling its `InnerException` if a transaction is invalid according to the `BlockChain<T>.Policy`.  Although such behavior is not a bug, but it makes hard to investigate why it is invalid.

It would be better if it includes the original exception made by `IBlockPolicy<T>.ValidateNextBlockTx()` as its `InnerException`.  See also:

<https://github.com/planetarium/libplanet/blob/be4dd0195d48b7eb08d85735af660d20fc2ac01f/Libplanet/Blockchain/BlockChain.cs#L888-L893>

colibrishin:
This PR resolves #1830.

> `TxPolicyViolationException` and `InvalidTxException` take an inner-exception by  nullable parameter. If there's no exception to be caught, set `inner`, the inner-exception parameter, null. 

Also, I think `inner` shouldn't be null default parameter for raising attention. which is to double-check the existence of an exception that should be caught.

Following `TxPolicyViolationException` and `InvalidTxException`  inherited exceptions changes are also fixed in this PR.

_( + InvalidTxIdException is never used.)_

### DIFF
--- a/Libplanet/Tx/InvalidTxException.cs
+++ b/Libplanet/Tx/InvalidTxException.cs
@@ -15,13 +15,15 @@ namespace Libplanet.Tx
         /// <see cref="Transaction{T}.Id"/>.  It is automatically included to
         /// the <see cref="Exception.Message"/> string.</param>
         /// <param name="message">A descriptive error message for programmers.
+        /// </param> <param name="inner">A Inner-exception.
         /// </param>
-        protected InvalidTxException(TxId txid, string message)
-            : base($"{txid.ToHex()}: {message}")
+#nullable enable
+        protected InvalidTxException(TxId txid, string message, Exception? inner)
+            : base($"{txid.ToHex()}: {message}", inner)
         {
             TxId = txid;
         }
-
+#nullable disable
         /// <summary>
         /// The <see cref="TxId"/> of the invalid <see cref="Transaction{T}"/>.
         /// </summary>
--- a/Libplanet/Tx/InvalidTxException.cs
+++ b/Libplanet/Tx/InvalidTxException.cs
@@ -16,8 +16,28 @@ namespace Libplanet.Tx
         /// the <see cref="Exception.Message"/> string.</param>
         /// <param name="message">A descriptive error message for programmers.
         /// </param>
-        protected InvalidTxException(TxId txid, string message)
+        protected InvalidTxException(
+            TxId txid,
+            string message)
             : base($"{txid.ToHex()}: {message}")
+        {
+            TxId = txid;
+        }
+
+        /// <summary>
+        /// Instantiates a new exception object with proper metadata.
+        /// </summary>
+        /// <param name="txid">The invalid <see cref="Transaction{T}"/>'s
+        /// <see cref="Transaction{T}.Id"/>.  It is automatically included to
+        /// the <see cref="Exception.Message"/> string.</param>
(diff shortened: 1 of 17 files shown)


riemannulus (commented):
Is it some case `TxPolicyViolationException.InnerException` is `null`? I mean, according to #1830, there is a no way to empty `TxPolicyViolationException.InnerException`.

Libplanet/Tx/InvalidTxException.cs
```diff
@@ -15,13 +15,15 @@ public abstract class InvalidTxException : Exception
...
+        /// </param> <param name="inner">A Inner-exception.
         /// </param>
-        protected InvalidTxException(TxId txid, string message)
-            : base($"{txid.ToHex()}: {message}")
+#nullable enable
+        protected InvalidTxException(TxId txid, string message, Exception? inner)
+            : base($"{txid.ToHex()}: {message}", inner)
         {
             TxId = txid;
         }
```
greymistcube, line 25: Two things:
- It'd be better to have `InvalidTxExceptioin(TxId, string)` *and* `InvalidTxException(TxId, string, Exception)` separately.
  - This would get rid of unnecessary cluttered use of `null`s everywhere.
  - This would be in better accordance to built-in `Exception`s.
- Change the parameter name `inner` to `innerException`. Again, this is what built-in `Exception`s use. 😗

colibrishin:
> Is it some case `TxPolicyViolationException.InnerException` is `null`? I mean, according to #1830, IMO there is a no way to empty `TxPolicyViolationException.InnerException`.

If i understood correctly, No. there aren't any case that intentionally `TxPolicyViolationException.InnerException` is being `null`. 

My intention was if 'null' is default behavior of not having inner-exception, then one might could design giving null if there is no exception, for clarity.

_('Exception.InnerException')[https://docs.microsoft.com/ko-kr/dotnet/api/system.exception.-ctor?view=net-6.0#system-exception-ctor(system-string)] is null if it initialized without exception._

dahlia (changes requested):
Could you add unit tests to ensure if these new overloads actually initialize `InnerException` property (and the existing ones don't)?

Also these new APIs should be recorded in the changelog.

colibrishin:
> Could you add unit tests to ensure if these new overloads actually initialize `InnerException` property (and the existing ones don't)?
> 
> Also these new APIs should be recorded in the changelog.
> 
> You need to rebase your commits on the current main branch. Use [`git pull` command's `-r`/`--rebase` option](https://git-scm.com/docs/git-pull#Documentation/git-pull.txt--r).

Sure thing!

colibrishin:
I have Pull & Rebased PR repository. Plus, Added test codes for if TxPolicy. (Sorry for misleading commit titles... I am kinda bit dizzy right now. idk why though.)

First test case is with assumption that we have a blockchain which is somehow always throws Exception at ValidateNextBlockTx.
Second test case, "inner-exception test code", is the case that only catches ValidateNextBlockTx.

Libplanet/Tx/InvalidTxException.cs
colibrishin: Updated!

PR #1916, opened 2022-04-19
`TxPolicyViolationException` should contain `InnerException`

### LINKED ISSUE #1830: `TxPolicyViolationException` should contain `InnerException`

colibrishin:
This is squashed PR of #1912, and resolves #1830.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,12 +18,19 @@ To be released.
 
 ### Behavioral changes
 
+ -  `TxPolicyViolationException` and `InvalidTxException` now have a new
+    constructor that takes an `Exception` as their `Exception.InnerException`.
+    [[#1830], [#1912], [#1916]]
+
 ### Bug fixes
 
 ### Dependencies
 
 ### CLI tools
 
+[#1830]: https://github.com/planetarium/libplanet/issues/1830
+[#1912]: https://github.com/planetarium/libplanet/pull/1912
+[#1916]: https://github.com/planetarium/libplanet/pull/1916
 
 Version 0.33.0
 --------------
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -2005,10 +2005,85 @@ namespace Libplanet.Tests.Blockchain
             Assert.Equal(4, _blockChain.StagePolicy.Iterate(_blockChain, filtered: false).Count());
         }
 
+        [Fact]
+        private void CheckIfTxPolicyExceptionHasInnerException()
+        {
+            var policy = new NullPolicyButTxPolicyAlwaysThrows<DumbAction>(
+                x =>
+                {
+                    // ReSharper disable AccessToModifiedClosure
+                    // The following method calls should not throw any exceptions:
+                    x?.GetStates(new[] { default(Address) });
+                    x?.GetState(default);
+                    // ReSharper restore AccessToModifiedClosure
+                });
+            var store = new MemoryStore();
+            var stateStore = new TrieStateStore(new MemoryKeyValueStore());
+            Block<DumbAction> genesisWithTx = MineGenesis(
+                policy.GetHashAlgorithm,
+                GenesisMiner.PublicKey,
+                new[]
+                {
+                    Transaction<DumbAction>.Create(
+                        0,
+                        new PrivateKey(),
+                        null,
+                        Array.Empty<DumbAction>()
+                    ),
+                }
+            ).Evaluate(GenesisMiner, policy.BlockAction, stateStore);
+
+            Assert.Throws<TxPolicyViolationException>(() =>
+            {
+                try
+                {
+                    var chain = new BlockChain<DumbAction>(
+                        policy,
+                        new VolatileStagePolicy<DumbAction>(),
+                        store,
+                        stateStore,
+                        genesisWithTx
+                    );
+                }
+                catch (TxPolicyViolationException e)
+                {
+                    Assert.NotNull(e.InnerException);
+
+                    // This is visual representation of the test.
+                    // In actual use-case, this debugging lines do
+                    // not show.
+                    _logger.Debug("Exception: {Exception}", e.ToString());
+                    _logger.Debug("InnerException: {InnerException}", e.InnerException.ToString());
+                    throw e;
+                }
+            });
+        }
+
+        private class
+            NullPolicyButTxPolicyAlwaysThrows<T> : NullPolicyForGetStatesOnUninitializedBlockChain<
+                T>
+            where T : IAction, new()
+        {
+            public NullPolicyButTxPolicyAlwaysThrows(
+                Action<BlockChain<T>> hook)
+                : base(hook)
+            {
+            }
+
+            public override TxPolicyViolationException ValidateNextBlockTx(
+                BlockChain<T> blockChain,
+                Transaction<T> transaction
+            )
+            {
+                _hook(blockChain);
+                return new TxPolicyViolationException(transaction.Id, "Test Message");
+            }
+        }
+
         private class NullPolicyForGetStatesOnUninitializedBlockChain<T> : NullBlockPolicy<T>
             where T : IAction, new()
         {
-            private readonly Action<BlockChain<T>> _hook;
+            protected readonly Action<BlockChain<T>> _hook;
 
             public NullPolicyForGetStatesOnUninitializedBlockChain(
                 Action<BlockChain<T>> hook
--- a/Libplanet/Tx/InvalidTxException.cs
+++ b/Libplanet/Tx/InvalidTxException.cs
@@ -9,7 +9,9 @@ namespace Libplanet.Tx
     public abstract class InvalidTxException : Exception
     {
         /// <summary>
-        /// Instantiates a new exception object with proper metadata.
+        /// Instantiates a new exception object with proper metadata. Use
+        /// <see cref="InvalidTxException(TxId, string, Exception)"/> if innerException
+        /// is required.
         /// </summary>
         /// <param name="txid">The invalid <see cref="Transaction{T}"/>'s
         /// <see cref="Transaction{T}.Id"/>.  It is automatically included to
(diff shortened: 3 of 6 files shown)


riemannulus:
Need changelog fix 😢

colibrishin:
Commits are squashed and changelog finally has been fixed. 😊

Libplanet/Tx/InvalidTxException.cs
```diff
@@ -16,12 +16,32 @@ public abstract class InvalidTxException : Exception
...
 
+        /// <summary>
+        /// Instantiates a new exception object with proper metadata.
+        /// </summary>
+        /// <param name="txid">The invalid <see cref="Transaction{T}"/>'s
+        /// <see cref="Transaction{T}.Id"/>.  It is automatically included to
+        /// the <see cref="Exception.Message"/> string.</param>
+        /// <param name="message">A descriptive error message for programmers.
+        /// </param> <param name="innerException">The <see cref="Exception"/> for
+        /// <see cref="Exception.InnerException"/>.</param>
```
riemannulus, line 35: trivial, but...
```suggestion
        /// </param> 
        /// <param name="innerException">The <see cref="Exception"/> for
        /// <see cref="Exception.InnerException"/>.</param>
```

colibrishin: And like, this?

https://github.com/colibrishin/libplanet/blob/9c8fe16a2bb18b8e239130c2285c98e52e79af59/Libplanet/Tx/TxPolicyViolationException.cs#L32-L34

CHANGES.md
```diff
@@ -73,6 +73,9 @@ Released on April 19th, 2022.
     [[#1866], [#1891]]
  -  (Libplanet.Net) `IMessageCodec.Encode()` now requires a *matching*
     `PrivateKey` to be provided for `Peer`.  [[#1904], [#1905]]
+ -  `TxPolicyViolationException` and `InvalidTxException` now have a new
+    constructor that takes an `Exception` as their `Exception.InnerException`.
+    [[#1830], [#1912], [#1916]]
```
dahlia, line 78: Apparently these reference links are not [defined](https://spec.commonmark.org/0.30/#link-reference-definition).  Usually we place definitions below the section:

https://github.com/colibrishin/libplanet/blame/cb6b0b3091bbbc456e5b1f9f96dea2597931fc2c/CHANGES.md#L87-L92

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2005,10 +2005,82 @@ private async Task FilterLowerNonceTxAfterStaging()
...
+            }
+            catch (TxPolicyViolationException e)
+            {
+                Assert.NotNull(e.InnerException);
+
+                // This is visual representation of the test.
+                // In actual use-case, this debugging lines do
+                // not show.
+                _logger.Debug(e.ToString());
+                _logger.Debug(e.InnerException.ToString());
```
dahlia, line 2055: The first parameter of logging methods take a template string, rather than a bare string, which means it can misbehave if its first argument contains some placeholders like `{0}`.  I know `e.ToString()` and `e.InnerException.ToString()` probably does not contains such things at the moment, someone might change these methods in the future so that their return values contain such placeholder syntax.

```suggestion
                _logger.Debug("Exception: {0}", e.ToString());
                _logger.Debug("InnerException: {0}", e.InnerException.ToString());
```

Libplanet/Tx/TxPolicyViolationException.cs
```diff
@@ -23,5 +23,21 @@ public TxPolicyViolationException(TxId txid, string message)
             : base(txid, message)
         {
         }
+
+        /// <inheritdoc cref="Libplanet.Tx.TxPolicyViolationException" />
```
dahlia, line 27: What's your intention on this?  I'm curious why it inherits docs from class `TxPolicyViolationException`.

colibrishin: I was thinking that `TxPolicyViolationException` have same summary naively. 
Now think about this, It will be better if one can notice that also there's a constructor for innerException from summary.

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2005,10 +2005,82 @@ private async Task FilterLowerNonceTxAfterStaging()
             Assert.Equal(4, _blockChain.StagePolicy.Iterate(_blockChain, filtered: false).Count());
         }
 
+        [Fact]
+        private void CheckIfTxPolicyExceptionHasInnerException()
+        {
+            IValue value = null;
```
limebell, line 2011:
```suggestion
```
This variable is only assigned, not used in anywhere.

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2005,10 +2005,82 @@ private async Task FilterLowerNonceTxAfterStaging()
...
+            var policy = new NullPolicyButTxPolicyAlwaysThrows<DumbAction>(
+                c =>
+                {
+                    // ReSharper disable AccessToModifiedClosure
+                    // The following method calls should not throw any exceptions:
+                    c?.GetStates(new[] { default(Address) });
+                    value = c?.GetState(default);
+                    // ReSharper restore AccessToModifiedClosure
```
limebell, line 2019:
```suggestion
                    // The following method calls should not throw any exceptions:
                    c?.GetStates(new[] { default(Address) });
                    c?.GetState(default);
```

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2005,10 +2005,82 @@ private async Task FilterLowerNonceTxAfterStaging()
...
+            }
+            catch (TxPolicyViolationException e)
+            {
+                Assert.NotNull(e.InnerException);
+
+                // This is visual representation of the test.
+                // In actual use-case, this debugging lines do
+                // not show.
+                _logger.Debug("Exception: {0}", e.ToString());
+                _logger.Debug("InnerException: {0}", e.InnerException.ToString());
```
limebell, line 2055:
```suggestion
                _logger.Debug("Exception: {Exception}", e.ToString());
                _logger.Debug("InnerException: {InnerException}", e.InnerException.ToString());
```
Named properties is suggested in Serilog structured logging instead of positional ones.

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2005,10 +2005,82 @@ private async Task FilterLowerNonceTxAfterStaging()
...
+            {
+                var chain = new BlockChain<DumbAction>(
+                    policy,
+                    new VolatileStagePolicy<DumbAction>(),
+                    store,
+                    stateStore,
+                    genesisWithTx
+                );
+            }
+            catch (TxPolicyViolationException e)
```
limebell, line 2047: This `catch` block is called only when `try` block throws exception, but it's not guaranteed. How about using `Assert.Throws<T>()` to insure exception is thrown.
```csharp
Assert.Throws<TxPolicyViolationException>(() => {
    try
    {
        ....
    }
    catch (TxPolicyViolationException e)
    {
        ....

        throw;
    }
});
```

dahlia (commented):
Could you squash these three commits into one?  I believe those three patches have no reason to be distinct each other:

- 7e50cf463a627c3f72adb4b9f090655df50d8643
- 3930262716e144131a05cc4fed44323cf79b192d
- 08d5dc23e3927f919f4bb10fb31dd4c524f37fad

Libplanet/Tx/InvalidTxException.cs
```diff
@@ -16,12 +16,33 @@ public abstract class InvalidTxException : Exception
         /// the <see cref="Exception.Message"/> string.</param>
         /// <param name="message">A descriptive error message for programmers.
         /// </param>
-        protected InvalidTxException(TxId txid, string message)
+        protected InvalidTxException(
+            TxId txid,
+            string message)
```
limebell, line 21: This change seems not required.

Libplanet/Tx/TxPolicyViolationException.cs
```diff
@@ -11,7 +11,9 @@ namespace Libplanet.Tx
     public class TxPolicyViolationException : InvalidTxException
     {
         /// <summary>
-        /// Creates a new <see cref="TxPolicyViolationException"/> instance.
+        /// Creates a new <see cref="TxPolicyViolationException"/> instance. Use
+        /// <see cref="TxPolicyViolationException(TxId, string, Exception)"/> if innerException
+        /// is required.
```
limebell, line 16: It would be great if we add this statement to the document of `InvalidTxException` too.

limebell (commented):
Great work. I've added some style-related comments. :)

Libplanet.Tests/Blockchain/BlockChainTest.cs
colibrishin: I should get used to lambda function, the `Action` (in C#?). 😅

colibrishin:
Following suggestions have been applied. 🤖

dahlia (approved):

limebell (dismissed):
LGTM :)

dahlia:
@riemannulus Do you have anything you want @colibrishin to change more?

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2005,10 +2005,85 @@ private async Task FilterLowerNonceTxAfterStaging()
...
+        {
+            var policy = new NullPolicyButTxPolicyAlwaysThrows<DumbAction>(
+                c =>
+                {
+                    // ReSharper disable AccessToModifiedClosure
+                    // The following method calls should not throw any exceptions:
+                    c?.GetStates(new[] { default(Address) });
+                    c?.GetState(default);
+                    // ReSharper restore AccessToModifiedClosure
+                });
```
riemannulus, line 2019: It's trivial but we use the variable `x` in situation like this.

limebell (approved):

dahlia (approved):

riemannulus (approved):
+1

PR #1917, opened 2022-04-19
Improve `BlockCandidateTable<T>`

### LINKED ISSUE #1892: The chain follows the shortest neighbor
Since #1699, the logic about block sync is changed like that.

```mermaid
sequenceDiagram
autonumber
  participant NW as Network
  participant MH as MessageHandler()
  participant FB as FillBlocksAsync()
  participant BCT as BlockCandidateTable
  participant CBC as ConsumeBlockCandidates()
  NW->>MH: Add BlockDemand
  loop download blocks
    FB-)MH: Request all BlockDemand
    MH->>FB: Return all BlockDemand
    par download block with parallel
      FB->>FB: Download Blocks
      FB->>BCT: Add Blocks to BlockCandidateTable
    end
  end
  loop Append blocks
    CBC-)BCT: Get blocks by current tip
    BCT->>CBC: Return blocks
    loop Append block to BlockChain<T>
      CBC->>CBC: Append block to BlockChain<T>
    end 
  end
```

And the following flowchart is about `[download block with parallel]`.

```mermaid
flowchart
    MS[Start] -- BlockDownload start by n --> P1[Parallel 1] 
    P1 --> A[Download blocks until n + 10 from node A] & B[Download blocks until n + 20 from node B]
    A -- Download blocks at 10 sec --> P2[Parallel 2]
    B -- Download blocks at 20 sec --> P2
    P2 --> C[Try to add blocks to BlockCandidatesTable]
    C --> D[Check already downloaded blocks using current tip as a key]
    D --> Q1{is it?}
    Q1 --> |Yes| Ignore
    Q1 --> |No| E[Add to BlockCandidatesTable]
    E --> F[BlockCandidatesTable]
    Ignore -- loop --> MS
    E -- loop --> MS
    subgraph Add to BlockChain
        F -- Get blocks using current BlockChain tip as a key --> SA[Add blocks to BlockChain]
        SA -- Finish Add --> SB[Remove outdated candidates]
        SB -- loop --> F
    end 
    style P1 fill:black,color:black
    style P2 fill:black,color:black
```

`BlockCandidateTable` is a key-value table and it is use a `BlockChain<T>.Tip` as a key when starting download blocks.
When n is `BlockChain<T>.Tip.Index` and n is 10, download blocks from node A from 10 to 20 and from node B from 10 to 30.
In this case, node A's chain downloaded way faster than node B's chain.

So, in this case, start to `Add to BlockChain` flow with node A's chain. and nod B's chain is deprecated on the `Remove outdated candidates` process.
Because blocks downloaded from node B are calculated locator from n, not n + 10.

In this case, you may have some questions.
1. Why are dropped all blocks is the problem?
    - Imagine if some node, we called it node A in here, sending the shortest blocks so cannot start to `Add to BlockChain` flow if the longest blocks download from node B?
    And if it is repeated? the current node cannot follow node B. It causes a sync problem.
    - The most important thing is until append blocks from node A to the current chain, download blocks from node B from current tip 'n'. So we cannot use downloaded blocks until block append is finished.
2. Why dropped all blocks? why don't use node B's left blocks?
    - If the chain is not reorged, can use node B's left blocks. (e.g from n + 10 to n + 20) But what if node A's chain occurred reorg and node B's chain is not? 
    For example, download blocks from node A from n - 10 to n + 10 and from node B from n to n + 30. In this case, at first, node A is synced with reorg.
    after then, try syncing node B but it cannot, node B does not compatible with node A's chain. Cannot find branchpoint too.
    - IMO, it may cause deep reorg because the current chain is too late to recognize the chain is fallen local optima.

colibrishin:
This resolves #1892.

I'd like to recommend checking the original issue. The following details are well-documented and discussed in the issue.

### Abstract
This improvement is based on the assumption that the current `Swarm<T>` misses the opportunity of getting the longest chain. 

For example, `Swarm<T>` got two branches in parallel, which are oriented from the same ancestor (BranchPoint), from two nodes for each. One's has 10 blocks, and the other has 20 blocks. Now, 10 blocks come in first (e.g.. Downloaded or Processed first) after then, those blocks are appended into `Swarm<T>` and change `Swarm<T>.Blockchain.Tip` to the Tip of 10 blocks.  

20 blocks will be considered lost due to the current Tip policy. Because the `PreviousHash` of `Root` of 20 blocks `Branch` is not the same as `Tip` in `Swarm<T>` anymore. 

If we could trace the branch and `Tip` where we can connect, then `Swarm<T>` might get additional 10 blocks.

### Word
- `Branch` : A set of `Block<T>`s that every `Block<T>`s have their pair that has the same `Hash` and `PreviousHash`, except the first element. This is used in updating a `Tip` of `Blockchain<T>`.
- `Root` : A `Block<T>` that is the first element in a `Branch`.
- `OldTip` : A `Tip` of `Blockchain<T>` currently have.
- `NewTip` : A `Block<T>` that is the last element in a `Branch`. the _New_ prefix appears because if this branch is appended with a `Blockchain<T>`, then this `Block<T>` will be the new `Tip` of `Blockchain<T>`.
- `Branchpoint` : A `Block<T>` where the `PreviousHash` of `Root` of `Branch` is same as `OldTip`.

### What's new

- `UpdatePath<T>`
  - A path from `Blockchain<T>`. `Blocks` is `Block<T>` to `Tip` which are continuous `Block<T>`s in `PreviousHash`-wise for each `Block<T>`.

- `BlockCandidateTable.Add`
   - `Add()` takes branch only if `CandidateBranch<T>` has a higher comparer value than `Blockchain<T>` and Blocks has their PreviousHash Blocks except for the first block of the list. 
   - Because of where `BlockCandidateTable.Add()` is called, `BlockCandidateTable.Update()` is called with `UpdatePath<T>(oldTip, oldTip)` when `BlockCandidateProcess()` is failed.

- `BlockCandidateTable.Update`
  Before going through the explanation, we need to make sure if the `CandidateBranch.Tip` is not higher than the current tip of the `Blockchain` branch `TotalDifficulty`, then the branch will be discarded. 
For convenience, I will use the term `Tip` as the current `Blockchain` Tip. any other `Tip` will be explained with its term.

  - Case 1. The reorganization did not happen (Branchpoint is equal to the current `Blockchain` branch)
    - A. We define a `LongestBranch`, which is a branch that has the highest `TotalDifficulty` in `Branches`.
    - B. Pick a `CandidateBranch<T>`, Collect all the blocks from it, In the condition of having a high `Block.Index` then `Tip.Index`.
    - C. If `TotalDifficulty` of blocks (`Branch`) from B is higher than the current `LongestBranch`, then replace `LongestBranch`.
    - D. Go back to B and repeat until visit the all the other `Branches`.

  - Case 2. The reorganization did happen (Branchpoint is not equal to the current `Blockchain` branch)
    - A. We define `LongestBranch` the same as Case 1.
    - B.  Pick a `CandidateBranch<T>`, Find the longest block from `Root`  (= its most previous block) to `BranchTip`
    - C. If the previous block of `BranchRoot` is the same as `Tip`, then insert `BranchPoint` into the branch as `BranchRoot`.
    - D. If `TotalDifficulty` of blocks (`Branch`) from B is higher than the current `LongestBranch`, then replace `LongestBranch`.
    - E. Go back to B and repeat until visit the all the other `Branches`.
  - In conclusion, if there's a branch that can be "connected" with `Tip`, and if `TotalDifficulty` of the branch is high enough, then we've got a new `Tip`.
    - On the other hand, If it is "connected" with another block but if it is more past block, then we've got a new branch and new Tip.
    - Any other case that isn't illustrated cannot be connected with the current `Blockchain`. those will be discarded. However, it will be stored short amount of moment, if any incoming block can be "connected" with any block AND branch in `BlockCandidateTable` and `Blockchain` then we can use the "cached" branch.

To be updated (2022/05/23)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,12 +32,18 @@ To be released.
 
 ### Behavioral changes
 
+ -  (Libplanet.Net) `BlockCandidateTable<T>` updated for reducing the chance
+    of dropping the possible longest chain while `ProcessBlockDemandAsync()`
+    and `PullBlocksAsync()` are in progress. [[#1892], [#1917]]
+
 ### Bug fixes
 
 ### Dependencies
 
 ### CLI tools
 
+[#1892]: https://github.com/planetarium/libplanet/issues/1892
+[#1917]: https://github.com/planetarium/libplanet/pull/1917
 [#2024]: https://github.com/planetarium/libplanet/pull/2024
 
 
--- a/Libplanet.Net/BlockCandidateTable.cs
+++ b/Libplanet.Net/BlockCandidateTable.cs
@@ -22,16 +22,11 @@ namespace Libplanet.Net
             Branches = new ConcurrentBag<CandidateBranch<T>>();
         }
 
-       /// <summary>
-        /// Reference to <see cref="BlockChain{T}.Tip"/>.
-        /// </summary>
-        public Block<T>? Tip { get; }
-
         /// <summary>
-        /// Retrieves the <see cref="CandidateBranch{T}"/> with the best <see cref="Tip"/>, i.e.
+        /// Retrieves the <see cref="CandidateBranch{T}"/> with the best Tip, i.e.
         /// the highest <see cref="Block{T}.TotalDifficulty"/>.
         /// </summary>
-        public CandidateBranch<T>? BestBranch { get; }
+        public CandidateBranch<T>? BestBranch { get; private set; }
 
         public long Count
         {
@@ -39,10 +34,11 @@ namespace Libplanet.Net
         }
 
         /// <summary>
-        /// An <see cref="IEnumerable{T}"/> of <see cref="CandidateBranch{T}"/>es that can be appended
+        /// An <see cref="IEnumerable{T}"/> of
+        /// <see cref="CandidateBranch{T}"/>es that can be appended
         /// to the <em>current</em> <see cref="BlockChain{T}"/>.
         /// </summary>
-        private ConcurrentBag<CandidateBranch<T>> Branches { get; }
+        private ConcurrentBag<CandidateBranch<T>> Branches { get; set; }
 
         /// <summary>
         /// Adds an <see cref="CandidateBranch{T}"/> to <see cref="Branches"/>.
@@ -50,15 +46,11 @@ namespace Libplanet.Net
         /// <param name="branch">The <see cref="CandidateBranch{T}"/>
         /// to add to <see cref="Branches"/>.
         /// </param>
-        /// <remarks>
-        /// Note that <see cref="Tip"/> could have changed while downloading <see cref="Block{T}"/>s
-        /// to create <paramref name="branch"/>.  This should be dealt with accordingly in order
-        /// to ensure concurrency and enforce the stated properties in the interface summary.
-        /// In particular, the first point may no longer be <c>true</c> even if
-        /// <paramref name="branch"/> can still be "appended".
-        /// </remarks>
         public void Add(CandidateBranch<T> branch)
         {
+            BestBranch ??= branch;
+            BestBranch = CompareBranch(BestBranch, branch);
+
             Branches.Add(branch);
         }
 
@@ -67,7 +59,7 @@ namespace Libplanet.Net
         /// an <see cref="UpdatePath{T}"/>.
         /// </summary>
         /// <param name="path">The <see cref="UpdatePath{T}"/> representing a path from the previous
-        /// <see cref="Tip"/> to the newly changed <see cref="Tip"/>.</param>
+        /// Tip to the newly changed Tip.</param>
         /// <param name="predicate"><see cref="Func{TResult}"/> for predicate which
         /// <see cref="Branches"/> are no more needed.</param>
         /// <remarks>
@@ -79,13 +71,49 @@ namespace Libplanet.Net
         /// <para>
         /// Also, while updating each <see cref="CandidateBranch{T}"/>,
         /// if an <see cref="CandidateBranch{T}"/> is no longer needed,
-        /// i.e. if <see cref="CandidateBranch{T}.Tip"/>'s <see cref="Block{T}.TotalDifficulty"/> is
-        /// no longer higher than that of <see cref="UpdatePath{T}.NewTip"/>, it should be discarded.
+        /// i.e. if <see cref="CandidateBranch{T}.Tip"/>'s
+        /// <see cref="Block{T}.TotalDifficulty"/> is
+        /// no longer higher than that of <see cref="UpdatePath{T}.NewTip"/>,
+        /// it should be discarded.
         /// </para>
         /// </remarks>
         public void Update(UpdatePath<T> path, Func<IBlockExcerpt, bool> predicate)
         {
-            return;
+            var newBag = new ConcurrentBag<CandidateBranch<T>>();
+            CandidateBranch<T>? longestBranch = null;
+
+            // if OldTip is equals to BranchPoint, means it is not reorg.
+            if (path.OldTip.Equals(path.BranchPoint))
+            {
+                foreach (var branch in Branches)
+                {
+                    if (!predicate(branch.Tip))
+                    {
+                        continue;
+                    }
+
+                    var newBlocks =
+                        branch.Blocks.FindAll(x => x.Index > path.NewTip.Index).ToList();
+                    var newBranch = new CandidateBranch<T>(
+                        newBlocks,
+                        newBlocks.First(),
+                        newBlocks.Last());
+
+                    longestBranch ??= newBranch;
+                    longestBranch = CompareBranch(longestBranch, newBranch);
+                }
+            }
+            else
+            {
+            }
+
+            if (longestBranch is { })
+            {
+                newBag.Add(longestBranch);
+                BestBranch = longestBranch;
+            }
+
+            Branches = newBag;
         }
 
         /// <summary>
@@ -93,9 +121,13 @@ namespace Libplanet.Net
         /// </summary>
         public void Clear()
         {
-            return;
+            Branches = new ConcurrentBag<CandidateBranch<T>>();
+            BestBranch = null;
         }
 
         public bool Any() => Branches.Any();
+
+        private CandidateBranch<T> CompareBranch(CandidateBranch<T> lf, CandidateBranch<T> rf)
+            => lf.Tip.TotalDifficulty < rf.Tip.TotalDifficulty ? rf : lf;
     }
 }
--- a/Libplanet.Net/BlockCandidateTable.cs
+++ b/Libplanet.Net/BlockCandidateTable.cs
@@ -85,7 +85,7 @@ namespace Libplanet.Net
             // if OldTip is equals to BranchPoint, means it is not reorg.
             if (path.OldTip.Equals(path.BranchPoint))
             {
-                foreach (var branch in Branches)
+                foreach (CandidateBranch<T> branch in Branches)
                 {
                     if (!predicate(branch.Tip))
                     {
@@ -105,6 +105,42 @@ namespace Libplanet.Net
             }
             else
             {
+                foreach (CandidateBranch<T> branch in Branches)
+                {
+                    if (!predicate(branch.Tip))
+                    {
+                        continue;
+                    }
+
+                    var newBlocks = branch.Blocks;
+                    Block<T> index = branch.Root;
+                    while (index.PreviousHash != null &&
+                           path.Blocks.Contains(index))
+                    {
+                        try
+                        {
+                            index = path.Blocks.Single(x => x.Hash.Equals(index.PreviousHash));
+                            newBlocks.Insert(0, index);
+                        }
+                        catch (ArgumentNullException)
+                        {
+                            break;
+                        }
+                    }
+
+                    if (newBlocks.First().PreviousHash.Equals(path.OldTip.Hash))
+                    {
+                        newBlocks.Insert(0, path.BranchPoint);
+                    }
+
+                    var newBranch = new CandidateBranch<T>(
+                        newBlocks,
+                        newBlocks.First(),
+                        newBlocks.Last());
+
+                    longestBranch ??= newBranch;
+                    longestBranch = CompareBranch(longestBranch, newBranch);
+                }
             }
 
             if (longestBranch is { })
--- a/Libplanet.Net/BlockCandidateTable.cs
+++ b/Libplanet.Net/BlockCandidateTable.cs
@@ -92,13 +92,22 @@ namespace Libplanet.Net
                         continue;
                     }
 
-                    var newBlocks =
-                        branch.Blocks.ToList().FindAll(x => x.Index > path.NewTip.Index)
-                            .ToList();
-                    var newBranch = new CandidateBranch<T>(newBlocks);
+                    try
+                    {
+                        var newBlocks =
+                            branch.Blocks.ToList().FindAll(x => x.Index > path.NewTip.Index)
+                                .ToList();
+                        var newBranch = new CandidateBranch<T>(newBlocks);
 
-                    longestBranch ??= newBranch;
-                    longestBranch = CompareBranch(longestBranch, newBranch);
+                        longestBranch ??= newBranch;
+                        longestBranch = CompareBranch(longestBranch, newBranch);
+                    }
+                    catch (ArgumentNullException)
+                    {
+                        // FIXME: This has to be done to prevent stopping Update() by
+                        // exception thrown from FindAll().
+                        continue;
+                    }
                 }
             }
             else
@@ -120,7 +129,9 @@ namespace Libplanet.Net
                             index = path.Blocks.Single(x => x.Hash.Equals(index.PreviousHash));
                             newBlocks.Insert(0, index);
                         }
-                        catch (ArgumentNullException)
+                        catch (Exception e) when (
+                            e is ArgumentException ||
+                            e is InvalidOperationException)
                         {
                             break;
                         }
--- a/Libplanet.Net/BlockCandidateTable.cs
+++ b/Libplanet.Net/BlockCandidateTable.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using Libplanet.Action;
 using Libplanet.Blockchain;
 using Libplanet.Blocks;
-using Serilog;
 
 namespace Libplanet.Net
 {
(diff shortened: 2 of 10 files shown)


Libplanet.Net/UpdatePath.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <typeparam name="T">An <see cref="IAction"/> type.  It should match
+    /// to <see cref="Block{T}"/>'s type parameter.</typeparam>
+    public class UpdatePath<T>
+        where T : IAction, new()
+    {
+        public UpdatePath(
+            IEnumerable<Block<T>> blocks,
+            Block<T> oldTip,
+            Block<T> branchPoint,
+            Block<T> newTip)
```
longfin, line 44: As like `CandidateBranch`, it would be better if we omit unnecessary arguments to prevent representing illegal state.

colibrishin:
Updated ctor. closed and First() and Last() applied for Root and Tip each.
I didn't close ctor. of `UpdatePath<T>` yet. Somehow, the definition of `BranchPoint` is weak.

riemannulus (commented):
👍

Libplanet.Net/UpdatePath.cs
```diff
@@ -37,22 +38,30 @@ namespace Libplanet.Net
...
-            Blocks = blocks;
+            Blocks = new List<Block<T>>(blocks);
             OldTip = oldTip;
             BranchPoint = branchPoint;
-            NewTip = newTip;
         }
 
         /// <summary>
-        /// The <see cref="IEnumerable{T}"/> of <see cref="Block{T}"/> in the path.
+        /// The <see cref="List{T}"/> of <see cref="Block{T}"/> in the path.
```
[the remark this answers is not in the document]
colibrishin: Ah, My bad. I didn't made a roll-back after testing with another types.

Libplanet.Net/CandidateBranch.cs
```diff
@@ -15,31 +16,49 @@ public class CandidateBranch<T>
...
+        public Block<T> Tip => Blocks.Last();
+
+        /// <summary>
+        /// The method to append the block <see cref="Block{T}"/> into
+        /// <see cref="CandidateBranch{T}"/>.
+        /// </summary>
+        /// <param name="block">A block that continuous to <see cref="CandidateBranch{T}.Root"/>
+        /// by <see cref="Block{T}.PreviousHash"/>.
+        /// </param>
+        public void Push(Block<T> block) => Blocks.Insert(0, block);
```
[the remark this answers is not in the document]
colibrishin: Could you elaborate _"CandidateBranch does not need to insert a block"_? and I'd like to hear your thinking about defining `Blocks` as `readonly`.

colibrishin:
I still have concerns and confusions about removing `OldTip` from `UpdatePath<T>`. Correct me If I'm wrong.

First of all, arrow denotes _`PreviousHash` is equals to `Hash` of ..._, and Branch is _a set of `Block`s_.

### Not Reorganization

- Let's say **Node A** has encountered two branches. _Branch_A_ and _Branch_B_, where two branches have same ancestor block and ancestor block is same as Tip of **Node A**,
```mermaid
flowchart TD
    Branch_B --> Tip
    Branch_A --> Tip
```
This is not Reorganization.

Consider this case, 

### Reorganization
```mermaid
flowchart TD
    BlockFromPast_Tip'--> Tip
    Branch_B --> BlockFromPast_Tip'
    Branch_A --> Tip
```
- Let's say _Branch_A_ is Canonical Chain (Branch) from **Node A**'s perspective, and **Node A** has encountered _Branch_B_ that has higher `TotalDifficulty` than _Branch_A_  which is followed by past block of **Node A**, Not `Tip`. 
```latex
TotalDifficulty(Tip + Branch_A) < TotalDifficulty(BlockFromPast_Tip' + Branch_B)
```

This is Reorganization.

Therefore, To determine whether this branch will occurs reorganization, we need _BlockFromPast_Tip'_, _Tip_, and the `Root` of _Branch_B_.

If we compute `OldTip` from `Blocks` in `UpdatePath<T>`, then we lose _BlockFromPast_Tip'_. Even if we say _Tip_ can be acquired naively.

colibrishin:
This will fail in test however, I'll push for sharing.

greymistcube:
From an informational perspective, `UpdatePath` needs to contain the following information:

- An iterable list of `Block`s from `OldTip` to `NewTip`.

It needs *only* the above information in the sense that by giving it an explicit structure, `OldTip`, `BranchPoint`, and `NewTip` *can be derived* from a simple list of `Block`s for the following reasons:

- `NewTip` should always have a higher `TotalDifficutly` than that of `OldTip`.
- `BranchPoint` should be the `Block` in the list with unique minimal `Index` / `TotalDifficulty`.

Properties such as `OldTip`, `BranchPoint`, and `NewTip` are there only for interfacing reasons.

It needs *all* the above information in the sense that it is required to update the `Branch`es in `BlockCandidateTable` according to its spec for different possible cases.

I would suggest either
- `UpdatePath()` take a single `IEnumerable<T>` of `Block`s for its constructor where the path is already constructed externally and its legality is determined inside the constructor.
- `UpdatePath()` take two `IEnumerable<T>` of `Block`s with a single common `Block`, namely the "roots", and join them to construct a proper path inside the constructor.

In either case, a traversable path from the `OldTip` to `BranchPoint` is needed initially to construct an `UpdatePath`.

colibrishin:
God, I've force-pushed, what like 4 times today. sorry for the mess.

### Changes
- Preventing possible illegal states by closing `UpdatePath<T>` and `CandidateBranch<T>` within the context.
- `CandidateBranch<T>.Blocks` is not editable from externally. (`Blocks` returns `IEnumerable<Block<T>>`.)
- Minor changes
  - Using expression body for `BlockCandidateTable<T>.Count`
  - Supress value type boxing warning while Iterating blocks in `BlockCandidateTable<T>.Update()`

colibrishin:
```mermaid
graph TD
    A[BlockCandidateProcess] --> B(takes CandidateBranch)
    C[BlockCandidateTable.BestBranch] --> B
    D[BlockCandidateDownload] --> C
    B --> E{Reorg?}
    E --> |Yes| F[ForkedChain, UpdatePath]
    E --> |No| G[ExtendedChain, UpdatePath]
    F --> H{Is Cannonical?}
    G --> H
    H --> |Yes| I[Swap BlockChain]
    H --> |No| J[Discard Chain]
    K[BlockCandidatTable.Update] --> C
    L("where we are!") --> K
```

### Usage of `CandidateBranch<T>`

`CandidateBranch<T>` is initially created from **`BlockCandidateDownload(...)`**.
https://github.com/planetarium/libplanet/blob/acfe8ddbc418cf91d2037c690d5cf29a8790754d/Libplanet.Net/Swarm.BlockCandidate.cs#L371-L405

`CandidateBranch<T>` will have,
- `branch` as `Blocks`
- `Root` as `Blocks.First()`
- `Tip` as `Blocks.Last()`

After the process, `CandidateBranch<T>` will be added into `BlockCandidateTable<T>`.

### Usage of `UpdatePath<T>`


We get `BranchPoint` from **`FindBranchpoint(...)`**.
https://github.com/planetarium/libplanet/blob/acfe8ddbc418cf91d2037c690d5cf29a8790754d/Libplanet.Net/Swarm.BlockCandidate.cs#L129

And `FindBranchpoint(Block<T> oldTip, Block<T> newTip, List<Block<T>> newBlocks)` behaves in following codes,
https://github.com/planetarium/libplanet/blob/acfe8ddbc418cf91d2037c690d5cf29a8790754d/Libplanet.Net/Swarm.BlockCandidate.cs#L227-L286

- `oldTip` is the current `Blockchain<T>.Tip`.
- `newTip` is the `Tip` of `Branch` we got from (`newBlocks.First()` in other word). 
- `Branch`, following blocks that we got from a node, are in `newBlocks`.

After then, `UpdatePath<T>` is used in **`AppendPreviousBlocks(...)`**.

## If there's no reorganization, code goes to,
https://github.com/planetarium/libplanet/blob/acfe8ddbc418cf91d2037c690d5cf29a8790754d/Libplanet.Net/Swarm.BlockCandidate.cs#L132-L142

## If there's a reorganization, code goes to,
https://github.com/planetarium/libplanet/blob/acfe8ddbc418cf91d2037c690d5cf29a8790754d/Libplanet.Net/Swarm.BlockCandidate.cs#L160-L175

`UpdatePath<T>` will have following values in both of cases.
- `Blockchain<T>.Tip` as `OldTip`. 
- `newBlocks` as `Blocks`.
- `newBlock.First()` as `NewTip` .

**`AppendPreviousBlocks(...)`** returns forked (or extended) `Blockchain<T>`. And `BlockCandidateProces(...)` determine whether branch should be appended by given `UpdatePath<T>` and `Blockchain<T>`.

https://github.com/planetarium/libplanet/blob/acfe8ddbc418cf91d2037c690d5cf29a8790754d/Libplanet.Net/Swarm.BlockCandidate.cs#L58-L113

riemannulus:
Do you think this will be difficult to finish?

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,155 @@ namespace Libplanet.Net
...
+        /// to add to <see cref="Branches"/>.
+        /// </param>
+        public void Add(CandidateBranch<T> branch)
         {
-            return _blocks.TryGetValue(thisRoundTip, out var blocks)
-                ? blocks
-                : null;
-        }
+            BestBranch ??= branch;
+            BestBranch = CompareBranch(BestBranch, branch);
```
riemannulus, line 52: This is might be the wrong behavior. If the new branch is a reorg branch and branchpoint is missing, `CompareBranch` is cannot return correctly. 

Plus, `Add` is do not have other behavior except `Add`. This is might be a cause of the mistake. (Even do not note this behavior on the docstring.)

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,155 @@ namespace Libplanet.Net
...
+                            try
+                            {
+                                index = path.Blocks.Single(
+                                    x => x.Hash.Equals(index.PreviousHash.Value));
+                                newBlocks.Insert(0, index);
+                            }
+                            catch (ArgumentNullException)
+                            {
+                                break;
+                            }
```
riemannulus, line 131: If then, `Branch.Root` is the longest but do not compatible current chain?

Libplanet.Net/BlockCandidateTable.cs
colibrishin: After removing `BestBranch` comparing, it needs to be call update at some point to get a valid `BestBranch` not `null`. However, the problem is, `Update()` requires a fork path `UpdatePath<T>` from Blockchain. 

If node is up or new Blockchain, `Update()` cannot be called because there's no  `UpdatePath<T>` to use. because `UpdatePath<T>` hasn't created yet.

colibrishin:
I think It's now ready for review

CHANGES.md
```diff
@@ -21,12 +21,18 @@ To be released.
 
 ### Behavioral changes
 
+ -  (Libplanet.Net) Reintroduces `BlockCandidateTable<T>` for reducing the chance
+    of dropping the possible longest chain while `ProcessBlockDemandAsync()`
+     and `PullBlocksAsync()` are in progress. [[#1892], [#1917]]
```
greymistcube, line 26: Over 80 character limit. 🙄

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,238 @@ namespace Libplanet.Net
...
+        /// <summary>
+        /// An <see cref="IEnumerable{T}"/> of
+        /// <see cref="CandidateBranch{T}"/>es that can be appended
+        /// to the <em>current</em> <see cref="BlockChain{T}"/>.
+        /// </summary>
+        private ConcurrentBag<CandidateBranch<T>> Branches { get; set; }
 
         /// <summary>
-        /// Adds a <see cref="Block{T}"/>s to the table.
+        /// Adds an <see cref="CandidateBranch{T}"/> to <see cref="Branches"/>.
```
greymistcube, line 49:
```suggestion
        /// Adds a <see cref="CandidateBranch{T}"/> to <see cref="Branches"/>.
```

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,238 @@ namespace Libplanet.Net
...
+        /// </para>
+        /// </remarks>
+        public void Update(UpdatePath<T> path)
         {
-            return _blocks.TryGetValue(thisRoundTip, out var blocks)
-                ? blocks
-                : null;
+            lock (_lock)
+            {
+                var newBag = new ConcurrentBag<CandidateBranch<T>>();
```
greymistcube, line 115: No need for `ConcurrentBag<T>` type if this is only used locally inside a lock.

Libplanet.Net/UpdatePath.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <typeparam name="T">An <see cref="IAction"/> type.  It should match
+    /// to <see cref="Block{T}"/>'s type parameter.</typeparam>
+    public class UpdatePath<T>
+        where T : IAction, new()
+    {
+        public UpdatePath(IEnumerable<Block<T>> blocks, Block<T> oldTip)
+        {
+            Blocks = blocks;
+            OldTip = oldTip;
+        }
```
greymistcube, line 45: This particular implementation does not match the original specification in the documentation.
Either the documentation or the implementation should be changed to match each other.

Either way, there should be some additional sanity checks for this constructor.

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,238 @@ namespace Libplanet.Net
...
-            catch (ArgumentException e)
-            {
-                Log.Debug(
-                    "Among the previous blocks of BlockHeader " +
-                    "#{Index} {BlockHash}, there is a block with the wrong index. " +
-                    "InnerMessage: {InnerMessage}",
-                    blockHeader.Index,
-                    blockHeader.Hash,
-                    e.Message);
+                // If BestBranch is not initialized, set incoming branch as BestBranch.
```
greymistcube, line 66: I think the logic is simple enough to remove this comment. 😗

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,238 @@ namespace Libplanet.Net
...
+                if (BestBranch is null)
+                {
+                    if (AreBlocksContinuous(branch.Blocks))
```
greymistcube, line 69: Usually a check for sanity for an argument at the start of the method, if possible, makes more sense. Also, it would make even more sense if branch continuity is checked by `CandidateBranch<T>` during its creation so we can remove `AreBlocksContinuous` method. 😛

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,238 @@ namespace Libplanet.Net
...
+        /// <param name="branch">A series of <see cref="Block{T}"/>.
+        /// </param>
+        /// <returns>
+        /// returns true if <paramref name="branch"/> is continuous or its count is 1; returns false
+        /// if <paramref name="branch"/> is not continuous or its count is 0.
+        /// </returns>
+        private static bool AreBlocksContinuous(IEnumerable<Block<T>> branch)
         {
-            foreach (var blockHeader in _blocks.Keys)
+            var enumerable = branch as Block<T>[] ?? branch.ToArray();
```
greymistcube, line 220: Why is casting needed here? 🤔

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -18,79 +17,238 @@ namespace Libplanet.Net
...
         }
+
+        // FIXME: This method is duplicated with Swarm<T>.IsBlockNeeded()
+        private bool IsBlockNeeded(IBlockExcerpt block, IBlockExcerpt tip) =>
+            _canonicalChainComparer.Compare(block, tip) > 0;
+
+        private CandidateBranch<T> CompareBranch(
+            CandidateBranch<T> lf,
+            CandidateBranch<T> rf)
+            => _canonicalChainComparer.Compare(lf.Tip, rf.Tip) > 0 ? lf : rf;
```
greymistcube, line 252: This is problematic. The logic is not argument order agnostic, i.e. `CompareBranch(a, b)` is not necessarily equal to `CompareBranch(b, a)`. This should be made clearer by using a different name for the method or via renaming of parameters. Think of a scenario where `CompareBranch(a, b)` might be equal to `0`.

colibrishin: Yes, I've noticed where the `a == b` boundary exists while implementing `IsBlockNeeded()` and I somehow miss the same thing applied to `CompareBranch()`.😅

I'm not sure what would be the best name for this method. I've tried and commented on additional documentation.

https://github.com/planetarium/libplanet/blob/dedae5e3f869925a9d59afb9bfb1186369461197/Libplanet.Net/BlockCandidateTable.cs#L238-L259

Libplanet.Net/BlockCandidateTable.cs
colibrishin: I do casts this purely because of multiple enumeration warning. It seems unnecessary casting because CandidateBranch<T> isn't belong to any parallel situation. I'll try to fix with moving AreBlocksContinuous() into CandidateBranch<T> constructor.

colibrishin:
> Seems conflict with main branch 😢 Can you please rebase?

Okay, I'll rebase with `UpdatePath<T>` and `CandidateTable<T>` changes.

limebell (approved):

Libplanet.Net/BlockCandidateTable.cs
colibrishin: https://github.com/planetarium/libplanet/blob/2d826486ed23614e840c6b14e0d89c0e8c7da084/Libplanet.Net/CandidateBranch.cs#L22-L55

Now the block order sanity will be checked when a `CandidateBranch<T>` is made. 
And `CandidateBranch<T>` will be empty if the blocks are not in order.

Libplanet.Net/BlockCandidateTable.cs
colibrishin: Solved with moving AreBlocksContinuous() to the constructor of `CandidateBranch<T>`

Libplanet.Net/CandidateBranch.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Libplanet.Net
+{
+    /// <summary>
+    /// A class to store an appendable <see cref="CandidateBranch{T}"/> for a
+    /// <see cref="BlockChain{T}"/>. <see cref="CandidateBranch{T}.Blocks"/> can be empty if
+    /// <see cref="CandidateBranch{T}"/> was created with non-continuous blocks.
+    /// </summary>
+    /// <typeparam name="T">An <see cref="IAction"/> type. It should match
+    /// to <see cref="Block{T}"/>'s type parameter.</typeparam>
```
greymistcube, line 16: It is generally better to not allow creation of a `Branch` by throwing an exception rather than allowing a "special" type of branch (i.e. an empty branch) if invalid argument is passed on to a constructor. 😕

CHANGES.md
```diff
@@ -18,12 +18,19 @@ To be released.
 
 ### Behavioral changes
 
+ -  (Libplanet.Net) `BlockCandidateTable<T>` updated for reducing the chance
+    of dropping the possible longest chain while `ProcessBlockDemandAsync()`
+     and `PullBlocksAsync()` are in progress. [[#1892], [#1917]]
```
greymistcube, line 23:
```suggestion
 -  (Libplanet.Net) `BlockCandidateTable<T>` updated for reducing the chance
    of dropping the possible longest chain while `ProcessBlockDemandAsync()`
    and `PullBlocksAsync()` are in progress. [[#1892], [#1917]]
```

colibrishin:
> Hmm... I thought there were locks for `BlockCandidateTable` at some point. Why did it get removed? 🤔

Oh I have misunderstood... 😖 You were saying that ConcurrentBag is not needed in lock, Right?

Libplanet.Net/UpdatePath.cs
colibrishin: https://github.com/planetarium/libplanet/blob/3da5ba314e5c368e77b795433b4fb7ebbc1425ba/Libplanet.Net/CandidatePath.cs#L8-L36

The following properties, `NewTip`, and `Blocks` that are unique and in order, will be checked by creating `UpdatePath<T>` with `CandidateBranch<T>` that is appended successfully.

https://github.com/planetarium/libplanet/blob/17f4e97415ebeac341b9c623a18ad0b508fb5be1/Libplanet.Net/UpdatePath.cs#L9-L28

Libplanet.Net/CandidateBranch.cs
colibrishin: I see, I wasn't sure whether throwing exception in constructor is expectable. 🤔 I replaced empty branch strategy to throw `CreateCandidateBranchException` and caught `Exception` where the `CanddiateBranch<T>` is created.

Libplanet.Net/UpdatePath.cs
colibrishin: I don't have any good idea for this... 

By using Distinct(), for checking uniqueness in `Blocks` costs `O(N)`. However, The properties are held by creating `UpdatePath<T>` with the appended `CandidateBranch<T>` to current `BlockChain<T>`, which means, they are in order, and unique.

PR #1919, opened 2022-04-21
New message for PBFT

riemannulus:
There are more messages for PBFT.

### DIFF
--- a/Libplanet.Net/Messages/ConsensusCommit.cs
+++ b/Libplanet.Net/Messages/ConsensusCommit.cs
@@ -1,0 +1,17 @@
+﻿namespace Libplanet.Net.Messages
+{
+    public class ConsensusCommit : ConsensusMessage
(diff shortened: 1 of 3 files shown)


dahlia (commented):
We may need to keep them `internal` until they indeed function together.  It's better for not forgetting to write changelogs, because we can be aware when they become `public`.

riemannulus:
> We may need to keep them `internal` until they indeed function together. It's better for not forgetting to write changelogs, because we can be aware when they become `public`.

This PR is targeted to `pbft` subbranch, not `main` or `maintenance`. Also. We use those messages on the [posplanet](https://github.com/planetarium/posplanet)(now private). So left to `public` is needed.

riemannulus:
This is a just test so you don't worry about it. I and @limebell will code cleaning before making merge PR into main.

PR #1925, opened 2022-04-27
Remove rethrowing of caught exception

jeongsoolee09:
Rethrowing a caught exception causes an error on .NET 6, so I would like to remove this from the test code.

I guess the `catch` branch intended to give some textual demonstration of the test, thus intercepted the received exception `e`. It seems better to me that the demonstration be given elsewhere (maybe as a separate documentation).

### DIFF
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -2048,9 +2048,7 @@ namespace Libplanet.Tests.Blockchain
                 catch (TxPolicyViolationException e)
                 {
                     Assert.NotNull(e.InnerException);
-
-                    _logger.Debug("Exception: {Exception}", e.ToString());
-                    _logger.Debug("InnerException: {InnerException}", e.InnerException.ToString());
+                    throw;
                 }
             });
         }
--- a/Libplanet.Tests/Blockchain/BlockChainTest.cs
+++ b/Libplanet.Tests/Blockchain/BlockChainTest.cs
@@ -2049,12 +2049,8 @@ namespace Libplanet.Tests.Blockchain
                 {
                     Assert.NotNull(e.InnerException);
 
-                    // This is visual representation of the test.
-                    // In actual use-case, this debugging lines do
-                    // not show.
                     _logger.Debug("Exception: {Exception}", e.ToString());
                     _logger.Debug("InnerException: {InnerException}", e.InnerException.ToString());
-                    throw e;
                 }
             });
         }


Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2049,12 +2049,8 @@ private void CheckIfTxPolicyExceptionHasInnerException()
                 {
                     Assert.NotNull(e.InnerException);
 
-                    // This is visual representation of the test.
-                    // In actual use-case, this debugging lines do
-                    // not show.
```
dahlia, line 2054: Should these lines be removed?  What did you intend?

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2049,12 +2049,8 @@ private void CheckIfTxPolicyExceptionHasInnerException()
...
                     _logger.Debug("Exception: {Exception}", e.ToString());
                     _logger.Debug("InnerException: {InnerException}", e.InnerException.ToString());
-                    throw e;
                 }
```
dahlia, line 2054: You probably intended to get rid of only `e`, instead of the whole line, didn't you?

Libplanet.Tests/Blockchain/BlockChainTest.cs
jeongsoolee09: My thought was:
- The contents of the comments are ambiguous; e.g. why should the lines not appear?
- Also, the following `Debug` lines are quite self-explanatory: they print the exception.

Therefore, the comments deemed as unnecessary.

Libplanet.Tests/Blockchain/BlockChainTest.cs
jeongsoolee09: I think it really depends on how the original test-writer thought. If the writer wanted to throw that exception anyways, then `throw` would the correct one. However, if the `Debug` code suffices for recording that a `TxPolicyViolationException` took place, then the whole `throw` statement would be unnecessary.

colibrishin: I'm the trouble-maker in this situation. I really appreciate that you've caught this. Thanks! 😊

This `TxPolicyViolationException` is used in `BlockPolicy` and `BlockChain<T>.ValidateNextBlockTx()`. 

In one of tests I wrote, this `CheckIfTxPolicyExceptionHasInnerException()`, I wanted to make sure new `BlockChain<T>`, with full dummy policies excepts TxPolicy, throws `TxPolicyViolationException` and only `TxPolicyViolationException`. That's why I've made this test code as followed.

> The contents of the comments are ambiguous; e.g. why should the lines not appear?
> Also, the following Debug lines are quite self-explanatory: they print the exception.

Yes, now that I think about it, that comment was added unnecessarily. Most test cases I've seen were not written with `_logger.Debug(...)` stdout, At that moment I thought comment should be needed somewhat. that's why the comment was added. Also I want to see myself how stacktrace would be looked like. In short, those lines are not that really necessary.

jeongsoolee09: Ah, nice! Guess I was a mind-reader then 👀

colibrishin: Could you restore `throw;` back for the context of this test and remove `_logger.Debug(...)` lines for now? or Let me know if you have better way to test this. 🙂

jeongsoolee09: @colibrishin Hmm... I guess that's the best option for now. I will carry them out.

jeongsoolee09:
I have:
- restored `throw`, but without designating the exception `e`
- removed `Debug` codes and their accompanying comments.

colibrishin:
One last thing, could you squash them into one commit?

dahlia:
> One last thing, could you squash them into one commit?

@jeongsoolee09 Never mind!  We could squash all commits into one using the merge button provided by GitHub web.  (If only some commits should be squashed while other ones should remain distinct, we would ask you to do so then.)  Thanks for your first contribution.

dahlia (approved):

colibrishin (approved):

PR #1935, opened 2022-05-04
Abstract `DataModel` feature

### LINKED ISSUE #1932: If possible, reduce `Bencodex` exposure to game client
Directly dealing with `Bencodex`'s `Encode()` and `Decode()` methods and `IValues` may be a point of high friction for using [Libplanet] for game developers. Other than for very primitive and simple game logic, it is more likely than not each `IValue` gets decoded to a native datatype and gets re-encoded back to an `IValue`. I'm not sure what approach would be the most appropriate for hiding `Bencodex` from the client side. 😶

greymistcube:
Closes #1932.

## Currently implemented spec:

- The complete list of allowed property types are as follows:
  - Primitive types: `bool`, `int`, `long`, `BigInteger`, `ImmutableArray<byte>`, `Address`, and `string`.
  - Special types: Any type inherited from `DataModel`.
  - Collective types:
    - `ImmutableList<T>` where `T` is a primitive type.
    - `ImmutableDictionary<K, V>` where
      - `K` is `ImmutableArray<byte>`, `Address`, or `string`.
      - `V` is a primitive type.
- Nullable value types such as `int?`, `ImmutableArray<byte>?` are **not supported**.
- Assignment of `null` to reference types is **not supported**.

## Example:

The following example shows how this class can be used:

```csharp
public class CharacterData : DataModel
{
    /// <summary>
    /// Name of the character.
    /// </summary>
    public string Name { get; private set; }

    /// <summary>
    /// Current level of the character.
    /// </summary>
    public int Level { get; private set; }

    /// <summary>
    /// Inventory of the character.
    /// </summary>
    public InventoryData Inv { get; private set; }

    public CharacterData(string name, int level, InventoryData inv)
        : base()
    {
        Name = name;
        Level = level;
        Inv = inv;
    }

    public CharacterData(Bencodex.Types.Dictionary encoded)
        : base(encoded)
    {
    }
}

public class InventoryData : DataModel
{
    /// <summary>
    /// The amount of gold in the inventory.
    /// </summary>
    public int Gold { get; private set; }

    public InventoryData(int gold)
        : base()
    {
        Gold = gold;
    }

    public InventoryData(Bencodex.Types.Dictionary encoded)
        : base(encoded)
    {
    }
}
```

Then the concrete model defined above can be used as shown below:

```csharp
CharacterData characterData = new CharacterData("John", 5, new InventoryData(100));
Bencodex.Types.Dictionary encoded = characterData.Encode()
```

This would result in `encoded` in the following format:

```
Bencodex.Types.Dictionary {
    "Name": "John",
    "Level": 5,
    "Inv": {
        "Gold": 100,
    }
}
```

To decode this back into an instance, simply use it as shown below:

```csharp
CharacterData decoded = new CharacterData(encoded);
```

Then `decoded.Name`, `decoded.Level`, and `decoded.Inv.Gold` should be `"John"`, `5`, and `100` respectively.

## Footnote

- Edit on May 10th: Usage example simplified.


Libplanet/Store/DataModel.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <returns>An encoded <see cref="BTypes.Dictionary"/> instance.</returns>
+        /// <exception cref="NotSupportedException">Thrown when <c>nullable</c> value type property
+        /// or <c>null</c> value inside a reference type is encountered.</exception>
+        /// <exception cref="ArgumentException">Thrown when an unknown invalid property type is
+        /// encountered.</exception>
+        [Pure]
+        public BTypes.Dictionary Encode()
+        {
+            BTypes.Dictionary result = BTypes.Dictionary.Empty;
+            PropertyInfo[] properties = this.GetType().GetProperties();
```
longfin, line 221: I guess `GetProperties()` can be called earlier in its constructor. (generally, reflection is a heavy work)

Libplanet/Store/DataModel.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <returns>A decoded instance of type <typeparamref name="T"/> as
+        /// <see cref="DataModel"/>.  Additional casting to <typeparamref name="T"/> may be
+        /// necessary.</returns>
+        /// <exception cref="TargetInvocationException">Thrown when failed to create an
+        /// instance of type <typeparamref name="T"/> with <paramref name="encoded"/>.
+        /// Its reason for failure is stored in thrown <see cref="TargetInvocationException"/>'s
+        /// <see cref="Exception.InnerException"/>.</exception>
+        /// <exception cref="NullReferenceException">Thrown when <paramref name="encoded"/> is
+        /// <c>null</c> or <c>null</c> reference is returned by inner instantiation.</exception>
+        public static DataModel Decode<T>(BTypes.Dictionary encoded)
```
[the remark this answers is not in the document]
greymistcube: Yup, seems like sloppy work on my part. 😅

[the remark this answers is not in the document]
greymistcube: Eh, relevant code is gone since `Decode<T>()` is now removed. 🙄

Libplanet/Store/DataModel.cs
greymistcube: I've given this some more thought while implementing as suggested. Depending on its usage, calling `GetProperties()` during construction might be worse for performance. For one thing, I've settled on using `ImmutableArray<T>`, `ImmutableList<T>`, and `ImmutableDictionary<K, V>` instead of `T[]`, `List<T>`, and `Dictionary<K, V>`, to suggest classes inherited from `DataModel` should be immutable (see #1943).

Of course, there is no mechanism to actually prevent game developers from bypassing this restriction by reassigning values to internally, but if a class is modeled to be immutable, which in my opinion is a better practice, with any "mutating" method returning a new instance, then `GetProperties()` will get called unnecessarily often.

Also there is an issue of taxing performance when data only needs to be read from the chain. 😶

greymistcube:
This not being DRY, there might still be some mistakes here and there. 😐

Libplanet/Store/DataModel.cs
greymistcube: As per additional discussion, this is implemented via caching. See 7b2ae7ea8f11bbf35784d4756ec2a019441e723e.

longfin (approved):

PR #1938, opened 2022-05-09
Add Height to ConsensusMessages

colibrishin:
Add new Height field in message


riemannulus (approved):
👍 commit with `[skip changelog]` message

PR #1939, opened 2022-05-09
Merge posplanet into libplanet

riemannulus:
# Rationale
- As our Q2 target, we need to change to PoS from PoW.
- In this PR, implement basic PBFT message and state machine on the libplanet.
- This code is rooted from [posplanet](https://github.com/planetarium/posplanet)(private repo)


limebell:
Build seems to fail

limebell (approved):

colibrishin:
Squashed mistaken commits. It's ready for review. @limebell

limebell (approved):

osamadeep:
is that means there is no community mining ever?

longfin:
> is that means there is no community mining ever?

As Libplanet dev perspective, PoW (including permissioned comunity mining) isn't a main target to develop because it has definite weekeness (lack of finality). so we're preparing PoS like as this PR. 

One challenge we're having is that we're not a big enough team to do all of this in parallel. Since PoS conversion itself is already a big deal, @planetarium/libplanet is concentrating on this part, and in terms of importance, research and development work for PoS conversion should precede community mining. (At least for Libplanet development)

But it doesn't mean that 9c will also skip community mining. we are (slowly) testing community mining on the previewnet show up in Q1, and we plan to review it along with other features to be released this quarter.

colibrishin (approved):

Issue #1953, opened 2022-05-13
Make a pair of `(IStore, IStateStore)` referrable by a simple string (preferrably URI)


dahlia:
This will be also useful for achieving <https://github.com/planetarium/libplanet/issues/1946>.

It addresses <https://github.com/planetarium/libplanet/issues/1359> & <https://github.com/planetarium/libplanet/issues/1573> as well.

Maybe related?: <https://github.com/planetarium/libplanet/pull/1023#discussion_r499520404>

Issue #1359, opened 2021-06-28
Encapsulate the relation between `IStore` & `IStateStore`


dahlia:
Currently, `BlockChain<T>()` constructor takes `IStore` and `IStateStore` through distinct parameters, but two stores tend to be placed together filesystem-wise in most cases.  For example, Nine Chronicles place its state store inside its store directory:

```csharp
var store = new RocksDBStore(storePath);
var stateStore = new TrieStateStore(
    new RocksDBKeyValueStore(Path.Join(storePath, "states")),
    new RocksDBKeyValueStore(Path.Join(storePath, "state_hashes"))
);
```

The above names for subdirectories (`states` & `state_hashes`) can be considered as kind of leaky abstraction, because these subdirectories should not have other names, and if names are changed Nine Chronicles won't work anymore.

Instead of `BlockChain<T>()` constructor taking two distinct parameters, I suggest to define a new simple type like `struct BlockChainStore { IStore Store; IStateStore StateStore; }` to encapsulate both stores, and let `BlockChain<T>` take one `BlockChainStore`.

PR #1955, opened 2022-05-13
WIP: Load `(IStore, IStateStore)` from URI

dahlia:
Addresses #1953 and indirectly #1359.


longfin:
> This introduces [StoreLoaderAttribute] for connecting a URI scheme to a pair of IStore & IStateStore implementations. By default, the following URI schemes are provided:

I can't convince that we need attribute based dispatching because we have only three types for URI scheme now and variety of them wouldn't increase easily... is there any possibility about that?

Libplanet/Store/StoreLoaderAttribute.cs
```diff
@@ -0,0 +1,146 @@
...
+                    return m.GetCustomAttributes<StoreLoaderAttribute>()
+                        .Select(attr => (attr.UriScheme, m));
+                });
+        }
+
+        private static MethodInfo? FindStoreLoader(string uriScheme) =>
+            ListStoreLoaderMethods()
+                .Cast<(string, MethodInfo)?>()
+                .FirstOrDefault(pair =>
+                    pair?.Item1?.Equals(uriScheme, StringComparison.InvariantCulture) == true)
```
longfin, line 143:
```suggestion
                    pair?.Item1?.Equals(uriScheme, StringComparison.InvariantCulture))
```

greymistcube:
> I can't convince that we need attribute based dispatching because we have only three types for URI scheme now and variety of them wouldn't increase easily... is there any possibility about that?

Indeed the use of URI schemes and attributes are quite appropriate for technical reasons, but this seems a bit of an overkill. 🙂

Libplanet/Store/StoreLoaderAttribute.cs
dahlia: As the left operand's type is `bool?` (not `bool`), it's a type error without `== true`.  😵

dahlia:
> I can't convince that we need attribute based dispatching because we have only three types for URI scheme now and variety of them wouldn't increase easily... is there any possibility about that?

It actually doesn't matter how many types are there currently, it's just for adding a new type from outside of *Libplanet* code (e.g., *Libplanet.RocksDBStore*) or without changing *Libplanet* code.  We may be able to maintain a global dictionary declared in *Libplanet* and request every client code to register `"rocksdb" => RocksDBStore.LoadStore`… at runtime, but I'm not sure whether it would be really usable in such way.  I tried to use [static constructors] to automatically register loaders from outside, but I gave up because they were not executed before the first instantiation of the type.

Or you might think that it would be okay if we simply maintain the static table inside *Libplanet* code, but it's simply impossible as it would require circular dependencies between *Libplanet* package and *Libplanet.RocksDBStore* package.

Nevertheless, if there is any concrete example of a better alternative I would consider to introduce that.

[static constructors]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors

longfin:
> It actually doesn't matter how many types are there currently, it's just for adding a new type from outside of Libplanet code (e.g., Libplanet.RocksDBStore) or without changing Libplanet code. 

Ah, I see. well... we can add some API to "register" for external loader but... OK, it wouldn't very different from current implementation.

longfin (dismissed):

dahlia:
@planetarium/libplanet I rebased changes on the current main, and fixed the tests.  Could you take a look one more time?

longfin (approved):

Libplanet/Store/StoreLoaderAttribute.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// Loads a pair of <see cref="IStore"/> and <see cref="IStateStore"/> from the specified
+        /// <paramref name="storeUri"/>.
+        /// </summary>
+        /// <param name="storeUri">A URI referring to a store.</param>
+        /// <returns>A pair of loaded <see cref="IStore"/> and <see cref="IStateStore"/>.</returns>
+        public static (IStore Store, IStateStore StateStore)? LoadStore(Uri storeUri)
+        {
+            const BindingFlags flags =
+                BindingFlags.Public |
+                BindingFlags.NonPublic |
```
greymistcube, line 59: Just curious, why also `NonPublic`?

dahlia: Without this, `internal`/`protected`/`private` methods are not discovered.

greymistcube: Yes, I understand that. Is this because due to already implemented access level? Not a big fan of rummaging through `internal` and `private` unless absolutely necessary. 😕

dahlia: I didn't wanted add more unnecessary (these loader methods would never be called directly, as there is no reason to prefer them over regular constructors) public methods, because we should maintain them after once they become public.

greymistcube: Well, if the intention is to specifically hide them, then I can get behind NonPublic flag. But in that case, maybe we can get rid of Public? 🤔

dahlia: I indeed considered to get rid of it, but decided to let it discover any methods conforming the proper signature regardless of their visibility, because it would be very difficult to troubleshoot if someone write their own loader method with wrong visibility… We may be able to build a Roslyn analyzer to warn such mistakes in the future.

greymistcube (approved):

PR #1961, opened 2022-05-13
Unncessary/inappropriate uses of timeout removed

greymistcube:
Session timeouts are different from network timeouts and should be implemented accordingly, if we want such features that is. Simply passing along timeouts willy-nilly is bad. 😕

For example the call sequence for `timeout` is

- `FillBlocksAsync()` -> `ProcessBlockDemandAsync()` -> `BlockCandidateDownload()` -> `GetBlockHashes()`.

Multiple issues here:

- From `FillBlocksAsync()`, it is unclear what `timeout` is for.
- I don't think anyone is fully aware of the consequences of overriding the default `GetBlockHashes()` timeout period provided by `SwarmOptions.BlockHashRecvTimeout` as it is not immediately apparent this is overridden with different values from several places.

If we want a finer control for `GetBlockHashes()` from different long running processes (then again, I don't know what the purpose of `SwarmOptions.BlockHashRecvTimeout` would be in that case) it should be more clearly documented and implemented with care.

----

Additional removal of parameters resulted in an API change. Both `millisecondsSingleSessionTimeout` and `singleSessionTimeout` for `BlockCompletion<P, A>.Complete()` removed. Single session length is already capped by
the `ITransport` in use. Some ancient code was modified, so @dahlia should take a closer look. 😗

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,9 @@ To be released.
 ### Backward-incompatible API changes
 
  -  `BlockCompletion<TPeer, TAction>.Complete()` no longer accepts
-    an argument to regulate a single session.  [[#1961]]
+    neither parmeter `TimeSpan singleSessionTimeout` nor
+    `int millisecondsSingleSessionTimeout` to regulate a single session length.
+    [[#1961]]
 
 ### Backward-incompatible network protocol changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  `BlockCompletion<TPeer, TAction>.Complete()` no longer accepts
+    an argument to regulate a single session.  [[#1961]]
+
 ### Backward-incompatible network protocol changes
 
 ### Backward-incompatible storage format changes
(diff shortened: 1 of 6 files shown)


Libplanet.Net.Tests/BlockCompletionTest.cs
```diff
@@ -337,10 +337,12 @@ public async Task CompleteWithNonRespondingPeers()
                     // Peer A does not respond and Peer B does respond.
                     if (peer == 'A')
                     {
-                        while (true)
-                        {
-                            await Task.Delay(5000, yield.CancellationToken);
-                        }
+                        int transportTimeout = 3_000;
```
dahlia, line 340:
```suggestion
                        const int transportTimeout = 3_000;
```

CHANGES.md
```diff
@@ -10,6 +10,9 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  `BlockCompletion<TPeer, TAction>.Complete()` no longer accepts
+    an argument to regulate a single session.  [[#1961]]
```
dahlia, line 14: It would be better if it mentions the removed parameter names.

Libplanet.Net.Tests/BlockCompletionTest.cs
greymistcube: See ad0012a19e9530016f1f45de0baee110fec65f3d.

CHANGES.md
greymistcube: See ad0012a19e9530016f1f45de0baee110fec65f3d.

dahlia (approved):

Issue #1933, opened 2022-04-29
Implement a more structured `SwarmOptions`


greymistcube:
Currently, `SwarmOptions` is simply a flat set of properties without any clear criteria as to why certain options are present and some are not, (see #1922). If we consider `Swarm<T>` as an orchestrating entity for `BlockChain<T>`, `ITransport<T>`, and `IProtocol`, it'd be better if this is also reflected in the structure of `SwarmOptions`. Hierarchical structure is not a necessity, depending on how we implement it, but from a usability perspective, I think it'd be better to have a single configurable class.

In any case, as a mock example,

```
Swarm<PolymorphicAction<ActionBase>> swarm = new Swarm<PolymorphicAction<ActionBase>>(
    blockChain: blockChain,
    privateKey: privateKey,
    host: host,
    listenPort: port,
    iceServers: iceServers,
    appProtocolVersion: appProtocolVersion,
    trustedAppProtocolVersionSigners: trustedAppProtocolVersionSigners,
    differentAppProtocolVersionEncountered: differentAppProtocolVersionEncountered,
    options: swarmOptions);
```

should look more like

```
Swarm<PolymorphicAction<ActionBase>> swarm = new Swarm<PolymorphicAction<ActionBase>>(
    blockChain: blockChain,
    privateKey: privateKey,
    transport: transport,
    protocol: protocol,
    options: swarmOptions,
    appProtocolVersionHandler: appProtocolVersionHandler);
```

or

```
Swarm<PolymorphicAction<ActionBase>> swarm = new Swarm<PolymorphicAction<ActionBase>>(
    privateKey: privateKey,
    options: swarmOptions
    appProtocolVersionHandler: appProtocolVersionHandler);
```

greymistcube:
Eh, since daily is coming up, I'm dumping this preliminary draft, or more like
a pre-draft, here. This needs to be organized and/or filled a lot more.
Tracking down informational units/arguments in the codebase and checking
when/why/how they are used itself is taking a lot more time than I thought.
There will be future edits (hopefully). 😗

----

I tried to be a tad bit more descriptive with terms instead of
simply using argument names. These are informational units necessary and/or
used when running a node. Some are stored as a part of `SwarmOptions`,
some are directly passed on to methods from an application, and some are
even **hardcoded** internally.

## Application level

These are handled outside `Swarm<T>`.

- Chain storage type: Needed to load storage before `Swarm<T>()`.
[...]

PR #1962, opened 2022-05-15
Added `TimeoutOptions` to `SwarmOptions`

### LINKED ISSUE #1957: Group timeout related options
See #1933 and its comments.

Seems like a good starting point. For the most part, as far as I know, even [9c-headless] doesn't use variable arguments for timeout options. 🙄

[9c-headless]: https://github.com/planetarium/NineChronicles.Headless

greymistcube:
Closes #1957.

Had to update certain parts of `Swarm<T>` API to make it less non-sensical. 🙄
There's still a long way to go to mitigate #1933.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@ To be released.
 
 ### Deprecated APIs
 
+ -  (Libplanet.Net) Unused property `SwarmOptions.PollInterval` removed.
+    [[#1962]]
+
 ### Backward-incompatible API changes
 
  -  `BlockCompletion<TPeer, TAction>.Complete()` no longer accepts
@@ -31,6 +34,9 @@ To be released.
         `Swarm<T>.PreloadAsync(TimeSpan?, IProgress<PreloadState>, bool
         long, CancellationToken)` changed to `Swarm<T>.PreloadAsync(
         TimeSpan, IProgress<PreloadState>, bool, long, CancellationToken)`
+ -  (Libplanet.Net) `TimeoutOptions` property added to `SwarmOptions` with
+    all timeout related options moved from `SwarmOptions` to `TimeoutOptions`.
+    [[#1957], [#1962]]
 
 ### Backward-incompatible network protocol changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,23 @@ To be released.
     neither parmeter `TimeSpan singleSessionTimeout` nor
     `int millisecondsSingleSessionTimeout` to regulate a single session length.
     [[#1961]]
+ -  (Libplanet.Net) General API changes made to `Swarm<T>.BootstrapAsync()`,
+    `Swarm<T>.PreloadAsync()`, and `Swarm<T>.StartAsync()`.  [[#1962]]
+     -  `Swarm<T>.BootstrapAsync(IEnumerable<Peer>, double, double,
+        int, CancellationToken)` replaced with
+        `Swarm<T>.BootstrapAsync(IEnumerable<Peer>, int, CancellationToken)`
+        which uses default values provided by `SwarmOptions` for
+        `pingSeedTimeout` and `findPeerTimeout`.
+     -  `Swarm<T>.StartAsync(int, int, int CancellationToken)`
+        replaced with `Swarm<T>.StartAsync(CancellationToken)`
+        which uses default values provided by `SwarmOptions` for
+        `millisecondsDialTimeout`, `millisecondsBroadcastBlockInterval`
+        and `millisecondsBroadcastTxInterval`.
+     -  `Swarm<T>.PreloadAsync(IProgress<PreloadState>,
+        bool, long, CancellationToken)` overload method added and
+        `Swarm<T>.PreloadAsync(TimeSpan?, IProgress<PreloadState>, bool
+        long, CancellationToken)` changed to `Swarm<T>.PreloadAsync(
+        TimeSpan, IProgress<PreloadState>, bool, long, CancellationToken)`
 
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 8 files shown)


longfin:
Well... `TimeoutOptions` class seemed that classify its format. IMHO, many of classification is more natural for me when classify by semantics. for example, Containing `.GetBlockHashesTimeout` and `.GetBlocksBaseTimeout` in the same class is natural for me, but `.BootstrapDialTimeout` and `.GetBlockHashesTimeout` isn't... 😕 

Is there any useful case for grouping all timeouts in one class? (Ignoring the usage and wanting to manipulate only the whole timeout)

greymistcube:
@longfin
I think it depends on how we interpret it. Yes, `BootstrapDialTimeout` is kind of different from `GetBlockHashesTimeout`, but these are ultimately all used **only** to be passed on to `ITransport.SendMessageAsync()`, and in that sense they are also kind of the same. As there are three phases for `Swarm<T>`, namely `BootstrapAsync()`, `PreloadAsync()`, and `StartAsync()`, and each phase has different set of `Message` types that they deal with, if we take this to the extreme, we might get something like:

- `SwarmOptions`
  - `TimeoutOptions`
    - `BootstrapOptions`
      - `PingTimeout`
      - `FindNeighborsTimeout`
    - `PreloadOptions`
      - `PingTimeout`
      - ...

You get the idea.

Another way to organize this would be:

- `SwarmOptions`
  - `BootstrapOptions`
    - `TimeoutOptions`
      - `PingTimeout`
      - `FindNeighborsTimeout`
    - `PreloadOptions`
     - `PingTimeout`
     - ...

In this case, yes, it kind of feels unnatural to have `BootstrapDialTimeout` in the same hierarchy as `GetBlockHashesTimeout`. I do agree the latter design would probably be better further down the road, but I decided to go with the former for the following reasons:

- I was lazy; I didn't want to create more than two layers at this point.
- It seemed silly to have `BootstrapDialTimeout` in `SwarmOptions`.
- Again, I was lazy; separating out `SwarmOptions`, `BootstrapOptions`, `PreloadOptions`, `AsyncOptions` seemed like too much work. Couldn't figure out how there'd be a nice sensible workable unit for an iterative process.

> Is there any useful case for grouping all timeouts in one class? (Ignoring the usage and wanting to manipulate only the whole timeout)

Not really. 😗
Joking aside, from usage perspective, for the most part, I thought it'd be better to just **dump everything** during `Swarm<T>()` and make it just work if possible. I just had to stick it in **somewhere** so I could work towards `Swarm<T>.BootstrapAsync()`, `Swarm<T>.PreloadAsync()`, etc. without any arguments.

longfin:
> I was lazy; I didn't want to create more than two layers at this point.

Agree, I also think that we need to flatten hierarchy as possible. so more comportable structure in my mind was...

- SwarmOptions
  - BootstrapOptions
    - PingTimeout
    - FindNeighborsTimeout
  - PreloadOptions
    - PingTimeout
  - BlockSyncOptions(?)
    - GetBlockHashesTimeout 

But... well, OK I realize that, there is too many `dialTimeout`, `pingTimeout` and any other timeouts in [`Swarm<T>` API](https://docs.libplanet.io/0.34.0/api/Libplanet.Net.Swarm-1.html). In this situation, I think it might not be strange to tie them together in terms of waiting for responses from other peers. so I also think it might be a good idea to group them by request/response message type. (I think the options are organized in this PR with this perspective)

However on the other side, I still haven't found a reason to separate these timeouts by layering them in `SwarmOption` (`Options.TimeoutOptions.DialTimeout` instead of `Options.DialTimeout`), other than a visual cleanup of the code...

greymistcube:
> However on the other side, I still haven't found a reason to separate these timeouts by layering them in SwarmOption (Options.TimeoutOptions.DialTimeout instead of Options.DialTimeout), other than a visual cleanup of the code...

It's just a starting point to have at least some semblance of organization. Neither am I particularly happy with the current result at the moment. 😐

I wanted to track down each option and:

- Figure out what it does and what it is meant to do as to clarify its usage
- Fix internal inconsistency, inappropriate uses, and/or remove redundancy

However, doing this to a whole of `SwarmOptions` turns out to be lot more tedious and time consuming than it seems. I just needed to have some "scope" to work with, and put them somewhere. This probably will be temporary and ultimately `TimeoutOptions` might be short lived and I also agree there is a bit of code smell, I really don't think it stinks any more than the current state `SwarmOptions` is in; in all honesty, I consider it to be a complete mess anyway in its current form.

On the other hand, I don't quite agree with keeping `SwarmOptions` completely flat. 😕

longfin (dismissed):

riemannulus (approved):

CHANGES.md
```diff
@@ -48,21 +69,19 @@ To be released.
 
  -  The following store URI schemes are deprecated:  [[#1573], [#1955]]
 
-     - `default`: Use `default+file` instead.
-     - `rocksdb`: Use `rocksdb+file` instead.
-
```
dahlia, line 53: Why were these lines removed?

CHANGES.md
```diff
@@ -48,21 +69,19 @@ To be released.
...
+[#1957]: https://github.com/planetarium/libplanet/issues/1957
 [#1959]: https://github.com/planetarium/libplanet/issues/1959
 [#1960]: https://github.com/planetarium/libplanet/pull/1960
 [#1961]: https://github.com/planetarium/libplanet/pull/1961
+[#1962]: https://github.com/planetarium/libplanet/pull/1962
 [#1966]: https://github.com/planetarium/libplanet/pull/1966
 
 
 Version 0.34.0
---------------
```
dahlia, line 65: You probably made some mistakes during rebasing.

dahlia (changes requested):

CHANGES.md
greymistcube: I'm fixing it currently. I accidently removed some links also while rebasing. 😅

CHANGES.md
```diff
@@ -27,6 +44,10 @@ To be released.
  -  Added `NameValueCollectionExtensions` static class.  [[#1955]]
  -  Added `TimeSpanExtensions.Multiply()` static method.  [[#1966]]
  -  Type support for `Guid` added to `DataModel`.  [[#1959], [#1960]]
+ -  `TimeSpanExtensions.Multiply()` method added.  [[#1966]]
```
greymistcube, line 47: Still needs a fix. 😑

CHANGES.md
greymistcube: Should be all fixed I think...? Can you check again for me? 😅

dahlia (approved):

PR #1968, opened 2022-05-18
Introduce `Vote` and `VoteSet`

riemannulus:
## Context
For migration PoS, We need some struct for recording who participated in the block validation.

## Rationale
```mermaid
classDiagram
    Block~T~ o-- Commit
    Commit *-- Vote
    VoteSet *-- Vote
    class Block~T~ {
        +Commit LastCommit
    }

    class Commit {
        +Vote[] Votes
    }
    
    class Vote {
        + long Height
        + int Round
        + PublicKey Validator
        + TimeStmap timeStamp 
    }

    class VoteSet {
        -Vote[] _prevote
        -Vote[] _commit
    }
```

PR #1969, opened 2022-05-18
Introduce `BlockCommit`

riemannulus:
continue on the #1968 

## Context
For migration PoS, We need some struct for recording who participated in the block validation.

## Decision
```mermaid
classDiagram
    Block~T~ o-- Commit
    Commit *-- Vote
    VoteSet *-- Vote
    class Block~T~ {
        +Commit LastCommit
    }

    class Commit {
        +Vote[] Votes
    }
    
    class Vote {
        + long Height
        + int Round
        + PublicKey Validator
        + TimeStmap timeStamp 
    }

    class VoteSet {
        -Vote[] _votes
    }
```

### DIFF
--- a/Libplanet/Blocks/BlockHeader.cs
+++ b/Libplanet/Blocks/BlockHeader.cs
@@ -150,6 +150,8 @@ namespace Libplanet.Blocks
         /// <inheritdoc cref="IBlockMetadata.TxHash"/>
         public HashDigest<SHA256>? TxHash => _preEvaluationBlockHeader.TxHash;
 
+        public BlockCommit? LastCommit => _preEvaluationBlockHeader.LastCommit;
+
         /// <inheritdoc cref="IBlockHeader.Signature"/>
         public ImmutableArray<byte>? Signature { get; }
 
--- a/Libplanet.Tests/Blocks/BlockCommitTest.cs
+++ b/Libplanet.Tests/Blocks/BlockCommitTest.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+using Libplanet.Blocks;
+using Libplanet.Consensus;
+using Libplanet.Crypto;
+using Libplanet.Tests.Store;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Libplanet.Tests.Blocks
+{
+    public class BlockCommitTest
+    {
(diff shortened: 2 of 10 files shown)


Libplanet/Blocks/BlockHeader.cs
```diff
@@ -150,6 +150,8 @@ BlockHash Hash
         /// <inheritdoc cref="IBlockMetadata.TxHash"/>
         public HashDigest<SHA256>? TxHash => _preEvaluationBlockHeader.TxHash;
 
+        public BlockCommit? LastCommit => _preEvaluationBlockHeader.LastCommit;
```
limebell, line 153: Is `LastCommit` required in both `Block` and `BlockHeader`?

riemannulus: Recording to #1492, `BlockHeader` and `Block<T>` are implementations of `IBlockHeader` and `IBlock<T>`. 
`LastCommit` is a property of `IBlockMetadata` so it is required separately.

limebell: But we have `BlockHeader` property in `Block<T>`. You should use like
```csharp
    public BlockCommit? LastCommit => BlockHeader.LastCommit;
```
see also https://github.com/planetarium/libplanet/blob/9e874b24a0c2b0cf188adbf9bf52f20a1cfdb613/Libplanet/Blocks/Block.cs

riemannulus: Aha, you mean https://github.com/planetarium/libplanet/pull/1969/files/4c8ed9c8645051edf523af07a5cb7d73a84825cc#diff-81f06fcc7609e3eaf7fa94e98ce460fdaf9ec8923766e1dee100c7e03af88aebR146 here? I'm confused you made a claim on the `BlockHeader` lol. Yes, you're right. I mistake it. I'll amend it.

limebell: Oh I should've leave this comment in `Block.cs` :)

riemannulus: c6f5c3f2e070fa222a0ea450e13590a695973840 amended

riemannulus:
PTAL 🙏  @limebell

limebell (approved):

PR #1978, opened 2022-05-20
`UntypedBlock` & `UntypedTransaction`

### LINKED ISSUE #1974: `UntypedBlock` & `UntypedTransaction`
Sometimes we need to deal with serialized blocks and transactions with unknown type `T : IAction` at compile time.  However, `Block<T>` and `Transaction<T>` require their type parameter `T` to be concrete before instantiating their objects.

Unless we need to actually evaluate actions, we don't have to know what these action types are in reality.  So, I suggest non-generic version of `Block<T>` & `Transaction<T>`: `UntypedBlock` & `UntypedTransaction`.  Although they cannot evaluate their actions, they would be still useful as fields other than `Actions` would be validated at least.

dahlia:
*Closes #1974, and partially addresses <https://github.com/planetarium/libplanet/issues/1164#issuecomment-1135297670>.*

This introduces `UntypedBlock` and `UntypedTransaction` under a new project named *Libplanet.Node*.


Libplanet/Tx/ITxExcerpt.cs
```diff
@@ -0,0 +1,16 @@
+namespace Libplanet.Tx
+{
+    /// <summary>
+    /// Similar to <see cref="ITxMetadata"/> except that it has <see cref="TxId"/> as well.
+    /// Note that this does not contain actions or signature.
+    /// </summary>
+    public interface ITxExcerpt : ITxMetadata
```
greymistcube, line 7: Can we clarify it's use a bit more?

As I understand it, seems like this is to "group" `UntypedTransaction` and `Transaction<T>` for validating nonces. In any case, the relation between `ITxExcerpt` and `ITxMetadata` should mirror that of `IBlockExcerpt` and `IBlockMetadata` if possible. 😕

dahlia: I defined the term *excerpt* as a set of metadata functioning as a proof without complete data (usually lacking its body like actions and transactions).  That's why `IBlockExcerpt` has `Hash`, and `ITxExcerpt` has `Id`.

greymistcube: > I defined the term excerpt as a set of metadata

I see *excerpts* are some sort of metadata. On the other hand, `ITxMetadata` and `IBlockMetadata` have analogous specific uses. I'm still confused as to what this is "proving" really and if `ITxExcerpt : ITxMetadata` is necessary.

- If the term *excerpt* is to *extend* metadata interfaces with additional proof data, than inheritance is granted.
- If the term *excerpt* is to be used for naming an auxiliary `interface` with minimal amount of information than there shouldn't be an inheritance relation.

I'm suggesting that either we have `IBlockExcerpt : IBlockMetadata` and `ITxExcerpt : ITxMetadata` (this is the latter of the above) or not have inheritance relations between them (the former of the above). The split design seems confusing.

dahlia: Ah, I see.

greymistcube: Not to pester, but is this being worked on? I'm done with ~skimming~ reviewing the code, but I wasn't sure about your last comment. 😶

dahlia: I tried to make `ITxMetadata` to inherit nothing, but it wasn't that easy as `TxExcerptExtensions.ValidateTxNonces()` used several properties of `ITxMetadata`.  Instead, I'm going to rename its name later.

greymistcube: 👌

greymistcube (approved):

Libplanet/Tx/ITxExcerpt.cs
dahlia: <https://github.com/planetarium/libplanet/issues/1997>

PR #1979, opened 2022-05-20
Using `Vote` and `VoteSet` on the PBFT

riemannulus:
continue on the https://github.com/planetarium/libplanet/pull/1968

## Context
Previously, voting was conducted based on the peer's public key. However, this has the following problems:

1. It can be trusted that the message comes from the node that signed the message, but it cannot be verified that the message is a signed message with the validator's key.
2. In order to save the Vote to the Block, it was necessary to directly sign the Vote.

## Decision
For the same reason as above, it has been implemented to ship and send `Vote` in the message.

```mermaid
classDiagram
    ConsensusContext o-- RoundContext~T~
    RoundContext~T~ o-- VoteSet
    ConsensusVote *-- Vote
    VoteSet *-- Vote
    ConsensusCommit *-- Vote
    class ConsensusContext {
        -Dict~< long, RoundContext< T>>~ _roundContexts
    }

    class RoundContext~T~ {
        -VoteSet _voteSet
        +Vote Voting(VoteFlag flag)
        +long Vote(Vote vote)
    }

    class ConsensusVote {
        + ConsensusVote(Vote)
    }

    class ConsensusCommit {
        + ConsensusCommit(Vote)
    }
    
    class Vote {
        + long Height
        + int Round
        + PublicKey Validator
        + TimeStmap timeStamp 
    }

    class VoteSet {
        -Vote[] _votes
        + bool Add()
    }
```

### DIFF
--- a/Libplanet.Net/Consensus/PreVoteState.cs
+++ b/Libplanet.Net/Consensus/PreVoteState.cs
@@ -43,7 +43,6 @@ namespace Libplanet.Net.Consensus
                 return null;
             }
 
-            roundContext.ResetVote();
             roundContext.State = new PreCommitState<T>();
             return new ConsensusCommit(roundContext.Voting(VoteFlag.Commit));
         }
--- a/Libplanet.Net/Consensus/PreVoteState.cs
+++ b/Libplanet.Net/Consensus/PreVoteState.cs
@@ -1,4 +1,5 @@
 using Libplanet.Action;
+using Libplanet.Consensus;
 using Libplanet.Net.Messages;
 
 namespace Libplanet.Net.Consensus
@@ -35,7 +36,7 @@ namespace Libplanet.Net.Consensus
             }
 
             RoundContext<T> roundContext = context.CurrentRoundContext;
-            roundContext.Vote(vote.Remote?.Address);
+            roundContext.Vote(vote.ProposeVote);
 
             if (!roundContext.HasTwoThirdsAny())
             {
@@ -44,11 +45,7 @@ namespace Libplanet.Net.Consensus
 
             roundContext.ResetVote();
             roundContext.State = new PreCommitState<T>();
-            return new ConsensusCommit(
-                context.NodeId,
-                context.Height,
-                context.Round,
-                roundContext.BlockHash);
+            return new ConsensusCommit(roundContext.Voting(VoteFlag.Commit));
         }
     }
 }
--- a/Libplanet.Net/Consensus/RoundContext.cs
+++ b/Libplanet.Net/Consensus/RoundContext.cs
@@ -99,24 +99,22 @@ namespace Libplanet.Net.Consensus
             lock (_lock)
             {
                 Log.Debug(
-                    "Vote({Before}/{Total}) NodeID: {Id}, Validator: {Address}, " +
-                    "Height: {Height}, Round: {Round}",
-                    _voteSet.Votes.Length,
+                    "Vote({Vote}/{Commit}/{Total}) NodeID: {Id}, " +
+                    "Validator: {Address}, Height: {Height}, Round: {Round}",
+                    VoteCount,
+                    CommitCount,
                     _numberOfValidator,
                     NodeId,
                     vote.Validator.ToAddress(),
                     Height,
                     Round);
-            }
-
-            lock (_lock)
-            {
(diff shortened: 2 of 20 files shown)


Libplanet.Net/Consensus/PreVoteState.cs
```diff
@@ -44,11 +45,7 @@ public class PreVoteState<T> : IState<T>
 
             roundContext.ResetVote();
```
limebell, line 46: I think this is not required because it became to distinguish votes by its certain property?

Libplanet.Net/Consensus/RoundContext.cs
```diff
@@ -52,46 +57,69 @@ public class RoundContext<T>
...
+                CurrentNodePublicKey,
+                flag,
+                NodeId,
+                Encoding.Default.GetBytes("sign").ToImmutableArray());
+        }
 
-            if (address is null)
+        public long Vote(Vote vote)
+        {
+            lock (_lock)
```
limebell, line 99: This lock block is not required. How about move this logging code into the next lock block?

Libplanet.Net/Consensus/RoundContext.cs
```diff
@@ -52,46 +57,69 @@ public class RoundContext<T>
...
             {
-                throw new ArgumentNullException(nameof(address));
+                Log.Debug(
+                    "Vote({Before}/{Total}) NodeID: {Id}, Validator: {Address}, " +
+                    "Height: {Height}, Round: {Round}",
+                    _voteSet.Votes.Length,
```
limebell, line 104: This value is always equal to `_numberOfValidator` because `Votes` is an array.

Libplanet.Net/Consensus/RoundContext.cs
```diff
@@ -52,46 +57,69 @@ public class RoundContext<T>
...
             {
-                if (!_votes.Contains(address.Value))
+                if (_voteSet.Add(vote))
                 {
-                    _votes.Add(address.Value);
                     Log.Debug(
                         "Vote Success({After}/{Total}) NodeID: {Id}, " +
                         "Validator: {Address}, Height: {Height}, Round: {Round}",
-                        _votes.Count,
+                        _voteSet.Votes.Length,
```
limebell, line 119: This value is always equal to `_numberOfValidator` because `Votes` is an array.

Libplanet.Net/Messages/ConsensusCommit.cs
```diff
@@ -1,19 +1,35 @@
...
+
+        public override IEnumerable<byte[]> DataFrames
+        {
+            get
+            {
+                var frames = new List<byte[]>
+                {
+                    CommitVote.ByteArray,
+                };
+                return frames;
```
limebell, line 29: May be missing `base.DataFrames`?

Libplanet.Net/Messages/ConsensusVote.cs
```diff
@@ -1,19 +1,35 @@
...
+
+        public override IEnumerable<byte[]> DataFrames
+        {
+            get
+            {
+                var frames = new List<byte[]>
+                {
+                    ProposeVote.ByteArray,
+                };
+                return frames;
```
limebell, line 29: May be missing `base.DataFrames`?

Libplanet.Net/Messages/ConsensusCommit.cs
riemannulus: https://github.com/planetarium/libplanet/pull/1979/files#diff-8f7496d7becdca3f0362db3cbd3c09f4c0cc16923436b0f4aeec5e094be984bcR14-R15 You should check out here. `base.DataFrames` is unnecessary. it initialized on the constructor.

Libplanet.Net/Messages/ConsensusVote.cs
riemannulus: https://github.com/planetarium/libplanet/pull/1979/files#diff-8f7496d7becdca3f0362db3cbd3c09f4c0cc16923436b0f4aeec5e094be984bcR14-R15 You should check out here. base.DataFrames is unnecessary. it initialized on the constructor.

riemannulus:
I amend it. PTAL! @limebell

limebell (approved):
Good

PR #2002, opened 2022-05-25
Block Synchronization (refactored)

colibrishin:
## Summary
- `ConsensusPropose` will not broadcast 'ConsensusVote` if Block is not present in the store. (= Hold the vote)
- `PreVoteState<T>` will hold the node's vote until the proposed block is present in the store.
- When a node doesn't have a proposed block from `ConsensusMessage`, it will request the proposed block from the sender.

## Further notices
- Currently, `CommitBlock()` will be called even if the Block does not exist in the store.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -103,8 +103,8 @@ namespace Libplanet.Net.Consensus
                     hash,
                     NodeId);
 
-                // TODO: There's no coping mechanism when the proposed block is not present
-                // in commit stage.
+                // TODO: Needs additional block synchronization and recommit sequence if proposed
+                // block is not present in commit stage.
                 Block<T> block = _blockChain.Store.GetBlock<T>(
                     _blockChain.Policy.GetHashAlgorithm,
                     hash);
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -72,6 +72,8 @@ namespace Libplanet.Net.Consensus
         /// </summary>
         public long NodeId { get; internal set; }
 
+        public HashAlgorithmGetter HashAlgorithm => _blockChain.Policy.GetHashAlgorithm;
+
         public RoundContext<T> CurrentRoundContext => RoundContextOf(Round);
 
         // FIXME: Storing all voteset on memory is not required. Leave only 1~2 votesets.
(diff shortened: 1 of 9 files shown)


Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -87,6 +94,8 @@ public void CommitBlock(long height, BlockHash hash)
                     hash,
                     NodeId);
 
+                // TODO: There's no coping mechanism when the proposed block is not present
+                // in commit stage.
```
limebell, line 98:
```suggestion
                // TODO: There's no copying mechanism when the proposed block is not present
                // in the commit stage.
```

Libplanet.Net/Consensus/ConsensusReactor.cs
```diff
@@ -127,12 +131,98 @@ private async Task ProcessMessageHandler(Message message)
...
+                    _logger.Debug(
+                        "{MethodName}: Received Block {BlockHash} from {Remote}",
+                        nameof(RequestBlockNotExistsAsync),
+                        unmarshalBlock.Hash,
+                        message.Remote);
+
+                    if (_context.IsVoteOnHold)
+                    {
+                        HandleMessage(
+                            new ConsensusVote(_context.CurrentRoundContext.Voting(VoteFlag.Absent)));
```
limebell, line 188: Line must be no longer than 100 characters.

Libplanet.Net/Consensus/ConsensusReactor.cs
```diff
@@ -127,12 +131,98 @@ private async Task ProcessMessageHandler(Message message)
...
+                _logger.Debug(
+                    "{MethodName}: Received Block request {BlockHash} from {Remote}",
+                    nameof(RequestBlockNotExistsAsync),
+                    message.BlockHashes.First(),
+                    message.Remote);
+
+                await _transport.ReplyMessageAsync(sending, CancellationToken.None);
+            }
+
+            await ReplyPongAsync(message);
```
limebell, line 219: I don't think this is required because a reply `sending` already sent? I guess you should make an `else` code block and return `Pong`?

limebell (changes requested):

Libplanet.Net/Consensus/ConsensusContext.cs
colibrishin: It was meant to say coping, however, the words seem too broad and vague. I've replaced the sentence with better context. How about this?

https://github.com/planetarium/libplanet/blob/5c60fd2283435b83f34ff871847ac9985ad4e0f6/Libplanet.Net/Consensus/ConsensusContext.cs#L97-L98

limebell: Sorry my bad. `Coping` is the right word ;)

riemannulus:
> Currently, CommitBlock() will be called even if the Block does not exist in the store.
Then, if the block is broadcasted later than `CommitBlock()`. what happened?

colibrishin:
> > Currently, CommitBlock() will be called even if the Block does not exist in the store.
> 
> Then, if the block is broadcasted later than `CommitBlock()`. what happened?

The State will be frozen into `PreCommitState<T>` and at the height and round and the other `ConsensusMessage` will be discarded. The missed block is requested nowhere, however, the next block for height and round is still be requested and stored.

riemannulus (commented):
LGTM but could you rebase PR?

PR #2003, opened 2022-05-25
Verify signature when received PBFT vote message

riemannulus:
# Context
Previously, each node did not validate the `Vote` message when they received it.
This vulnerability causes counterfeit of the `Vote` message from the attacker.

# Decision
For the same reason as above, this PR is implemented to verify and sign about `Vote`.  

```mermaid
sequenceDiagram
    State ->> RoundContext: call `Voting()`.
    RoundContext ->> State: return `Vote` with unsigned.
    State ->> ConsensusContext: call `SignVote()`.
    ConsensusContext ->> State: return `Vote` with signed.
    State ->> VoteSet: call `Add()`
    VoteSet ->> VoteSet: verify and validate `Vote`
    VoteSet ->> State: return bool which `Add()` is success
```

### DIFF
--- a/Libplanet.Net.Tests/TestUtils.cs
+++ b/Libplanet.Net.Tests/TestUtils.cs
@@ -14,10 +14,15 @@ namespace Libplanet.Net.Tests
 {
     public static class TestUtils
     {
-        public static readonly Peer Peer0 = new Peer(
-            new PublicKey(
+        public static readonly PrivateKey Peer0Priv =
+            new PrivateKey(
                 ByteUtil.ParseHex(
-                    "0204b09e833560269d0dc1100c221442b9969e74a3949ca306aecda20a6a2afba4")));
+                    "b17c919b07320edfb3e6da2f1cfed75910322de2e49377d6d4d226505afca550"));
+
+        public static readonly Peer Peer0 = new Peer(
+            new PrivateKey(
+                ByteUtil.ParseHex(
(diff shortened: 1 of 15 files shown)


Libplanet.Net.Tests/TestUtils.cs
```diff
@@ -112,6 +119,21 @@ public static class TestUtils
                 id,
                 ImmutableArray<byte>.Empty);
 
+        public static Vote SignVote(Vote vote, PrivateKey privateKey)
```
limebell, line 122: I think this should be included in `Vote`, like `Vote.Sign()`. We can consider making vote as `class` (was `struct`).

riemannulus: [4afb5c6](https://github.com/planetarium/libplanet/pull/2003/commits/4afb5c612cee8bb7841d2b5a73fb45f6a358cf6d) Amended.

limebell: Do we still need this test function? How about replace them with `.Sign()`?

riemannulus: To be honest. it was bothering me... Eh, I'll fix it up.

riemannulus: [4e61652](https://github.com/planetarium/libplanet/pull/2003/commits/4e61652f941f5be7477e835df5b456184def17a3) Amended!

PR #2009, opened 2022-05-26
Store voteset on memory when commit new block and access from reactor

limebell:
When commiting a new block, the votes for the block is stored in order to make next block. The votes collected will be included in `BlockHeader.LastCommit` field.

Additionally, fixed a bug where context committed block even commit count does not met the condition (2/3+).


riemannulus:
I rebase your PR.

riemannulus (approved):
LGTM

PR #2013, opened 2022-05-27
✨ Introduce `NodeUtils` class

greymistcube:
Serves two purposes:
- Removes the need to directly access "raw" class constructors as much as possible.
- Provide common basic functionality necessary to writing a boilerplate code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@ To be released.
 
 ### Added APIs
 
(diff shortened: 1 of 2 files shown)


greymistcube:
I thought writing changelogs might not be necessary (or it's just me being lazy) as the API is prone to change rapidly. We should, however, take care in writing changelogs before merging `libplanet-node` to `main`. 😶

dahlia (approved):

Issue #2015, opened 2022-05-28
Allow `0x` prefix for hexstring on `Address()` constructor


longfin:
Many case, we're using `0x` prefix when representing `Address` as hexstring. 

e.g. https://github.com/planetarium/libplanet/blob/2086fb9250884b4b12ff7a2af740166d78a175df/Libplanet/Address.cs#L243

So I think that it would be better `Address()` constructor allows `0x` prefixed address as well.

kasumielf:
I'll take it.

PR #2018, opened 2022-05-28
Allow 0x prefix for Address class

kasumielf:
It addresses #2015.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@ To be released.
 ### Added APIs
 
 ### Behavioral changes
+ - Allow `0x` prefix for `Address` class constructor.
+   `DeriveAddress` method will remove '0x' prefix string. [[#2015]] 
 
 ### Bug fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@ To be released.
 
 ### Behavioral changes
  - Allow `0x` prefix for `Address` class constructor.
-   `DeriveAddress` method will remove '0x' prefix string. [[#2015]] 
+    `DeriveAddress` method will remove '0x' prefix string. [[#2015], [#2018]]
 
 ### Bug fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,7 @@ To be released.
 
 ### CLI tools
 
+[#2015]: https://github.com/planetarium/libplanet/issues/2015
 
 Version 0.36.1
 --------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@ To be released.
 ### Added APIs
 
 ### Behavioral changes
- - Allow `0x` prefix for `Address` class constructor.
+ -  Allow `0x` prefix for `Address` class constructor.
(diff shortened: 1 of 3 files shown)


Libplanet/Address.cs
```diff
@@ -320,6 +320,13 @@ private static byte[] DeriveAddress(string hex)
                 throw new ArgumentNullException(nameof(hex));
             }
 
+            int pos = hex.IndexOf('x');
+
+            if (pos >= 0)
+            {
+                hex = hex.Remove(0, pos + 1);
+            }
+
```
longfin, line 329: I guess, it would be better if we have sanity check before this removal because we don't have to consider about other case than length is 42.

```suggestion
            if (hex.Length == 42)
            {
                int pos = hex.IndexOf('x');
                if (pos >= 0)
                {
                    hex = hex.Remove(0, pos + 1);
                }
            }
```

CHANGES.md
```diff
@@ -17,13 +17,16 @@ To be released.
 ### Added APIs
 
 ### Behavioral changes
+ - Allow `0x` prefix for `Address` class constructor.
+   `DeriveAddress` method will remove '0x' prefix string. [[#2015]]
```
longfin, line 21:
```suggestion
   `DeriveAddress` method will remove '0x' prefix string. [[#2015], [#2018]]
```
Also, we need to a link for #2018 too 😅

CHANGES.md
```diff
@@ -17,13 +17,16 @@ To be released.
 ### Added APIs
 
 ### Behavioral changes
+ - Allow `0x` prefix for `Address` class constructor.
```
greymistcube, line 20:
```suggestion
 -  Allow `0x` prefix for `Address` class constructor.
```

CHANGES.md
greymistcube:
```suggestion
    `DeriveAddress` method will remove '0x' prefix string. [[#2015], [#2018]]
```
Minor, but for spacing consistency. 😗

Libplanet/Address.cs
greymistcube: Eh, I think we need to explicitly check if the prefix is `0x`. 😶
This matches regex `.+x`. Although the if statement below would only let `.x`, this is still not the same as removing `0x`.

CHANGES.md
```diff
@@ -17,13 +17,16 @@ To be released.
...
+   `DeriveAddress` method will remove '0x' prefix string. [[#2015]] 
 
 ### Bug fixes
 
 ### Dependencies
 
 ### CLI tools
 
+[#2015]: https://github.com/planetarium/libplanet/issues/2015
```
greymistcube, line 29:
```suggestion
[#2015]: https://github.com/planetarium/libplanet/issues/2015
[#2018]: https://github.com/planetarium/libplanet/pull/2018
```

Libplanet/Address.cs
longfin: Agree for that. I think it's some kind of room(?") for first time contributer😅. we may need to follow for that soon.

longfin (approved):

riemannulus (approved):

limebell (dismissed):
LGTM

Issue #2012, opened 2022-05-26
Make `PrivateKey()` also accept hex `string`


greymistcube:
Although internally within [libplanet] `PrivateKey` as `byte` array might be more useful, on the client facing side, direct conversion from and to a hex `string` could be more intuitive and useful. 🙂

[libplanet]: https://github.com/planetarium/libplanet

PR #2022, opened 2022-05-29
Add PrivateKey constructor for hexadecimal string.

kasumielf:
- Add PrivateKey(string) constructor
- Create GenerateBytesFromHexString(string) private method.
- Modify FromString method.
- Add PrivatKey(string) constructor test.

It addresses #2012.


longfin (dismissed):

riemannulus (dismissed):
👍

PR #2023, opened 2022-05-30
Recomitting Failed Round / Preparing Swarm

colibrishin:
- Block synchronization has been replaced with Event. (`VoteHolding`, `RequestBlockInVoteHolding`)
- If committing was failed before, getting block and recommit will be retried in methodically. (`CommitFailed`, `RecommittingFailedRound`). This is not guaranteed in `IState` level yet.
  - Answering `Blocks` is removed for letting Swarm answering (TODO: Swarm answers messages?). Above implementations should work without Swarm If response method with `Blocks` is added.
- Some Preparation for migration to Swarm.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -3,13 +3,13 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Text.Json;
 using System.Timers;
-using Bencodex;
 using Libplanet.Action;
 using Libplanet.Blockchain;
 using Libplanet.Blocks;
 using Libplanet.Consensus;
 using Libplanet.Crypto;
 using Libplanet.Net.Messages;
+using Nito.AsyncEx;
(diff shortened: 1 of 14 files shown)


Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -103,12 +117,22 @@ public void CommitBlock(long height, BlockHash hash)
...
-                // TODO: Needs additional block synchronization and recommit sequence if proposed
-                // block is not present in commit stage.
                 Block<T> block = _blockChain.Store.GetBlock<T>(
                     _blockChain.Policy.GetHashAlgorithm,
                     hash);
-                _blockChain.Append(block);
+
+                // TODO: So if the validation of block is done in vote, then treating the
+                // BlockPolicyException, InvalidBlockException, InvalidTxNonceException is not
+                // needed?
```
riemannulus, line 126: I guess we validate those `Exception`s when `Voting`.

riemannulus (approved):

PR #2024, opened 2022-05-30
`BootstrapOptions` added to `SwarmOptions`

greymistcube:
Mainly to reduce clutter on higher layers. 😶

This intermediary `SwarmOptions` seems quite untidy at the moment, but it'll get better soon after I add `PreloadOptions`. 😗

### DIFF
--- a/Libplanet.Explorer.Executable/Program.cs
+++ b/Libplanet.Explorer.Executable/Program.cs
@@ -17,6 +17,7 @@ using Libplanet.Explorer.Executable.Exceptions;
 using Libplanet.Explorer.Interfaces;
 using Libplanet.Explorer.Store;
 using Libplanet.Net;
+using Libplanet.Net.Protocols;
 using Libplanet.Store;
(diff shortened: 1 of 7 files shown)


Libplanet.Explorer.Executable/Program.cs
```diff
@@ -370,6 +371,7 @@ private static BlockPolicy<T> LoadBlockPolicy<T>(Options options)
                 await swarm.BootstrapAsync(
```
limebell, line 371: How about using newly introduced `Swarm<T>.BootstrapAsync(CancellationToken)` here?

greymistcube: Done. See https://github.com/planetarium/libplanet/pull/2024/commits/7e5fe076f5fb24593319d8d7c825bd2b8559ffd3. We probably need to overhaul `Program.cs` at some point. 🙃

limebell (approved):
👍

Issue #1995, opened 2022-05-24
Change `IceServer` to accept either single URL or parse URLs internally


greymistcube:
As seen [here][example1], `IceServer` allows a single username and credential pair for multiple URLs.

- There is no real use for allowing multiple URLs. This is even being "abused" in most cases as seen [here][example2].
- Class name suggests that it is a single entity.

If the naming is to suggest handling of ICE protocol, then a single instance of `IceServer` should be used to create a `TurnClient`, not a list of `IceServer`s.

There are two ways to go about this:

- Either change an `IceServer` instance to accept only single URL.
- Accept multiple URLs but parse internally.

[example1]: https://github.com/planetarium/libplanet/blob/7d6132750efd6767d323b350cefe9cc046ecd476/Libplanet.Net/IceServer.cs#L20-L28
[example2]: https://github.com/planetarium/libplanet/blob/7d6132750efd6767d323b350cefe9cc046ecd476/Libplanet.Explorer.Executable/Options.cs#L126-L156

PR #2026, opened 2022-05-30
IceServer now accepts only one URI.

Akamig:
Addresses https://github.com/planetarium/libplanet/issues/1995,
Also changed test conditions accordingly.

In my point of view, accepting multiple URI seems redundant, and may cause serious confusion. so I choose to accept only one URI. however, this might cause some side effects to the current codebase. I tested some use cases but cannot assure about that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,7 @@ To be released.
 [#2024]: https://github.com/planetarium/libplanet/pull/2024
 [#2026]: https://github.com/planetarium/libplanet/pull/2026
 
+
 Version 0.36.1
 --------------
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@ To be released.
     CancellationToken)` and made non-optional.  [[#2024]]
  -  (Libplanet.Net) `Swarm<T>.BootstrapAsync(IEnumerable<Peer>, depth,
     CancellationToken)` removed.  [[#2024]]
+ -  (Libplanet.Net) Parameter name `Urls` changed to `Url` for `IceServer`
+    and no longer accepts multiple Urls for single instance. [[#2026]]
 
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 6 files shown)


greymistcube:
Looks like almost done! We just need some changelog added to `CHANGES.md`. 🙂

CHANGES.md
```diff
@@ -39,7 +41,7 @@ To be released.
 ### CLI tools
 
 [#2024]: https://github.com/planetarium/libplanet/pull/2024
-
+[#2026]: https://github.com/planetarium/libplanet/pull/2026
```
greymistcube, line 44:
```suggestion
[#2026]: https://github.com/planetarium/libplanet/pull/2026

```

Libplanet.Net/IceServer.cs
```diff
@@ -10,24 +9,26 @@ namespace Libplanet.Net
...
-            IEnumerable<string> urls,
+            string url,
             string? username = null,
             string? credential = null)
-            : this(urls.Select(u => new Uri(u)), username, credential)
         {
+            Url = new Uri(url);
+            Username = username;
+            Credential = credential;
         }
```
greymistcube, line 19: Any reason redirection to `IceServer(Uri, string?, string?)` was removed? Redirection seems more DRY. 😗

Akamig: You mean this code is repetitive with the codes between line [21-29](https://github.com/Akamig/libplanet/blob/b863d36fcb100fb782b1f7b1ccb8dc9cb3c5f02f/Libplanet.Net/IceServer.cs#L21)?
I tried
```       
public IceServer(
            string url,
            string? username = null,
            string? credential = null)
            : this(url, username, credential)
        {
        }
```
But it didn't worked. how can I redirect string type url initializer to that Initializer?

greymistcube: You'd want something like

```csharp
public IceServer(
            string url,
            string? username = null,
            string? credential = null)
            : this(new Uri(url), username, credential)
        {
        }
```

Otherwise, you are recursively referencing the same constructor as there is no signature change. 🙃

Akamig: Yeap, this is what I wanted.

greymistcube (approved):

greymistcube (approved):

PR #2036, opened 2022-05-31
Newly styled up docs page.

Akamig:
Screenshot 2022-05-31 at 22-59-21 Libplanet Libplanet
Screenshot 2022-05-31 at 22-59-45 Class BlockChain T Libplanet

I'm pretty sure yall like this!

- Widen sidebar (270px -> 285px)
- Changed fonts (Pretendard, Roboto Mono)
- Planetarium BI Colored!
- New Code Highlight scheme!

### DIFF
--- a/Docs/theme/styles/main.css
+++ b/Docs/theme/styles/main.css
@@ -1,5 +1,5 @@
 @import url('https://fonts.googleapis.com/css2?family=Roboto+Mono:ital,wght@0,400;0,700;1,400;1,700&display=swap');
-@import url('https://cdn.jsdelivr.net/gh/orioncactus/pretendard/dist/web/static/pretendard-dynamic-subset.css');/* COLOR VARIABLES*/
+@import url("https://cdn.jsdelivr.net/gh/orioncactus/pretendard/dist/web/variable/pretendardvariable.css");/* COLOR VARIABLES*/
 :root {
     --header-bg-color: #3e2a8c;
     --header-ft-color: #f3f3f3;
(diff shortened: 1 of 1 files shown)


greymistcube:
Hmm... I'll try it later, but going by the screenshots, for whatever reason, font spacing still seems uneven and a bit off (unlike seen in [this page](https://cactus.tistory.com/306)). 😕

Akamig:
Hmm, Then I want to ask something. I used Pretendard solely because it supports Hangul and Eng without frustration between fonts. but after I look around docs there wasn't single hangul in out pages. Can I completely drop  hangul support  and change it to any Latin fonts not causing alignment issue?

Akamig:
Screenshot_20220601_140504

Alright, I checked really hard this time, and seems like container's floating point width (83.33333333%) is affecting all of contents, and since this is docfx's default stylesheet, I'd better not engage with it (layout breaks). but mediated condition by engaging variable font.

also, I really focused on 'Bound', becuase In my screen they seemed seperated normally to me, Turns out, Firefox's internal screenshot function don't capture some of it's rendering effects seen on users. so this time I used different software to capture the screen, and It seems okay for now.... I guess? at least it's same as what I look on browser. so yeah. 

If this doesn't work then I'm going to change main font to [Inter](https://fonts.google.com/specimen/Inter?category=Sans+Serif&subset=latin)

greymistcube:
I've checked on my local build and it seems fine. I guess original defects were probably due to quirks in taking screenshots. The latest screenshot still looks pretty bad though. 🙃

While you are at it, I feel like there is too much wasted screen real estate on both sides. Can we have less horizontal margins like [this](https://docs.microsoft.com/en-us/dotnet/api/system.console.writeline?view=net-6.0)?

Akamig:
Screenshot from 2022-06-02 13-59-13
Screenshot from 2022-06-02 13-59-02
Before

After
Screenshot from 2022-06-02 13-57-43
@greymistcube Can-do!

- NOW THIS PAGE USES FULL WIDTH WITHOUT MARGIN!
- ToC no longer overflows!
- Now you can SCROLL ToC!

riemannulus:
You can just add [skip changelog] to commit message for skip `changelog` check.

riemannulus (approved):

PR #2037, opened 2022-06-01
Improve logging about consensus

riemannulus:
## Context
- `RoundContext` is do not use `Serliog`.
- We want to filter only `Consensus` logs.

## Rationale
- Use `Serliog` in the `RoundContext`.
- Add some tagging to all `Consesnsus` logic.


riemannulus:
Oops, I missopend PR with wrong branch 😓

Issue #2054, opened 2022-06-07
Validate proposer of the block


limebell:
A proposer (which was 'miner') should be in the validator set and is valid for the height and round.

PR #2056, opened 2022-06-07
Validate `LastCommit` field of the block

limebell:
Add validation for `Block<T>.LastCommit`, added new exception `InvalidBlockLastCommitException` which inherits `InvalidBlockException`.

Implements #2054


Libplanet/Blocks/PreEvaluationBlockHeader.cs
```diff
@@ -183,6 +185,67 @@ ImmutableArray<byte> preEvaluationHash
...
+                                    return true;
+                                }
+
+                                if (vote.Flag == VoteFlag.Null || vote.Flag == VoteFlag.Unknown)
+                                {
+                                    return true;
+                                }
+
+                                return false;
+                            }))
```
riemannulus, line 226: We need to make an issue for the BLS signature instead Schnorr 🍈

riemannulus (approved):

Issue #2065, opened 2022-06-09
Provide `Schema` object instead of `Query`s


longfin:
(cont. from https://github.com/planetarium/libplanet/issues/2016#issuecomment-1151096535)

It would be useful when we provide `Schema` object (e.g., [StandaloneSchema](https://github.com/planetarium/NineChronicles.Headless/blob/d8db549c5eb64bf9ebd7d983f470dd98de99c766/NineChronicles.Headless/GraphTypes/StandaloneSchema.cs) in 9c) instead of `Query`s (e.g., [BlockQuery](https://github.com/planetarium/libplanet/blob/817782f3ebbca036be28888d535d18bac6562810/Libplanet.Explorer/Queries/BlockQuery.cs)) because we can simplify installation steps as bellow...

```csharp
    builder.Services
        .AddGraphQL(builder =>
        {
            builder
                .AddSchema<LibplanetExplorerSchema>()
                .AddGraphTypes(typeof(LibplanetExplorerSchema).Assembly)
                .AddSystemTextJson();
        })
        .AddCors()
        .AddSingleton<LibplanetExplorerSchema>()
        .AddSingleton<GraphQLHttpMiddleware<LibplanetExplorerSchema>>()

    using WebApplication app = builder.Build();
    app.UseRouting();
    app.UseEndpoints(endpoints =>
    {
        endpoints.MapGraphQLPlayground();
    });
    app.UseGraphQL<LibplanetExplorerSchema>();

    app.Run();
```

Note: In current Libplanet(0.38.0), we should define `Query` and `Schema` on the app side like ([1](https://github.com/planetarium/planet-node/blob/a74e17dc1c7637ef1eaa6db3f4d745f49645d47c/PlanetNode/GraphTypes/PlanetNodeQuery.cs#L13), [2](https://github.com/planetarium/planet-node/blob/a74e17dc1c7637ef1eaa6db3f4d745f49645d47c/PlanetNode/GraphTypes/PlanetNodeSchema.cs))

Thus, this issue was about `Schema` object, not schema file. 😅

longfin:
@moreal Do you have any idea for it? 🙏

kasumielf:
I want take this issue.

longfin:
# Motivation

[...]

PR #2074, opened 2022-06-13
Added GraphQL type descriptions

### LINKED ISSUE #1247: Explorer: GraphQL type descriptions
Our custom GraphQL types have no description.  We could provide short descriptions for schema introspection.

Akamig:
This PR resolves #1247, and "Tried but failed" to mediate #2065, but failed due to Generics of each queries in explorer.
Is there any good Idea to circumvent each IActions required by all queries?

### DIFF
--- a/Libplanet.Explorer/GraphTypes/BlockType.cs
+++ b/Libplanet.Explorer/GraphTypes/BlockType.cs
@@ -14,6 +14,7 @@ namespace Libplanet.Explorer.GraphTypes
         public BlockType()
         {
             #pragma warning disable SA1118
+            // We need multiple row of description for clearer, not confusing explanation of field.
             Field<NonNullGraphType<IdGraphType>>(
             "Hash",
             description: "A block's hash.",
--- a/Libplanet.Explorer/GraphTypes/BlockType.cs
+++ b/Libplanet.Explorer/GraphTypes/BlockType.cs
@@ -29,7 +29,8 @@ namespace Libplanet.Explorer.GraphTypes
             resolve: x => x.Source.Difficulty);
             Field<NonNullGraphType<BigIntGraphType>>(
             name: "TotalDifficulty",
-            description: "The total mining difficulty since the genesis including the block's Difficulty.",
+            description: @"The total mining difficulty since the genesis including
+            the block's Difficulty.",
             resolve: x => x.Source.TotalDifficulty);
             Field<NonNullGraphType<ByteStringType>>(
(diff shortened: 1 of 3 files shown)


longfin (changes requested):
@Akamig Sadly, the build is failing. 😢there are two main reasons.

1. Lint is complaining "Line must be no longer than 100 characters"
2. `StandaloneSchema` needs a type argument `T`.
  - but... I don't have confidence about `StandaloneSchema` itself. (because I guess it's too ambigus naming)
   - how about opening another PR for #2065 ?

Akamig:
@longfin 
Alright, I changed :
- Remove StandaloneQuery and Schema, which requires <T>.
- Shorten descriptions fit under 100 characters. Now it resolves only #1247.

longfin (changes requested):
@Akamig Oops, lint is now complaining that "The parameter spans multiple lines". 😢 

I guess it would be better checking `dotnet build` on your local env before commit & push. 🙏

Akamig:
...Which fails because of error SA1118: The parameter spans multiple lines. 
Maybe should I shorten description to fit under 100 characters? 
but It may confuse users because of lack of additional description.

longfin:
> but It may confuse users because of lack of additional description.

It seems that make sense. then, disabling lint using `#progma warning` (like [this](https://github.com/planetarium/libplanet/blob/ff57748bd58351a4b01da3bdc9a497996ace8a1f/Libplanet.Net.Tests/BoundPeerTest.cs#L62)) can be solution. cc @planetarium/libplanet

longfin (dismissed):

Libplanet.Explorer/GraphTypes/BlockType.cs
```diff
@@ -13,45 +13,78 @@ public class BlockType<T> : ObjectGraphType<Block<T>>
     {
         public BlockType()
         {
+            #pragma warning disable SA1118
```
riemannulus, line 16: Is it necessary? if that. IMO we need more detailed comments about [SA1118](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1118.md).

Akamig: @riemannulus 
You mean we need some justification message under that pragma disable?
Is commenting out the reason `//like this` would be enough?

riemannulus (approved):

PR #2079, opened 2022-06-14
Libplanet.Tools key command now Import/Exports Web3 Formatted json.

### LINKED ISSUE #1807: Allow importing `ProtectedPrivateKey` from Web3 Secret Storage directly
## Motivation
Users should be able to import their protected private key stored from the key file rather than storing an unprotected private key. This can improve usability and security (since users don't need to manage their private keys in an unprotected state).

## Proposal
- [~~Implement a way to construct `ProtectedPrivateKey ` from a valid Web3 Secret Storage key file.~~](https://docs.libplanet.io/0.27.4/api/Libplanet.KeyStore.ProtectedPrivateKey.html#Libplanet_KeyStore_ProtectedPrivateKey_FromJson_System_String_)
- (Optional, but feature wanted from the launcher) Allow users to load their key file in CLI interface.

## Reference
* https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition
* [`Libplanet.KeyStore.ProtectedPrivateKey`](https://docs.libplanet.io/0.27.4/api/Libplanet.KeyStore.ProtectedPrivateKey.html)
* [`Libplanet.KeyStore.Web3KeyStore`](https://docs.libplanet.io/0.27.4/api/Libplanet.KeyStore.Web3KeyStore.html)

Akamig:
this resolves #1807.
- key create --json outputs newly created key as json
- key import --json import Web3 formatted json. (path optional.)
- key export --json exports Web3 formatted json. (default path provided)

### DIFF
--- a/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
+++ b/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
@@ -38,7 +38,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             [Option(Description = "Do not add to the key store, but only show the created key.")]
             bool dryRun = false,
             [Option(Description = "Path to export key as Web3 Secret Storage Format")]
-            string path = "key.json"
+            string path = ""
         ) =>
             Add(string.Empty, passphrase, json, dryRun, path, true);
 
@@ -100,7 +100,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             [Option(Description = "Export a Web3 Secret Storage Formatted json to path.")]
             bool json = false,
             [Option(Description = "Path to export json key.")]
-            string path = "key.json"
+            string path = ""
         )
         {
             PrivateKey key = UnprotectKey(keyId, passphrase);
@@ -110,9 +110,9 @@ namespace Libplanet.Extensions.Cocona.Commands
                 using Stream stdout = Console.OpenStandardOutput();
                 stdout.Write(rawKey, 0, rawKey.Length);
             }
-            else if (json && path != string.Empty)
+            else if (json)
             {
-                FileStream fs = File.Create(path);
+                Stream fs = PathHandler(path);
                 var ppk = KeyStore.Get(keyId);
                 ppk.WriteJson(fs, keyId);
                 fs.WriteByte(0x0a);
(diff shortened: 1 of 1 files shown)


longfin:
It seems change `key export` behaviour from printing (to stdout) to storing as file. it would break other codes that assuming `key export` command's previous behviour. (e.g., [1](https://github.com/planetarium/9c-launcher/blob/713847355cda09a513b773eee0413f505c9c2ba8/src/main/headless/key-store.ts#L50-L58) [2](https://github.com/planetarium/planet-node/blob/a74e17dc1c7637ef1eaa6db3f4d745f49645d47c/TUTORIAL.md?plain=1#L164-L168))

In general, I prefer stdin/stdout other than file I/O because it can be easily combine through [Unix pipe](https://en.wikipedia.org/wiki/Pipeline_(Unix)), and also I believe that many POSIX software relying [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) doing this way,

@planetarium/libplanet @moreal @BasixKOR Have you any opinion about it?

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -248,24 +244,50 @@ public KeyCommand()
...
             }
             else
             {
-                PrintKeys(new[] { (keyId, ppk) });
+                if (json)
+                {
+                    if (File.Exists(path))
+                    {
+                    ProtectedPrivateKey ppk = ProtectedPrivateKey.FromJson(File.ReadAllText(path));
+                    Guid keyId = dryRun ? Guid.NewGuid() : KeyStore.Add(ppk);
```
limebell, line 277:
```suggestion
                        ProtectedPrivateKey ppk = ProtectedPrivateKey.FromJson(File.ReadAllText(path));
                        Guid keyId = dryRun ? Guid.NewGuid() : KeyStore.Add(ppk);
```
Just for coding style agreement in out products :)

Akamig:
Now output to stdout became default behaviour.
regardless of that, I fixed a bug previously requiring passphrase for json import (it doesn't do anything sice it's predetermined)
duplicated code inserted result of that, but I defend myself for better behavior for users.

I tested out some edge cases for path and file creation, and I think it's okay for now.

riemannulus (approved):

PR #2081, opened 2022-06-16
Unit Tests for new PBFT algorithm

colibrishin:
WIP, Round-Skip and Timeout, NIL vote +2/3 need to be tested.

### DIFF
--- a/Libplanet.Net.Tests/TestUtils.cs
+++ b/Libplanet.Net.Tests/TestUtils.cs
@@ -1,12 +1,18 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
 using Libplanet.Blockchain;
 using Libplanet.Blockchain.Policies;
 using Libplanet.Blocks;
 using Libplanet.Consensus;
 using Libplanet.Crypto;
+using Libplanet.Net.Consensus;
+using Libplanet.Net.Messages;
 using Libplanet.Net.Protocols;
+using Libplanet.Net.Transports;
 using Libplanet.Store;
 using Libplanet.Store.Trie;
 using Libplanet.Tests.Common.Action;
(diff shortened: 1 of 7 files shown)


Libplanet.Net.Tests/TestUtils.cs
```diff
@@ -85,5 +95,84 @@ public static PrivateKey GeneratePrivateKeyOfBucketIndex(Address tableAddress, i
...
+                TestUtils.Policy,
+                new VolatileStagePolicy<DumbAction>(),
+                store,
+                new TrieStateStore(new MemoryKeyValueStore()),
+                fx.GenesisBlock);
+
+            return blockChain;
+        }
+
+        public static (ITransport, ConsensusContext<DumbAction>, PrivateKey)
```
limebell, line 112: `PrivateKey` is argument of the function and is returned. And also for the transport, how about pass it to the argument and make creator method for `ITransport` in `TestUtils.cs`?

Libplanet.Net.Tests/TestUtils.cs
```diff
@@ -85,5 +95,90 @@ public static PrivateKey GeneratePrivateKeyOfBucketIndex(Address tableAddress, i
...
+        public static (ITransport, ConsensusContext<DumbAction>)
+            CreateStandaloneConsensusContext(
+            BlockChain<DumbAction> blockChain,
+            TimeSpan newHeightDelay,
+            long nodeId = 0,
+            string host = "localhost",
+            int port = 18192,
+            PrivateKey? privateKey = null,
+            List<PublicKey>? validators = null)
+        {
```
limebell, line 134:
```suggestion
        public static ConsensusContext<DumbAction>
            CreateStandaloneConsensusContext(
            BlockChain<DumbAction> blockChain,
            TimeSpan newHeightDelay,
            ITransport transport,
            long nodeId = 0,
            PrivateKey? privateKey = null,
            List<PublicKey>? validators = null)
        {
```
I guess if we need to use `ITransport` both inside and outside, it looks better to create from outside using `CreateNetMQTransport()` and handle it to this method's argument.

colibrishin: Oops, I see. I'll fix it with next push.

colibrishin:
Timeout Test is half-way done.
- Propose (Done)
- Vote
- PreCommit

Round-skipping (validRound condition) is not done yet. Also, it needs to be checked whether message is correctly broadcasting.

limebell (approved):

PR #2086, opened 2022-06-16
`StaticPeers` are now `ConsensusPeers`

riemannulus:
## Context
1. Recent `libplanet` development ecosystem, we don't use `StaticPeers` nowadays.
2. Instead, PBFT needs its own `PeerTable` and `ITransport` for consensus validator.

## Rationale
- Remove `StaticPeers` related logic. 
- Rename `StaticPeers` to `ConsensusPeers`

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusReactor.cs
+++ b/Libplanet.Net/Consensus/ConsensusReactor.cs
@@ -19,7 +19,7 @@ namespace Libplanet.Net.Consensus
     {
         private ITransport _consensusTransport;
         private ILogger _logger;
-        private IImmutableSet<BoundPeer>? _validatorPeers;
+        private IImmutableSet<BoundPeer> _validatorPeers;
         private List<PublicKey>? _validators;
         private ConsensusContext<T> _consensusContext;
         private BlockChain<T> _blockChain;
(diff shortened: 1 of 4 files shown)


Libplanet.Net/Consensus/ConsensusReactor.cs
```diff
@@ -37,7 +37,8 @@ public class ConsensusReactor<T> : IReactor
             _consensusTransport = consensusTransport;
             _validators = validators;
             _validatorPeers = validatorPeers?
-                .Where(peer => !peer.PublicKey.Equals(privateKey.PublicKey)).ToImmutableHashSet();
+                .Where(peer => !peer.PublicKey.Equals(privateKey.PublicKey)).ToImmutableHashSet()
```
limebell, line 40: Now, `BroadcastMessage` sends messages to self anyway, instead of  excluding self from here, how about remove the code that adds self  in `BroadcastMessage` [#L107](https://github.com/planetarium/libplanet/pull/2086/files#diff-be45a2ea18eeeb43a9ddca740fd2581952b2b148cbc5b79856771caef762de98R107).

riemannulus: 23f3d69de84051e2abaa537c64ff052941ee6df8 amended!

limebell (approved):

Issue #1696, opened 2022-01-05
`new PrivateKey()` returns a key with shorter length than expected 


longfin:
Sometimes re-encoding it fails because the `new PrivateKey()` constructor returns a private key that is less than 32 bytes in length.

```csharp
var key = new PrivateKey();
var encoded = key.ToByteArray();
var decoded = new PrivateKey(encoded);  // throws ArgumentOutOfRangeException
```

Related to https://github.com/planetarium/libplanet/pull/1571

zeta6:
✋

longfin:
@zeta6 

- [`PrivateKey()` relies `ECPrivateKeyParameters` on `GenerateKeyParam()` private method][1].
- [The length of `PrivateKey` be determined by `KeyParam.D`][2].
- `KeyParam.D` is `Org.BouncyCastle.Math.BigInteger`, and it has `.ToByteArrayUnsigned()` method and `.BitLength` property. I think we can check it before return on `GenerateKeyParam()` method.

[1]: https://github.com/planetarium/libplanet/blob/2086fb9250884b4b12ff7a2af740166d78a175df/Libplanet/Crypto/PrivateKey.cs#L367
[2]: https://github.com/planetarium/libplanet/blob/2086fb9250884b4b12ff7a2af740166d78a175df/Libplanet/Crypto/PrivateKey.cs#L345

longfin:
Related workaround: https://github.com/planetarium/9c-launcher/pull/1555/files

longfin:
It seemed that resolved by #2091

PR #2091, opened 2022-06-18
fix: issue #1696

zeta6:
Modify GenerateKeyParam to return a new value
when .D.ToByteArrayUnsigned().Length != 32 Repeat until == 32


limebell:
Seems like newly added test `Libplanet.Tests.Crypto.PrivateKeyTest.PrivateKeyShorterThan32Byte` is failing. 😢

limebell (approved):

PR #2100, opened 2022-06-21
Narrow down PBFT unit tests to its methods

colibrishin:
# Summary 
This is part 1 of https://github.com/planetarium/libplanet/pull/2099, which includes Round-skip unit tests.

+ Narrow down unit tests to its methods.


Libplanet.Net.Tests/Consensus/ContextTest.cs
```diff
@@ -867,6 +869,603 @@ public async void PreCommitBlockToEndCommit()
...
+                0,
+                blockChain.Tip.Index + 1,
+                privateKeys[3],
+                validators);
+            var block = await blockChain.MineBlock(privateKeys[3], append: false);
+            var timeoutOccurred = new AsyncManualResetEvent();
+            context.TimeoutOccurred += (sender, tuple) => timeoutOccurred.Set();
+
+            _ = transport.StartAsync();
+            await transport.WaitForRunningAsync();
```
riemannulus, line 909: it might be called` ITransport.StartAsync()` twice in one instance, Is it intended?

Libplanet.Net.Tests/Consensus/ContextTest.cs
```diff
@@ -867,6 +869,603 @@ public async void PreCommitBlockToEndCommit()
...
+        {
+            var (validators, privateKeys) = GetRandomValidators();
+
+            using ITransport transport =
+                TestUtils.CreateNetMQTransport(privateKeys[3], port: Port + 13);
+            var codec = new Codec();
+            BlockChain<DumbAction> blockChain =
+                TestUtils.CreateDummyBlockChain((MemoryStoreFixture)_fx);
+
+            var consensusContext =
```
riemannulus, line 981:
```suggestion
            ConsensusContext<DumbAction> consensusContext =
```

Libplanet.Net.Tests/Consensus/ContextTest.cs
```diff
@@ -867,6 +869,603 @@ public async void PreCommitBlockToEndCommit()
...
+                TestUtils.CreateStandaloneConsensusContext(
+                    blockChain,
+                    transport,
+                    newHeightDelay,
+                    3,
+                    blockChain.Tip.Index + 1,
+                    port: Port + 14,
+                    privateKey: privateKeys[3],
+                    validators: validators,
+                    watchConsensusMessage: IsPreCommitSent);
```
riemannulus, line 1096: IMO this is repeated in another test case. How about make this logic into a new method?

Libplanet.Net.Tests/Consensus/ContextTest.cs
colibrishin: I'm sorry but, I don't understand when will `transport` called twice. I had to call `StartAsync()` without async and await `WaitForRunningAsync()` instead because awaiting `StartAsync()` would block the test. 😅

colibrishin: Oh.. I see, my mistake.

limebell (approved):

PR #2106, opened 2022-06-22
Context process messages in single task

limebell:
## Rationale
When more than two messages processed simultaneously, an if statement can be `true` but it actually should be false due to timing issue of the context change.


riemannulus:
Could you rebase this PR?

colibrishin (approved):

riemannulus (commented):
LGTM

riemannulus (approved):

Issue #2107, opened 2022-06-22
Implement `planet key derive`


Akamig:
Currently `planet key derive` command is not implemented. It would be useful if we can derive pubkey and address from private key input.

https://github.com/Akamig/libplanet/blob/3f29c38e5433db74667423bee7ebf89884df702c/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs#L13-L14

Issue #2017, opened 2022-05-28
Ask storing path for keystore


longfin:
[`key create` command](https://github.com/planetarium/libplanet/blob/2086fb9250884b4b12ff7a2af740166d78a175df/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs#L32) assumes that storing keystore file to default path. it may would be better if we provide the chance to change it while creating. (like ssh-keygen)

longfin:
But on the other hand, it may harms consistency about sub-commands in `key`... 🤔

Akamig:
I think we can provide path changing and not harm consistency by introducing the `--path` option generally, but not mandatorily. (it's currently only used when exporting Web3 Secret Storage JSON). 
also introducing console message such as 
```
Key Created. 
Adding to default path.

Key Created.
Exporting to provided path (or print provided path here)
(and fallback to default path if fails)
```
could help users without breaking change.

On the other hand, for the long term, I think Create command should be merged with Generate command. Is there a specific reason these two commands are separated?

PR #2108, opened 2022-06-23
Add --path flag in `key` command, and implemented `key derive`

Akamig:
this addresses #2017 and  #2107.

- `key derive [private key]` introduced.
- now most of key command accepts --path or --store-path flag to specify the KeyStore path.
- because of that, key export --path behaves differently now, not exporting file to that path, but use that path as keystore to import 'from'
- key export behavior changed 
    - BEFORE : with --json and --path flag, JSON file has been exported to specified path.
    - AFTER  : always outputs as stdout, regardless of json or raw hex key.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,7 +28,7 @@ To be released.
     [2.0.\*][Cocona.Lite 2.0.0].  [[#2101]]
 
  - Implemented *planet key derive*, now you can get public key and
-   address from private key directly! [[#2108]]
+   address from private key directly!  [[#2108]]
 
 [#2101]: https://github.com/planetarium/libplanet/pull/2101
 [#2108]: https://github.com/planetarium/libplanet/pull/2108
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,7 +27,11 @@ To be released.
  - (Libplanet.Extensions.Cocona) Upgraded *Cocona.Lite* from 1.6.\* to
     [2.0.\*][Cocona.Lite 2.0.0].  [[#2101]]
 
+ - Implemented *planet key derive*, now you can get public key and
+   address from private key directly! [[#2108]]
+
 [#2101]: https://github.com/planetarium/libplanet/pull/2101
+[#2108]: https://github.com/planetarium/libplanet/pull/2108
 [Cocona.Lite 2.0.0]: https://www.nuget.org/packages/Cocona.Lite/2.0.0
 
 
--- a/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
+++ b/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
@@ -37,7 +37,7 @@ namespace Libplanet.Extensions.Cocona.Commands
         public void Create(
             PassphraseParameters passphrase,
             [Option(
-                Description = "print created private key as Web3 Secret Storage format."
+                Description = "Print created private key as Web3 Secret Storage format."
             )]
             bool json = false,
             [Option(Description = "Do not add to the key store, but only show the created key.")]
--- a/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
+++ b/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
@@ -66,7 +66,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             PassphraseParameters passphrase,
             [Option(Description = "Remove without asking passphrase.")]
             bool noPassphrase = false,
-            [Option(Description = "Path to key store")]
+            [Option(Description = "Path to key store.")]
             string? path = null
         )
         {
--- a/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
+++ b/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
@@ -25,11 +25,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             )]
             string? path = null)
         {
-            if (path != null)
-            {
-                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
-            }
-
+            ChangeKeyStorePath(path);
             PrintKeys(KeyStore.List().Select(t => t.ToValueTuple()));
         }
 
@@ -46,11 +42,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             string? path = null
         )
         {
-            if (path != null)
-            {
-                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
-            }
-
+            ChangeKeyStorePath(path);
             string passphraseValue = passphrase.Take("Passphrase: ", "Retype passphrase: ");
             PrivateKey pkey = new PrivateKey();
             ProtectedPrivateKey ppk = ProtectedPrivateKey.Protect(pkey, passphraseValue);
@@ -78,13 +70,9 @@ namespace Libplanet.Extensions.Cocona.Commands
             string? path = null
         )
         {
+            ChangeKeyStorePath(path);
             try
             {
-                if (path != null)
-                {
-                    KeyStore = new Web3KeyStore(Path.GetFullPath(path));
-                }
-
                 if (!noPassphrase)
                 {
                     UnprotectKey(keyId, passphrase);
@@ -117,11 +105,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             string? path = null
         )
         {
-            if (path != null)
-            {
-                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
-            }
-
+            ChangeKeyStorePath(path);
             if (json)
             {
                 try
@@ -162,11 +146,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             string? path = null
         )
         {
-            if (path != null)
-            {
-                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
-            }
-
+            ChangeKeyStorePath(path);
             PrivateKey key = UnprotectKey(keyId, passphrase);
             byte[] rawKey = publicKey ? key.PublicKey.Format(true) : key.ToByteArray();
             using Stream stdout = Console.OpenStandardOutput();
@@ -244,11 +224,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             string? storePath = null
         )
         {
-            if (storePath != null)
-            {
-                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
-            }
-
+            ChangeKeyStorePath(storePath);
             PrivateKey key = UnprotectKey(keyId, passphrase);
 
             byte[] message;
@@ -290,7 +266,7 @@ namespace Libplanet.Extensions.Cocona.Commands
                 "PRIVATE-KEY",
                 Description = "A raw private key to import."
             )]
-            string key,
+            string key
         )
         {
             PrivateKey privateKey = ValidateRawHex(key);
@@ -378,6 +354,18 @@ namespace Libplanet.Extensions.Cocona.Commands
             {
                 Utils.Error("This file does not exist.");
                 return string.Empty;
+            }
+        }
+
+        private void ChangeKeyStorePath(string? path)
+        {
+            if (path != null)
+            {
+                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
+            }
+            else
+            {
+                KeyStore = Web3KeyStore.DefaultKeyStore;
             }
         }
     }
--- a/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
+++ b/Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
@@ -101,7 +101,7 @@ namespace Libplanet.Extensions.Cocona.Commands
             bool json = false,
             [Option(Description = "Do not add to the key store, but only show the created key.")]
             bool dryRun = false,
-            [Option(Description = "Path to key store")]
+            [Option(Description = "Path to key store.")]
             string? path = null
         )
         {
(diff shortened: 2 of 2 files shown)


Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -66,23 +98,50 @@ public KeyCommand()
...
         }
 
-        [Command(Description = "Import a raw private key.")]
+        [Command(Description = "Import a raw private key or Web3 Secret Storage.")]
         public void Import(
             [Argument(
                 "PRIVATE-KEY",
-                Description = "A raw private key to import in hexadecimal string."
+                Description = @"A raw private key in hexadecimal string, or path to Web3 Secret
+                Storage to import"
```
moreal, line 106: In C# language, the `@` keyword makes us able to use all strings until the end character `"` as raw. For examples:

```csharp
Console.WriteLine(@"Hello
World");
```

The above code will print:

```
Hello
World
```

And in the indented case, it will print all of them without modification.

```csharp
Console.WriteLine(@"Hello
       World");
```

```
Hello
       World
```

So I saw the help message like the below text and I thought it should be checked once.

```
Usage: dotnet key import [--passphrase <PASSPHRASE>] [--passphrase-file <FILE>] [--json] [--dry-run] [--path <String>] [--help] private-key

Import a raw private key or Web3 Secret Storage.

Arguments:
  0: private-key    A raw private key in hexadecimal string, or path to Web3 Secret
                Storage to import (Required)

Options:
  -p, --passphrase <PASSPHRASE>    Take passphrase through this option instead of prompt.  Mutually exclusive with --passphrase-file option.
  --passphrase-file <FILE>         Read passphrase from the specified file instead of taking it through prompt.  Mutually exclusive with -p/--passphrase option.  For standard input, use a hyphen (`-').  For an actual file named a hyphen, prepend `./', i.e., `./-'.  Note that the trailing CR/LF is trimmed.
  --json                           Import Web3 Secret Storage key.
  --dry-run                        Do not add to the key store, but only show the created key.
  --path <String>                  Path to key store
  -h, --help                       Show help message
```

If you didn't want the newline and indents and you used it to avoid line length limitation, how about using `+` operator to split string? like...
```suggestion
                Description = "A raw private key in hexadecimal string, or path to Web3 Secret " +
                              "Storage to import"
```

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -97,26 +156,29 @@ public KeyCommand()
...
             }
             else if (json)
             {
-                Stream fs = PathHandler(path);
                 var ppk = KeyStore.Get(keyId);
-                ppk.WriteJson(fs, keyId);
-                fs.WriteByte(0x0a);
-                fs.Close();
+                ppk.WriteJson(stdout, keyId);
+                stdout.WriteByte(0x0a);
```
moreal, line 181: Maybe this seems like a break-change. However, it may be okay on the Libplanet side 🤔 I don't remember whether the Libplanet has a duty to guarantee backward compatibility or not well.

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -177,9 +239,16 @@ public KeyCommand()
...
-            string? binaryOutput = null
+            string? binaryOutput = null,
+            [Option(Description = "Path to key store to use key from.")]
+            string? storePath = ""
         )
         {
+            if (storePath != null)
+            {
+                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
+            }
```
moreal, line 250: Though it introduced a new argument `string?` typed `storePath`, the default value seems like `""`. Maybe it will be always not `null`. What do you think about it?

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -215,6 +284,25 @@ public KeyCommand()
...
+        )
+        {
+            PrivateKey privateKey = ValidateRawHex(key);
+            string passphraseValue = passphrase.Take("Passphrase: ");
+            string addr = privateKey.ToAddress().ToString();
+            string pub = ByteUtil.Hex(privateKey.PublicKey.Format(compress: true));
+            ProtectedPrivateKey ppk = ProtectedPrivateKey.Protect(
+                privateKey, passphraseValue);
+            Utils.PrintTable(("Public Key", "Address"), new[] { (pub, addr) });
+        }
```
moreal, line 304: If I understood well, the `passphrase` variable and the `ppk` variable seem not useless in the `derive` command. Because it doesn't print the protected private key or store it. What do you think about it?

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -323,16 +371,17 @@ private PrivateKey ValidateRawHex(string rawKeyHex)
...
-                return System.Console.OpenStandardOutput();
+                string json = new StreamReader(jsonPath).ReadToEnd();
+                return json;
             }
-            else
+            catch (Exception)
             {
-                string path = Path.GetFullPath(pathString);
-                return File.Open(path, FileMode.OpenOrCreate);
+                Utils.Error("This is not valid json file or file does not exists.");
```
moreal, line 383: The [`System.IO.StreamReader`][stream-reader] doesn't validate whether the content is compatible with JSON format or not. The original code had used `ProtectedPrivateKey.FromJson` so the error message was right.
So this error message should be corrected like...

```suggestion
                Utils.Error("This file does not exist.");
```

[stream-reader]: https://docs.microsoft.com/ko-kr/dotnet/api/system.io.streamreader?view=net-6.0

moreal (commented):
I'll review furthermore after the reviews became resolved well. Because I worried that it can be seen as too big review 😓

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
Akamig: fixed as `string? storePath = null`.

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
Akamig: Yep, totally unused. removed unused ppk and passphrase.

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -66,23 +98,50 @@ public KeyCommand()
...
+            bool dryRun = false,
+            [Option(Description = "Path to key store")]
+            string? path = null
         )
         {
-            Add(key, passphrase, json, dryRun, key, false);
+            if (path != null)
+            {
+                KeyStore = new Web3KeyStore(Path.GetFullPath(path));
+            }
```
moreal, line 123: The `--path` option and the logic to override `KeyStore` seem used in almost all commands. How about abstracting it with `ICommandParameterSet` like `PassphraseParameters`?

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -215,6 +284,21 @@ public KeyCommand()
             }
         }
 
+        [Command(Description = "Derive Public key and Address from Private key.")]
+        public void Derive(
+            [Argument(
+                "PRIVATE-KEY",
+                Description = "A raw private key to import."
+            )]
+            string key,
```
moreal, line 293: Maybe the trailing comma should be removed.

```suggestion
            string key
```

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -4,55 +4,87 @@
...
 
-        public IKeyStore KeyStore { get; }
+        public IKeyStore KeyStore { get; set; }
 
         [PrimaryCommand]
         [Command(Description = "List all private keys.")]
-        public void List() =>
+        public void List(
+            [Option(
+                Description = "Specify KeyStore path to list."
```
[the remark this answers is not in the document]
Akamig: No, I think nouns in descriptions should be strictly consistent through out whole command, and I strongly agree on unifying them.

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
Akamig: In my perspective, abstracting as `ICommandParameterSet` may harm the consistency of behaviour DefaultStore being as a singleton, and bit complicated for simple overriding, but It's also true that codes are duplicated, so I just made the `ChangeKeyStorePath(path)` function and put the logic inside of it. this way we can ignore both scope issue of using interface and duplicated code.

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
Akamig: applied with the latter commit.

Akamig:
Ah, and I forgot to emphasize, thanks for your time and consideration put in an in-depth review, @moreal!

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -4,53 +4,73 @@
...
-using Libplanet.Extensions.Cocona.Commands.Key;
 using Libplanet.KeyStore;
 
 namespace Libplanet.Extensions.Cocona.Commands
 {
-    [HasSubCommands(
-        typeof(DerivationCommand),
-        "derive",
-        Description = "Derive the address or public key from private key."
-    )]
```
moreal, line 16: If you deprecated the `planet key derive` command and introduce a new command, should `DerivationCommand` be also removed?

CHANGES.md
```diff
@@ -27,7 +27,11 @@ To be released.
  - (Libplanet.Extensions.Cocona) Upgraded *Cocona.Lite* from 1.6.\* to
     [2.0.\*][Cocona.Lite 2.0.0].  [[#2101]]
 
+ - Implemented *planet key derive*, now you can get public key and
+   address from private key directly! [[#2108]]
```
moreal, line 31:
```suggestion
 - Implemented *planet key derive*, now you can get public key and
   address from private key directly!  [[#2108]]
```

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -4,53 +4,73 @@
...
 
         [Command(Aliases = new[] { "rm" }, Description = "Remove a private key.")]
         public void Remove(
             [Argument(Name = "KEY-ID", Description = "A key UUID to remove.")]
             Guid keyId,
             PassphraseParameters passphrase,
             [Option(Description = "Remove without asking passphrase.")]
-            bool noPassphrase = false
+            bool noPassphrase = false,
+            [Option(Description = "Path to key store")]
```
moreal, line 69:
```suggestion
            [Option(Description = "Path to key store.")]
```

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -323,16 +343,29 @@ private PrivateKey ValidateRawHex(string rawKeyHex)
...
+            try
+            {
+                string json = new StreamReader(jsonPath).ReadToEnd();
+                return json;
+            }
+            catch (Exception)
+            {
+                Utils.Error("This file does not exist.");
+                return string.Empty;
+            }
```
moreal, line 357: If there is no file, it throws `System.IO.FIleNotFoundException`. And `Utils.Error` return `CommandExitedException`
```suggestion
            catch (System.IO.FileNotFoundException)
            {
                throw Utils.Error("This file does not exist.");
            }
```

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
```diff
@@ -66,23 +86,46 @@ public KeyCommand()
...
+            {
+                try
+                {
+                    ProtectedPrivateKey ppk = ProtectedPrivateKey.FromJson(ValidateJsonPath(key));
+                    PrintKeys(new[] { (Add(ppk, dryRun), ppk) });
+                }
+                catch (Exception)
+                {
+                    Utils.Error("Couldn't load ppk from json.");
+                }
```
moreal, line 119: It should pass if the `CommandExitedException` was thrown. And it should throw `CommandExitedException`  for the exit code. If it doesn't throw, it'll return `0` like successful execution.

```suggestion
                catch (CommandExitedException)
                {
                    throw;
                }
                catch (Exception)
                {
                    throw Utils.Error("Couldn't load ppk from json.");
                }
```

moreal (commented):
I realized the `planet key derive` doesn't work though I implemented in #1268, because there was no `[Command]` attribute 😅

moreal:
> * `key derive [public key]` introduced.

It seems incorrect because the implementation seems like `key derive [private key]`.

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
Akamig: Wh.... just wow. I just couldn't find the DerivationCommand file from commands folder, so I was like "Well, it says not implemented." Hmm... what do you think? I think in consiceness your implementation is way better, but for simplicity, mine seems better. what's better between yours and mine?

Akamig:
> > * `key derive [public key]` introduced.
> 
> It seems incorrect because the implementation seems like `key derive [private key]`.

~~I think I was totally drunk at this point~~

Libplanet.Extensions.Cocona/Commands/KeyCommand.cs
moreal: I think your implementation seems more simple to use. I think yours is better.

Issue #2105, opened 2022-06-22
Cache validation


limebell:
Validating blocks every time a new message is received is redundant. Should validate only once when propose is received, or cache the result of the validation.

PR #2111, opened 2022-06-24
feat: add valid block hash cache

OnedgeLee:
## Changes
- Add `_validBlockHashCache` field on `Context<T>`
- Add `IsCached` method on `Context<T>`
- Add caching logic on `Context<T>.IsValid`

## Related issue
- Resolves https://github.com/planetarium/libplanet/issues/2105

### DIFF
--- a/Libplanet.Net/Consensus/Context.cs
+++ b/Libplanet.Net/Consensus/Context.cs
@@ -8,6 +8,7 @@ using System.Threading.Channels;
 using System.Threading.Tasks;
 using Bencodex;
 using Bencodex.Types;
+using Caching;
 using Libplanet.Action;
 using Libplanet.Blockchain;
 using Libplanet.Blocks;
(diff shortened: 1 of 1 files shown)


Libplanet.Net/Consensus/Context.cs
```diff
@@ -50,6 +50,7 @@ public class Context<T> : IDisposable
         private Block<T>? _validValue;
         private int _validRound;
         private int _requestCount;
+        private BlockHash? _validBlockHashCache;
```
limebell, line 53: How about cache result of the `IsValid`?
```suggestion
        private (BlockHash, bool)? _validBlockHashCache;
```

OnedgeLee: @limebell 
May be I've comitted some mistakes.
We have to cache multiple blockhashes, not just one, right?
If so, I think I have to maintain some `Dictionary<BlockHash, bool>`.
Am I got the point?

OnedgeLee: @limebell I don't know why but diff has been so large, I'll reopen the PR.

PR #2112, opened 2022-06-24
Fill `Block<T>.LastCommit` by getting voteset from last block's consensus context

limebell:
By storing `Vote`s from previous round, chain can be validated.


riemannulus (approved):
LGTM

Issue #2058, opened 2022-06-07
Account for multiple `:`s in `Uri` for `IceServer`


greymistcube:
As seen in [this comment](https://github.com/planetarium/libplanet/blob/eb9f478ca8cfb02b77bdee1f5b693f71334ac0d8/Libplanet.Net/IceServer.cs#L17-L35), if `Uri.UserInfo` contains more than one `:`, it should throw an `Exception` instead of parsing the `string` improperly.

PR #2116, opened 2022-06-27
fix: Prevent improper parsing on 'Uri.UserInfo' with exception

OnedgeLee:
## Changes
- Replace improper handling on invalid url with throwing exception

## Related issue
- Resolves https://github.com/planetarium/libplanet/issues/2058

### DIFF
--- a/Libplanet.Net/IceServer.cs
+++ b/Libplanet.Net/IceServer.cs
@@ -16,21 +16,24 @@ namespace Libplanet.Net
 
         public IceServer(Uri url)
         {
-            // FIXME: This doesn't account for multiple colons.
 #pragma warning disable S1121, S3358, SA1316
             Url = url;
             Func<string, string, bool, char, (string, string, bool)> parser =
-                (username, credential, isUsernameChar, c) =>
+                (username, credential, colonFound, c) =>
                     c == ':'
-                        ? (username, credential, !isUsernameChar)
-                        : isUsernameChar
-                            ? (username += c, credential, isUsernameChar)
-                            : (username, credential += c, isUsernameChar);
-            (string username, string credential, bool isUsernameChar) seed =
+                        ? colonFound
+                            ? (username, credential, !colonFound)
+                            : throw new ArgumentException(
+                                $"Uri.UserInfo '{url.UserInfo}' have to contain single colon " +
+                                "as a delimiter between username and credential.")
+                        : colonFound
+                            ? (username += c, credential, colonFound)
+                            : (username, credential += c, colonFound);
+            (string username, string credential, bool colonFound) seed =
                 (string.Empty, string.Empty, true);
             (Username, Credential, _) = url.UserInfo
-                .Aggregate(seed, (prev, c) =>
-                    parser(prev.username, prev.credential, prev.isUsernameChar, c));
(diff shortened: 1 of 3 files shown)


OnedgeLee:
[First CI Test](https://app.circleci.com/pipelines/github/planetarium/libplanet/2240/workflows/44115420-7f1a-44d2-9b64-653ef73c0bbe/jobs/13634) on `ci/circleci: linux-netcore-test-fr-FR` was failed, so I've [reran the test](https://app.circleci.com/pipelines/github/planetarium/libplanet/2240/workflows/98d98b51-59b7-432f-92c6-bde905596bd2/jobs/13637).
I couldn't figure out the reason why the first test was failed, but in my opinion, it's not related with this PR.

OnedgeLee:
Do I have to update `CHANGES.md`?

OnedgeLee:
I'll try to track [test failure case](https://app.circleci.com/pipelines/github/planetarium/libplanet/2240/workflows/44115420-7f1a-44d2-9b64-653ef73c0bbe/jobs/13634), and raise an issue if it's needed

greymistcube:
https://github.com/planetarium/libplanet/blob/6146ed48f427d08bbc80eb9eef908c29839b470f/Libplanet.Net/IceServer.cs#L19

This comment should be removed. 😗

greymistcube:
> Do I have to update `CHANGES.md`?

Should be added under `Bug fixes`.

Libplanet.Net/IceServer.cs
```diff
@@ -22,7 +22,11 @@ public IceServer(Uri url)
             Func<string, string, bool, char, (string, string, bool)> parser =
                 (username, credential, isUsernameChar, c) =>
                     c == ':'
-                        ? (username, credential, !isUsernameChar)
+                        ? isUsernameChar
+                            ? (username, credential, !isUsernameChar)
+                            : throw new ArgumentException(
+                                $"Uri.UserInfo '{url.UserInfo}' have to contain single colon " +
+                                "as a delimiter between username and credential.")
```
greymistcube, line 29: Using something like `int colonCount` instead of `bool isUsernameChar` might make this more readable. 🙂

OnedgeLee: @greymistcube Updated PR according to your comments :)

greymistcube: Actually, you can remove `isUsernameChar` argument. It's redundant information with `colonCount`. 😗

OnedgeLee: @greymistcube I think it's better to use `bool isBeforeColon` instead of `int colonCount`, since it will have only 0~1 range, any opinion for this?

greymistcube: Either only using `bool colonCounted` or `int colonCount` is both fine I guess. Now that I look at it, perhaps `isUsernameChar` was a bad name. 😐

OnedgeLee:
@greymistcube
1. Use single variable `bool colonFound` to check multiple colons and split `username` and `credential`
2. Add changelog to specify `where` in the API has "changed" 😄

OnedgeLee:
Rebased onto `main` branch

greymistcube (approved):

limebell (approved):

PR #2117, opened 2022-06-27
Consensus option cleanup

riemannulus:
## Rationale
The consensus option is hard-wired to `Swarm<T>`. It causes non-validator nodes cannot to join the network.

## Feature
- Cleanup consensus option
- Cleanup transport parameter


Libplanet.Net/Transports/InitTransport.cs
```diff
@@ -0,0 +1,89 @@
...
+                param.TrustedAppProtocolVersionSigners,
+                param.Workers ?? 50,
+                param.Host,
+                param.ListenPort,
+                param.IceServers ?? new IceServer[0],
+                param.DifferentAppProtocolVersionEncountered,
+                param.MessageTimestampBuffer);
+        }
+
+        public static TcpTransport Tcp(
```
limebell, line 48: Name seems general and used in this class only. How about make it private?

Libplanet.Net/SwarmOptions.cs
```diff
@@ -118,7 +93,8 @@ public enum TransportType : byte
         /// <summary>
         /// The type of <see cref="ITransport"/> used in <see cref="Swarm{T}"/>.
         /// </summary>
-        public TransportType Type { get; set; } = TransportType.TcpTransport;
+        public InitTransport.TransportType Type { get; set; }
```
limebell, line 96: I think this change makes user import `InitTransport` class when creating `SwarmOptions` with option. Is it intended?

colibrishin (approved):

limebell (approved):

Issue #2121, opened 2022-06-28
`SwarmTest.BroadcastBlockWithSkip` is flaky and sometimes fails


OnedgeLee:
I've met `Assert.Equal()` Failure two times([1](https://app.circleci.com/pipelines/github/planetarium/libplanet/2240/workflows/44115420-7f1a-44d2-9b64-653ef73c0bbe/jobs/13634), [2](https://app.circleci.com/pipelines/github/planetarium/libplanet/2254/workflows/a403ce9b-3f97-4d91-88fc-e6d454f2f395/jobs/13816)) on `SwarmTest.BroadcastBlockWithSkip` during CI test on [PR#2116](https://github.com/planetarium/libplanet/pull/2116), and suspected current `main` branch have same problem.
I've wrapped this test with for loop, enlarged timeout threshold, and could reproduce same result(failure) on `main` branch.

In this test, 4 `DumbAction`s are recorded on `recieverChain`, and rendering occurs 4 times
(during block appending on `recieverChain`, by `AppendPreviousBlocks` method, called by Task `Swarm.ConsumeBlockCandidates`), 
so variable `renderCount` have to be 4 at the end of the test, but 2 and 3 were found during tests.

I suspect, `RecordingActionRenderer.RenderAction` called less than 4 times, 
even though `LoggedActionRenderer.RenderAction` called 4 times. 
(I've checked the log from failed test, and there were 4 lines of `Invoked {MethodName}() for an action {ActionType} at block #{BlockIndex}`)

OnedgeLee:
Certain and Uncertain

Certain
- `recieverChain` has 3 `Blocks`
  - evidence
    - Test code waits for tip of `recieverChain` to be equal to the last block appended to `blockChain`.
    - I've never met the assertion on `Assert.Equal(3, recieverChain.Count)` during whole test.
Uncertain
- `recieverChain` has 4 `DumbActions`
  - expected evidence : I've checked every `DumbAction` in `blockChain` and `recieverChain`, when encountered an error.
  - restricted : After adding log on `DumbAction`s on `recieverChain`, couldn't reproduce problem yet.

OnedgeLee:
[...]

PR #2125, opened 2022-06-29
fix: fix flaky test `SwarmTest.BroadcastBlockWithSkip`

OnedgeLee:
## Changes
- Enlarge time interval between blockchain tip check and `renderCount` assertion check

## Background
- `SwarmTest.BroadcastBlockWithSkip` sometimes fails, and it's due to early checking of `renderCount`
- `renderAction` occurs after blockchain tip change, and interval `100` between blockchain tip check and `renderCount` assertion could be so small that makes test flaky
- Enlarging time interval to `1000` between blockchain tip check and `renderCount` assertion check won't make this test 100% deterministic, but `almost` deterministic
- There are some ways to make this test deterministic, but those will contain some changes on functionality on `libplanet`, and those changes seems to not mendatory for main purpose of `libplanet`
- Before fix, almost every time, within 7 minutes, test failure has been occured, when wrapped this test with loop
- After fix, no failure has been found on this test yet

## Related issue
- https://github.com/planetarium/libplanet/issues/2121

### DIFF
--- a/Libplanet.Net.Tests/SwarmTest.Broadcast.cs
+++ b/Libplanet.Net.Tests/SwarmTest.Broadcast.cs
@@ -696,7 +696,8 @@ namespace Libplanet.Net.Tests
 
                 await AssertThatEventually(
                     () => receiverChain.Tip.Equals(block2),
-                    5_000
+                    5_000,
+                    1_000
                 );
                 Assert.Equal(3, receiverChain.Count);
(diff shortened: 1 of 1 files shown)


limebell (dismissed):

limebell:
Please rebase on our main branch

limebell (approved):

PR #2130, opened 2022-06-30
Add GQL mutations & queries to Libplanet.Explorer

longfin:
This PR introduces new GQL APIs to provide transaction staging and siging features. (see also CHANGES.md 🙏 )

I know that `explorer` isn't suitable place to these features so we may need to rename `Libplanet.Explorer`. 🤔  (but I also believe seperating PR would be better.)


riemannulus (dismissed):
LGTM

limebell (approved):

Libplanet.Explorer/Mutations/TransactionMutation.cs
```diff
@@ -0,0 +1,39 @@
...
+        private readonly IBlockChainContext<T> _context;
+
+        public TransactionMutation(IBlockChainContext<T> context)
+        {
+            _context = context ?? throw new ArgumentNullException(nameof(context));
+
+            Field<TransactionType<T>>(
+                "stage",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<StringGraphType>> { Name = "payload" }
```
moreal, line 24: I think it'll be better to describe its format as `base64`.

Libplanet.Explorer/Mutations/TransactionMutation.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class TransactionMutation<T> : ObjectGraphType
+        where T : IAction, new()
+    {
+        private readonly IBlockChainContext<T> _context;
+
+        public TransactionMutation(IBlockChainContext<T> context)
+        {
+            _context = context ?? throw new ArgumentNullException(nameof(context));
+
+            Field<TransactionType<T>>(
```
moreal, line 21: If I understood the logic well, it could return a non-null transaction value. Can you let me know what part you worried about?

moreal, line 21: When I looked the [`BlockChain<T>.StageTransaction(...)`](https://github.com/planetarium/libplanet/blob/6e0b23883040d3acfe5f3679f146b3d10e6ef587/Libplanet/Blockchain/BlockChain.cs#L653-L667), it seems return boolean value. Or it should return the `tx` variable if you intended to return the transaction.

```suggestion
            Field<BooleanGraphType>(
```

Libplanet.Explorer/Queries/TransactionQuery.cs
```diff
@@ -107,6 +117,95 @@ public TransactionQuery()
...
+                            chain.Genesis.Hash,
+                            new[] { action }
+                        );
+                    return unsignedTransaction.Serialize(false);
+                }
+            );
+
+            Field<NonNullGraphType<StringGraphType>>(
+                name: "attachSignature",
+                deprecationReason: "Use signTransaction",
```
moreal, line 166: It seems this code is from NineChronicles.Headless but the explorer seems not to have the `signTransaction` GraphQL field, right?

moreal (commented):
When I had worked for the `planet key sign` feature, I learned about base64's variants and I applied it in also `NineChronicles.Headless` code. When you see the `signTransaction` field in `NineChronicles.Headless`, it also uses hexadecimal strings as inputs. So I want to suggest using hexadecimal strings as inputs.

https://github.com/planetarium/libplanet/pull/1920#discussion_r860391794

Libplanet.Explorer/Mutations/TransactionMutation.cs
longfin: `ArgumentNullException` isn't related to `Field<TransactionType<T>>` but is relate to `context` argument. 😅

longfin: Ah, thanks for pointing out. originally, I've intended returning `tx` but on second thought, returning `bool` is better.

longfin: Hmm, I'll change to returning `tx` and throwing exception when `BlockChain<T>.StageTransaction()` returns `false`.

Libplanet.Explorer/Queries/TransactionQuery.cs
longfin: Right, it would be better copying `signTransaction` instead of it haha.

longfin: But... well, `attachSignature` may be a better name than `signTranscation` because `signTransaction` doesn't perform sign, but perform attach the given signature. 🤔 

@moreal @planetarium/libplanet What do you think about it?

Akamig: I’m also +1 on `attatchSignature` for better explanation for following code, but if I had to name it I’d call it `bindSignature` for shorter naming.

longfin:
it seems that there was already [a discussion](https://github.com/planetarium/libplanet/pull/1920#discussion_r860391794) about action/trasnaction format from `planet` CLI. 

but I more prefer base64 (exactly, [RFC 4648 vanilla](https://datatracker.ietf.org/doc/html/rfc4648#section-4)) than hex as API's input because of the compression efficiency. [^1] In that case, I'll open another PR to add `--base64` option for usability.

@planetarium/libplanet @moreal 

[^1]: Of course, if we go further, it might be better to use multi-part instead of worrying about how to encode the binary 😢

longfin:
> it seems that there was already [a discussion](https://github.com/planetarium/libplanet/pull/1920#discussion_r860391794) about action/trasnaction format from `planet` CLI.
> 
> but I more prefer base64 (exactly, [RFC 4648 vanilla](https://datatracker.ietf.org/doc/html/rfc4648#section-4)) than hex as API's input because of the compression efficiency. [1](#user-content-fn-1-bd71cf4f516ae22661f5637c883b2cd9) In that case, I'll open another PR to add `--base64` option for usability.

After considering with #2132, compression efficiency isn't big deal for me too. thus, I'll convert base64 to hex in this PR. (This PR may become obsolete, but there are related tasks such as planet-node and chrono, so I will keep it open).

longfin:
Turned to draft due to [a temporary fixing](https://github.com/planetarium/libplanet/pull/2130/commits/9c5429956c902c68a7cecf4293440fc2bacfb0ea) that shoud be removed.

longfin:
This PR still contains [a temporary fix commit](https://github.com/planetarium/libplanet/pull/2130/commits/9c5429956c902c68a7cecf4293440fc2bacfb0ea). but I think that it's okay to merge because 

- [That issue](https://github.com/planetarium/libplanet/pull/2142) already [had been solved at main branch](https://github.com/planetarium/libplanet/pull/2142/commits/4f506dcea7f762ea07c19db99aa99e40bbf6e0e9).
- And, we'll merge this PR by squash.

PTAL 🙇  @planetarium/libplanet

CHANGES.md
```diff
@@ -19,6 +19,9 @@ To be released.
 ### Backward-incompatible storage format changes
 
 ### Added APIs
+ - (Libplanet.Explorer) Added `TransactionMutation<T>` class.  [[#2130]]
+ - (Libplanet.Explorer) Added `unsignedTransaction`, `bindSignature` and
+   `transactionResult` GraphQL fields to `TransactionQuery<T>`.  [[#2130]]
```
dahlia, line 24:
```suggestion
 -  (Libplanet.Explorer) Added `TransactionMutation<T>` class.  [[#2130]]
 -  (Libplanet.Explorer) Added `unsignedTransaction`, `bindSignature` and
    `transactionResult` GraphQL fields to `TransactionQuery<T>`.  [[#2130]]
```

Libplanet.Explorer/Queries/TransactionQuery.cs
```diff
@@ -107,6 +122,164 @@ public TransactionQuery()
...
+                            chain.Genesis.Hash,
+                            new[] { action }
+                        );
+                    return unsignedTransaction.Serialize(false);
+                }
+            );
+
+            Field<NonNullGraphType<StringGraphType>>(
+                name: "bindSignature",
+                #pragma warning disable MEN002
```
dahlia, line 171: How about suppressing it for the whole method?

Libplanet.Explorer/GraphTypes/TxResultType.cs
```diff
@@ -0,0 +1,28 @@
...
+                nameof(TxResult.BlockIndex),
+                description: "The block index which the target transaction executed.",
+                resolve: context => context.Source.BlockIndex
+            );
+
+            Field<StringGraphType>(
+                nameof(TxResult.BlockHash),
+                description: "The block hash which the target transaction executed.",
+                resolve: context => context.Source.BlockHash
+            );
```
moreal, line 25: In my thought, the `TxResult.ExceptionName` and the `TxResult.ExceptionMetadata` should be also exposed (Or in the next PR).

Libplanet.Explorer/Queries/TransactionQuery.cs
longfin: It may be a kind of choice between code verbosity and keeping warning areas...
In my point of view, the code verbosity of this sort can be bearable, but warning diabling isn't.

but on the other hands, it's totally personal opinion and increasing region isn't critical for even me. 😄 

@planetarium/libplanet any ideas?

Libplanet.Explorer/GraphTypes/TxResultType.cs
moreal: I think hexadecimal strings seems suit. Because there are many bencodex implementations and it can be handled with them in the client.

longfin: Added https://github.com/planetarium/libplanet/pull/2130/commits/cbc18e8fb9b5657f7bf505d355309f526a816978 to address [it](https://github.com/planetarium/libplanet/pull/2130#discussion_r936131723).

Libplanet.Explorer/GraphTypes/IValueType.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public override object? Serialize(object? value)
+        {
+            if (value is Bencodex.Types.IValue iv)
+            {
+                return _codec.Encode(iv);
+            }
+
+            return value;
```
moreal, line 24: How about throwing if unexpected types came?

You can see the `GuidGraphType` as an example.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/Scalars/GuidGraphType.cs

```suggestion
            return ThrowSerializationError(value);
```

Libplanet.Explorer/GraphTypes/IValueType.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public override object? ParseLiteral(IValue? value)
+        {
+            if (value is StringValue)
+            {
+                return ParseValue(value.Value);
+            }
+
+            return null;
```
moreal, line 47: How about throwing errors instead of `null` if unexpected types came?

You can see the `GuidGraphType` as an example.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/Scalars/GuidGraphType.cs

```suggestion
            return ThrowLiteralConversionError(value);
```

moreal (commented):
Using custom scalar types seems good to me. I left a review about it.

Libplanet.Explorer/GraphTypes/IValueType.cs
```diff
@@ -0,0 +1,50 @@
+using System;
+using Bencodex;
+using GraphQL.Language.AST;
+using GraphQL.Types;
+
+namespace Libplanet.Explorer.GraphTypes
+{
+    public class IValueType : StringGraphType
```
moreal, line 8: I think this type can inherit `ScalarGraphType`.

longfin: afik `StringGraphType` has been inherited from `ScalarGraphType`. did you mean that inheriting `ScalarGraphType` is better than `StringGraphType`? (Actually, I referenced `PublicKeyType` 😅 )

Libplanet.Explorer/GraphTypes/IValueType.cs
longfin: @moreal 

`ThrowSerializationError()` seems require that `value` isn't nullable. ([see also](https://github.com/graphql-dotnet/graphql-dotnet/blob/4.7.1/src/GraphQL/Types/Scalars/GuidGraphType.cs#L49-L51)).

I think that allowing null value can be fair enough because `Serialize()` has already opened the possibility. (on the other hands, we may need to care about other cases using `ThrowSerializationError()` 😓 )

Libplanet.Explorer/GraphTypes/IValueType.cs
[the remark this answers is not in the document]
longfin: Renamed https://github.com/planetarium/libplanet/pull/2130/commits/c2ca39c7730c810261973d178bde19932a2f9df9

longfin:
I realized that there are too many commits will be squashed. 😅 (I also want to avoid rebasing as much as possible becasue this commit had been shared for other project as submodule)
so I suggest that separating new PR for additional refactoring work. 🙏 cc @moreal @dahlia

Libplanet.Explorer/GraphTypes/BencodexValueType.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Libplanet.Explorer.GraphTypes
+{
+    public class BencodexValueType : StringGraphType
+    {
+        private static readonly Codec _codec = new ();
+
+        public BencodexValueType()
+        {
+            Name = "IValue";
```
moreal, line 14: I think this should be also renamed 👀

Libplanet.Explorer/GraphTypes/IValueType.cs
moreal: It's my mistake 😅 . Close this thread.

Libplanet.Explorer/GraphTypes/IValueType.cs
moreal: Yes, it was because the document example does and I don't know what parts of the `StringGraphType` are needed 🤔

moreal:
> I realized that there are too many commits will be squashed. 😅 (I also want to avoid rebasing as much as possible becasue this commit had been shared for other project as submodule)
so I suggest that separating new PR for additional refactoring work. 

I agree 👍🏻

Libplanet.Explorer/GraphTypes/IValueType.cs
longfin: On my understanding, `StringGraphType` can be natural because encoded value will be expressed as (hex) string value. (but I don't have strong opinion, I copied that from `PublicKeyType` in our project 😕)

longfin:
@moreal PTAL 🙏 (discussing about `StringGraphType` seems fine to delaying to the new issue or PR).

moreal (approved):
Though I'm no longer a maintainer of this project, I submitted an approval.
About the `BencodexValueType`, we should know about `ScalarGraphType` and what side effects when using `StringGraphType`.

PR #2137, opened 2022-07-11
Refactor `DataModel`

### LINKED ISSUE #1956: Refactor `DataModel` decoding
Since I couldn't find a right solution in time, decoding has been implemented using a long list of `if` and `else if` statements to find the right type to decode to. However, `DecodeFromListIValue()` can be replaced by something akin to the following:

```csharp
private static object DecodeFromListIValue(BTypes.List list, Type type)
{
	if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(ImmutableList<>))
	{
		Type[] genericTypes = type.GetGenericArguments();
		Type genericType = genericTypes[0];

		List<object> tempList = list
			.Select(x => DecodeFromIValue(x, genericType))
			.ToList();

		switch (tempList)
		{
			case List<object> listBool when genericType == typeof(bool):
				return listBool.Cast<bool>().ToImmutableList();
			case List<object> listInt when genericType == typeof(int):
				return listInt.Cast<int>().ToImmutableList();
			case List<object> listLong when genericType == typeof(long):
				return listLong.Cast<long>().ToImmutableList();
			case List<object> listBigInteger when genericType == typeof(BigInteger):
				return listBigInteger.Cast<BigInteger>().ToImmutableList();
			case List<object> listBytes when genericType == typeof(ImmutableArray<byte>):
				return listBytes.Cast<ImmutableArray<byte>>().ToImmutableList();
			case List<object> listAddress when genericType == typeof(Address):
				return listAddress.Cast<Address>().ToImmutableList();
			case List<object> listString when genericType == typeof(string):
				return listString.Cast<string>().ToImmutableList();
			default:
				throw new ArgumentException();
		}
	}
	else
	{
		throw new ArgumentException(
			$"Invalid target property type {type} encountered.");
	}
}
```

I doubt that there may be any significant performance hit (or this might even be faster), but once we observe that the speed is comparable to the one implemented in #1935, `DecodeFromListIValue()` and `DecodeFromDictionaryIValue()` should be cleaned up.

greymistcube:
Closes #1956.

Revisiting #1980.

To make sure that there will be no substantial performance, for reference, I'm leaving a benchmark result here.

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19044
11th Gen Intel Core i7-1185G7 3.00GHz, 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.413
  [Host]     : .NET Core 3.1.19 (CoreCLR 4.700.21.41101, CoreFX 4.700.21.41603), X64 RyuJIT
  DefaultJob : .NET Core 3.1.19 (CoreCLR 4.700.21.41101, CoreFX 4.700.21.41603), X64 RyuJIT

|             Method |       Mean |     Error |    StdDev |
|------------------- |-----------:|----------:|----------:|
|    EncodeRootModel | 5,661.5 us | 107.83 us | 110.74 us |
|    EncodeLeafModel | 1,645.6 us |  28.50 us |  25.26 us |
| EncodeRawLeafModel | 1,246.0 us |  22.13 us |  20.70 us |
|    DecodeRootModel | 3,246.3 us |  64.89 us |  88.82 us |
|    DecodeLeafModel | 1,034.1 us |  20.03 us |  18.73 us |
| DecodeRawLeafModel |   854.3 us |  16.65 us |  17.10 us |

### DIFF
--- a/Libplanet.Benchmarks/DataModel/DataModelBenchmark.LeafModel.cs
+++ b/Libplanet.Benchmarks/DataModel/DataModelBenchmark.LeafModel.cs
@@ -9,12 +9,12 @@ namespace Libplanet.Benchmarks.DataModel
 {
     public partial class DataModelBenchmark
     {
-        private class LeafModel : StoreDataModel
+        public class LeafModel : StoreDataModel
         {
             public LeafModel()
                 : base()
             {
-                System.Random random = new System.Random();
+                System.Random random = new System.Random(2);
                 BigList = Enumerable
                     .Range(0, 1000)
                     .Select(_ => random.Next())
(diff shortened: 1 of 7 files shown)


greymistcube:
Eh, there is slight change in `Exception` type being thrown (more of a bugfix), but I think it's a small enough change to skip changelog. 😗

dahlia (approved):

PR #2148, opened 2022-07-18
Slightly improve CI/CD experience

dahlia:
- `dotnet nuget push` had been failed when we re-run a job triggered by a tag push (due to its flakiness).  Now such failure would not happen as it ignores NuGet packages of already uploaded versions.
- Test report XMLs hadn't been collected for tests without code coverage.  (I omitted it by mistake.)  It's now fixed.
- @colibrishin & @OnedgeLee are added to the auto-assign list.

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -2,11 +2,12 @@ addReviewers: true
 addAssignees: author
 
 reviewers:
+- colibrishin
 - dahlia
-- longfin
-- limebell
-- riemannulus
 - greymistcube
-#- kfangw
+- limebell
+- longfin
+- OnedgeLee
+- riemannulus
 
 numberOfReviewers: 0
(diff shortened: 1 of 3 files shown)


.github/auto_assign.yml
```diff
@@ -2,11 +2,12 @@ addReviewers: true
 addAssignees: author
 
 reviewers:
+- colibrishin
```
riemannulus, line 5: 👋

.github/auto_assign.yml
```diff
@@ -2,11 +2,12 @@ addReviewers: true
...
 - dahlia
-- longfin
+- greymistcube
 - limebell
+- longfin
+- OnedgeLee
```
riemannulus, line 10: 👋

riemannulus (approved):

.github/auto_assign.yml
OnedgeLee: 👋

OnedgeLee (approved):

.github/auto_assign.yml
colibrishin: 👋

Issue #2155, opened 2022-07-19
Timestamp-related operations fail when BlockChainTest is run with RocksDBStoreFixture


tkiapril:
When I add `Libplanet.RocksDBStore` reference to `Libplanet.Tests`, copy `RocksDBStoreFixture.cs` from `Libplanet.RocksDBStore.Tests`, use `RocksDBStoreFixture` for `_fx` and run `BlockChainTest` tests, two kinds of failures occur:

- `BlockPerception.PercievedTime` loses precision under millisecond when `BlockChain<T>.PerceiveBlock()` is called on the same block multiple times (https://github.com/planetarium/libplanet/blob/main/Libplanet.Tests/Blockchain/BlockChainTest.cs#L140)
```
Assert.Equal() Failure
Expected: 2022-07-19T08:07:38.3575230+00:00
Actual:   2022-07-19T08:07:38.3570000+00:00
```
- `RocksDBStore.PutTransaction<T>` throws exception when dividing `timestamp` with `_txEpochUnitSeconds`, while called from `BlockChain<T>.Append()`
  - https://github.com/planetarium/libplanet/blob/main/Libplanet.RocksDBStore/RocksDBStore.cs#L705
  - https://github.com/planetarium/libplanet/blob/main/Libplanet.Tests/Blockchain/BlockChainTest.Append.cs#L73
  - https://github.com/planetarium/libplanet/blob/main/Libplanet.Tests/Blockchain/BlockChainTest.Internals.cs#L114-L119
  - https://github.com/planetarium/libplanet/blob/main/Libplanet.Tests/Blockchain/BlockChainTest.cs#L824

From a crude investigation, these issues seem to occur due to problems related to how RocksDBStore handles time. Specifically, the first failure occurs due to the fact that time is saved in milliseconds to the DB, but returns the original time object when the block is perceived for the first time. (https://github.com/planetarium/libplanet/blob/main/Libplanet.RocksDBStore/RocksDBStore.cs#L1039)

tkiapril:
This issue affects `RocksDBStoreBlockChainTest` added in #2156.

PR #2156, opened 2022-07-19
Consider nonexistent chain as an empty chain

### LINKED ISSUE #2047: Fix a `ChainIdNotFoundException` when transaction was never run
https://github.com/planetarium/libplanet/blob/main/Libplanet/Blockchain/BlockChain.cs#L1192
If it ran PreloadAsync after creating a clean blockchain where no transactions have been created, `ChainIdNotFoundException` occurs.
When no transactions are created, I think need to skip

tkiapril:
Resolves #2047.
With this PR, nonexistent chain ids in `IStore` are considered to be chain ids with empty chains, thus fixing the bug when empty block collections or `TxNonce` collections are to be forked.


tkiapril:
A concern for `RocksDBStore`: In `ForkBlockIndexes`, as with when `branchpoint` is same as the `genesisHash`, when `genesisHash` is null (there are no blocks associated with the chain id,) I made it so that the function does nothing and return. I believe this is logically okay as it didn't break any (previously not broken) tests, but I was wondering if I would have had to do create `PreviousChainId`, `PreviousChainIndex`, `IndexCount`, `ChainId` and do `AddFork` for the (both empty) chain ids. It wouldn't be a problem since other operations on nonexistent chain id should return appropriate values indicating nonexistent chain (be it an empty collection, `null`, or `0`,) but I was wondering if there may be some places with codes that do not work accordingly when there are no `PreviousChainId` etc.

greymistcube:
Also can we add test cases for using `DefaultStore` and `RocksDBStore` to `StoreTest.cs`?

tkiapril:
> Also can we add test cases for using `DefaultStore` and `RocksDBStore` to `StoreTest.cs`?
Good idea, though tests using `RocksDBStore` is broken atm. (#2155) Should I be adding Libplanet.RocksDBStore as a dependency to Libplanet.Tests or does that make any sense?

dahlia (approved):

dahlia (approved):

tkiapril:
#2159 should be merged first for the tests to not fail.

dahlia:
#2159 has just been merged.  Could you rebase it on the current *main*?  There seems a conflict on *CHANGES.md* file.

tkiapril:
@dahlia Done!

dahlia (approved):

longfin (approved):

dahlia (approved):

PR #2159, opened 2022-07-20
Fix timestamp-related problems in RocksDBStore

### LINKED ISSUE #2155: Timestamp-related operations fail when BlockChainTest is run with RocksDBStoreFixture

tkiapril:
Resolves #2155.

The overflow was due to trying to cast `timestamp` to `int`, whereas `timestamp` can be anything in the range of `Int64`. Specifically, in `BlockChainTest`, the timestamp used for txs are set to Jan 1st 0, which is well outside `Int32`. `int` cast was removed.

`RocksDBStore` was using `DateTimeOffset.ToUnixMilliseconds` in `PutBlock`, which reduces the precision to at most miliseconds. As such, tests were failing for `PerceivedTime` since the default value for it was derived by `DateTimeOffset.UtcNow` which has more precision. Now, the default value for `PerceivedTime` has millisecond precision.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,9 @@ To be released.
 
 ### Behavioral changes
 
+ -  `BlockChain<T>.PerceiveBlock()` method now uses millisecond precision for
+    perceive time of newly perceived blocks. [[#2155], [#2159]]
+
 ### Bug fixes
 
 ### Dependencies
@@ -33,6 +36,8 @@ To be released.
 
 [#1762]: https://github.com/planetarium/libplanet/issues/1762
 [#2140]: https://github.com/planetarium/libplanet/pull/2140
+[#2155]: https://github.com/planetarium/libplanet/issues/2155
+[#2159]: https://github.com/planetarium/libplanet/pull/2159
 
 
 Version 0.39.0
--- a/Libplanet.RocksDBStore/RocksDBStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBStore.cs
@@ -702,7 +702,7 @@ namespace Libplanet.RocksDBStore
             }
 
             long timestamp = tx.Timestamp.ToUnixTimeSeconds();
-            string txDbName = $"epoch{(int)timestamp / _txEpochUnitSeconds}";
+            string txDbName = $"epoch{timestamp / _txEpochUnitSeconds}";
             _rwTxLock.EnterWriteLock();
             try
             {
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -25,6 +25,9 @@ To be released.
 
 ### Behavioral changes
 
+ -  `BlockChain<T>.PerceiveBlock` now uses millisecond precision for perceive
+    time of newly perceived blocks. [[#2155], [#2159]]
```
longfin, line 29:
```suggestion
 -  `BlockChain<T>.PerceiveBlock()` method now uses millisecond precision for 
    perceive time of newly perceived blocks. [[#2155], [#2159]]
```

longfin (approved):

longfin (approved):

dahlia (approved):

PR #2171, opened 2022-07-22
Introduce `IStore.GetCanonicalGenesisBlock<T>()` method

### LINKED ISSUE #2162: Introduce `IStore.GetCanonicalGenesisBlock()` method
Currently, we need to write like as [this](https://github.com/planetarium/NineChronicles.Headless/blob/036c161db28661b5d53f29298ee9183f4bd23a81/NineChronicles.Headless.Executable/Store/IStoreExtensions.cs#L12-L28) below for getting the genesis block from given store. Since `IStore` can directly obtain the original chain ID information in storage via `IStore.GetCanonicalChainId()`, it would be convenient if a method such as `IStore.GetCanonicalGenesisBlock()` also be provided.

But, current implmentation as `IStoreExtensions` has been claimed [due to performance reason](https://github.com/planetarium/NineChronicles.Headless/pull/1429/). so we need to tune the code when implement it on `IStore` implementations.

p.s. It can be another story, but I'm still wondering about `IStore`'s scope for chain IDs. (and chain ID concept itself). however, the scope of design decisions is too broad, so I'll break this into a separate discussion.

tkiapril:
Resolves #2162.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@ To be released.
 ### Backward-incompatible API changes
 
  -  Removed `ChainIdNotFoundException` class. [[#2047], [#2156]]
+ -  Added `IStore.GetCanonicalGenesisBlock(HashAlgorithmGetter)` method.
+    [[#2162], [#2171]]
 
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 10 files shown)


CHANGES.md
```diff
@@ -24,6 +24,8 @@ To be released.
    - Added `BlockChain<T>.FindNextHashes(BlockLocator, BlockHash?, int)` method.
    - Added `BlockChain<T>.Fork(BlockHash, bool)` method.
    - Added `BlockChain<T>.GetBlockLocator(int)` method.
+ -  Added `IStore.GetCanonicalGenesisBlock(HashAlgorithmGetter)` method.
+    [[#2162], [#2171]]
```
dahlia, line 28: Since adding a new method to an existing interface enforces classes implementing it to implement the new method, this change should be list in the backward-incompatible API changes.

dahlia (approved):

longfin (approved):

PR #2173, opened 2022-07-22
Bump target framework to 6.0 for test projects

### LINKED ISSUE #2090: Test projects require .NET Core 3.1

### LINKED ISSUE #2164: Bump `TargetFramework` of Libplanet.Test and other test projects
Currently, all test proejcts are targetting `.netcoreapp3.1`. but NET Core 3.1 will be dropped as supporting at the [end of 2022](https://dotnet.microsoft.com/en-us/platform/support/policy), and above all, .NET Core SDK 3.1 doesn't have Aarch support, so it's not available for Apple Silicon users.

Therefore, it is suggested to upgrade all the versions of the projects to .NET 6.

tkiapril:
Resolves #2164. Resolves #2090.

To fully drop 3.1, `Libplanet`, `Libplanet.Analyzers`, `Libplanet.Net`, `Libplanet.Stun`, and `Libplanet.Tools` projects also has to be bumped to 6.0.

Note: When `Libplanet.Explorer` project is set to target `net6.0`, it requires `6.x` version of `Microsoft.AspNetCore.Mvc.NewtonsoftJson` package. However, since MSBuild bundled with Mono cannot correctly interpret `net6.0` moniker, it is impossible to build the project with mono. For this, a new target `ReleaseMono` was set up to specifically exclude `Libplanet.Explorer*` projects from being built with mono.

Note 2: With `netcoreapp3.1`, only the executable produced with `dotnet publish` was marked executable. However, with `net6.0`, all the produced binaries are marked executable. Because of this, `bin_name` had to be produced in another way, which at the moment is looking whether it has `o+x` (or `g+x`) permission.


tkiapril:
Something to decide: In net6.0, using `--runtime` without `--(no-)self-contained` results in a [warning](https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/runtimeidentifier-self-contained). Need to decide which option to use.

longfin (approved):

.circleci/config.yml
```diff
@@ -119,15 +119,15 @@ commands:
     - run: |
         msbuild \
           -t:Restore \
-          -p:Configuration=Release \
+          -p:Configuration=ReleaseMono \
```
dahlia, line 122: Does this purpose to merely have a separated output directory?

tkiapril: This is actually for exclusing `Libplanet.Explorer*` projects from being compiled on Mono target. This is implemented via having a new `ReleaseMono` build target on the solution and excluding specific projects from the target, and then using the build target on Mono.

longfin (dismissed):

.circleci/config.yml
dahlia: Okay, could you mention that in *CONTRIBUTING.md* file?

dahlia (approved):

longfin (approved):

PR #2178, opened 2022-07-27
Refactor `Context` for thread safety and covering additional edge cases

### LINKED ISSUE #2193: Refactor PBFT Consensus Algorithm
When thought of as a state machine, `Context` transitions its state whenever certain [upon rule] conditionals are satisfied.

However, current implementation of `Context` does not account for satisfying additional [upon rule] conditionals as a result of processing an [upon rule]. Moreover, although a timeout trigger may change the `Step` and/or the `Round` of the `Context`, this also does not trigger checks on [upon rule]s.

There should be a mechanism to trigger running [upon rule]s whenever there is a state change for a `Context` object.

[upon rule]: https://arxiv.org/abs/1807.04938

greymistcube:
Closes #2193.

Refactors `Context.cs`.

## Overview
Largely tries to achieve the following two points:
- Better thread safety.
- Cover more cases.

### Thread Safety
While message consumption was single threaded, spawning side tasks for timeout events were problematic. As triggered timeout tasks needed to access the message log as a side task, race condition could happen as the main message consuming task could add messages to the message log while a timeout task was accessing the message log. Use of `Concurrency` type itself does not ensure concurrency, especially when conditional checks on the collection in question and `LINQ` expressions are involved.

## Edge Cases
The original intent was to additionally trigger processing tasks (i.e. apply upon rules) when timeout events occur, but as mentioned above, as timeout events are side tasks, simply calling the processing method, previously named `ProcessMessage()`, is even more dangerous as it would've meant processing methods running in parallel. So in order to
have additional calls to `ProcessMessage()`, calls were needed to be managed somehow.

## Approach
The gist of it is to have a separate long running task called `MutationConsumerTask()` that runs any process that **might potentially mutate** the state of the `Context` object. Note that adding a `ConsensusMessage` to the `MessageLog`[^1] is **mutating** the `Context` in question. Hence, it is important that `MessageConsumerTask()` **does not directly add** an incoming message to the `MessageLog`. Instead it queues up a process that adds the message to the `MessageLog` which `MutationConsumerTask()` can run.

Similarly, timeout events occurring outside the main thread, i.e. the long running `MutationConsumerTask()`, also do not call methods processing state changes directly, but simply queue them for `MutationConsumerTask()` to consume at some point.

Handling of edge cases are covered my spawning additional processes when `MutationConsumerTask()` consumes a mutation and observes a state change.

[^1]: There is no such exposed property called `MessageLog` for `Context`, but this term is used to simply assist in the understanding of the schematics described above.

### DIFF
--- a/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
@@ -17,7 +17,6 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
 {
     public class ConsensusContextNonProposerTest : ConsensusContextTestBase
     {
-        private const int Timeout = 60_000;
         private readonly ILogger _logger;
 
         public ConsensusContextNonProposerTest(ITestOutputHelper output)
@@ -241,8 +240,6 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
             await heightThreeStepChanged.WaitAsync();
             // Commit ends
             await heightTwoStepChanged.WaitAsync();
-            // GST, PreVote -> Propose (NewHeight and NewRound started)
-            await NewHeightDelayAssert(3);
             // Propose -> PreVote (message consumed)
             await heightThreeStepChanged.WaitAsync();
             Assert.Equal(3, ConsensusContext.Height);
--- a/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
@@ -137,7 +137,10 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
         {
             ConsensusPropose? propose = null;
             var heightTwoStepChanged = new AsyncAutoResetEvent();
-            var heightThreeStepChanged = new AsyncAutoResetEvent();
+            var heightTwoStepChangedToPreVote = new AsyncAutoResetEvent();
+            var heightTwoStepChangedToPreCommit = new AsyncAutoResetEvent();
+            var heightTwoStepChangedToEndCommit = new AsyncAutoResetEvent();
+            var heightThreeStepChangedToPreVote = new AsyncAutoResetEvent();
             var proposed = new AsyncAutoResetEvent();
 
             var codec = new Codec();
@@ -165,11 +168,22 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
                 throw new NullException(propose);
             }
 
-            ConsensusContext.Contexts[2].StepChanged += (sender, tuple) =>
+            ConsensusContext.Contexts[2].StateChanged += (sender, state) =>
             {
-                heightTwoStepChanged.Set();
+                if (state.Step == Step.PreVote)
+                {
+                    heightTwoStepChangedToPreVote.Set();
+                }
+                else if (state.Step == Step.PreCommit)
+                {
+                    heightTwoStepChangedToPreCommit.Set();
+                }
+                else if (state.Step == Step.EndCommit)
+                {
+                    heightTwoStepChangedToEndCommit.Set();
+                }
             };
-            await heightTwoStepChanged.WaitAsync();
+            await heightTwoStepChangedToPreVote.WaitAsync();
             Assert.Equal(2, ConsensusContext.Height);
             Assert.Equal(Step.PreVote, ConsensusContext.Step);
 
@@ -196,7 +210,7 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
                     });
             }
 
-            await heightTwoStepChanged.WaitAsync();
+            await heightTwoStepChangedToPreCommit.WaitAsync();
             Assert.Equal(Step.PreCommit, ConsensusContext.Contexts[2].Step);
 
             foreach (var privateKey in TestUtils.PrivateKeys)
@@ -232,16 +246,18 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
                     codec.Encode(blockHeightThree.MarshalBlock()),
                     -1));
 
-            ConsensusContext.Contexts[3].StepChanged += (sender, tuple) =>
+            ConsensusContext.Contexts[3].StateChanged += (sender, state) =>
             {
-                heightThreeStepChanged.Set();
+                if (state.Step == Step.PreVote)
+                {
+                    heightThreeStepChangedToPreVote.Set();
+                }
             };
-            // Propose -> PreVote (by consuming the message)
-            await heightThreeStepChanged.WaitAsync();
+
             // Commit ends
-            await heightTwoStepChanged.WaitAsync();
+            await heightTwoStepChangedToEndCommit.WaitAsync();
             // Propose -> PreVote (message consumed)
-            await heightThreeStepChanged.WaitAsync();
+            await heightThreeStepChangedToPreVote.WaitAsync();
             Assert.Equal(3, ConsensusContext.Height);
             Assert.Equal(Step.PreVote, ConsensusContext.Step);
         }
--- a/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Immutable;
+using System.Linq;
 using Bencodex;
 using Bencodex.Types;
 using Libplanet.Blocks;
@@ -37,7 +38,7 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
         public async void NewHeightWithLastCommit()
         {
             AsyncAutoResetEvent tipChanged = new AsyncAutoResetEvent();
-            AsyncAutoResetEvent proposed = new AsyncAutoResetEvent();
+            AsyncAutoResetEvent heightTwoProposed = new AsyncAutoResetEvent();
             ConsensusPropose? propose = null;
             var codec = new Codec();
             BlockChain.TipChanged += (sender, tuple) =>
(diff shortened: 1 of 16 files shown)


Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
```diff
@@ -233,18 +230,18 @@ public async void HandleMessageFromHigherHeight()
...
-            await heightThreeStepChanged.WaitAsync();
+
             // Commit ends
-            await heightTwoStepChanged.WaitAsync();
-            // GST, PreVote -> Propose (NewHeight and NewRound started)
-            await NewHeightDelayAssert(3);
+            await heightTwoStepChangedToEndCommit.WaitAsync();
             // Propose -> PreVote (message consumed)
-            await heightThreeStepChanged.WaitAsync();
+            await heightThreeStepChangedToPreVote.WaitAsync();
```
colibrishin, line 244: If I understood correctly, This does not check GST time but checking change of next height step change?

Libplanet.Net/Consensus/Context.cs
```diff
@@ -293,15 +256,23 @@ public VoteSet VoteSet(int round)
...
         {
-            Interlocked.Increment(ref _requestCount);
             _messageRequests.Writer.WriteAsync(message);
         }
 
+        /// <summary>
+        /// Adds a mutating <see cref="System.Action"/> to the mutation queue.
+        /// </summary>
+        /// <param name="mutation">A <see cref="System.Action"/> to be processed.</param>
+        public void ProduceMutation(System.Action mutation)
```
limebell, line 271: Is there any reason for this method to be public? If not, how about make it private and move it into `Context.Mutate.cs` because consumer task code is written there?

Libplanet.Net/Consensus/Context.Async.cs
```diff
@@ -0,0 +1,189 @@
...
+                    nextState.Round,
+                    nextState.Step.ToString());
+                StateChanged?.Invoke(
+                    this, (nextState.MessageLogSize, nextState.Round, nextState.Step));
+
+                // FIXME: This is to avoid an exception.
+                // Methods accessing message log should be changed instead.
+                if (_messagesInRound.ContainsKey(Round))
+                {
+                    ProduceMutation(() => ProcessGenericUponRules());
```
limebell, line 131: Why produce mutation instead of just call it directly? Seems like `ConsumeMutation` is single threaded.

Libplanet.Net/Consensus/Context.Async.cs
```diff
@@ -0,0 +1,189 @@
...
+namespace Libplanet.Net.Consensus
+{
+    public partial class Context<T>
+    {
+        /// <summary>
+        /// Starts round #0 of consensus for <see cref="Height"/>.
+        /// </summary>
+        /// <param name="lastCommit">A <see cref="Block{T}.LastCommit"/> from previous block.
+        /// </param>
+        public void StartAsync(BlockCommit? lastCommit = null)
```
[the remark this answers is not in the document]
greymistcube: I've pondered a bit about this. It depends on what you want "Async" to mean. From my perspective, the usage of keyword `async` together with return type `Task` would indicate that you get an `await`able asynchronous `Task`. Without return type signature, there would be no indication that this method starts an "async" process in the background.

Libplanet.Net/Consensus/Context.cs
greymistcube: It's an oversight. I'll change it to `private`. As for moving to `Context.Mutate.cs`, I tried to make `Context.Mutate.cs` **only** contain methods that can **actually** mutate the `Context` object. I think `Context.Async.cs` would be a better place for both `ProduceMessage()` and `ProduceMutation()`. 🙄

Libplanet.Net/Consensus/Context.Async.cs
greymistcube: Well, running `ProcessGenericUponRules()` can result in additional state changes, so no, that would possibly result in a different behavior unless we run a loop. 😶

Libplanet.Net/Consensus/Context.Async.cs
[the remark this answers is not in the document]
greymistcube: Hmm.. but now that I think of it, maybe `Task` without `async` might be more apt? 🤔

Libplanet.Net/Consensus/Context.Async.cs
limebell: But isn't `ProcessGenericUponRules` just put mutation into the mutastion channel?

greymistcube: I don't think I understand the question.

- `ProduceMutation()` puts a mutation, in this case `ProcessGenericUponRules()`, into the mutation channel.
- `MutationConsumerTask()` takes it out of the queue and consumes it at some point.
- `ProcessGenericUponRules()` actually mutates `Context` when certain conditions are met.

greymistcube: Now that I look at it again, there is an issue with split conditional and body. 😥
At the time of actually running `ProcessGenericUponRules()` does not guarantee `_messagesInRound` contains `Round`. Should be easy enough to fix I think. 😗

greymistcube: Arg... seems like this isn't that easy to fix. Probably have to deal with `FIXME` comment right above the highlighted body. 😥

Libplanet.Net/Consensus/Context.cs
greymistcube: See de40c5485070e83b0df34b289da2ec9494bf86c2.

Libplanet.Net/Consensus/Context.Async.cs
greymistcube: See de40c5485070e83b0df34b289da2ec9494bf86c2.

Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
```diff
@@ -16,97 +15,99 @@ public class ContextNonProposerTest : ContextTestBase
...
+            };
+            ConsensusMessageSent += (observer, message) =>
             {
-                if (step == Step.PreCommit)
+                if (message is ConsensusCommit commit)
                 {
-                    stepChanged.Set();
+                    Assert.Equal(commit.BlockHash, blockHash);
+                    commitSent.Set();
                 }
```
colibrishin, line 79: I remember of conversation about whether this catches exception internally. Is it okay to have this?

Libplanet.Net/Consensus/Context.Async.cs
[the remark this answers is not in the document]
limebell: This will go as it is?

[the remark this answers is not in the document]
greymistcube: Just changed the name to `StartTask()`. 😗

Libplanet.Net/Consensus/Context.Mutate.cs
```diff
@@ -0,0 +1,361 @@
...
+
+            lock (_messagesInRoundLock)
+            {
+                if (!_messagesInRound.ContainsKey(message.Round))
+                {
+                    _messagesInRound.TryAdd(message.Round, new HashSet<ConsensusMessage>());
+                }
+
+                // TODO: Prevent duplicated messages adding.
+                _messagesInRound[message.Round].Add(message);
```
colibrishin, line 130: Is `HashSet` added for preventing duplicated message?

limebell (approved):

Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
greymistcube: Due to ordering, `Assert` coming before `Set()` would break the test as failing `Assert` would result in `Set()` never being run, i.e. event never happening. As for using `Assert`, it is not much different from using `if` statement from unit test framework's perspective.

Libplanet.Net/Consensus/Context.Mutate.cs
greymistcube: `HashSet` is not added. It was there already. We might have to look into this further in another issue. 😗

Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
```diff
@@ -233,18 +230,18 @@ public async void HandleMessageFromHigherHeight()
...
+                }
             };
-            // Propose -> PreVote (by consuming the message)
-            await heightThreeStepChanged.WaitAsync();
+
             // Commit ends
-            await heightTwoStepChanged.WaitAsync();
-            // GST, PreVote -> Propose (NewHeight and NewRound started)
-            await NewHeightDelayAssert(3);
+            await heightTwoStepChangedToEndCommit.WaitAsync();
```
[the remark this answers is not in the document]
greymistcube: From what I understand, GST isn't something you can actually test. 🤔

Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
colibrishin: Ah, I have misunderstood with exception being caught internally in `Task`.

Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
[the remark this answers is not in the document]
colibrishin: Correction: There needs to be test of `Task.Delay()`, when `OnBlockChainTipChanged()` invoked by committing.

colibrishin (approved):

Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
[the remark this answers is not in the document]
greymistcube: See 9cdded85ab559fbda372d4b0b93ca9d4f73ee1c3.

Issue #1965, opened 2022-05-16
Fix SwarmBenchmark IndexOutOfRangeException Error


Atralupus:
Detail log: https://github.com/planetarium/libplanet/runs/5986779803?check_suite_focus=true
<img width="1450" alt="스크린샷 2022-05-16 오후 1 47 23" src="https://user-images.githubusercontent.com/30599098/168521604-e052ae69-1b39-4206-9430-b00f67518d04.png">

seems to be a problem here
```csharp
// Libplanet.Tests/TestUtils.cs 39 line
public static readonly PrivateKey[] AdjacentKeys = new PrivateKey[10]
```

kasumielf:
I was discover this bug not a caused by PrivateKey Array. When I attach to debug, IndexOutOfRangeException was excepted from https://github.com/planetarium/libplanet/blob/e19221ab6e59f07684c58b1f68f6ea8247a61f48/Libplanet.Benchmarks/SwarmBenchmark.cs#L62.

I'd fix this issue but not a make PR yet because SwarmBenchmark takes too much running times(over 10~15 minutes.) when I ran it.

I'm not sure this situation is not a specific. If it have no problem, I'll make PR.

PR #2181, opened 2022-07-28
Fix SwarmBenchmark IndexOutOfRangeException

kasumielf:
- Add debug support in Program.cs
- Switch object initialize in IterationSetup.
- Fix array index access in MakeGenesisBlock.

It addresses https://github.com/planetarium/libplanet/issues/1965


kasumielf:
I guess need to run benchmark and check a problem again.

PR #2185, opened 2022-07-31
Change 'Message' names

binnini:
Add postfix Msg on all Message concrete classes.

### DIFF
--- a/Libplanet.Net.Tests/Messages/BlockHashesTest.cs
+++ b/Libplanet.Net.Tests/Messages/BlockHashesTest.cs
@@ -23,10 +23,10 @@ namespace Libplanet.Net.Tests.Messages
         public void Constructor()
         {
             Assert.Throws<ArgumentException>(() =>
-                new BlockHashes(null, new[] { default(BlockHash) })
+                new BlockHashesMsg(null, new[] { default(BlockHash) })
             );
             Assert.Throws<ArgumentException>(() =>
-                new BlockHashes(123, new BlockHash[0])
+                new BlockHashesMsg(123, new BlockHash[0])
             );
         }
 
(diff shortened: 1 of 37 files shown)


longfin (changes requested):
@binnini Thanks for first contritution! There're two points we have to check. 😄 

1. CI is complaining about change log (`CHANGES.md`) now. but IMHO, there is no public interface changes for library users. so adding `[skip changelog]` can help on this situation. cc @planetarium/libplanet 

2. [We require CLA agrement to first time contributor](https://github.com/planetarium/libplanet/pull/2185#issuecomment-1200324615). please check [it](https://cla-assistant.io/planetarium/libplanet?pullRequest=2185) and sign (just login as GitHub account 😅 ) if no matter

OnedgeLee:
@longfin I'm not sure, so asking for opinion.
There are some magic bytes that identifies messages on [Message.cs](https://github.com/planetarium/libplanet/blob/main/Libplanet.Net/Messages/Message.cs), and they're referenced by `Type` property on methods.
Do we have to modify them, also? Actually there are no chance to confuse about their name, but for consistency, may be they're also have to be changed.

longfin:
> @longfin I'm not sure, so asking for opinion. There are some magic bytes that identifies messages on [Message.cs](https://github.com/planetarium/libplanet/blob/main/Libplanet.Net/Messages/Message.cs), and they're referenced by `Type` property on `Message`s. Do we have to modify them, also? Actually there are no chance to confuse about their name, but for consistency, may be they're also have to be changed.

Well, consistency can be matter as you said. but I think that `MessageType.Ping` seems natrual for me even if we changed `Ping` to `PingMsg` because `MessageType` is implying that its fields are message. 
Of course we also have namespaces(i.e., `Libplanet.Net.Messages`) for types like `PingMsg`, but we tend to use only name of type instead of fully-qualified name. 😅  cc @planetarium/libplanet 

On the other hands, if we may need more discussion about it, separating new issue seems good for progress. @OnedgeLee @binnini

OnedgeLee (approved):
LGTM

longfin (approved):
LGTM!

PR #2200, opened 2022-08-05
Implement `Currency` maximum supply and total supply tracking

### LINKED ISSUE #915: Currency should be able to have its maximum cap
So that minters cannot mint money at their will.

tkiapril:
Resolves #915.
I'm unsure if there are enough tests to ensure the correct mechanism, but I'm kind of lost which kinds of tests might be needed. Would you reviewers be able to have an emphasis on which kinds of tests might be needed while reading through the code please?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,13 +12,66 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  Removed `Currency(string, byte, IImutableSet<Address>?)` constructor.
+    [[#2200]]
+ -  Removed `Currency(string, byte, Address?)` constructor.
+    [[#2200]]
+ -  Added static methods of `Currency` that defines different kinds of
+    `Currency`.  [[#2200]]
+     -  Added `Currency.Capped(string, byte, (BigInteger, BigInteger),
+        IImutableSet<Address>?)` static method which defines an instance of
+        `Currency` with a hard limit on the maximum minted supply.
+     -  Added `Currency.Capped(string, byte, (BigInteger, BigInteger),
+        Address)` static method which defines an instance of `Currency` with a
+        hard limit on the maximum minted supply.
+     -  Added `Currency.Uncapped(string, byte, IImutableSet<Address>?)` static
+        method which defines an instance of `Currency` without an enforced
+        maximum supply limit.
+     -  Added `Currency.Uncapped(string, byte, Address)` static method which
+        defines an instance of `Currency` without an enforced maximum supply
+        limit.
+     -  *OBSOLETE, ONLY FOR LEGACY SUPPORT*: Added `Currency.LegacyUntracked(
+        string, byte, IImutableSet<Address>?)` static method which defines a
+        legacy `Currency` instance which is compatible with `Currency`
+        instances defined before total supply tracking support was introduced.
+     -  *OBSOLETE, ONLY FOR LEGACY SUPPORT*: Added `Currency.LegacyUntracked(
+        string, byte, Address)` static method which defines a legacy `Currency`
+        instance which is compatible with `Currency` instances defined before
+        total supply tracking support was introduced.
+     -  *NOTE:* if you already have some `Currency` instances defined in prior
+        to the addition of total supply tracking on a live chain, you cannot
+        modify the already-defined `Currency` instances as a capped or uncapped
+        `Currency` but have to define them with `Currency.LegacyUntracked()` as
+        the new Currency kinds are internally backwards-incompatible with the
+        legacy `Currency`.
+
 ### Backward-incompatible network protocol changes
 
 ### Backward-incompatible storage format changes
 
 ### Added APIs
 
+ -  Added `Currency.MaximumSupply` property.  [[#915], [#2200]]
+ -  Added `Currency.TotalSupplyTrackable` field.  [[#915], [#2200]]
+
 ### Behavioral changes
+
+ -  Bencodex related methods in `Currency` now accounts for the maximum
+    supply and total supply tracking.  [[#915], [#2200]]
+     -  For capped currencies, `Currency.Serialize()` method stores the `major`
+        and `minor` values of the maximum supply as `Integer` values under the
+        keys `maximumSupplyMajor` and `maximumSupplyMinor`. For uncapped and
+        legacy untracked currencies, the entries are omitted.
+     -  `Currency(IValue)` constructor now looks for the maximum supply and
+        total supply trackability in the given dictionary and restores them if
+        found.
+ -  `Currency`'s implementation of `ISerializable` now accounts for the maximum
+    supply and total supply tracking.  [[#915], [#2200]]
+     -  `Currency`'s implementation of `ISerializable.GetObjectData` now stores
+         the maximum supply if the `Currency` is capped.
+     -  `Currency(SerializationInfo, StreamingContext)` constructor now looks
+        for the maximum supply and total supply trackability and restores it if
+        found.
 
 ### Bug fixes
 
@@ -26,7 +79,9 @@ To be released.
 
 ### CLI tools
 
+[#915]: https://github.com/planetarium/libplanet/issues/915
 [#2216]: https://github.com/planetarium/libplanet/pull/2216
+[#2200]: https://github.com/planetarium/libplanet/pull/2200
 
 
 Version 0.40.0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,8 +44,23 @@ To be released.
         `Currency` but have to define them with `Currency.LegacyUntracked()` as
         the new Currency kinds are internally backwards-incompatible with the
         legacy `Currency`.
+ -  Added `IAccountStateDelta.TotalSupplyUpdatedCurrencies` property.
+    [[#915], [#2200]]
+ -  Added `IAccountStateView.GetTotalSupply(Currency)` method.
+    [[#915], [#2200]]
+ -  Added `IBlockChainStates<T>.GetTotalSupply(Currency, BlockHash,
+    TotalSupplyStateCompleter<T>` method which gets the total supply of
+    a `Currency` in `FungibleAssetValue` from the state, and if not found,
+    returns null.  [[#915], [#2200]]
 
 ### Backward-incompatible network protocol changes
+
+ -  The `Block<T>.CurrentProtocolVersion` is bumped from 3 to 4.  [[#2200]]
+     -  `IAccountStateDelta.GetTotalSupply` returns the current total supply
+        of a `Currency` if the total supply of`Currency` is trackable from
+        version 4, and throws `NotSupportedException` below version 4.
+     -  `IAccountStateDelta.MintAsset` and `IAccountStateDelta.BurnAsset`
+        tracks the total supply if trackable from version 4.
 
 ### Backward-incompatible storage format changes
 
@@ -53,6 +68,12 @@ To be released.
 
  -  Added `Currency.MaximumSupply` property.  [[#915], [#2200]]
  -  Added `Currency.TotalSupplyTrackable` field.  [[#915], [#2200]]
+ -  Added `SupplyOverflowException` class.  [[#915], [#2200]]
+ -  Added `TotalSupplyGetter` delegate.  [[#915], [#2200]]
+ -  Added `TotalSupplyStateCompleter<T>` delegate.  [[#915], [#2200]]
+ -  Added `TotalSupplyStateCompleters<T>` static class.  [[#915], [#2200]]
+ -  Added `StateCompleterSet<T>.TotalSupplyStateCompleter` property.
+    [[#915], [#2200]]
 
 ### Behavioral changes
 
@@ -72,6 +93,14 @@ To be released.
      -  `Currency(SerializationInfo, StreamingContext)` constructor now looks
         for the maximum supply and total supply trackability and restores it if
         found.
+ -  `IAccountStateDelta.MintAsset(Address, FungibleAssetValue)` and
+    `IAccountStateDelta.BurnAsset(Address, FungibleAssetValue)` methods now
+    track the total supply if the total supply of the `Currency` is trackable.
+    [[#915], [#2200]]
+     -  `IAccountStateDelta.MintAsset(Address, FungibleAssetValue)` method now
+        throws `SupplyOverflowException` if the sum of current total supply and
+        the value to be minted exceeds the maximum supply of the `Currency`
+        instance.
 
 ### Bug fixes
 
--- a/Libplanet/Action/AccountStateDeltaExtensions.cs
+++ b/Libplanet/Action/AccountStateDeltaExtensions.cs
@@ -32,6 +32,18 @@ namespace Libplanet.Action
                 )
             ).ToImmutableDictionary();
 
+        internal static IImmutableDictionary<Currency, FungibleAssetValue>
+            GetUpdatedTotalSupplies(this IAccountStateDelta delta) =>
+            delta.TotalSupplyUpdatedCurrencies
+                .Where(currency => currency.TotalSupplyTrackable)
+                .Select(currency => (currency, totalSupply: delta.GetTotalSupplyImpl(currency)))
+                .Where(t => t.totalSupply.HasValue)
+                .Select(t =>
+                    new KeyValuePair<Currency, FungibleAssetValue>(
+                        t.currency,
+                        t.totalSupply!.Value))
+                .ToImmutableDictionary();
+
         internal static IImmutableDictionary<string, IValue?> GetUpdatedRawStates(
             this IAccountStateDelta delta) =>
             delta.GetUpdatedStates()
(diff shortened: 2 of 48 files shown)


Libplanet/Action/AccountStateDeltaExtensions.cs
```diff
@@ -32,6 +32,16 @@ internal static class AccountStateDeltaExtensions
...
             ).ToImmutableDictionary();
 
+        internal static IImmutableDictionary<Currency, FungibleAssetValue>
+            GetUpdatedTotalSupplies(this IAccountStateDelta delta) =>
+            (
+                from currency in delta.TotalSupplyUpdatedCurrencies
+                where currency.TotalSupplyTrackable
+                let totalSupply = delta.GetTotalSupplyImpl(currency)
+                where totalSupply.HasValue
+                select new KeyValuePair<Currency, FungibleAssetValue>(currency, totalSupply.Value)
```
dahlia, line 42: It's never justified, but we haven't used SQL-style LINQ expressions. It might be better to keep avoiding it for consistency, but I'm not really sure.  @planetarium/libplanet Opinions?

Libplanet/Assets/Currency.cs
```diff
@@ -61,19 +62,67 @@ namespace Libplanet.Assets
...
+        /// <param name="maximumSupply">The uppermost quantity of currency allowed to exist.
+        /// See also <see cref="MaximumSupply"/> field which corresponds to this.
+        /// Must be <value>null</value> when <paramref name="totalSupplyTrackable" /> is
+        /// <value>null</value>.</param>
         /// <param name="minters">The <see cref="Address"/>es who can mint the currency.
         /// See also <see cref="Minters"/> field which corresponds to this.</param>
+        /// <param name="totalSupplyTrackable">Whether the total supply of this currency is
+        /// trackable. This parameter is for backwards compatibility with
+        /// <see cref="Currency">Currencies</see> defined before maximum supply was introduced,
+        /// and must be set to <value>false</value> in that case, and should be <value>true</value>
```
dahlia, line 88: It's my first time to meet `<value>` in XML comments.  Could you give a reference about this feature?

Libplanet/Assets/Currency.cs
```diff
@@ -61,19 +62,67 @@ namespace Libplanet.Assets
...
-        public Currency(string ticker, byte decimalPlaces, IImmutableSet<Address>? minters)
+        /// is an empty string, or when <paramref name="totalSupplyTrackable"/> is
+        /// <value>false</value> but <paramref name="maximumSupply"/> is not <value>null</value>,
+        /// or when either the Major or the Minor values of <paramref name="maximumSupply"/> is
+        /// a negative number, or when the given Minor unit for the <paramref name="maximumSupply"/>
+        /// is too big for the given <paramref name="decimalPlaces"/>.</exception>
+        public Currency(
+            string ticker,
+            byte decimalPlaces,
+            (BigInteger Major, BigInteger Minor)? maximumSupply,
```
dahlia, line 99: It might be good to have one more overload taking only `maximumSupplyMajor`.

Libplanet/Assets/Currency.cs
```diff
@@ -61,19 +62,67 @@ namespace Libplanet.Assets
...
+        /// See also <see cref="MaximumSupply"/> field which corresponds to this.
+        /// Must be <value>null</value> when <paramref name="totalSupplyTrackable" /> is
+        /// <value>null</value>.</param>
         /// <param name="minters">The <see cref="Address"/>es who can mint the currency.
         /// See also <see cref="Minters"/> field which corresponds to this.</param>
+        /// <param name="totalSupplyTrackable">Whether the total supply of this currency is
+        /// trackable. This parameter is for backwards compatibility with
+        /// <see cref="Currency">Currencies</see> defined before maximum supply was introduced,
+        /// and must be set to <value>false</value> in that case, and should be <value>true</value>
+        /// for new <see cref="Currency">Currencies</see>.</param>
```
dahlia, line 89: As its parameter type is unusual, users could be confused what's major and what's minor and what values should be passed to them.  How about providing an example?  Like this:

> For example, `maximumSupply` in `new Currency("ABC", 2, (123, 45))` means token ABC can be minted up to 123.45 ABC.

Libplanet/Assets/Currency.cs
tkiapril: Initially when I skimmed through [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#value) document, I thought it was about marking raw values, however now that I read it again it seems like I had a misunderstanding about the feature. 😅 Gotta remove it.

Libplanet/Action/AccountStateDeltaImpl.cs
```diff
@@ -118,6 +131,19 @@ Address signer
...
+        public virtual FungibleAssetValue GetTotalSupply(Currency currency)
+        {
+            if (((IAccountStateDelta)this).GetTotalSupplyImpl(currency) is { } totalValue)
+            {
+                return totalValue;
+            }
+
+            throw new TotalSupplyNotTrackableException(
+                currency,
+                $"The total supply of currency {currency} is not trackable.");
```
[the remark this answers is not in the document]
tkiapril: 👌

Libplanet/Assets/Currency.cs
tkiapril: That would make more sense. 👍

Libplanet/Assets/Currency.cs
dahlia: Thought about this again, and such overload would confuse code readers, because with no parameter name specified, its order of magnitude could be quite ambiguous, say, whether `1000` means `(1000, 0)` or `(10, 0)` when `decimalPlaces` is 2.

So, now I'm not sure if having such overload is really better.  Let's keep going as it is.

tkiapril: 👌

CHANGES.md
```diff
@@ -12,21 +12,107 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  Constructors of `Currency` class now take `(BigInteger Major),
+    BigInteger Minor)?` as the maximum supply parameter, and a boolean value to
```
greymistcube, line 16:
```suggestion
 -  Constructors of `Currency` class now take `(BigInteger Major,
    BigInteger Minor)?` as the maximum supply parameter, and a boolean value to
```

CHANGES.md
```diff
@@ -12,21 +12,107 @@ To be released.
...
+    mark whether the `Currency` is trackable.  [[#915], [#2200]]
+    -  *NOTE:* if you already have defined some `Currency` instances in prior
+       to the addition of maximum supply, and if it is on a live chain, you
+       cannot modify the already-defined `Currency` instances to have maximum
+       supply caps since it would make those `Currency` instances to be
+       effectively different ones compared to the original ones, as it would
+       change their `Hash`es. You must pass a `null` as their maximum supply
+       parameter. Also, you must set the `totalSupplyTrackable` parameter to
+       false.
```
greymistcube, line 25: Any reason why  this should be *explicitly* set to `false`? Is there a case where we don't want to track but have total supply capped for new `Currency`?

CHANGES.md
```diff
@@ -12,21 +12,107 @@ To be released.
...
+    -  Removed `Currency(string, byte, IImutableSet<Address>?)` constructor.
+    -  Removed `Currency(string, byte, Address?)` constructor.
+    -  Added `Currency(string, byte, (BigInteger Major, BigInteger Minor)?,
+       IImutableSet<Address>?, bool)` constructor.
+    -  Added `Currency(string, byte, (BigInteger Major, BigInteger Minor)?,
+       Address?, bool)` constructor.
```
greymistcube, line 48: I think this is a copy?

CHANGES.md
tkiapril: Actually, three configurations are possible:

- `maximumSupply is null` and `totalSupplyTrackable = false`, in which case no maximum supply cap nor total supply tracking is available;
- `maximumSupply is null` and `totalSupplyTrackable = true`, in which case no maximum supply cap is present but total supply tracking is available;
- `maximumSupply is { }` and `totalSupplyTrackable = true`, in which case maximum supply cap is set in place.

In case of `maximumSupply is { }` and `totalSupplyTrackable = false`, the constructor throws an [`ArgumentException`](https://github.com/planetarium/libplanet/pull/2200/commits/beb3e7b064fe36d7cf4419dcfa22c7e595404164#diff-4ba5d5e78e585f64e217441ed95e634f2d1b5c5df16ac3d419834d040882d7c4R103-R109).

CHANGES.md
tkiapril: Oopsie 😶

CHANGES.md
```diff
@@ -12,21 +12,107 @@ To be released.
...
+ -  Added `TotalSupplyStateCompleters<T>` static class.  [[#915], [#2200]]
+ -  Added `StateCompleterSet<T>.TotalSupplyStateCompleter` property.
+    [[#915], [#2200]]
+
 ### Behavioral changes
 
+ -  Bencodex related methods in `Currency` now accounts for the maximum
+    supply and total supply tracking.  [[#915], [#2200]]
+    -  `Currency.Serialize()` method now stores the maximum supply as
+       non-nullable Bencodex `Integer` values under the keys
```
greymistcube, line 85: This is a bit misleading. I think "non-nullable" should be removed or the surrounding context should be rephrased as Bencodex dictionary values are type agnostic.

Libplanet/Action/AccountStateDeltaExtensions.cs
greymistcube: I think consistency is more important at this point. I'm sure most, if not all, of us are comfortable enough with using method syntax. Until we come up with some sort of guideline when to use query syntax and when to use method syntax, I think it is better to postpone mixing different types of syntaxes. 😶

CHANGES.md
tkiapril: Would something like:

-  If an instance of `Currency` has a maximum supply value, `Currency.Serialize()` method now stores it as
    `Integer` values under the keys
    `maximumSupplyMajor` and `maximumSupplyMinor`. Otherwise, the entries are omitted.

be ok?

Libplanet/Action/AccountStateDeltaExtensions.cs
tkiapril: 🆗

Libplanet/Action/AccountStateDeltaImpl.cs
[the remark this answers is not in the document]
greymistcube: Without "(Libplanet 0.41.0)" phrase perhaps, just in case? 🤔

CHANGES.md
greymistcube: So it is for backward compatibility reason. Now I understand the context more. Then again, I'm wondering if it'd be possible to discourage using `(null, false)`, i.e. legacy `Currency`, more strongly in general. 🙄

tkiapril: One of the efforts currently in place to avoid having `totalSupplyTrackable = false` is having `totalSupplyTrackable = true` as the default parameter, but now that I think about it, how would it be having public constructor for "new" `Currency` only, and having a private constructor that creates the legacy `Currency`, and having a `[Obsolete] public static Currency.CreateNonTrackableCurrency()`? cc @dahlia

tkiapril:
I've applied the suggestions. PTAL 🙇 @dahlia @greymistcube

CHANGES.md
```diff
@@ -12,21 +12,105 @@ To be released.
...
 
+ -  Removed `Currency(string, byte, IImutableSet<Address>?)` constructor.
+    [[#2200]]
+ -  Removed `Currency(string, byte, Address?)` constructor.
+    [[#2200]]
+ -  Added static methods of `Currency` that defines different kinds of
+    `Currency`.  [[#2200]]
+    -  Added `Currency.Capped(string, byte, (BigInteger, BigInteger),
+       IImutableSet<Address>?)` static method which defines an instance of
+       `Currency` with a hard limit on the maximum minted supply.
```
greymistcube, line 23: Indentation is off for nested list items. 😶

Libplanet/Action/TotalSupplyGetter.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <para>Gets the total supply of the given <paramref name="currency"/>.
+    /// </para>
+    /// </summary>
+    /// <param name="currency">The currency type to query the total supply.
+    /// </param>
+    /// <returns>The total supply value of the <paramref name="currency"/> in
+    /// <see cref="FungibleAssetValue"/>, or null if there were no former total supply value.
+    /// </returns>
+    /// <remarks>Note that you must not return the value of <paramref name="currency"/> * 0 if
+    /// <see cref="Currency.TotalSupplyTrackable"/> is <value>true</value> and there is no former
```
greymistcube, line 16:
```suggestion
    /// <see cref="Currency.TotalSupplyTrackable"/> is <c>true</c> and there is no former
```
Or `<see langword="true"/>`. We should really decide which should be the standard. 😗

tkiapril: Oof. Missed that one.

Libplanet/Assets/Currency.cs
```diff
@@ -24,12 +24,25 @@ namespace Libplanet.Assets
...
+    /// do not have the said limit. To define a <see cref="Currency"/> you may call either of the
+    /// following.
+    /// <ul>
+    /// <li><see
+    /// cref="Capped(string,byte,ValueTuple{BigInteger, BigInteger},IImmutableSet{Address}?)"/>
+    /// </li>
+    /// <li><see cref="Capped(string,byte,ValueTuple{BigInteger, BigInteger},Address)"/></li>
+    /// <li><see cref="Uncapped(string,byte,IImmutableSet{Address}?)"/></li>
+    /// <li><see cref="Uncapped(string,byte,Address)"/></li>
+    /// </ul>
```
greymistcube, line 44: I don't think this works? See other lists in XML docs. Also [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#list).

Libplanet/Action/TotalSupplyGetter.cs
tkiapril: IMO, for single keywords `<see langword="blah"/>` seems more semantical, but there are many uses of `<c/>` in the codebase as well. 🤔

Libplanet/Assets/Currency.cs
tkiapril: Thanks for pointing out. Though, not that I prefer the usage of <ul/> and <li/> tags, but they actally does work on Rider (apparently XML docs are still XML, so regular tags do work.) Regardless <list> seems recommended so I'll use that.

Libplanet/Action/TotalSupplyGetter.cs
greymistcube: `<see langword="foo"/>` is indeed more appropriate, but sadly, DocFX [does not support it properly](https://github.com/dotnet/docfx/issues/7625) at the moment.

tkiapril: 😢

Libplanet/Assets/Currency.cs
greymistcube: Seems like it doesn't work in VSCode. Also I was more concerned about generated documentation. 😗

Libplanet/Assets/Currency.cs
```diff
@@ -149,9 +122,92 @@ public Currency(IValue serialized)
...
+                    + " omitted or be non-negative integers.",
+                    nameof(serialized)
+                );
+            }
+
             Ticker = ticker;
             DecimalPlaces = (byte)(long)decimals;
 
+            if (_maximumSupply is var (_, minor) && minor > 0 &&
+                (int)Math.Floor(BigInteger.Log10(minor) + 1) > DecimalPlaces)
```
greymistcube, line 204: Shouldn't
```csharp
Math.Floor(BigInteger.Log10(minor)) >= DecimalPlaces
```
be enough? 🤔

Libplanet/Blockchain/KeyConverters.cs
```diff
@@ -12,5 +12,8 @@ internal static class KeyConverters
 
         internal static string ToFungibleAssetKey((Address, Currency) pair) =>
             ToFungibleAssetKey(pair.Item1, pair.Item2);
+
+        internal static string ToTotalSupplyKey(Currency currency) =>
+            "__" + ByteUtil.Hex(currency.Hash.ByteArray).ToLowerInvariant();
```
greymistcube, line 17: I'm sure there is a reason, but just out of curiosity, what's with the magic string `"__"`? 😶

greymistcube (dismissed):
I've skimmed through twice, but I'd double check with @dahlia. 🙄

Libplanet/Blockchain/KeyConverters.cs
tkiapril: It's out of thought that if the key for FAV is `_(address)_(currency)`, having the total supply of a currency (which might be as well as thought of the "global" balance) might as well as be `_(nothing)_currency`. I've also thought about having something like `_system_(currency)`, but after @dahlia and I had a conversation about whether we need a system namespace for state, we thought that it might be unnecessary.

tkiapril:
@greymistcube Thank you so much for your time 🙇

Libplanet/Assets/Currency.cs
tkiapril: That seems much less convoluted!

dahlia (approved):

Issue #1934, opened 2022-05-02
Introduce an easy way to get peer string for `Swarm<T>.BootstrapAsync()`


longfin:
`BoundPeer.ToString()` returns [`Address`.`EndPoint`.`PublicIPAddress`](https://github.com/planetarium/libplanet/blob/515dbe10fc1f92db1781432f82dd05941796aa1c/Libplanet.Net/BoundPeer.cs#L159) (e.g. `"0x7d904Bf4A84A92e7C4f976369361aD834b7dCA5a.Unspecified/localhost:31234."`). 

but we need to its public key to create `BoundPeer` directly. (Even `BoundPeer` has a factory method `.ParsePeer()` which creates from an already formatted string. 😇 )

greymistcube:
Just to elaborate and give it more context, this heavily depends on the running context.

- Is `PrivateKey` provided?
- Is `Swarm<T>` using `host` or `iceServers`?
  - If `Swarm<T>` is using `host`, is `port` provided?

The earliest stage as to when we can obtain the peer string for each combination is as follows:
- `PrivateKey`, `host`, and `port` are all provided: Before `Swarm<T>` instantiation.
- All other possible combinations: After `Swarm<T>.BootstrapAsync()`.

The latter is due to binding happening at unnecessarily late, and I mean **really** late, during `Swarm<T>.StartAsync()` phase, although this is probably possible during `Swarm<T>()`. This is due to a design flaw, and to properly resolve this issue, #1612 must be resolved first. 🙄

As we have discussed, as for needing it only for development/deployment purpose of getting a fixed seed peer string, this **should not** require an `Swarm<T>` instance anyway as the decision for what `Swarm<T>`'s `BoundPeer` should be is being made at a higher level. I'll be putting in a quick and dirty method somewhere under `Libplanet.Node`. 😗

greymistcube:
Rudimentary functionality implemented in [UniLibplanet](https://github.com/planetarium/UniLibplanet).

Issue #1612, opened 2021-11-24
Encapsulate `ITransport` constructors


greymistcube:
Due to `async` components required for fully initializing `ITransport` implementations, initialization should be done in a two step process. In particular, it makes more sense to have allocation of the endpoint host and port to be finished during the initialization step, but this is currently done inside `ITransport.StartAsync()`. Possible advantages are:

- `Swarm<T>` would be bound as a `Peer`, i.e. as a `BoundPeer`, when exiting `Swarm<T>()`.
- `Swarm<T>` can fail fast at an appropriate point.

I suggest lowering access to constructors and making use of `static` `async` creators instead.

PR #2215, opened 2022-08-15
♻️ Refactor `NetMQTransport`'s initialization

greymistcube:
Related to #1612, #1934.
Some preparatory work to streamline uses of `ITransport` and `Swarm<T>`.

This is mostly to move up `ITransport`'s binding phase more in line with the rest of the `Libplanet.Net`'s (seemingly) intended design. 🙄

There are some problematic uses of `async` and `CancellationToken` handling here and there, which should be dealt with shortly. 😐

Hopefully with this, we might be ready to combine `Peer` and `BoundPeer` into a single class.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@ To be released.
 
 ### Deprecated APIs
 
+ -  (Libplanet.Net) Removed `NetMQTransport()` constructor.  Use
+    `NetMQTransport.Create()` instead.  [[#2215]]
  -  Unused `TcpMessageCodec` class removed.  [[#2216]]
  -  (Libplanet.Stun) Removed `TurnClient.IsConnectable()` method.  [[#2219]]
  -  (Libplanet.Stun) Removed `TurnClient.BindProxies()` method.
@@ -112,7 +114,8 @@ To be released.
         instance.
  -  (Libplanet.Net) `NetMQTransport`'s general behavior has changed.  [[#2215]]
      -  `NetMQTransport` is now able to send requests and receive
-        replies as soon as it is created.
+        replies as soon as it is created through `NetMQTransport.Create()`
+        factory method.
      -  `NetMQTransport.StartAsync()` enables a `NetMQTransport` instance
         to recieve requests and send replies.
      -  `NetMQTransport.StopAsync()` only disables a `NetMQTransport` instance
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,8 @@ To be released.
     TotalSupplyStateCompleter<T>` method which gets the total supply of
     a `Currency` in `FungibleAssetValue` from the state, and if not found,
     returns null.  [[#915], [#2200]]
+ -  (Libplanet.Net) `ITransport.AsPeer` and `Swarm<T>.AsPeer` type changed from
+    `Peer` to `BoundPeer`.  [[#2215]]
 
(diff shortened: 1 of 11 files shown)


longfin:
I'm somewhat skeptical of the need to force sync for `ITransport` (and `Swarm<T>`) creation. of course C# doesn't support `async` for constructor so that we need to additional async factory method. but in other hands, async factory method gives the user the freedom to choose how they want to handle the wait.

It may feel like this kind of freedom is at odds with usability, but unless we're making the decision to eliminate delays in the creation of these objects,  at least we need to provide a way to deal with them by default. and if necessary, it may be more appropriate to provide a framework-level helper such as DI extensions or UniLibplanet instead of providing a synchronous constructor.

That could be an entirely different story, but it's also a topic worth considering if we really need to get rid of the "uninitialized state". At least I agree with the problem that there is something more confusing and awkward about it at this point, but considering many ecosystems around C#/.NET(like DI) that already assumes, we have options as belows

- (as suggested in this PR) force to wait on synchornous APIs. 
  - or, shift the responsibility of initializing it to the user. [^1]
- admit "unintilalzed state" 😇.

In that case, I can't keep my agreement about "unintilalzed state" easily because flexibility about blocking is also matter IMHO.😢 

[^1]: when using async factory method in DI, it will be implemented with this pattern.

cc @planetarium/libplanet

greymistcube:
I agree, in general, this is a bad practice. My opinion is that, however, this is causing way more trouble than it should. Ideally, `async` static factory method for creating an `ITransport` (with more detailed specification) and handing the responsibility over to the user would be better. This, however, is no small task. Design-wise, I also agree having a user create an "initialized" `ITransport` and handing it over to `Swarm<T>` would make a lot more sense, but there are too many things to consider to "simply" change to a better design (which is why this issue keeps getting kicked down the road):

- Separating out `Initialize()` from `StartAsync()`
- Moving up `Initialize()` to earlier stage
- Enforcing a call to `Initialize()` **outside** `Swarm<T>`
- Creation of new `Swarm<T>` being part of synchronous code (at least for 9c-headless)

I really didn't want to (nor do I have the time to) open a new PR with 1,000+ lines of code change.

As for admitting an "uninitialized state", I think creation of `TurnClient` is "low-level" enough not to expose it to an end-user. I'm also not happy with how it looks at the moment.

I'm really not happy with how it looks at the moment, but I want to keep at it really, as now `NetMQTransport` is getting also used for PBFT gossip. This is just a **really awkward** compromise while trying to unentangle `Swarm<T>` and `ITransport`. There is really no reason why `Swarm<T>` should be responsible for initializing an `ITransport`. Also `Swarm<T>.Initialize()` was a no go for me. So I forcefully (and awkwardly) stuck it in `Swarm<T>()`. 🙄

Perhaps, this is too sloppy I suppose. I'll try to incorporate some design-choices as to hint more at what the end goal is. 😗

greymistcube:
@longfin 
So as discussed, I've tried to keep things `async` as much as possible. 😶

We still have ugly and awkward `GetAwaiter().GetResult()` stuck in `Swarm<T>()` constructor, but this should at least make it easier to extract it out from `Swarm<T>()` later down the road (although a proper handling shared `PrivateKey` and some other stuff between `Swarm<T>` and `ITransport` might be little tricky).

longfin (approved):

Issue #2218, opened 2022-08-16
Implement synchronous initialization for `TurnClient`


greymistcube:
When creating an `ITransport` instance, although the use of `async` "initialization" could be considered more generic, this seems to create more problems than solving unspecified and unrelated issues.

As far as `Swarm<T>` is concerned, before preloading and before peer discovery, `Swarm<T>` needs to have its transport ready to send and receive messages. As binding of ports and creating a TURN client involves network communication, although it would be tempting (and maybe even proper given all dependent design are in line with the transport layer's design) to implement the initialization step, apart from constructor phase, as constructors cannot be `async`, as a separate step. However, in doing so, this would require separate management of an explicit call to initializing method, which would leak into `Swarm<T>`'s design, i.e. `Swarm<T>` would be required to have a separate initializing method. In any case, currently implemented design is ill suited as can be seen in #2215 (which should be dealt with separately).

Contagiousness of `async` originates from `IceServer.CreateTurnClient` and this get seeped into `ITransport` and in turn into `Swarm<T>` (and probably all the way up to the client layer, if the overall design is to be consistent). IMO, either this should be resolved at the lowest level, i.e. `IceServer.CreateTurnClient`, or managed at the highest level, i.e. the client side. From a usability point, I believe the former would be better suited for our current situation, even if this would make things more restrictive. I don't see [Libplanet] needing to handle failure of `TurnClient` initialization in any foreseeable future.

[Libplanet]: https://github.com/planetarium/libplanet

greymistcube:
Closed by(?) #2219. After a discussion with @longfin, we've decided to take `async static` factory method route for now.

PR #2219, opened 2022-08-16
Refactor `IceServer.CreateTurnClient()`

greymistcube:
Related to #2218.

Two points:
 - Not sure if `ProbeConnectableLength` is appropriate.
 - ~Still wondering if there is a more elegant way of calling `async` from `sync`.~ 🙄

~P.S. I know forcefully calling `async` from `sync` is bad practice, but IMO this is still better than `async` code spreading uncontrollably without proper design support.~ 😐

### DIFF
--- a/Libplanet.Stun/Stun/TurnClient.cs
+++ b/Libplanet.Stun/Stun/TurnClient.cs
@@ -17,6 +17,7 @@ namespace Libplanet.Stun
     public class TurnClient : IStunContext, IDisposable
     {
         public const int TurnDefaultPort = 3478;
+        private const int ConnectableProbeLength = 2;
         private const int AllocateRetry = 5;
         private const int ProxyCount = 16;
 
@@ -37,7 +38,7 @@ namespace Libplanet.Stun
         private CancellationTokenSource _turnTaskCts;
         private List<Task> _turnTasks;
 
-        public TurnClient(
+        internal TurnClient(
             string host,
             string username,
             string password,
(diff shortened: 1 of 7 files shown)


Libplanet.Stun/Stun/TurnClient.cs
```diff
@@ -281,28 +282,32 @@ public async Task ReconnectTurn(int listenPort, CancellationToken cancellationTo
...
 
-        public async Task<bool> IsConnectable(
-            CancellationToken cancellationToken = default)
+        public async Task<bool> IsConnectable(int probeLength = ConnectableProbeLength)
         {
+            CancellationTokenSource cts = new CancellationTokenSource();
             try
             {
-                using var client = new TcpClient();
+                using TcpClient client = new TcpClient();
```
longfin, line 290: I think that `var` is fine in this case because type is clear due to`new TcpClient()` 😅

Libplanet.Stun/Stun/TurnClient.cs
```diff
@@ -281,28 +282,32 @@ public async Task ReconnectTurn(int listenPort, CancellationToken cancellationTo
             return !_control.Client.LocalEndPoint.Equals(mapped);
         }
 
-        public async Task<bool> IsConnectable(
-            CancellationToken cancellationToken = default)
+        public async Task<bool> IsConnectable(int probeLength = ConnectableProbeLength)
         {
+            CancellationTokenSource cts = new CancellationTokenSource();
```
longfin, line 287:
```suggestion
            using var cts = new CancellationTokenSource();
```

we can omit `finally` with `using` IMHO 😄

longfin (approved):
My opinion about sync consturctor had been written at https://github.com/planetarium/libplanet/pull/2215#issuecomment-1218147153 so I left other trivial comments here.🙏

greymistcube:
Completely different PR now.

- Keeps `async` for `TurnClient` "initialization".
- `IceServer.CreateTrunClient()` moved to `TurnClient.Create()`.
- Added `IIceServer` as a contract for `Libplanet.Net` to use `TurnClient.Create()`.

greymistcube:
As a side-note, decided to have `IIceServer` in `Libplanet.Stun` instead of moving `IceServer` to `Libplanet.Stun` from `Libplanet.Net` in order to avoid [lib9c] directly depending on `Libplanet.Stun` when parsing ICE server `string`. 😶

[lib9c]: https://github.com/planetarium/lib9c

longfin (approved):

PR #2223, opened 2022-08-18
Introduce BLS digital signature

colibrishin:
This PR introduces a new digital signature, BLS, which has the aggregable private key, public key, and signature. Due to this property, it has an advantage in verifying multiple signatures with multiple public keys by reducing signatures with aggregation. The implementation uses Minimal-pubkey-size[^1] (G1 as Public key and G2 as Signature).

## How fast it is?
| Method | 10 Signatures  | 100 Signatures |
| ------------- | ------------- | ------------- |
| BLS Signature.Aggregate & FastAggregateVerify  | 6 ms  | 72.2 ms |
| BLS Signature.Aggregate & AggregateVerify | 16.6 ms | 127.3 ms |
| Secp256k1 Sign & Verify  | 81.4 ms | 462.1 ms |

- Test signs and verifies the message.
- Average in 10 runs. 
- Note that `FastAggregateVerify`[^2] is for the case where the same message but different signatures, and `AggregateVerify`[^3] is for the case where different messages and different signatures. 
- The BLS library has a first initialization time of 423ms. the above bls run-time is the subtracted time.

## Classes
```mermaid
classDiagram
    class BlsPublicKey{
      byte[] KeyBytes
      Verify()
      VerifyProofOfPossession()
    }
    class BlsSignature{
      byte[] KeyBytes
      Aggregate()
      FastAggregateVerify()
      AggregateVerify()
      static MultiVerify()
    }
    class BlsPrivateKey{
      byte[] KeyBytes
      BlsSignature ProofOfPossession
      Sign()
    }
```

```mermaid
classDiagram
    PrivateKey <|-- IPrivateKey
    PublicKey <|-- IPublicKey
    BlsPrivateKey <|-- IPrivateKey
    BlsPublicKey <|-- IPublicKey
    DefaultCryptoBackend <|-- ICryptoBackend
    BlsCryptoBackend <|-- ICryptoBackend
    class BlsPrivateKey{
      -IReadonlyList<byte> KeyBytes => ByteArray
      -BlsPublicKey? publicKey
      +BlsPublicKey PublicKey
      +BlsSignature ProofOfPossession
      +ImmutableArray<byte> ByteArray
      +static PrivateKey FromString(string)
      +byte[] Sign(byte[])
      +ImmutableArray<byte> Sign(ImmutableArray<byte>)
      +static byte[] GenerateBytesFromHexString(string)
    }
    class PrivateKey{
      -IReadonlyList<byte> KeyBytes => ByteArray
      -PublicKey? publicKey
      +PublicKey PublicKey
      -ECPrivateKeyParameters KeyParam
      +ImmutableArray<byte> ByteArray
      +static PrivateKey FromString(string)
      +byte[] Sign(byte[])
      +ImmutableArray<byte> Sign(ImmutableArray<byte>)
      +byte[] Decrypt(byte[])
      +ImmutableArray<byte> Decrypt(ImmutableArray<byte>)
      +SymetricKey ExchangeKey(PublicKey)
      +static ECDomainParameters GetECParameters()
      +static ECDomainParameters GetECParameters(string)
      +static ECPrivateKeyParameters GenerateKeyParam()
      +static ECPrivateKeyParameters GeenrateKeyFromBytes(byte[])
      +static byte[] GenerateBytesFromHexString(string)
      -ECPoint CalculatePoint(ECPublicKeyParameters)
    }
    class BlsPublicKey{
      -IReadonlyList<byte> _publicKey
      +IReadonlyList<byte> KeyBytes => ToImmutableArray
      +ImmutableArray<byte> ByteArray
      +byte[] Verify(IReadonlyList<byte>, BlsSignature)
      +bool VerifyProofOfPossession(BlsSignature)
      +byte[] ToByteArray()
      +ImmutableArray<byte> ToImmutableArray()
    }
    class PublicKey{
      -IReadonly<byte> KeyBytes => ByteArray
      -ECPublicKeyParameters KeyParam
      -IPublicKey.KeyParam => KeyParam
      +static PrivateKey FromString(string)
      +byte[] Format(bool compress)
      +ImmutableArray<byte> ToImmutableArray(bool compress)
      +byte[] Verify(byte[])
      +ImmutableArray<byte> Verify(ImmutableArray<byte>)
      +byte[] Encrypt(byte[])
      +ImmutableArray<byte> Encrypt(ImmutableArray<byte>)
      +string ToString()
      -static ECDomainParameters GetECParameters(byte[] bs)
    }
    class BlsCryptoBackend {
      +BlsSignature Sign(HashDigest<T>, IPrivateKey)
      +bool Verify(HashDigetst<T>, byte[], IPublicKey)
      +bool AggregateVerify(BlsSignature, BlsPublicKey[])
      +bool FastAggregateVerify(BlsSignature, BlsPublicKey[])
      +byte[] GeneratePrivateKey()
      +BlsPublicKey GetPublicKey(BlsPrivateKey)
      +BlsSignature GetProofOfPossession(BlsPrivateKey)
      +bool VerifyPoP(BlsPublicKey, BlsSignature)
      +herumi.bls.SecretKey ValidateGetNativePrivateKey(BlsPrivateKey)
      +herumi.bls.PublicKey ValidateGetNativePublicKey(BlsPublicKey)
      +herumi.bls.Signature ValidateGetNativeSignature(BlsSignature)
      +herumi.bls.Msg ValidateGetNativeMsg(IReadonly<byte>)
      +Msg ConvertHashDigestToNativeMessage(HashDigest<T>)
    }
    class IPrivateKey{
      IReadonlyList<byte> KeyBytes
      IPublicKey PublicKey
      Sign(byte[])
    }
    class IPublicKey{
      IReadonlyList<byte> KeyBytes
      Verify(IReadOnlyList<byte> message, IReadOnlyList<byte> signature)
    }
    class ICryptoBackend{
      Sign(HashDigest<T>, IPrivateKey)
      Verify(HashDigetst<T>, byte[], IPublicKey)
    }
    class DefaultCryptoBackend{
      Sign(HashDigest<T>, IPrivateKey)
      Verify(HashDigetst<T>, byte[], IPublicKey)
    }
    class static CryptoConfig{
      -static ICryptoBackend<SHA256> _cryptoBackend
      -static BlsCryptoBackend<SHA256> _consensusCryptoBackend
      +static ICryptoBackend<SHA256> CryptoBackend
      +static BlsCryptoBackend<SHA256> ConsensusCryptoBackend
    }
```

## Use cases
This BLS digital signature will be used in during the PBFT consensus for verifying multiple votes from nodes. There will be another key to store and manage.

```mermaid
sequenceDiagram
    ConsensusReactor->>ConsensusContext: Gets a propose message
    ConsensusContext->>+Context: Validates the propose
    Context->>+BlsPublicKey.Verify: Verifies the proposer's signature and its public key
    BlsPublicKey.Verify->>-Context: Returns the validity
    Context->>+ValidateNextBlock: Validates the proposed block
    ValidateNextBlock->>-Context: Returns the validity
    Context->>Context: Votes the block by result
    Context->>+Vote.Sign: Signs the vote
    Vote.Sign->>+BlsPrivateKey.Sign: Signs the vote
    BlsPrivateKey.Sign->>-Vote.Sign: Returns a signature
    Vote.Sign->>-Context: Returns a signature
    Context->>-ConsensusContext: Requests to send an vote message
    ConsensusContext->>ConsensusReactor: Sends the vote message
```

The validation of the block will be the aggregation of `LastCommit`. The same procedure can be done with the above (the verification of votes).

- Proposer side
```mermaid 
sequenceDiagram
    ConsensusReactor->>+ProposeBlock: Proposes a new block with the last commits
    loop until hits the last index of LastCommits
    ProposeBlock->>+BlsPublicKey.Verify: Validates a vote and its signature
    BlsPublicKey.Verify->>-ProposeBlock: Adds to lastCommit if the vote and signature of the vote are okay
    end
    ProposeBlock->>-ConsensusReactor: Throw if valid LastCommits are not +2/3 or return a Block
```
- Non-Proposer side
```mermaid
sequenceDiagram
    ConsensusReactor->>ConsensusContext: Gets a propose message
    ConsensusContext->>+Context: Validates the propose
    Context->>+ValidateNextBlock: Validates the block
    ValidateNextBlock->>ValidateNextBlock: Gets the LastCommit from BlockMetadata
    loop until hits the last index of LastCommit
    ValidateNextBlock->>+BlsSignature.Aggregate: Aggregates the signatures
    BlsSignature.Aggregate->>-ValidateNextBlock: Returns an aggregated signature
    end
    ValidateNextBlock->>+BlsSignature.AggregateVerify: Validates the aggregated signature and public keys
    BlsSignature.AggregateVerify->>-ValidateNextBlock: Returns true if the signature and public keys are valid
    ValidateNextBlock->>-Context: Returns the validity
    Context->>-ConsensusContext: Votes the block by result
    ConsensusContext->>ConsensusReactor: Sends the vote message
```

In a bonding process of the validator, the public key should be validated with Proof of Possession[^4] created from its private key, for preventing Rouge Key Attack, which is simply saying, publishing the aggregated public key and not using the true public key of one's private key.

## Further discussion
 - There is an ongoing discussion of forbidding zero value private key and public key (See https://github.com/supranational/blst/issues/11) due to the soundness and producing the infinite public key and signature. For now, This is allowed for the sake of standardization (IETF standard has no mention of blocking neither zero-value private key nor public key.)

 - The behavior of the Infinite Public key and Signature should be further investigated. For more details, See also [Security of BLS batch verification](https://ethresear.ch/t/security-of-bls-batch-verification/10748).

- The Proof of Possession is for validating whether the public key is not aggregated and proving there is a pair of private key. It should be enough checking once and that moment is when the non-validator peer is trying to bonding as a validator.

 - The byte size of `Vote` and `LastCommit` will be increased after the replacement of the Public Key and Signature to BLS. 

 - This PR does not cover `ProtectedPrivateKey`, `Web3KeyStore`, `Address` and `AddressExtensions` modification.

The following discussion should be concerned and the thread is made at https://github.com/planetarium/libplanet/discussions/2247.

[^1]: https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#section-2.1-2.2.1
[^2]: https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-fastaggregateverify
[^3]: https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-aggregateverify
[^4]: https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-proof-of-possession


greymistcube:
This needs to be rebased. 😗

greymistcube:
> There is an ongoing discussion of forbidding zero value private key and public key (See https://github.com/supranational/blst/issues/11) due to the soundness and producing the infinite public key and signature. For now, This is allowed for the sake of standardization (IETF standard has no mention of blocking neither zero-value private key nor public key.)

Eh, seems like newer IETF drafts such as [4] and [5] do not allow 0 SK. 🙄

[4]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04
[5]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05

Libplanet/Crypto/PublicKey.cs
```diff
@@ -52,6 +52,10 @@ internal PublicKey(ECPublicKeyParameters keyParam)
             KeyParam = keyParam;
         }
 
+        public IReadOnlyList<byte> KeyBytes => Format(false);
+
+        ECPublicKeyParameters IPublicKey.KeyParam => KeyParam;
+
         internal ECPublicKeyParameters KeyParam { get; }
```
greymistcube, line 59: I think this can be made public without redirection of interface method above.

Libplanet/Crypto/BlsPublicKey.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            _publicKey = publicKey;
+
+            _ = CryptoConfig.ConsensusCryptoBackend.ValidateGetNativePublicKey(this);
+        }
+
+        public IReadOnlyList<byte> KeyBytes => ToImmutableArray();
+
+        public ECPublicKeyParameters KeyParam => throw new NotSupportedException(
+            "Currently setting key with BLS key parameter is not supported.");
```
greymistcube, line 71: Setting may be disallowed, but what about accessing it?

Libplanet/Crypto/BlsPrivateKey.cs
```diff
@@ -0,0 +1,280 @@
...
+            byte[] bytes = ByteUtil.ParseHex(hex);
+            if (bytes.Length != KeyByteSize)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(hex),
+                    $"Expected {KeyByteSize * 2} hexadecimal digits."
+                );
+            }
+
+            return bytes;
```
greymistcube, line 277:
```suggestion
            return bytes.Length == KeyByteSize
                ? bytes
                : throw new ArgumentOutOfRangeException(
                    nameof(hex),
                    $"Expected {KeyByteSize * 2} hexadecimal digits.");
```

Libplanet/Crypto/IPrivateKey.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// The <see cref="byte"/> representation of private key.
+        /// </summary>
+        IReadOnlyList<byte> KeyBytes { get; }
+
+        /// <summary>
+        /// The key parameter, <see cref="ECPrivateKeyParameters"/> of private key.
+        /// </summary>
+        ECPrivateKeyParameters KeyParam { get; }
```
greymistcube, line 24: As discussed during daily, if this property can't be shared by implementations, I recommend removing it from the interface. 😗

Libplanet/Crypto/BlsSignature.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(signature),
+                    $"The key must be {KeyByteSize} bytes."
+                );
+            }
+
+            _signature = signature;
+            _ = CryptoConfig.ConsensusCryptoBackend.ValidateGetNativeSignature(ToByteArray());
+        }
```
greymistcube, line 70: There is some difference between `IReadOnly` and `ImmutableArray` / `ImmutableList`. Either this should accept immutable type or copy `signature` passed in as argument. Also I think `_signature` should also be an immutable type.

Libplanet/Consensus/BlsVote.cs
```diff
@@ -0,0 +1,229 @@
...
+                   Validator.Equals(other.Validator) &&
+                   Flag == other.Flag &&
+                   Nullable.Equals(Signature, other.Signature);
+        }
+
+        /// <inheritdoc />
+        public override bool Equals(object? obj)
+        {
+            return obj is Vote other && Equals(other);
+        }
```
greymistcube, line 215:
```suggestion
        public override bool Equals(object? obj) => obj is Vote other && Equals(other);
```

Libplanet/Crypto/BlsSignatureExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// messages are identical.
+        /// </summary>
+        /// <param name="signatures">The signatures to aggregate.</param>
+        /// <returns>Returns an aggregated signature of given signatures.</returns>
+        /// <remarks>Note that if any given signature is <b>infinite</b> value, then the
+        /// aggregated signature will be the aggregated signature without <b>infinite</b> signature.
+        /// This should be considered before aggregates the signature.</remarks>
+        [Pure]
+        public static BlsSignature AggregateAll(this BlsSignature[] signatures) =>
+            signatures.Aggregate((x, y) => x.AggregateSignature(y));
```
greymistcube, line 28: This would depend on its usage, but there may be a better suited enumerable type instead of `[]`.

Libplanet/Crypto/IPrivateKey.cs
colibrishin: Amended at https://github.com/planetarium/libplanet/pull/2223/commits/95eeeb2124d08a54c32b03c490b5cc98723c29fe!

Libplanet/Consensus/BlsVote.cs
colibrishin: Amended at https://github.com/planetarium/libplanet/pull/2223/commits/1cf6cf8a2886cc45a72567ebcd6e956ed63d7a46!

Libplanet/Crypto/BlsSignatureExtensions.cs
colibrishin: https://github.com/planetarium/libplanet/blob/bcf03ace27e8621d2f8c2a4603e230d2928371c8/Libplanet/Crypto/BlsSignatureExtensions.cs#L10-L11

How about this?

greymistcube: Depends on usage. Usually enforcing an immutable type is safer but with worse performance. I don't know enough about the exact context in which this will be used, but from the looks of it, this might get called on a "pool" of signatures that can change often. The `IEnumerable` signature signals that this method should be used synchronously or with some sort of locking mechanism from the caller.

colibrishin:
> > There is an ongoing discussion of forbidding zero value private key and public key (See [supranational/blst#11](https://github.com/supranational/blst/issues/11)) due to the soundness and producing the infinite public key and signature. For now, This is allowed for the sake of standardization (IETF standard has no mention of blocking neither zero-value private key nor public key.)
> 
> Eh, seems like newer IETF drafts such as [4](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04) and [5](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05) do not allow 0 SK. 🙄

>  A non-identity point is required because the identity public key has the property that the corresponding secret key is equal to zero, which means that the identity point is the unique valid signature for every message under this key. A malicious signer could take advantage of this fact to equivocate about which message he signed. While non-equivocation is not a required property for a signature scheme, equivocation is infeasible for BLS signatures under any nonzero secret key because it would require finding colliding inputs to the hash_to_point function, which is assumed to be collision resistant.  Prohibiting SK == 0 eliminates the exceptional case, which may help to prevent equivocation-related security issues in protocols that use BLS signatures.[^1]

Thanks for catching this out. I have added zero value check in `PrivateKey(IReadOnlyList<byte>)` constructor. https://github.com/planetarium/libplanet/pull/2223/commits/883c431cab849326d45602a2e0c828982f4be675

[^1]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-5.2

greymistcube:
Also, I'm not if it is exactly relevant, but there was [this] a while back. 🙄

[this]: https://www.cryptomathic.com/news-events/blog/explaining-the-java-ecdsa-critical-vulnerability

Libplanet/Crypto/BlsSignatureExtensions.cs
colibrishin: I think this should be okay. The `BlsSignature`s are converted to a native signature type, which is a separate object with `BlsSignature` and `BlsSignature` is only used for a 'referencing'.

https://github.com/planetarium/libplanet/blob/7080503a74570ae1774297eacf35d554d8896f65/Libplanet/Crypto/BlsCryptoBackend.cs#L117-L138

Libplanet/Crypto/BlsPrivateKey.cs
colibrishin: Could this change be postponable? I'd like to keep the code style with `PrivateKey`.

Libplanet/Crypto/BlsSignature.cs
colibrishin: I've changed the type of `_signature` and other `_privatekey` and `_publickey` to `ImmutableArray<byte>`, and also `IPublic/PrivateKey.KeyBytes`

Libplanet/Crypto/BlsSignatureExtensions.cs
colibrishin: - I think this can be used in asynchronous and synchronous in both situation. `BlsSignature` is immutable. The `AggregateSignature` internally creates new two native signatures from BLS library, and aggregates signatures, which is separate object from given `BlsSignature`s, and the results is packed into new `BlsSignature` and returns it.

- Does _`IEnumerable` is signaling the method should be used synchronously_ mean that LINQ does not support asynchronous operation?

limebell (approved):

PR #2227, opened 2022-08-19
GraphQL APIs for native tokens

dahlia:
List of native tokens:

```graphql
query {
  policy {
    nativeTokens {
      ticker
      decimalPlaces
      minters
      maximumSupply { quantity }
      totalSupplyTrackable
      hash
    }
  }
}
```

Total supplies:

```graphql
query {
  stateQuery {
    totalSupply(
      currencyHash: "0000000000000000000000000000000000000000",
      offsetBlockHash: "0000000000000000000000000000000000000000000000000000000000000000"
    ) {
      currency {
        ticker
        hash
      }
      quantity
    }
  }
}
```

Someone's balance:

```graphql
query {
  stateQuery {
    balance(
      owner: "0000000000000000000000000000000000000000",
      currencyHash: "0000000000000000000000000000000000000000",
      offsetBlockHash: "0000000000000000000000000000000000000000000000000000000000000000"
    ) {
      currency {
        ticker
        hash
      }
      quantity
    }
  }
}
```

### DIFF
--- a/Libplanet.Explorer.Tests/Queries/StateQueryTest.cs
+++ b/Libplanet.Explorer.Tests/Queries/StateQueryTest.cs
@@ -75,6 +75,80 @@ public class StateQueryTest
         Assert.Contains("native token", result.Errors[0].Message);
     }
 
+    [Fact]
+    public async Task TotalSupply()
+    {
+         var currency = Currency.Uncapped("ABC", 2, minters: null);
+#pragma warning disable CS0612  // LegacyUntracked, which is obsolete, is the only way to test this:
+         var legacyToken = Currency.LegacyUntracked("LEG", 0, null);
+#pragma warning restore CS0612
+         (IBlockChainStates<NullAction>, IBlockPolicy<NullAction>) source = (
+            new MockChainStates<NullAction>(),
+            new BlockPolicy<NullAction>(
+                nativeTokens: ImmutableHashSet.Create(currency, legacyToken)
+            )
+        );
+        ExecutionResult result = await ExecuteQueryAsync<StateQuery<NullAction>>(@"
+        {
+            totalSupply(
+                 currencyHash: ""84ba810ca5ac342c122eb7ef455939a8a05d1d40"",
+                 offsetBlockHash:
+                     ""01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b""
+            ) {
+                currency { ticker, hash }
+                sign
+                majorUnit
+                minorUnit
+                quantity
+                string
+            }
+        }
+        ", source: source);
+        Assert.Null(result.Errors);
+        ExecutionNode resultData = Assert.IsAssignableFrom<ExecutionNode>(result.Data);
+        IDictionary<string, object> resultDict =
+            Assert.IsAssignableFrom<IDictionary<string, object>>(resultData!.ToValue());
+        IDictionary<string, object> totalSupplyDict =
+            Assert.IsAssignableFrom<IDictionary<string, object>>(resultDict["totalSupply"]);
+        IDictionary<string, object> currencyDict =
+            Assert.IsAssignableFrom<IDictionary<string, object>>(totalSupplyDict["currency"]);
+        Assert.Equal("ABC", currencyDict["ticker"]);
+        Assert.Equal("84ba810ca5ac342c122eb7ef455939a8a05d1d40", currencyDict["hash"]);
+        Assert.Equal(1, Assert.IsAssignableFrom<int>(totalSupplyDict["sign"]));
+        Assert.Equal(10000, Assert.IsAssignableFrom<BigInteger>(totalSupplyDict["majorUnit"]));
+        Assert.Equal(0, Assert.IsAssignableFrom<BigInteger>(totalSupplyDict["minorUnit"]));
+        Assert.Equal("10000", totalSupplyDict["quantity"]);
+        Assert.Equal("10000 ABC", totalSupplyDict["string"]);
+
+        result = await ExecuteQueryAsync<StateQuery<NullAction>>(@"
+        {
+            totalSupply(
+                 currencyHash: ""0000000000000000000000000000000000000000"",
+                 offsetBlockHash:
+                     ""01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b""
+            ) {
+                quantity
+            }
+        }
+        ", source: source);
+        Assert.Single(result.Errors);
+        Assert.Contains("native token", result.Errors[0].Message);
+
+        result = await ExecuteQueryAsync<StateQuery<NullAction>>(@"
+        {
+            totalSupply(
+                 currencyHash: ""d9d45abf192b6c337027063cda2d4a61e76e66d8"",
+                 offsetBlockHash:
+                     ""01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b""
+            ) {
+                quantity
+            }
+        }
+        ", source: source);
+        Assert.Single(result.Errors);
+        Assert.Contains("not trackable", result.Errors[0].Message);
+    }
+
     private class MockChainStates<T> : IBlockChainStates<T>
         where T : IAction, new()
     {
--- a/Libplanet.Explorer.Tests/Queries/StateQueryTest.cs
+++ b/Libplanet.Explorer.Tests/Queries/StateQueryTest.cs
@@ -1,0 +1,103 @@
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Numerics;
+using System.Threading.Tasks;
+using Bencodex.Types;
+using GraphQL;
+using GraphQL.Execution;
+using Libplanet.Action;
+using Libplanet.Assets;
+using Libplanet.Blockchain;
+using Libplanet.Blockchain.Policies;
+using Libplanet.Blocks;
+using Libplanet.Explorer.Queries;
+using Xunit;
+using static Libplanet.Explorer.Tests.GraphQLTestUtils;
+
+namespace Libplanet.Explorer.Tests.Queries;
+
+public class StateQueryTest
+{
+    [Fact]
+    public async Task Balance()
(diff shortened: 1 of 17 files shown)


dahlia:
@moreal Could you review this?

Libplanet.Explorer/Queries/StateQuery.cs
```diff
@@ -0,0 +1,126 @@
...
+                new QueryArgument<NonNullGraphType<ByteStringType>> { Name = "currencyHash" },
+                new QueryArgument<NonNullGraphType<IdGraphType>> { Name = "offsetBlockHash" }
+            ),
+            resolve: ResolveTotalSupply
+        );
+    }
+
+    private static Currency GetNativeTokenFromHash(
+        IResolveFieldContext<(IBlockChainStates<T> ChainStates, IBlockPolicy<T> Policy)> context,
+        string name
```
tkiapril, line 44: Would it matter for this parameter to have a more descriptive name?

Libplanet.Explorer/Queries/StateQuery.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        byte[] currencyHashBytes = context.GetArgument<byte[]>(name);
+        HashDigest<SHA1> currencyHash;
+        try
+        {
+            currencyHash = new HashDigest<SHA1>(currencyHashBytes);
+        }
+        catch (Exception e)
+        {
+            throw new ExecutionError("currencyHash must consist of 20 bytes.\n" + e.Message, e);
```
tkiapril, line 55: I'm wondering if the value of `name` is not `currencyHash`, would it be better to have this message to include the actual value of `name`.

Libplanet.Explorer/Queries/StateQuery.cs
```diff
@@ -0,0 +1,126 @@
...
+        IImmutableSet<Currency> nativeTokens = context.Source.Policy.NativeTokens;
+        try
+        {
+            return nativeTokens.First(token => token.Hash.Equals(currencyHash));
+        }
+        catch (InvalidOperationException e)
+        {
+            throw new ExecutionError(
+                "currencyHash must be a native token.  Native tokens are:\n\n" +
+                    string.Join("\n", nativeTokens.Select(token => token.Hash.ToString())),
```
tkiapril, line 67:
```suggestion
                "currencyHash must be a hash of a native token.  Native tokens are:\n\n" +
                    string.Join("\n", nativeTokens.Select(token => token.Ticker + ": " + token.Hash.ToString())),
```

It might be better to have the tickers of the tokens printed along with the hash. Also, maybe consider the case of `name` having different value other than `currencyHash`.

Libplanet.Explorer.Tests/Queries/StateQueryTest.cs
```diff
@@ -0,0 +1,158 @@
...
+        IDictionary<string, object> currencyDict =
+            Assert.IsAssignableFrom<IDictionary<string, object>>(totalSupplyDict["currency"]);
+        Assert.Equal("ABC", currencyDict["ticker"]);
+        Assert.Equal("84ba810ca5ac342c122eb7ef455939a8a05d1d40", currencyDict["hash"]);
+        Assert.Equal(1, Assert.IsAssignableFrom<int>(totalSupplyDict["sign"]));
+        Assert.Equal(10000, Assert.IsAssignableFrom<BigInteger>(totalSupplyDict["majorUnit"]));
+        Assert.Equal(0, Assert.IsAssignableFrom<BigInteger>(totalSupplyDict["minorUnit"]));
+        Assert.Equal("10000", totalSupplyDict["quantity"]);
+        Assert.Equal("10000 ABC", totalSupplyDict["string"]);
+
```
tkiapril, line 117: Maybe a test case with token.TotalSupplyTrackable = false (Currency.Legacy())?

Libplanet.Explorer/Queries/StateQuery.cs
dahlia: I'm going to rename it to `paramName`.

Libplanet.Explorer/Queries/StateQuery.cs
dahlia: Right.  I'm going to remove the hard-coded parameter name.

Libplanet.Explorer/Queries/StateQuery.cs
dahlia: Amended.

Libplanet.Explorer/Queries/StateQuery.cs
dahlia: Amended.

Libplanet.Explorer/Queries/StateQuery.cs
dahlia: This message now shows tickers too.  Amended the last commit.

Libplanet.Explorer.Tests/Queries/StateQueryTest.cs
dahlia: It now handles the case, and I added a test for it too.  Amended the last commit.

tkiapril (approved):

moreal (commented):
I'm not expert of the new `Currency` type (maybe), I left a review about GraphQL.NET.

Libplanet.Explorer/GraphTypes/CurrencyType.cs
```diff
@@ -0,0 +1,42 @@
...
+            "ticker",
+            "The ticker symbol, e.g., USD.",
+            resolve: context => context.Source.Ticker
+        );
+        Field<NonNullGraphType<UIntGraphType>>(
+            "decimalPlaces",
+            "The number of digits to treat as minor units (i.e., exponents).",
+            resolve: context => (uint)context.Source.DecimalPlaces
+        );
+        Field<ListGraphType<AddressType>>(
```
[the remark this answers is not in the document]
dahlia: Amended!

Libplanet.Explorer/GraphTypes/PolicyType.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Libplanet.Explorer.GraphTypes;
+
+public class PolicyType<T> : ObjectGraphType<IBlockPolicy<T>>
+    where T : IAction, new()
+{
+    public PolicyType()
+    {
+        Name = "Policy";
+        Field<NonNullGraphType<ListGraphType<CurrencyType>>>(
```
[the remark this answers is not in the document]
dahlia: Amended!

dahlia:
@tkiapril @moreal Could you take a look again?

tkiapril (dismissed):

Libplanet.Explorer/GraphTypes/PolicyType.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Linq;
+using GraphQL.Types;
+using Libplanet.Action;
+using Libplanet.Blockchain.Policies;
+
+namespace Libplanet.Explorer.GraphTypes;
+
+public class PolicyType<T> : ObjectGraphType<IBlockPolicy<T>>
```
longfin, line 9: "Policy" seems to mean a little bit broad. how about `BlockPolicy` (from `IBlockPolicy`), or `ChainPolicy` (newly introduced, but more specific).

longfin (dismissed):
I left few minor comments 🙏

moreal (approved):

longfin:
@dahlia it becomes to have conflicts now. 😢

Libplanet.Explorer/GraphTypes/PolicyType.cs
dahlia: Renamed in 3379ec73078f9994caa7a3c27798b13474cfca9a.

longfin (dismissed):

tkiapril (commented):
I guess you might want to rename [Libplanet.Explorer.Tests/GraphTypes/PolicyTypeTest.cs](https://github.com/planetarium/libplanet/pull/2227#diff-6bd693519367920abe33edc921b1a569a7361c43c4b761eaa8bd6eb7f01c0684) as well?

Libplanet.Explorer/Queries/StateQuery.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+        catch (Exception e)
+        {
+            throw new ExecutionError(paramName + " must consist of 20 bytes.\n" + e.Message, e);
+        }
+
+        IImmutableSet<Currency> nativeTokens = context.Source.Policy.NativeTokens;
+        try
+        {
+            return nativeTokens.First(token => token.Hash.Equals(currencyHash));
```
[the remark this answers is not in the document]
dahlia: It's trivial too, but I concern performance trade-offs between those two methods.  🤔 

AFAIK `Enumerable.Single()` does *not* immediately terminate even if it finds a satisfied element early, because it needs to ensure if there are no more satisfied items in the sequence.  In theory, it terminates the earliest (even though it would throws an exception) when two satisfied items appear in a row, e.g., `{ true, true, false, false, false }` as it needs to only two items.  However, if the sequence only one satisfied item and it's appeared early, `Enumerable.Single()` superfluously consumes the whole sequence, e.g., `{ true, false, false, false, false }`, even though it returns the satisfied item without any exception.  😞

On the other hand, `Enumerable.First()` always terminates the earliest possible immediately after it finds the first satisfied item.  If it consumes the whole sequence it's the case where the sequence has no satisfied item at all.

Therefore, in general, `Enumerable.Single()` has a performance penalty compared to `First()`.  If we should use `Enumerable.Single()` instead of `First()` it's the case we need to ensure only one item in the input sequence satisfies the predicate.  However, it's not the case here, because the input enumerable is already a set which guarantees contained `Currency` values does not overlap their `Hash`.

dahlia:
@tkiapril I amended the last commit!

tkiapril (approved):

dahlia:
@tkiapril @moreal @longfin Could you take a look again?

moreal (approved):

PR #2231, opened 2022-08-21
Remove suppression: remove the unused local variable 'x' (#664)

sindydwns:
resolved
- #664 
I removes the sonar analyzer suppression (remove unused local variable 'x') and resolves build errors that result from it.

### DIFF
--- a/Libplanet.Net/BlockCompletion.cs
+++ b/Libplanet.Net/BlockCompletion.cs
@@ -438,7 +438,7 @@ namespace Libplanet.Net
                     if (tasks.Any())
                     {
                         var tcs = new TaskCompletionSource<object>();
-                        using CancellationTokenRegistration ctr = cancellationToken.Register(
+                        cancellationToken.Register(
                             () => tcs.TrySetCanceled(),
                             useSynchronizationContext: false
                         );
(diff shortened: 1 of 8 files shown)


OnedgeLee (approved):

OnedgeLee (approved):

Libplanet.Net/BlockCompletion.cs
```diff
@@ -438,7 +438,7 @@ public async Task WaitAll(CancellationToken cancellationToken = default)
                     if (tasks.Any())
                     {
                         var tcs = new TaskCompletionSource<object>();
-                        using CancellationTokenRegistration ctr = cancellationToken.Register(
+                        cancellationToken.Register(
```
longfin, line 441: Declaring `ctr` (even no usage) seemed be intended due to auto disposing. 🤔 (see also: https://github.com/SonarSource/sonar-dotnet/issues/3137)

so I think we may need to keep previous with new comments to suppress lint.
cc @planetarium/libplanet

sindydwns: ok thanks. i will check it!

dahlia: As Sonar Analyzer does not provide any way to discard this pattern of usage yet, we should manually suppress the warning at this moment, like:

```csharp
#pragma warning disable S1481  // Though var _ is unused, the using statement is still needed
using CancellationTokenRegistration _ = cancellationToken.Register(
    () => tcs.TrySetCanceled(),
    useSynchronizationContext: false
);
#pragma warning restore S1481
```

dahlia (changes requested):
Could you update *Libplanet.ruleset* and *Libplanet.Tests.ruleset* as well?

Also, you need to add `[changelog skip]` to your commit message.

PR #2234, opened 2022-08-23
WIP : Implement base `PoS` functionalities with mockup actions

OnedgeLee:
- [ ] Validator creation
- [ ] Delegation
- [ ] Undelegation
- [ ] Redelegation
- [ ] Validator set selection
- [ ] Distribution


Libplanet/PoS/DelegationInfo.cs
```diff
@@ -0,0 +1,52 @@
+using System;
+using Bencodex.Types;
+
+namespace Libplanet.PoS
+{
+    [Serializable]
+    public class DelegationInfo
```
riemannulus, line 7: Is `XXXInfo` mean inner data of 'XXX'? 

IMO If the place with the subject of data does not use a name that represents the data, and the place that uses the data takes the name that represents the data, I think ambiguity about the name and behavior arises.

In this case, my suggestion is `Delegation` to `Delegate`, `DelegationInfo` to `Delegation`. Because the current `Delegation` class is a behavior class. so the verb is way more representative of this feature. not noun.

Libplanet/PoS/Delegation.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        internal Delegation(DelegationInfo delegationInfo)
+            : base(delegationInfo)
+        {
+        }
+
+        internal static (IAccountStateDelta, FungibleAssetValue) Bond(
+            IAccountStateDelta states,
+            FungibleAssetValue consensusToken,
```
riemannulus, line 27: IMO this method is not good at here. I suggest making a new file for this method.
Because `Bond` is not only used `Delegation.Delegate`, but also `Redelegaion.Redelegate`.
If you think `Delegation` is an inner concept about `Redelegation`, The design needs to be changed to better reflect that concept.

Libplanet/PoS/UnbondingSet.cs
```diff
@@ -0,0 +1,112 @@
+using System;
+using Bencodex.Types;
+using Libplanet.Action;
+
+namespace Libplanet.PoS
+{
+    internal class UnbondingSet : UnbondingSetInfo
```
riemannulus, line 7: How about making static this class and method? This class and methods do not have captured state.

riemannulus (commented):
First of all, I commented on the implementation method and concept of the big strand. After the revision, I will leave a more detailed implementation review.

Libplanet/PoS/DelegationInfo.cs
riemannulus: Additionally, this review is not limited to this but applies to most classes on a similar level.

OnedgeLee: > Is XXXInfo mean inner data of 'XXX'?

=> Right.
> In this case, my suggestion is Delegation to Delegate, DelegationInfo to Delegation. Because the current Delegation class is a behavior class. so the verb is way more representative of this feature. not noun.

=> Okay, I'll reflect your suggestion.

Libplanet/PoS/Delegation.cs
OnedgeLee: I agree that this structure doesn't seem so good.
But I don't think `Delegation` is an inner concept about `Redelegation`.
`Redelegation` uses methods `Delegation.Bond` and `Undelegation.Unbond`, so it's between those.
I think there are some better structure, but let me think about it more.

Libplanet/PoS/UnbondingSet.cs
OnedgeLee: For example, below method removes item from RedelegationAddressSet.
```c#
        internal IAccountStateDelta DelRedelegationAddressSet(
            IAccountStateDelta states, Address redelegationEntryAddress)
        {
            RedelegationAddressSet.Remove(redelegationEntryAddress);
            states = states.SetState(Address, Serialize());
            return states;
        }
```

riemannulus: Oh, yeah, my bad. I mean, if `RedelegateionAddressSet` makes a method property too, It can be a static method.

OnedgeLee: > if RedelegateionAddressSet makes a method property too, It can be a static method.
I think I'm not understood. Could you explain a bit more?

OnedgeLee:
Revised #1

Libplanet/PoS/Control/Bond.cs
```diff
@@ -0,0 +1,97 @@
...
+            Address delegationAddress)
+        {
+            // TODO: Failure condition
+            // 1. Validator does not exist
+            // 2. Exchange rate is invalid(validator has no tokens but there are outstanding shares)
+            // 3. Amount is less than the minimum amount
+            // 4. Delegator does not have sufficient consensus token (fail or apply maximum)
+            if (!consensusToken.Currency.Equals(Asset.ConsensusToken))
+            {
+                throw new ArgumentException();
```
riemannulus, line 22: IMO the exception was a bit more descriptive. like `InvalidBondCurrencyException(actual, expected)`.

Libplanet/PoS/Control/Bond.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            Validator validator = ValidatorCtrl.GetValidatorByAddr(states, validatorAddress);
+
+            // If validator share is zero, exchange rate is 1
+            // Else, exchange rate is validator share / token
+            FungibleAssetValue issuedShare
+                = ValidatorCtrl.ShareFromConsensusToken(states, validator.Address, consensusToken)
+                ?? throw new InvalidOperationException();
```
riemannulus, line 31: `null` return is code that may cause unintended behavior. How about throwing an exception in the `ValidatorCtrl.ShareFromConsensusToken`?

riemannulus, line 31: Plus, IMO this also needs a bit more descriptive...

Libplanet/PoS/Control/ValidatorCtrl.cs
```diff
@@ -0,0 +1,205 @@
...
+            Validator validator;
+            if (create && serializedValidator == null)
+            {
+                validator = new Validator(operatorAddress);
+                states = states.SetState(validator.Address, validator.Serialize());
+            }
+            else
+            {
+                validator = GetValidatorByAddr(states, validatorAddress);
+            }
```
riemannulus, line 41: How about this?

```suggestion
            Validator validator;
            if (states.GetState(validatorAddress) is { } value)
            {
                validator = new Validator(value);
            }
            else if(create)
            {
                validator = new Validator(operatorAddress);
                states = states.SetState(validator.Address, validator.Serialize());
            }
```

Libplanet/PoS/Control/ValidatorCtrl.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            IValue? serializedValidator = states.GetState(validatorAddress);
+            if (serializedValidator == null)
+            {
+                throw new InvalidOperationException();
+            }
+
+            Validator validator = new Validator(serializedValidator);
+            return validator;
+        }
```
riemannulus, line 22: How about something like this to keep the behavior similar to `IAccountStateDelta.GetState()`?
```suggestion
        internal static Validator? GetValidatorByAddr(
            IAccountStateDelta states,
            Address validatorAddress)
        {
            if (states.GetState(validatorAddress) is { } value)
            {
                return new Validator(value);
            }
            return null;
        }
```

Libplanet/PoS/Misc/SystemAction.cs
```diff
@@ -0,0 +1,320 @@
...
+    {
+        protected SystemAction()
+        {
+            Id = Guid.NewGuid();
+        }
+
+        public Guid Id { get; private set; }
+
+        public IValue PlainValue =>
+#pragma warning disable LAA1002
```
riemannulus, line 27: What is `LAA1002`? and have any reason not to follow the warning and ignore it?

Libplanet/PoS/Misc/SystemAction.cs
```diff
@@ -0,0 +1,320 @@
+using System;
```
riemannulus, line 1: It seems some refactor is needed... in the next PR.

Libplanet/PoS/Util/AddressComparer.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public int Compare(Address x, Address y)
+        {
+            int result
+                = (new BigInteger(x.ToByteArray())
+                - new BigInteger(y.ToByteArray())).Sign;
+
+            return -result;
+        }
+    }
```
riemannulus, line 16: How about using [partial class](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods)?

riemannulus (commented):
I cannot review the following things:

- [ ] Test code is pretty good cover basic scenario.
- [ ] Action logics are robust.
I'll review it next time about that. 🙏

Libplanet/PoS/Control/DelegateCtrl.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        internal static IAccountStateDelta Execute(
+           IAccountStateDelta states,
+           Address delegatorAddress,
+           Address validatorAddress,
+           FungibleAssetValue governanceToken)
+        {
+            if (!governanceToken.Currency.Equals(Asset.GovernanceToken))
+            {
+                throw new ArgumentException();
```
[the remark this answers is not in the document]
OnedgeLee: Yeah, I've left all of exceptions not concrete. These have to be handled with custom exceptions.

Libplanet/PoS/Util/AddressComparer.cs
OnedgeLee: Not clearly understood. You mean make class nested?

Libplanet/PoS/Control/DelegateCtrl.cs
```diff
@@ -0,0 +1,84 @@
...
+            Delegation delegation;
+            if (create && serializedDelegation == null)
+            {
+                delegation = new Delegation(delegatorAddress, validatorAddress);
+                states = states.SetState(delegation.Address, delegation.Serialize());
+            }
+            else
+            {
+                delegation = GetDelegationByAddr(states, delegationAddress);
+            }
```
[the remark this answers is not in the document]
OnedgeLee: I think it's a way better, but let me think about just keeping `GetDelegationByAddr` again.

Libplanet/PoS/Control/Bond.cs
OnedgeLee: I've made it like it because I thought there can be some cases we need to know if `Validator` does not have `ConsensusToken`, and use this information.
Actually, I've thought exception is just bound to `bond` logic, not `ShareFromConsensusToken` converting logic.
If we throw error on `ShareFromConsensusToken`, we can handle our logic with try-catch on above condition, but I thought handling with null would be better like `getState`.

Libplanet/PoS/Control/ValidatorCtrl.cs
OnedgeLee: Same as above, I prefer this one.

Libplanet/PoS/Control/ValidatorCtrl.cs
OnedgeLee: Seems better

Libplanet/PoS/Control/Bond.cs
riemannulus: Then how about this?
```suggestion
            if (!(ValidatorCtrl
                    .ShareFromConsensusToken(
                        states,
                        validator.Address,
                        consensusToken) is { } issuedShare))
            {
                throw new InvalidOperationException();
            }

```

riemannulus: And IMO this is an anti-pattern using the same type as `ShareToken` and `ConsensusToken`. It may cause confuse 😵‍💫

Libplanet/PoS/Util/AddressComparer.cs
riemannulus: NVM, I'm confused 😵‍💫

Libplanet/PoS/Control/Bond.cs
OnedgeLee: Seems nice, but could you explain the difference of those?
I've seen so many lines written like that, so I think it's better and robust way,
but don't know why.

OnedgeLee: Hmm, do we have to inherit FAV? actually, It's a bit confusing to me also, to use different FAVs in just one type. (not just `Share` and `ConsensusToken`, but also `GovernanceToken`)...But not sure there are better way to handle.

riemannulus: verbally explained.

Libplanet/PoS/Control/Bond.cs
OnedgeLee: Applied on commit 8045f68c000f18f2e9b8608ae7c2b33ccfc96cd0.

Libplanet/PoS/Control/DelegateCtrl.cs
```diff
@@ -0,0 +1,84 @@
...
+           FungibleAssetValue governanceToken)
+        {
+            if (!governanceToken.Currency.Equals(Asset.GovernanceToken))
+            {
+                throw new ArgumentException();
+            }
+
+            if (governanceToken > states.GetBalance(delegatorAddress, Asset.GovernanceToken))
+            {
+                throw new ArgumentException();
```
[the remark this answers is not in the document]
OnedgeLee: Applied on commit 8045f68c000f18f2e9b8608ae7c2b33ccfc96cd0.

Libplanet/PoS/Control/DelegateCtrl.cs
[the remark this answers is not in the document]
OnedgeLee: Applied on commit 8045f68c000f18f2e9b8608ae7c2b33ccfc96cd0.

Libplanet/PoS/Control/Bond.cs
OnedgeLee: Pattern matching applied on commit 8045f68c000f18f2e9b8608ae7c2b33ccfc96cd0, but Wrapping FAV wasn't applied yet. 
I've extract this issue as separated : https://github.com/planetarium/libplanet/issues/2267

OnedgeLee: Applied on commit 8045f68c000f18f2e9b8608ae7c2b33ccfc96cd0.

CHANGES.md
```diff
@@ -1,6 +1,66 @@
...
+
+### Behavioral changes
+
+### Bug fixes
+
+### Dependencies
+
+### CLI tools
+
+
```
riemannulus, line 63: In PBFT, it was difficult to find the added PR, so it did not specify the added PR, but it seems good to write down the added PR because it is clear in the case of PoS.

`[#2234]: https://github.com/planetarium/libplanet/pull/2234`

CHANGES.md
```diff
@@ -1,6 +1,66 @@
...
+### Backward-incompatible API changes
+
+### Backward-incompatible network protocol changes
+
+### Backward-incompatible storage format changes
+
+### Added APIs
+
+ -  (Libplanet.Pos) Added `Validator` class.  [[#PoS]]
```
riemannulus, line 19:
```suggestion
 -  (Libplanet.Pos) Added `Validator` class.  [#2234]
```

CHANGES.md
OnedgeLee: Applied on 837bbc1d2f041cc29e427fc55a0d270cc288c71e

CHANGES.md
OnedgeLee: Applied on 837bbc1d2f041cc29e427fc55a0d270cc288c71e

riemannulus (approved):

riemannulus:
LGTM!

Issue #2235, opened 2022-08-23
Implement VRF


OnedgeLee:
Verifiable random function implementation is needed for proposer selection.

PR #2237, opened 2022-08-23
Add verifiable random function

OnedgeLee:
### Purpose
Use VRF on proposer election

### Changes
Implements verifiable random function
Proposer election will be handled on another PR, since it's related to PoS work.

### Reference document
https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vrf-03#section-5

### Reference code
https://github.com/google/keytransparency/tree/master/core/crypto/vrf

~~Reference document and code have minor conflict. I've followed reference code.~~
Reference document and code have minor conflict. I've followed spec document

### Related issue
- Resolves https://github.com/planetarium/libplanet/issues/2235


OnedgeLee:
Test sometimes success, but sometimes fail on checksum equality verification.
I've tried to figure out inner variables in process, and found below problem
image

1. kG $\neq$ (s+cd)G
2. kH $\neq$ (s+cd)H
Algorithm assumes they're equal, so may be some kind of misimplementation exists on calculating there.
I suspect that I've forgot to add some modulus operations.

OnedgeLee:
The problem was from encoding & decoding of `c` and `s`.
There was some missing handling `BigInteger`.

OnedgeLee:
@riemannulus 
VRF implementation has been done, starts sortition implementation.
May be, it would be better to move on `PrivateKey` and `PublicKey`,
like `PrivateKey.Evaluate()` and `PublicKey.ProofToHash()`.

PR #2252, opened 2022-08-29
Chains are compared only using index

greymistcube:


Libplanet.Net.Tests/SwarmTest.cs
```diff
@@ -758,63 +758,6 @@ public async Task RenderInFork()
...
-            {
-                await StopAsync(miner1);
-                await StopAsync(miner2);
-                miner1.Dispose();
-                miner2.Dispose();
-            }
-        }
-
         [Fact(Skip = "This should be fixed to work deterministically.")]
         public async Task HandleReorgInSynchronizing()
```
colibrishin, line 762: I think a Reorganization test case is not needed due to its finality.

greymistcube: Probably removed in some later PR, possibly in #2254. 😶

PR #2281, opened 2022-09-01
Using BLS Signature in PBFT voting

colibrishin:
This is the following PR of https://github.com/planetarium/libplanet/pull/2223, the use-case implementation of BLS. 

### Changes
-  `Address`, `AddressExtensions`, and `BoundPeer` are now taking a public key with `IPublicKey` for supporting BLS and pre-existing `PublicKey` secp256k1.
-  `Address` can create an address from BLS. This is not intended to be used in `Transaction` but in `BoundPeer`.
-  For deserialization of public key in `BoundPeer`, static class `PublicKeyGetter` is added and used for determining which public key is given from peer string. The public key is determined by the length of a key.
-  The private key of `Transport` for signing a message is now taking a private key with `IPrivateKey`.
-  `Vote` is now using `BlsPublicKey` and `BlsPrivateKey` for signing and verifying vote messages.
-  Added BLS backend, `BlsCryptoBackend`, which is inherits `ICryptoBackend`, in as `CryptoConfig.BlsCryptoBackend`.

### Notes
-  **In Unity, The Libplanet with BLS** can be built and run but **should be cautiously used depending on IL2CPP**. Using BLS in Unity with IL2CPP[^1] is not available (Using _System.Reflections.Emit_). An action for separating should be done.

### Further discussion
-  This is the out-of-scope of this PR, however, The verification in `BlockCommit` should check whether the **last commit of a block does have +2/3 committed votes**. An arbitrary byzantine could propose a block with full _NULL_ vote last commit and bypass the verification.

[^1]: > Some platforms don’t support AOT compilation, so the IL2CPP backend doesn’t work on every platform. Other platforms support AOT and IL2CPP, but don’t allow JIT compilation, and so can’t support the Mono backend. When a platform can support both backends, Mono is the default. For more information, see [Scripting restrictions](https://docs.unity3d.com/Manual/ScriptingRestrictions.html). ...... An AOT platform cannot implement any of the methods in the System.Reflection.Emit namespace. The rest of System.Reflection is acceptable, as long as the compiler can infer that the code used via reflection needs to exist at runtime.

### DIFF
--- a/Libplanet.Net/BoundPeer.cs
+++ b/Libplanet.Net/BoundPeer.cs
@@ -1,6 +1,8 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Globalization;
+using System.Linq;
 using System.Net;
 using System.Runtime.Serialization;
 using Bencodex.Types;
@@ -21,12 +23,12 @@ namespace Libplanet.Net
         /// <summary>
         /// Initializes a new instance of the <see cref="BoundPeer"/> class.
         /// </summary>
-        /// <param name="publicKey">A <see cref="PublicKey"/> of the
-        /// <see cref="BoundPeer"/>.</param>
+        /// <param name="publicKey">A <see cref="IPublicKey"/> (e.g., <see cref="PublicKey"/>,
+        /// <see cref="BlsPublicKey"/>) of the <see cref="BoundPeer"/>.</param>
         /// <param name="endPoint">A <see cref="DnsEndPoint"/> consisting of the
         /// host and port of the <see cref="BoundPeer"/>.</param>
         public BoundPeer(
-            PublicKey publicKey,
+            IPublicKey publicKey,
             DnsEndPoint endPoint)
             : this(publicKey, endPoint, null)
         {
@@ -35,7 +37,7 @@ namespace Libplanet.Net
 #pragma warning disable SA1118 // The parameter spans multiple lines
         public BoundPeer(Bencodex.Types.Dictionary dictionary)
             : this(
-                new PublicKey(((Binary)dictionary[PublicKeyKey]).ByteArray),
+                PublicKeyGetter.Get(((Binary)dictionary[PublicKeyKey]).ByteArray),
                 new DnsEndPoint(
                     (Text)dictionary[EndPointHostKey], (Integer)dictionary[EndPointPortKey]),
                 dictionary[PublicIpAddressKey] is Text text
@@ -46,7 +48,7 @@ namespace Libplanet.Net
 #pragma warning restore SA1118
 
         internal BoundPeer(
-            PublicKey publicKey,
+            IPublicKey publicKey,
             DnsEndPoint endPoint,
             IPAddress? publicIPAddress)
         {
@@ -57,7 +59,8 @@ namespace Libplanet.Net
 
         private BoundPeer(SerializationInfo info, StreamingContext context)
         {
-            PublicKey = new PublicKey(info.GetValue<byte[]>(nameof(PublicKey)));
+            PublicKey =
+                PublicKeyGetter.Get(info.GetValue<byte[]>(nameof(PublicKey)));
             EndPoint = new DnsEndPoint(
                 info.GetString("end_point_host"),
                 info.GetInt32("end_point_port"));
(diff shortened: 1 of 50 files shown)


Libplanet/Crypto/IECPublicKey.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Libplanet.Crypto
+{
+    /// <summary>
+    /// The interface of Elliptic Curve Public Key. It allows to align the essential methods and
+    /// uses in <see cref="CryptoConfig"/> and <see cref="ICryptoBackend{T}"/>, mainly
+    /// <see cref="Verify"/> and <see cref="KeyBytes"/>. This should be
+    /// inherited if there's a new implementation for Elliptic Curve Public Key.
+    /// </summary>
+    // ReSharper disable once InconsistentNaming
+    public interface IECPublicKey
```
greymistcube, line 13: What's the reason for `IECPublicKey` instead of `IPublicKey`?
I'd say `EC` is too generic of a term for its meaning to be apparent. 🙄

Libplanet.Net/BoundPeer.cs
```diff
@@ -81,15 +81,20 @@ public static BoundPeer ParsePeer(string peerInfo)
             }
 
             string[] tokens = peerInfo.Split(',');
-            if (tokens.Length != 3)
+            if (tokens.Length != 4)
             {
                 throw new ArgumentException(
-                    $"'{peerInfo}', should have format <pubkey>,<host>,<port>",
+                    $"'{peerInfo}', should have format <pubkey>,<pubkeytype>,<host>,<port>",
```
greymistcube, line 87: As this would break the API anyway, I think type prefix might be better, i.e. `<pubkeyType>,<pubkey>,<host>,<port>`.

Libplanet/Crypto/IECPublicKey.cs
colibrishin: Amended at https://github.com/planetarium/libplanet/pull/2223!

Libplanet.Net/BoundPeer.cs
colibrishin: Amended! https://github.com/planetarium/libplanet/pull/2281/commits/3e40c02fca04d742595b0a3cb6e6083e1b806244

Libplanet/Crypto/PublicKeyGetter.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// A static getter class for determine which public key it is by given <see cref="byte"/>
+    /// representation and returns a concrete class type in <see cref="IPublicKey"/>.
+    /// </summary>
+    public static class PublicKeyGetter
+    {
+        /// <summary>
+        /// Gets a concrete class in <see cref="IPublicKey"/> by checking given <see cref="byte"/>
+        /// representation length.
```
limebell, line 14: How about specifying the length and its return type?

colibrishin: https://github.com/planetarium/libplanet/blob/1f83dee55e5ed0f64d72da1c5c620b01b17eb513/Libplanet/Crypto/PublicKeyGetter.cs#L12-L25
Amended!

Libplanet/Crypto/PublicKeyGetter.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Libplanet.Crypto
+{
+    /// <summary>
+    /// A static getter class for determine which public key it is by given <see cref="byte"/>
+    /// representation and returns a concrete class type in <see cref="IPublicKey"/>.
+    /// </summary>
+    public static class PublicKeyGetter
+    {
+        /// <summary>
+        /// <p>Gets a concrete class in <see cref="IPublicKey"/> by checking given <see cref="byte"/>
```
limebell, line 13:
```suggestion
        /// <p>Gets a concrete class in <see cref="IPublicKey"/> by checking given <see cref="byte"/>
```
Line length exceeds 100 characters 😢

limebell (changes requested):

Libplanet/Crypto/PublicKeyGetter.cs
colibrishin: Oops, Amended again :P

limebell (approved):

Issue #2260, opened 2022-08-30
Fix `validatorSet` structure that will be used on PBFT


OnedgeLee:
PBFT will use `validatorSet` written on tip block as consensus validator committee.
So, `validatorSet` have to be saved each block.

PR #2282, opened 2022-09-01
Add `BondedValidatorSet`

OnedgeLee:
### Description
Add `BondedValidatorSet` for PBFT Consensus validators.
Each `OnTipChanged` event, `ValidatorPowerSetCtrl.Update()` have to be called, and it will refresh `BondedValidatorSet`.
By calling `BondedValidatorSetCtrl.FetchBondedValidatorSet`, we can obtain validator set for consensus.
This class contains `ValidatorAddress`, `OperatorPublicKey`, `PowerFrac`.
`(Address)ValidatorAddress` is a address of validator.
`(PublicKey)OperatorPublicKey` is public key of validator operator, and can be used on verifying.
`(double)PowerFrac` is validator consensus token divided by total.

### Changes
- [x] Add `BondedValidatorSet` class
- [x] Add update logic for `BondedValidatorSet` on `ValidatorPowerSetCtrl.Update()`
- [x] Model classes that have many instances inherit `IEquatable`
- [x] `ValidatorPower` class inherits `IComparable`
- [x] Remove `ValidatorPowerComparer` and `AddressComparer` since it's duplicated

### Related issues
- Resolves https://github.com/planetarium/libplanet/issues/2260


Libplanet/PoS/Model/Delegation.cs
```diff
@@ -46,5 +57,32 @@ public IValue Serialize()
...
+            unchecked
+            {
+                foreach (byte b in Address.ToByteArray())
+                {
+                    code = (code * 397) ^ b.GetHashCode();
+                }
+            }
+
+            return code;
+        }
```
riemannulus, line 86: Could we use `ByteUtil`? 
```suggestion
        public override int GetHashCode() => ByteUtil.CalculateHashCode(ToByteArray());
```
and implement `ToByteArray()`

Libplanet/PoS/Model/Redelegation.cs
```diff
@@ -98,5 +110,37 @@ public IValue Serialize()
...
+            unchecked
+            {
+                foreach (byte b in Address.ToByteArray())
+                {
+                    code = (code * 397) ^ b.GetHashCode();
+                }
+            }
+
+            return code;
+        }
```
riemannulus, line 144: Same as previous review.

Libplanet/PoS/Model/RedelegationEntry.cs
```diff
@@ -106,5 +117,36 @@ public IValue Serialize()
...
+            unchecked
+            {
+                foreach (byte b in Address.ToByteArray())
+                {
+                    code = (code * 397) ^ b.GetHashCode();
+                }
+            }
+
+            return code;
+        }
```
riemannulus, line 150: Same as previous review.

Libplanet/PoS/Model/Delegation.cs
OnedgeLee: Seems much better 👍

OnedgeLee: I've overrided some other `GetHashCode()` of classes that using validator address as a key, with other prime number (not 397) to avoid hash collision between them.
Actually, I think it's not so much matters if hash collision occurs between different type of classes, but not sure since I don't know much about its use cases.
If hash codes used for just collections, I think it would be better to use same prime number(397) on other classes, and use `ByteUtil.CalculateHashCode()`.
Do you have any opinion for this?

riemannulus (approved):
LGTM

PR #2291, opened 2022-09-05
Check marshaled transactions size instead of entire block size

### LINKED ISSUE #2290: Replace Block size limit with transactions payload size limit
Although I'm not even sure whether a block size limit is absolutely necessary (as [BitCoin] has such limit but [Ethereum] does not), considering its purpose, I think we can relax the policy to only check for the size of a block's transactions payload. We can even keep the current policy as the policy will be checked against a strictly smaller number. As block's metadata (not literal `BlockMetadata`) is tightly controlled by protocol, I don't think there should be much of an issue.

There several points of concern (at least for me) as of now:

- As `Block<T>` layout will be overhauled from protocol version 4 and onward for PBFT, max block bytes policy will no longer be compatible for "legacy" blocks. Although there is no plan to validate legacy blocks as of now, this will get messy if we ever decide to for whatever reason in the future.
- `BlockMetadata` is *required* to create `BlockContent<T>`. This is entirely other way around. Any design complication arising from this is not minor as it also affects `PreEvaluationBlockHeader`, `PreEvaluationBlock<T>`, `BlockHeader`, and `Block<T>`.
- Also newer PBFT blocks will also include `BlockCommit`, which can vary widely (even more without BLS). This can eat up whatever amount usually allocated for `Block<T>`'s transactions payload.

I suggest preemptively relaxing the policy.

[BitCoin]: https://bitcoin.org/en/
[Ethereum]: https://ethereum.org/en/

greymistcube:
Closes #2290.

To be frank, this is only tangentially related to what I'm supposed to be working on. 😗

~~It's a bit of a stretch, but the name `GetMaxBlockBytes()` is kept intact so the API doesn't have to change.~~
See the discussion below.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,10 @@ To be released.
 
 ### Behavioral changes
 
+ -  `IBlockPolicy.GetMaxBlockBytes()` is used as an upper limit for the
+    encoded size of `Block<T>.Transactions` instead of `Block<T>`.
+    [[#2290], [#2291]]
+
 ### Bug fixes
 
 ### Dependencies
@@ -43,6 +47,8 @@ To be released.
 [#2273]: https://github.com/planetarium/libplanet/pull/2273
 [#2283]: https://github.com/planetarium/libplanet/pull/2283
 [#2189]: https://github.com/planetarium/libplanet/pull/2189
+[#2290]: https://github.com/planetarium/libplanet/issues/2290
+[#2291]: https://github.com/planetarium/libplanet/pull/2291
 [Bencodex 0.5.0]: https://www.nuget.org/packages/Bencodex/0.5.0
 
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,18 @@ To be released.
  -  Renamed `BlockChain<T>.MakeTransaction(PrivateKey, IEnumerable<T>,
     IImmutableSet<Address>, DateTimeOffset?)` method's `actions` parameter to
     `customActions`.  [[#2151], [#2273]]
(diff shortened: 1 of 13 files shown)


limebell:
Instead of replacing pre-existing property's feature, how about add new property or rename it for the marshaled transaction size?

greymistcube:
I suppose I can incorporate an API change in this PR, as the meaning has changed.

limebell:
IMO as the feature has changed, it would be better to cause an error due to API change in order for the user to check the changelog.

greymistcube:
> IMO as the feature has changed, it would be better to cause an error due to API change in order for the user to check the changelog.

Not strictly a change per se, but it is just a more relaxed condition, which should be compatible with all existing chains. However, I agree the discrepancy is rather icky. I'll also change the names then. 🙃

limebell:
> > IMO as the feature has changed, it would be better to cause an error due to API change in order for the user to check the changelog.
> 
> Not strictly a change per se, but it is just a more relaxed condition, which should be compatible with all existing chains. However, I agree the discrepancy is rather icky. I'll also change the names then. 🙃

Thanks for your understanding 🙇

greymistcube:
@limebell 

I did change the names in 9a33128, but seems like the resulting API change is larger than I'd hoped (as usual). 🙄

limebell (approved):

longfin (approved):
From Libplanet's perspective, it looks fair enough. but it can be a deal breaker for upgrade it on 9c, without code migrations.  (if we use same value for new `.GetMaxTransactionsBytes`, the policy seems fine since it's more broder than previous)

PR #2296, opened 2022-09-06
Use dynamic validator

riemannulus:
## Context
In DPoS, the current Validator Set must change dynamically according to a transaction.

## Implementation
1. Get `BondedValidatorSet` state from `IStateStore`.
2. Use `BondedValidatorSet` instead static validator set.


riemannulus:
Continue at https://github.com/planetarium/libplanet/pull/2319

PR #2311, opened 2022-09-13
Fix `NetMQTransport` disposal & Release 0.41.2

tkiapril:
Fixed a bug where NetMQTransport is not properly disposed of due to NetMQTransport._router being stopped before _router.Unbind() call. Will be released as 0.41.2.

### DIFF
--- a/Libplanet.Net/Transports/NetMQTransport.cs
+++ b/Libplanet.Net/Transports/NetMQTransport.cs
@@ -335,7 +335,6 @@ namespace Libplanet.Net.Transports
 
                 if (_router is { } router && !router.IsDisposed)
                 {
-                    _router.Unbind($"tcp://*:{_listenPort}");
                     _router.Dispose();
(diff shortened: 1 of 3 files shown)


greymistcube (dismissed):

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -335,7 +335,6 @@ public void Dispose()
 
                 if (_router is { } router && !router.IsDisposed)
                 {
-                    _router.Unbind($"tcp://*:{_listenPort}");
                     _router.Dispose();
```
longfin, line 338:
```suggestion
                    // We omitted _router.Unbind() with intention due to hangs.
                    // See also: https://github.com/planetarium/libplanet/pull/2311
                    _router.Dispose();
```

it may help future readers 😅 .

longfin (approved):

Issue #2315, opened 2022-09-14
Fix reward distribution to be more flexible, to handle multiple currencies


OnedgeLee:
Current reward distribution cannot handle 9c condition (reward is defined on lib9c)
It can be overcomed by using some address as reward pool, and transfer some `FAV`s to reward pool on application side.
During fixing this, updating distribution logic for handling multiple currencies will be also happened, to apply AP circulation system later.

PR #2316, opened 2022-09-14
Fix reward distribution to be more flexible, to handle multiple currencies

OnedgeLee:
### Related issues
- Resolves https://github.com/planetarium/libplanet/issues/2315

### DIFF
--- a/Libplanet/Action/IActionContext.cs
+++ b/Libplanet/Action/IActionContext.cs
@@ -1,3 +1,4 @@
+using System.Collections.Immutable;
 using System.Diagnostics.Contracts;
 using System.Security.Cryptography;
 using Libplanet.Assets;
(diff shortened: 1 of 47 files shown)


Libplanet/Action/IActionContext.cs
```diff
@@ -98,15 +99,11 @@ public interface IActionContext
...
-        /// chain's <see cref="Libplanet.Blockchain.Policies.IBlockPolicy{T}.NativeTokens"/>.
+        /// A set of <see cref="Currency"/>s defined as native tokens by chain's
+        /// <see cref="Libplanet.Blockchain.Policies.IBlockPolicy{T}.NativeTokens"/>.
         /// </summary>
-        /// <param name="currency">A token currency to check.</param>
-        /// <returns><see langword="true"/> if the specified <paramref name="currency"/> is a native
-        /// token, otherwise <see langword="false"/>.</returns>
-        /// <seealso cref="Libplanet.Blockchain.Policies.IBlockPolicy{T}.NativeTokens"/>
         [Pure]
-        bool IsNativeToken(Currency currency);
```
dahlia, line 109: It would be better to let `IsNativeToken()` left and give it `[Obsolete]` attribute instead.

Libplanet/Action/IActionContext.cs
```diff
@@ -98,15 +99,11 @@ public interface IActionContext
...
+        IEnumerable<Currency> NativeTokens { get; }
```
dahlia, line 106: I believe `IImmutableSet<Currency>` is more appropriate here.

Libplanet/Action/IActionContext.cs
OnedgeLee: Applying every required changes, safely removed `IsNativeToken()`.

Libplanet/Action/IActionContext.cs
OnedgeLee: Applied changes.

PR #2319, opened 2022-09-14
Introduce: Dynamic validator

riemannulus:
## tl;dr
Introduces the function of dynamically importing staked validators.

## Context
In DPoS, validators must be managed dynamically. The way to manage this is to update it according to the state at the end of every block. So, in `ActionEvaluator`, I made `PoSAction` that treats the same level as `BlockAction`, and in `PoSAction`, I updated `Validators` according to the state.


Libplanet/Blockchain/BlockChain.PoS.cs
```diff
@@ -0,0 +1,36 @@
...
+                bondedValidator.OperatorPublicKey);
+        }
+
+        public List<PublicKey> BondedValidatorPubKey(BlockHash? hash = null)
+        {
+            ValidatorSet validatorSet = GetBondedValidators(hash);
+            return validatorSet.Set.Select(x => x.OperatorPublicKey).ToList();
+        }
+    }
+}
```
OnedgeLee, line 36: I'm thinking about right place to generate proposer.
It can be done on various places, but seems to handle this on block action(same procedure as validator set election, right after it) would be suitable, since selecting validator set and proposer at the same time make sense,
and since VRF proof hash can be used in various ways, it would be better to inject VRF proof to `IActionContext`.
We can talk about it later, though. :)

OnedgeLee (approved):
Added several comments, but those will be discussed after merge.

PR #2320, opened 2022-09-15
Let `Address` and `TxId` implement `IEquatable<T>`

dahlia:
It's shocking that they haven't implemented `IEquatable<T>`.  Note that they've already had `Equals(T)` which could satisfy the interface anyway…


limebell (approved):

riemannulus (approved):
lol

Issue #1299, opened 2021-05-16
Order `Transaction<T>`s for evaluation from a receiving node


greymistcube:
In order to prevent gaming the system with known `PreEvaluationHash`, `PreEvaluationHash` is used to determine the order in which `Transaction<T>`s in a `Block<T>` are executed. However, on the receiving end, when a received `Block<T>` is deserialized, the order of `Transaction<T>`s received are not validated. This allows a malicious actor to try different ordering of `Transaction<T>`s before finalizing a `Block<T>` with `StateRootHash` and `Hash`. To prevent this issue, the same ordering process must be introduced when creating a `Block<T>` from deserialization.

PR #2321, opened 2022-09-16
More robust `Block<T>` and its subset-types structures

greymistcube:
Related to #1492.

In a sense, this is a continuation of #1492.

## Goals

- Prevent instantiation of an invalid `Block<T>` object and all its subset-types (as in types with subset of the information contained in `Block<T>` such as `BlockHeader`, `PreEvaluationBlock<T>`, `PreEvaluationBlockHeader`, `BlockContent<T>`, and `BlockMetadata`).
- Enforce consistency in design between similarly related objects.

### Problem 1: Mutability of `BlockMetadata` and `BlockContent<T>`

Mostly for convenience, `BlockMetadata` and `BlockContent<T>` were implemented to be mutable. There were mainly two issues:

- Constraints bestowed on these types were somewhat arbitrary and did not cover the full extent of all constraints that could be checked.
  - Some requirements such as `Index` having to be a non-negative number and `TotalDifficulty` having to be greater than or equal to `Difficulty` seem obvious.
  - On the other hand, `PrevousHash` being able to be `null` only for `Index` of zero was not being enforced.
- Some mutation logic were explained for properties, but weren't necessarily "natural".
  - It isn't obvious why adjusting `Difficulty` should also adjust `TotalDifficulty`; moreover adjusting `TotalDifficulty` after object creation would have no affect on `Difficulty`, which seems confusing.
  - For the reason above, empty constructor with assignments in different order could effectively result in different objects.

Hence, despite all these "rules", the validity checks of `BlockMetadata` and `BlockContent<T>` were postponed until after mining when creating a `PreEvaluationBlock<T>`.

### Problem 2: Too much reliance on constructor redirects with automated meddling of arguments

In particular, I believe #1299 may still be an issue. Also there seemed to be some issues where loading a `Block<T>` saved in storage would create a `Block<T>` with wrongly loaded (i.e. **generated**) field.

### Problem 3: Unsafe constructors for `BlockHeader` and `PreEvaluationBlockHeader`

Some of the constructors for these types had so-called `proof` arguments. Although the exposures were limited to `internal` and `private`, combined with problem 2 above, it was way too obfuscated and unintuitive in determining **which** constructor should be used for a safe operation of a node. This also was also made worse when combined with problem 1. As the name suggests, **unsafe** constructors **bypass** certain constraints, but as mentioned above, creation of invalid `BlockMetadata` and `BlockContent<T>` were possible, which is very dangerous design-wise requiring **extreme** caution and **prone to mistakes** when handling these structures.

### Problem 4: "Header" objects holding on to relatively heavy "Content" objects

Due to design limitations, `BlockContent<T>` was stored inside `PreEvaluationHeader.Metadata` which is of type `BlockMetadata` when creating a `Block<T>` instance. This means the workings of the entire library was wholly reliant on this quirky implementation.

### Immutability of objects

In order to prevent creating an invalid object as much as possible, the first order of business was to make everything immutable. There are two advantages:

- Every necessary conditions can be checked inside constructors.
- There is no need to worry about "accidently" mutating an object into an invalid state.

This mitigates problem 1.

### Remove much of shortcut-like constructors

There is only one left in place for `BlockContent<T>` where it isn't absolutely necessary for creation of a `BlockContent<T>` without any `Transaction<T>`s. This mitigates problem 2.

### Funnel all constructors so conditions are always checked

This mitigates problem 3. Note that as `proof`s must be checked again, there might be a slight performance hit. I've done my best to mitigate this issue.

### Stronger and clearer distinction between handling interfaces and concrete types

Although interfaces are contracts, in general, their properties are not to be trusted when passed as an argument. This acts as an additional safeguard to accidently creating an invalid instance.

### Composition over inheritance

This is mainly for design consistency and to mitigate problem 4. Notably, `BlockContent<T>` inherited `BlockMetadata` and `PreEvaluationBlock<T>` inherited `PreEvaluationBlockHeader` but `Block<T>` **did not** inherit `BlockHeader`. I haven't dug much deeper into as to why this might be (most likely because `Block<T>` has `Header` property?), but this was an additional cause for confusion when handling `Block<T>` related code.

### Similar relations between types in accordance to positional relation in the diagram in #1492.

Internally, a `Block<T>` holds its own `PreEvaluationBlock<T>` and `BlockHeader`, a `PreEvaluationBlock<T>` holds a `BlockContent<T>` and a `PreEvaluationBlockHeader`, and a `BlockContent<T>` holds a `BlockMetadata`. Moreover,
`BlockHeader` holds its own `PreEvaluationBlockHeader`, and `PreEvaluationBlockHeader` owns `BlockMetadata` (a new diagram might be more helpful, but please refer to the top diagram in #1492).

This makes "lower" types more easily reusable by having consistent paths in the code and making them more easily trackable. In particular, a `BlockMetadata` instance is either directly or indirectly owned by `Block<T>`, `BlockHeader`, `PreEvaluationBlock<T>`, `PreEvaluationBlockHeader`, and `BlockContent<T>`, but

- `Block<T>.Header.Header.Metadata`
- `Block<T>.PreEvaluationBlock.Header.Metadata`
- `Block<T>.PreEvaluationBlock.Content.Metadata`
- etc.

all refer to the same instance of `BlockMetadata`.


longfin (approved):
- As like description, it looks like a continuous work of #1492, so my following comments may be inaccurate due to lack of context. 
  - I think that @dahlia can check about #1492's assumptions. 🙏 
- Many of changes seem that to be written for forbidding some constructor and making their types to immutables. While that, there are no critical symantic changes or the matters about performance degradation. IMO
- Since this PR affects many block related parts, I think that it would be better we check it through the real-world application (i.e., Nine Chronicles), before merging .

greymistcube:
Just squashing some commits to combine related ones. 😶

dahlia (approved):
Well… actually I left quite many review comments and suggestions, but I lost everything by mistake.  Fortunately, there were no anathema at all, so… never mind! 😂

PR #2331, opened 2022-09-20
Validate existence of +2/3 committed votes in `LastCommit`

colibrishin:
## LastCommit

`LastCommit` is the votes from the previous height. It has the previous `Height`, `Round`, `BlockHash` - which is the committed block hash, and `Vote`s. Note that `Vote`s can have `Commit` as `VoteFlag` and also `Absent`, `Unknown` and `Null`.

### Existence range of LastCommit
`LastCommit` does exist from Block index No.2 (or PBFT genesis block index + 2). The reason is that the genesis block is not decided by consensus, so the `LastCommit` for genesis does not exist. And the following block (Index No.1 block) does not have `LastCommit` because there were no previously committed votes for the genesis block. The current implementation is *null* `LastCommit` for the genesis block and Block Index No.1.

### Current flow of LastCommit

- In Consensus

```csharp
// Context.Mutate.cs L84
private void AddMessage(ConsensusMessage message)
{
  try
  {
      if (message.Height != Height)
      {
          throw new InvalidHeightMessageException(
              "Height of message differs with working height.  " +
              $"(expected: {Height}, actual: {message.Height})",
              message);
      }

      if (message is ConsensusPropose propose)
      {
          if (!propose.Validator.Equals(Proposer(message.Round)))
          {
              throw new InvalidProposerProposeMessageException(
                  "Proposer for the height " +
                  $"{message.Height} and round {message.Round} is invalid.  " +
                  $"(expected: Height: {message.Height}, Round: {message.Round}, " +
                  $"Proposer: {message.Validator} / " +
                  $"actual: Height: {Height}, Round: {Round}, " +
                  $"Proposer: {Proposer(message.Round)})",
                  message);
          }

          if (message.BlockHash.Equals(default(BlockHash)))
          {
              throw new InvalidBlockProposeMessageException(
                  "Cannot propose a null block.",
                  message);
          }
      }

      if (message is ConsensusVote vote &&
          (!vote.Validator.Equals(vote.ProposeVote.Validator) ||
          !vote.ProposeVote.Verify() ||
          !_validators.Contains(vote.Validator)))
      {
          throw new InvalidValidatorVoteMessageException(
              "Received ConsensusVote message is made by invalid validator.",
              vote);
      }

      if (message is ConsensusCommit commit &&
          (!commit.Validator.Equals(commit.CommitVote.Validator) ||
          !commit.CommitVote.Verify() ||
          !_validators.Contains(commit.Validator)))
      {
          throw new InvalidValidatorVoteMessageException(
              "Received ConsensusCommit message is made by invalid validator.",
              commit);
      }

      // TODO: Prevent duplicated messages adding.
      _messageLog.Add(message);
      _logger.Debug(
          "{FName}: Message: {Message} => Height: {Height}, Round: {Round}, " +
          "Validator Address: {VAddress}, Remote Address: {RAddress}, " +
          "Hash: {BlockHash}, MessageCount: {Count}. (context: {Context})",
          nameof(AddMessage),
          message,
          message.Height,
          message.Round,
          message.Validator.ToAddress(),
          message.Remote?.Address,
          message.BlockHash,
          _messageLog.GetTotalCount(),
          ToString());
  }
```

The votes are validated before the `AddMessage()` and if the vote has invalid information, (e.g., signature, unexpected validator, different height, *null* `BlockHash`, invalid proposal) then the message is not added to `messageLog`. Other such as `Round`, different locked value (`BlockHash`), locked round is not checked for round skipping. The values can be different and it is part of process (e.g., if validator notices another round is processing by message +2/3 then change the locked round and value. if my locked value is different with other message, and if other message is +2/3 in PreVote, then change the my locked value.)

```csharp
// Context.cs L226
public VoteSet VoteSet(int round)
{
    (Block<T>? block, int? _) = GetPropose(round);
    VoteSet voteSet = new VoteSet(Height, round, block?.Hash, _validators);
    _messageLog.GetVotes(round).ForEach(vote => voteSet.Add(vote.ProposeVote));
    _messageLog.GetCommits(round).ForEach(commit => voteSet.Add(commit.CommitVote));
    return voteSet;
}
```

The `VoteSet` is created and adds PreVotes and PreCommits *in order* (to overwrite existing `Absent` vote if `Commit` exists.) The vote is once again validated with `VoteSet` while `VoteSet.Add()`.

```csharp
// ConsensusContext.cs NewHeight(long height) L135
BlockCommit? lastCommit = null;
if (_contexts.ContainsKey(Height))
{
    lastCommit = _contexts[Height].CommittedRound == -1
        ? (BlockCommit?)null
        : new BlockCommit(
            _contexts[Height].VoteSet(_contexts[Height].CommittedRound),
            _blockChain.Tip.Hash);

    _contexts[Height].Dispose();
    _contexts.Remove(Height);
}

ClearOldLastCommitCache(maxSize: LastCommitClearThreshold);

if (lastCommit != null)
{
    _logger.Debug(
        "Caching LastCommit of Height {Height}...", Height);
    _blockChain.Store.PutLastCommit(lastCommit.Value);
}
else
{
    BlockCommit? storedCommit = _blockChain.Store.GetLastCommit(Height);
    if (storedCommit != null)
    {
        lastCommit = storedCommit;
        _logger.Debug(
            "Found cached LastCommit of Height #{Height} " +
            "and Round #{Round}",
            lastCommit.Value.Height,
            lastCommit.Value.Round);
    }
}

Height = height;
```

The `LastCommit` variable is firstly initialized with *null* in `ConsensusContext`. If `Context[Height - 1]`, which is the height context of the previous consensus, does not exist, then it will start a new height with *null* as `LastCommit`, and if not, constructs a `LastCommit` from `Context[Height -1].VoteSet()` and put it into `Block<T>.LastCommit` when a validator proposes. If there's a `LastCommit` for previous height in store, then it will load and use. If the invalid `LastCommit` was proposed (See **Added Validation** for detail), then the block will be processed with *NIL* and discarded by consensus.

### BlockCommit

`Height` and `Round` should not have negative value. *default* `BlockHash` is not allowed. *default* use case is considered *null* as `BlockCommit?` (i.e., Genesis block and Index No.1 Block.) Unmarshaling constructor is redirected to internal Bencodex constructor and redirects to public constructor, which has the check for invalid value.

```csharp
[SuppressMessage(
  "StyleCop.CSharp.ReadabilityRules",
  "SA1118:ParameterMustNotSpanMultipleLines",
  Justification =
      "Multiple lines are Vote decoding. Redirect Bencodex value to default " +
      "constructor for checking not allowed values.")]
internal BlockCommit(Dictionary dictionary)
    : this(
        dictionary.GetValue<Integer>(HeightKey),
        dictionary.GetValue<Integer>(RoundKey),
        new BlockHash(dictionary.GetValue<Binary>(BlockHashKey).ByteArray),
        dictionary.ContainsKey(VotesKey)
            ? dictionary.GetValue<List>(VotesKey)
                .Select(vote => new Vote((Binary)vote))
                .ToImmutableArray()
            : ImmutableArray<Vote>.Empty)
```

```csharp
// Constructor for BlockCommit
if (height < 0)
{
    throw new ArgumentOutOfRangeException(
        nameof(height),
        $"Height must be non-negative: {height}");
}

Height = height;

if (round < 0)
{
    throw new ArgumentOutOfRangeException(
        nameof(round),
        $"Round must be non-negative: {round}");
}

Round = round;

if (hash.Equals(default))
{
    throw new ArgumentNullException(
        nameof(hash),
        "Block hash for BlockCommit must not be null.");
}
```

### BlockCommitExtensions

`BlockCommitExtensions.HasValidVotes(this BlockCommit, IEnumerable<PublicKey>)` checks whether `BlockCommit` has +2/3 `Commit` of `BlockCommit.BlockHash`.

`BlockCommitExtensions.HasSameValidators(this BlockCommit, IEnumerable<PublicKey>)` checks whether `BlockCommit` has votes from given all validators.

`BlockCommitExtensions.HasValidVotes(this BlockCommit)` checks whether `BlockCommit.Votes` with `VoteFlag.Absent` or `VoteFlag.Commit` vote has valid signature.

### PreEvaluationHeader

The following conditions are added. For block which has `metadata.ProtocolVersion > BlockMetadata.PoWProtocolVersion`:

- `(metadata.Index == 0 || metadata.Index == 1) && metadata.LastCommit is { }`
- `(metadata.Index > 1 && !(metadata.LastCommit is { })`

Firstly, the protocol version of a block is checked for skipping a validation in a Proof-of-Work block. This is needed to still be able to read PoW blocks. And the two index conditions are checked whether `LastCommit` exist when it is needed or contrary.

The validation of votes is moved to `BlockChain<T>.ValidateNextBlock()`, validating of votes is needed before appending.

### BlockChain<T>.ValidateNextBlock()

The above **PreEvaluationHeader** conditions are applied same here. The differences are:

- `!commit.HasSameValidators(Policy.GetValidators(block.Index - 1))`
- `!commit.HasTwoThirdCommits(Policy.GetValidators(block.Index - 1))`
- `!commit.HasValidVotes()`

`HasSameValidator` checks if a `LastCommit` has same validators as given validators. By this, votes made not by validator in `ValidatorSet` or duplicated, dropped, empty cases will filtered out. `HasTwoThirdCommits` checks if a `LastCommit` has +2/3 `Commit` of `LastCommit.BlockHash`, so that we can take a block considered committed by +2/3 validators. the committed validators will be same by checking `HasSameValidator()`. +2/3 or more `Null`  or not `Commit` case is removed by it. `HasValidVotes()` finally checks whether `Commit`s or `Absent`s have valid signature for votes.


Libplanet/Blocks/BlockCommitExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+                        return true;
+                    }
+
+                    if (vote.Flag == VoteFlag.Null || vote.Flag == VoteFlag.Unknown)
+                    {
+                        return true;
+                    }
+
+                    return false;
+                });
```
greymistcube, line 78: I think this can be compressed to a single line.

Libplanet/Blocks/PreEvaluationBlockHeader.cs
```diff
@@ -148,44 +146,27 @@ public PreEvaluationBlockHeader(IBlockMetadata metadata)
                         $"the previous block's hash {metadata.PreviousHash}.");
                 }
 
-                if (commit.Votes is { } votes)
+                if (commit.Votes is { } votes && !votes.All(vote => vote.Height == commit.Height))
```
greymistcube, line 149: This can be checked internally from `BlockCommit`. I think its better to have this as part of `BlockCommitExtensions`.

Libplanet/Blocks/BlockCommitExtensions.cs
colibrishin: Amended.
https://github.com/planetarium/libplanet/blob/15f67c376b0bade3c0a32b0ed3475d42fd44f160/Libplanet/Blocks/BlockCommitExtensions.cs#L54-L65

Libplanet/Blocks/PreEvaluationBlockHeader.cs
colibrishin: Amended.
https://github.com/planetarium/libplanet/blob/15f67c376b0bade3c0a32b0ed3475d42fd44f160/Libplanet/Blocks/BlockCommitExtensions.cs#L67-L74

Libplanet/Blocks/BlockCommitExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Checks whether <see cref="BlockCommit"/> has same height for every votes.
+        /// </summary>
+        /// <param name="commit">A <see cref="BlockCommit"/> to check.</param>
+        /// <returns>Returns <see langword="true"/> if <see cref="BlockCommit.Votes"/> have same
+        /// height, otherwise returns <see langword="false"/>.</returns>
+        public static bool HasVotesSameHeight(this BlockCommit commit) =>
+            commit.Votes.All(vote => vote.Height == commit.Height);
+    }
+}
```
OnedgeLee, line 76: I don't have any strong opinion, just asking for your opinion.
Each `HasTwoThirdsCommits()`, `HasSameValidators()`, `HasValidVotes()` will perform iteration, so if we want to validate, it will take 3 times of iteration.
Compared to single validation method with multiple exception cases, this implementation is much clear but may needs more computation.
If we don't have to perform individual validation, may be integrating them would be also an option I think, though not sure if it's a better design.
(Current design is obviously clearer for reading)

OnedgeLee, line 76: I was wondered why `BlockCommit` has `ImmutableArray<Vote> Votes` instead of `VoteSet Votes` as a property. 
(which is not change of this PR)
May I ask? I thought most of these methods are related to `Vote`s, and can be implemented on `VoteSet` class.

colibrishin: Yes, I do agree on the point this would have a performance hit, but I thought that getting a review before any optimization would be better.

OnedgeLee, line 76: May I ask for the intension of the separation between `BlockCommit` and `BlockCommitExtensions`? May be saparation between model and its functionality?

Libplanet.Tests/TestUtils.cs
```diff
@@ -34,6 +35,34 @@ public static class TestUtils
...
+            "b17c919b07320edfb3e6da2f1cfed75910322de2e49377d6d4d226505afca550");
+
+        public static readonly PrivateKey ConsensusPeer1PrivateKey = PrivateKey.FromString(
+                    "e5792a1518d9c7f7ecc35cd352899211a05164c9dde059c9811e0654860549ef");
+
+        public static readonly PrivateKey ConsensusPeer2PrivateKey = PrivateKey.FromString(
+                    "91d61834be824c952754510fcf545180eca38e036d3d9b66564f0667b30d5b93");
+
+        public static readonly PrivateKey ConsensusPeer3PrivateKey = PrivateKey.FromString(
+                    "91602d7091c5c7837ac8e71a8d6b1ed1355cfe311914d9a76107899add0ad56a");
```
greymistcube, line 48: Odd formatting. 🙃

Libplanet/Blocks/BlockCommitExtensions.cs
greymistcube: Feels like `VoteSet` is just there as an intermediary entity to construct `BlockCommit`. 🙄
`VoteSet` doesn't seem to have (neither does `BlockCommit` to be frank) a rigorous specs. This issue will be addressed in future PRs.

greymistcube: I'd say it's due to a lack of specification for `BlockCommit`. As it stands, it's more like just a chunk of data. Although you are right in that height and round checking should probably be part of `BlockCommit` as it can be checked internally.

PR #2332, opened 2022-09-20
Add `PreviousMiner` property on `IBlockMetadata`

OnedgeLee:
`AllocationReward` was misimplemented to calculate proposer reward with "current proposer" and "previous validators".
To fix it, additional information`PreviousMiner` is required, so added this information to `IBlockMetadata` and `IActionContext`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,11 +10,14 @@ To be released.
 
 ### Backward-incompatible API changes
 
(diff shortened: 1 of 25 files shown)


OnedgeLee:
Since this task was done in wrong background, this PR isn't needed any more.
Closing.

PR #2351, opened 2022-09-27
WIP: Allow providing updatedAddresses for UnsignedTransaction in Libplanet.Explorer

tkiapril:
Related: planetarium/pn-chrono#11, planetarium/pnscan-cloud#6

Currently, updatedAddresses cannot be provided for UnsignedTransaction Query. This implies that external applications that connect to nodes via RPC and sign locally cannot populate the updatedAddresses field, which means block explorers like planetarium/pnscan-cloud that use the updatedAddresses field for finding involved addresses in a tx can't do so in a transaction created in external applications. This PR provides the needed facility.


tkiapril:
Closed in favor of javascript library.

PR #2355, opened 2022-09-28
`BlockCommit.Votes` no longer nullable or default

greymistcube:
🔧

### DIFF
--- a/Libplanet/Blocks/BlockCommit.cs
+++ b/Libplanet/Blocks/BlockCommit.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Bencodex;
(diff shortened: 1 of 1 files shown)


Libplanet/Blocks/BlockCommit.cs
```diff
@@ -99,7 +99,7 @@ public bool Equals(BlockCommit other)
 
             return other.Votes != null && Votes != null && Height == other.Height &&
```
colibrishin, line 100:
```suggestion
            return Height == other.Height &&
```

colibrishin (changes requested):

Libplanet/Blocks/BlockCommit.cs
greymistcube: Fixed. See the latest commit. 🙂

colibrishin (approved):

Issue #2280, opened 2022-09-01
@planetarium/cli: Support `arm64` CPU architecture for `planet` command


moreal:
Currently, it shows the below error message when I ran `npm install -g @planetarium/cli`:

```
$ npm install -g @planetarium/cli 
npm ERR! code EBADPLATFORM
npm ERR! notsup Unsupported platform for @planetarium/cli@0.41.1: wanted {"os":"darwin,linux,win32","arch":"x64"} (current: {"os":"darwin","arch":"arm64"})
npm ERR! notsup Valid OS:    darwin,linux,win32
npm ERR! notsup Valid Arch:  x64
npm ERR! notsup Actual OS:   darwin
npm ERR! notsup Actual Arch: arm64

npm ERR! A complete log of this run can be found in:
```

longfin:
I think that we may have to deal with it after #2161.

boscohyun:
I had the same problem and solved it by installing another dotnet sdk version 3.1.x.

BasixKOR:
Workaround: Install .NET SDK 3.1, and install it by `dotnet tool install -g Libplanet.Tools`

dahlia:
Although this issue (#2280) and #2161 are apparently duplicated, I'm going to distinguish two by their ways to reproduce the problem.

#2161 is about the problem faced when you try `dotnet build --project Libplanet.Tools` or `dotnet tool install Libplanet.Tools` — it's about building *Libplanet.Tools* on arm64.

This issue (#2280), on the other hand, deals with the problem faced when you try the `planet` program from GitHub Releases or `npm install @planetarium/cli` — it's about how we provide official executable binaries of `planet`.

The key blocker of this as of September 2022 is that:

- `dotnet build` does not support crosscompile osx-arm64 binaries from other hosts than osx-arm64.
- GitHub-hosted runners of GitHub Actions do not provide Apple silicon environments.  It's tracked by <https://github.com/actions/runner-images/issues/2187>.

longfin:
> dotnet build does not support crosscompile osx-arm64 binaries from other hosts than osx-arm64.

I guess it's possible. The below log is from another .NET executable project built with the `osx-arm64` on my linux-x64 machine.

```bash
[longfin@longfin-desktop-arch .Libplanet]$ dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.109
 Commit:    58a93139d8

Runtime Environment:
 OS Name:     arch
 OS Version:  
[...]

PR #2365, opened 2022-10-01
Support osx-arm64

longfin:
This PR adds builds for Apple Silicon (a.k.a. M1/M2) to GitHub Releases and NPM. 

I'm not sure this is all that needs to be changed due to a lack of context. so please let me know if there's something else that should be considered.


dahlia:
The [built binary][1] actually does not work on Apple silicon Macs:

```console
$ 7z x dist-bin.zip
$ tar xvfJ planet-0.43.0-dev.20221001102841-osx-arm64.tar.xz 
./Libplanet.Extensions.Cocona.pdb
./Libplanet.Net.pdb
./Libplanet.Net.xml
./Libplanet.RocksDBStore.pdb
./Libplanet.Stun.pdb
./Libplanet.pdb
./Libplanet.xml
./libgflags.2.2.dylib
./liblz4.1.dylib
./librocksdb.dylib
./libsnappy.1.dylib
./libzstd.1.dylib
./planet
./planet.pdb
$ file *.dylib planet
libgflags.2.2.dylib: Mach-O 64-bit dynamically linked shared library arm64
liblz4.1.dylib:      Mach-O 64-bit dynamically linked shared library arm64
librocksdb.dylib:    Mach-O 64-bit dynamically linked shared library arm64
libsnappy.1.dylib:   Mach-O 64-bit dynamically linked shared library arm64
libzstd.1.dylib:     Mach-O 64-bit dynamically linked shared library arm64
planet:              Mach-O 64-bit executable arm64
$ ./planet 
Killed: 9
```

It's the exactly same issue that made me to give up the [previous attempt][2]…

See also: <https://github.com/planetarium/libplanet/issues/2280#issuecomment-1238870064>

[1]: https://github.com/planetarium/libplanet/suites/8565116426/artifacts/382975775
[2]: https://github.com/planetarium/libplanet/pull/1650

longfin:
I succeed to run an executable built from CI with an adhoc code signing.

```bash
bash-3.2$ ./planet
Killed: 9

longfin@longfinui-MacBookAir osx-arm64 % codesign -s - -f planet
longfin@longfinui-MacBookAir osx-arm64 % ./planet
objc[41846]: Class PlaceholderObject is implemented in both /Users/longfin/Downloads/osx-arm64/planet (0x1017393f0) and /Users/longfin/Downloads/osx-arm64/libSystem.Native.dylib (0x122140600). One of the two will be used. Which one is undefined.
Usage: planet [command]

planet

Commands:
  apv
  key
  mpt
  store
  stats

Options:
  -h, --help    Show help message
  --version     Show version
```

I think that [Gatekeeper](https://support.apple.com/en-US/guide/security/sec5599b66df/web) may disturb running... (idk why it's okay on x86-64 binary, but I guess it may be related to Rosseta2 🤔 )

Anyway, we need an integration for codesigning. I'll work on that when available.

longfin:
- After macOS 11 on Apple Silicon(arm64), Apple seems to require a codesign more strictly.
- Since [Adhoc signing](https://github.com/planetarium/libplanet/pull/2365/#issuecomment-1266343562) is only valid in the environment of the user who signed it, it cannot be used for CI.
- Some package managers (e.g., Homebrew, `dotnet tools`) seemed to be prepared for that requirement through Adhoc code signing. ([1], [2])
  - so If `planet` is to be installed only through such channels, we can omit code signing.
  - In that case, our downloader(i.e., `download.mjs`) also should be adjusted.
- On the other hand, in order to fundamentally eliminate concerns, it is also a possible option for someone in the project to represent and register for the Apple Developer Program and feed to CI.
  - However, who is appropriate to represent this needs to be discussed and may take time.

Unless otherwise noted, it seems like a viable option right away, so I'm working towards adding ad hoc signing.

@planetarium/libplanet Any ideas? 

[1]: https://github.com/Homebrew/brew/issues/9082
[2]: https://github.com/dotnet/runtime/issues/52335

dahlia:
@longfin IMHO, it would be the quickest and the most approachable at present to make *download.mjs* to ad-hoc sign it immediately after downloading the executable on macOS.

longfin:
Added adhoc signing on https://github.com/planetarium/libplanet/pull/2365/commits/670f8f30b704361cf145eede41dcb8d786d5fbf1.

PTAL 🙏  @dahlia (and other Apple Silicon users)

Libplanet.Tools/download.mjs
```diff
@@ -119,7 +143,26 @@ Report this issue from https://github.com/planetarium/libplanet/issues
 ${err}
 -------------------------------------------------------------------------------
     `);
-    throw e;
+    throw err;
+  }
+  if (platform === "darwin" && arch === "arm64")
```
dahlia, line 148: Wouldn't it better to ad-hoc sign the executable on Intel Macs too?

dahlia (approved):

Libplanet.Tools/download.mjs
longfin: makes sense, but I didn't have Intel Mac to test😅 .

dahlia: Okay, never mind then.  Anyway, could you resolve the conflict?

dahlia (approved):

dahlia (changes requested):
Oh, we almost forgot about docs.  Could you adjust the following part from *Libplanet.Tools/README.md*?  It would be shown on [docs.libplanet.io](https://docs.libplanet.io/) as well.

https://github.com/planetarium/libplanet/blob/dca2386f7ef33807491a2650dbb3e02dede576be/Libplanet.Tools/README.md#L45-L48

longfin:
> Oh, we almost forgot about docs. Could you adjust the following part from Libplanet.Tools/README.md? It would be shown on [docs.libplanet.io](https://docs.libplanet.io/) as well.

Added https://github.com/planetarium/libplanet/pull/2365/commits/c840c986168bdda2c4d9b140c8514b253bae752e, PTAL @dahlia 🙏

dahlia (approved):

dahlia (approved):

greymistcube:
@dahlia @longfin 
Eh... this breaks linux-mono-build (not tests). Can you take another look?

longfin:
> @dahlia @longfin Eh... this breaks linux-mono-build (not tests). Can you take another look?

Sure, I'll check again now.

longfin:
> > @dahlia @longfin Eh... this breaks linux-mono-build (not tests). Can you take another look?
> 
> Sure, I'll check again now.

Opened the new PR(#2462) about that. cc @greymistcube @dahlia

PR #2366, opened 2022-10-04
Implement nextNonce query in Libplanet.Explorer

### LINKED ISSUE #2356: Implement nextNonce query in Libplanet.Explorer
Currently there are no ways to retrieve next tx nonce for an account with GraphQL query. To use #2294 effectively we need a GQL query to retrieve the next tx nonce.

tkiapril:
Resolves #2356.

### DIFF
--- a/Libplanet.Explorer.Tests/Queries/TransactionQueryTest.cs
+++ b/Libplanet.Explorer.Tests/Queries/TransactionQueryTest.cs
@@ -1,7 +1,9 @@
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Threading.Tasks;
+using Bencodex.Types;
 using GraphQL;
 using GraphQL.Execution;
 using Libplanet.Action;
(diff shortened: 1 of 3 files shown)


dahlia (commented):
Could you write a test cast for this query?

tkiapril:
@dahlia Actually I was wondering how I might go with writing tests with this one. I could have a mock Blockchain<T> with a hard-coded case returned, but would this be sufficient? I thought if implemented this way it would serve little purpose but hide it from coverage report.

dahlia:
Although its implementation is very thin at present, but it may change in the future.  (In fact, we once considered to remove `BlockChain<T>.GetNextTxNonce()` and expose `IStagePolicy.GetNextTxNonce()` instead months ago.)  Test cases would be useful for maintaining API's stability and backward compatibility despite its internal changes.

See also: <https://github.com/planetarium/libplanet/blob/c2537a8f38a91ed57f6a012a12ce65bd381c4cda/Libplanet.Tests/Blockchain/BlockChainTest.cs#L1417-L1490>.

dahlia (approved):

Libplanet.Explorer.Tests/Queries/TransactionQueryTest.cs
```diff
@@ -82,6 +83,46 @@ public async Task BindSignatureWithSystemAction()
...
+        await _source.BlockChain.MineBlock(new PrivateKey());
+        await AssertNextNonce(2, key1.ToAddress());
+
+        var key2 = new PrivateKey();
+        await AssertNextNonce(0, key2.ToAddress());
+
+        _source.BlockChain.MakeTransaction(key2, ImmutableList<NullAction>.Empty.Add(new NullAction()));
+        await _source.BlockChain.MineBlock(new PrivateKey());
+        await AssertNextNonce(1, key2.ToAddress());
+        await AssertNextNonce(2, key1.ToAddress());
```
moreal, line 123: I think It'll be better to test staged transactions also because `BlockChain<T>.GetTxNonce()` includes staged transactions' nonce.

moreal (approved):

moreal (dismissed):

dahlia (approved):

dahlia (approved):

PR #2373, opened 2022-10-12
Fix flakiness in PBFT `ConsensusContext` test

colibrishin:
This fixes the flakiness of the `ConsensusContext` test. 

## Reasons
- The flakiness occurred when the asynchronous `Context` starts and receives messages before any `EventHandler` is attached and catching.
- Some of the PBFT timeout tests run long.

## Changes
- Expose timeout parameters from `Context` and `ConsensusContext` and set 1 second for every timeout test.
- Remove unnecessary `NetMQConfiguration` collection from the test.
- `HandleMessageFromHigherHeight` covers two cases. separated as `NewHeightDelay`.
- Removed two duplicated WaitHandle in `HandleMessageFromHigherHeight` which blocks the test.
- Simplified `UseLastCommitCacheIfHeightContextIsEmpty`.
- Shortened `StartAsync`'s PropagationDelay that triggers timeout in the CI test.

## Related PR

https://github.com/planetarium/libplanet/pull/2383 - This change fixes the after-attaching, for lack of a better term, EventHandler problem.
https://github.com/planetarium/libplanet/pull/2382 - This change is a ground work of fixing timeout in CI failure because of test duration.

### DIFF
--- a/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
@@ -26,7 +26,7 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
         public ConsensusContextNonProposerTest(ITestOutputHelper output)
         {
             const string outputTemplate =
-                "{Timestamp:HH:mm:ss:ffffffZ} - {Message}";
+                "{Timestamp:HH:mm:ss:ffffffZ} - {Message} {Exception}";
             Log.Logger = new LoggerConfiguration()
                 .MinimumLevel.Verbose()
                 .WriteTo.TestOutput(output, outputTemplate: outputTemplate)
--- a/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextNonProposerTest.cs
@@ -17,7 +17,6 @@ using Xunit.Sdk;
 
 namespace Libplanet.Net.Tests.Consensus.ConsensusContext
 {
-    [Collection("NetMQConfiguration")]
     public class ConsensusContextNonProposerTest
     {
         private const int Timeout = 30000;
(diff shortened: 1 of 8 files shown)


Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -52,14 +59,29 @@ public class ConsensusContext<T> : IDisposable
...
             TimeSpan newHeightDelay,
             Func<long, IEnumerable<PublicKey>> getValidators,
-            long lastCommitClearThreshold)
+            long lastCommitClearThreshold,
+            int timeoutProposeSecond = 5,
+            int timeoutPreVoteSecond = 5,
+            int timeoutPreCommitSecond = 5,
+            int timeoutProposeMultiplier = 1,
+            int timeoutPreVoteMultiplier = 1,
+            int timeoutPreCommitMultiplier = 1)
```
greymistcube, line 84: I think these should be grouped together as a `class` for following reasons:

- Moving away from `const` allows nonsensical values to be passed such as negative values.
- Adding six more parameters to what is already seemingly too cluttered.

Also this seems like a nice sensible unit to make a small separate PR. 🙂

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -142,7 +171,75 @@ public void Dispose()
...
+            : "No context";
+
+        internal void NewHeight(
+            long height,
+            EventHandler<Exception>? initialExceptionHandler,
+            EventHandler<(Step Step, TimeSpan TimeSpan)>? initialTimeoutHandler,
+            EventHandler<int>? initialTimeoutProcessHandler,
+            EventHandler<(int MessageLogSize, int Round, Step Step)>? initialStateHandler,
+            EventHandler<ConsensusMessage>? initialMessageHandler,
+            EventHandler<System.Action>? initialMutationHandler)
```
greymistcube, line 242: If I understand correctly, this together with `AttachHandlers` is to use a specific handler on a `Context` with a specific height. I think it'd be better to have something like event handlers in `ConsensusContext.Event.cs` akin to `Context.Event.cs` (doesn't necessarily have to be in a separate file, just my preference) that passes on the events from all `Context`s, if possible, as a proper channel.

limebell (approved):

colibrishin:
Before reviewing, please check and review PR https://github.com/planetarium/libplanet/pull/2383 first. 🙏

limebell (approved):

Issue #2377, opened 2022-10-12
Check version before throw different apv exception


Atralupus:
We discussed and draw out the results
Just append only check APV version https://github.com/planetarium/libplanet/blob/71e204f4dc9bc4ee27e55122cf0eaccb52d7cff7/Libplanet.Net/Messages/MessageValidator.cs#L157 above this line

Currently `AppProtocolVersion` values `Version` not other data (FYI @greymistcube 's `Intention of APV`)

_Originally posted by @Atralupus in https://github.com/planetarium/libplanet/discussions/2367#discussioncomment-3857168_

PR #2380, opened 2022-10-12
Implement different extra callbacks and ignore

Atralupus:
Solved https://github.com/planetarium/libplanet/issues/2377

I add `!message.Version.Version.Equals(appProtocolVersion.Version) || !trusted` condition above `DifferentAppProtocolVersionException`
it only checks apv version and accepts trusted signer

### DIFF
--- a/Libplanet.Net.Tests/Messages/MessageValidatorTest.cs
+++ b/Libplanet.Net.Tests/Messages/MessageValidatorTest.cs
@@ -57,92 +57,159 @@ namespace Libplanet.Net.Tests.Messages
             var version2 = 2;
             var extra1 = new Bencodex.Types.Integer(13);
             var extra2 = new Bencodex.Types.Integer(17);
-            var trustedApv1 = AppProtocolVersion.Sign(trustedSigner, version1, extra1);
-            var trustedApv2 = AppProtocolVersion.Sign(trustedSigner, version2, extra2);
-            var trustedApv3 = AppProtocolVersion.Sign(trustedSigner, version1, extra2);
-            var unknownApv1 = AppProtocolVersion.Sign(unknownSigner, version1, extra1);
-            var unknownApv2 = AppProtocolVersion.Sign(unknownSigner, version1, extra2);
-            ImmutableHashSet<PublicKey>? trustedApvSigners1 =
-                new HashSet<PublicKey>() { trustedSigner.PublicKey }.ToImmutableHashSet();
-            ImmutableHashSet<PublicKey>? trustedApvSigners2 =
-                new HashSet<PublicKey>() { }.ToImmutableHashSet();
-            ImmutableHashSet<PublicKey>? trustedApvSigners3 = null;
+
             DifferentAppProtocolVersionEncountered callback = (p, pv, lv) => { called = true; };
             var peer = new BoundPeer(trustedSigner.PublicKey, new DnsEndPoint("0.0.0.0", 0));
-            var trustedPing1 = new PingMsg() { Remote = peer, Version = trustedApv1 };
-            var trustedPing2 = new PingMsg() { Remote = peer, Version = trustedApv2 };
-            var trustedPing3 = new PingMsg() { Remote = peer, Version = trustedApv3 };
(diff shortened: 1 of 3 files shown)


Libplanet.Net.Tests/Messages/MessageValidatorTest.cs
```diff
@@ -124,19 +129,19 @@ public void ValidateAppProtocolVersion()
                 messageTimestampBuffer: null,
                 differentAppProtocolVersionEncountered: callback);
             messageValidator.ValidateAppProtocolVersion(trustedPing1);
-            exception = Assert.Throws<DifferentAppProtocolVersionException>(
-                () => messageValidator.ValidateAppProtocolVersion(unknownPing1));
-            Assert.True(exception.Trusted);
+            Assert.False(called);
```
Atralupus, line 132: I'm not sure should need this assert 😶

greymistcube: I think it's fine for now as a part of the spec until there is some need to inspect messages from unknown signer from an application. 😶

Libplanet.Net.Tests/Messages/MessageValidatorTest.cs
```diff
@@ -70,6 +71,7 @@ public void ValidateAppProtocolVersion()
             var peer = new BoundPeer(trustedSigner.PublicKey, new DnsEndPoint("0.0.0.0", 0));
             var trustedPing1 = new PingMsg() { Remote = peer, Version = trustedApv1 };
             var trustedPing2 = new PingMsg() { Remote = peer, Version = trustedApv2 };
+            var trustedPing3 = new PingMsg() { Remote = peer, Version = trustedApv3 };
             var unknownPing1 = new PingMsg() { Remote = peer, Version = unknownApv1 };
             var unknownPing2 = new PingMsg() { Remote = peer, Version = unknownApv2 };
```
greymistcube, line 76: It's really hard to keep track of what conditions are being tested. I would suggest one of the following:

- Use more descriptive variable names for different scenarios, as it isn't immediately apparent what is being tested around the `Assert` statements.
- Parameterize arguments.
- Add more comments below.
- Order tested scenarios below to match the order of variable declaration here.

Libplanet.Net/Messages/MessageValidator.cs
```diff
@@ -154,18 +154,21 @@ public class MessageValidator
...
-                    $"signature {ByteUtil.Hex(appProtocolVersion.Signature)} by " +
-                    $"signer {appProtocolVersion.Signer}\n" +
-                    $"Actual: APV {message.Version} with " +
-                    $"signature: {ByteUtil.Hex(message.Version.Signature)} by " +
-                    $"signer: {message.Version.Signer}\n" +
-                    $"Signed by a trusted signer: {trusted}",
-                    appProtocolVersion,
-                    message.Version,
-                    trusted);
+                if (!message.Version.Version.Equals(appProtocolVersion.Version) || !trusted)
```
greymistcube, line 157:
```suggestion
                if (!trusted || !message.Version.Version.Equals(appProtocolVersion.Version))
```

Two reasons:
- Code-wise, `trusted` is already computed.
- Semantically, we don't want to need to compare versions (at least for now) for untrusted messages.

Libplanet.Net.Tests/Messages/MessageValidatorTest.cs
Atralupus: I agree, I had written test code that was hard to understand
thanks, I'll coding your suggestion

Libplanet.Net/Messages/MessageValidator.cs
Atralupus: yes right!

Libplanet.Net.Tests/Messages/MessageValidatorTest.cs
Atralupus: I modified the tests... could you check this? 🙏

greymistcube (dismissed):

greymistcube:
Looks great! Just needs rebasing again 😅

greymistcube (approved):

Atralupus:
🚀

PR #2382, opened 2022-10-12
Option class for consensus timeout configuration

colibrishin:
This PR introduces a new class for the option of consensus timeout configuration.

## Base context
- For each step (Propose, PreVote, PreCommit), there are base timeouts and multipliers. Base timeouts are the initial values for timeout, and multipliers are for multiplying factors when a new round is started.
- This class guards from unexpected values (where value <= 0) and lets the values can be configurable.

### DIFF
--- a/Libplanet.Net/Consensus/ContextTimeoutOption.cs
+++ b/Libplanet.Net/Consensus/ContextTimeoutOption.cs
@@ -1,0 +1,85 @@
+using System;
+
+namespace Libplanet.Net.Consensus
+{
+    /// <summary>
+    /// A options class to configure <see cref="Context{T}"/> timeout for each <see cref="Step"/>.
+    /// </summary>
+    public class ContextTimeoutOption
+    {
+        public ContextTimeoutOption(
+            int proposeSecondBase = 5,
+            int preVoteSecondBase = 5,
+            int preCommitSecondBase = 5,
+            int proposeMultiplier = 1,
+            int preVoteMultiplier = 1,
+            int preCommitMultiplier = 1)
+        {
+            if (proposeSecondBase <= 0)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(proposeSecondBase),
+                    "ProposeSecondBase must be greater than 0.");
+            }
+
+            ProposeSecondBase = proposeSecondBase;
+
+            if (preVoteSecondBase <= 0)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(preVoteSecondBase),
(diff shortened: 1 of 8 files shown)


Libplanet.Net/Consensus/ContextTimeoutOption.cs
```diff
@@ -0,0 +1,119 @@
...
+                if (value <= 0)
+                {
+                    throw new ArgumentOutOfRangeException(
+                        nameof(value),
+                        "ProposeSecondBase must be greater than or equal to 0."
+                    );
+                }
+
+                _proposeSecondBase = value;
+            }
```
greymistcube, line 31: Might be just my personal preference, but I'm not a big fan of setters where mutation doesn't make much sense. 🙄
Is there a context in which we would want to change the option on the fly? If not, I'd prefer using default values for a constructor.

greymistcube (approved):

Libplanet.Net/Consensus/ContextTimeoutOption.cs
colibrishin: Hmm, I also don't know if there's any such use-case, and furthermore, if it should be permitted.

colibrishin: Changed to the constructor style.
https://github.com/planetarium/libplanet/commit/9568ed60af44d865baaa9ba196700cdbb5398253#diff-95065cc3ba907122e5352fa2ee987c580d32bfe93d8c8e5a71b70a23e2767aeb

Libplanet.Net/Consensus/ContextTimeoutOption.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        public int PreVoteSecondBase { get; } = 5;
+
+        public int PreCommitSecondBase { get; } = 5;
+
+        public int ProposeMultiplier { get; } = 1;
+
+        public int PreVoteMultiplier { get; } = 1;
+
+        public int PreCommitMultiplier { get; } = 1;
```
greymistcube, line 86: I think you can remove default assignments here as parameter-less default constructor is not available. 🤔

colibrishin: I see. One thing I noticed is that struct default sets all values to zero, but class default treats the same as null.
https://github.com/planetarium/libplanet/pull/2382/commits/7fa12b6a7634b75b2dae35b83a303e094d1b0cf7#diff-95065cc3ba907122e5352fa2ee987c580d32bfe93d8c8e5a71b70a23e2767aeb

greymistcube (approved):

PR #2383, opened 2022-10-13
Subscribing all `Context` EventHandler

colibrishin:
This PR adds `EventHandler`s in `ConsensusContext<T>` which are listening to their designated `EventHandler` from multiple `Context<T>`.

Solves after-attaching (Subscribe method to `EventHandler` after the `Context<T>` has started) problem.

## Changes
`ConsensusContext` has now its `EventHandler`s which are the same type as `Context<T>`'s. These will be attached to `Context<T>` when it is initialized (By `NewHeight()` or `HandleMessage`.)

## Failing tests
This PR is one of fixing flakiness in the PBFT test, but the groundwork for subscribing `EventHandler` before the `Context` starts.

- `HandleMessageFromHigherHeight` fails due to time check. target time was 1 sec, actual time delta was ~= 0.99 sec, 1.5sec in the worst case.
- Some of the tests have a long duration. Most tests are timeout, so the shortened timeout value is needed.

These reasons are out of the scope of this PR so the remaining failed test is fixed in https://github.com/planetarium/libplanet/pull/2373.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.Event.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.Event.cs
@@ -1,0 +1,54 @@
+using System;
+using Libplanet.Action;
+using Libplanet.Net.Messages;
+
+namespace Libplanet.Net.Consensus
+{
+    public partial class ConsensusContext<T>
+        where T : IAction, new()
+    {
+        /// <inheritdoc cref="Context{T}.ExceptionOccurred"/>
+        internal event EventHandler<(long Height, Exception)>? ExceptionOccurred;
+
+        /// <inheritdoc cref="Context{T}.TimeoutOccurred"/>
+        internal event EventHandler<
+                (long Height, int Round, Step Step, TimeSpan TimeSpan)>?
+            TimeoutOccurred;
+
+        /// <inheritdoc cref="Context{T}.TimeoutProcessed"/>
+        internal event EventHandler<(long Height, int Round)>? TimeoutProcessed;
+
+        /// <inheritdoc cref="Context{T}.StateChanged"/>
+        internal event EventHandler<
+                (long Height, int MessageLogSize, int Round, Step Step)>?
+            StateChanged;
+
+        /// <inheritdoc cref="Context{T}.MessageConsumed"/>
+        internal event EventHandler<(long Height, ConsensusMessage Message)>?
+            MessageConsumed;
+
+        /// <inheritdoc cref="Context{T}.MutationConsumed"/>
+        internal event EventHandler<(long Height, System.Action)>? MutationConsumed;
+
+        private void AttachEventHandlers(Context<T> context)
+        {
+            context.ExceptionOccurred += (sender, exception) =>
+                ExceptionOccurred?.Invoke(this, exception);
+
+            context.TimeoutOccurred += (sender, timeoutWait) =>
+                TimeoutOccurred?.Invoke(this, timeoutWait);
+
+            context.TimeoutProcessed += (sender, timeoutStart) =>
+                TimeoutProcessed?.Invoke(this, timeoutStart);
+
+            context.StateChanged += (sender, state) =>
+                StateChanged?.Invoke(this, state);
+
+            context.MessageConsumed += (sender, message) =>
+                MessageConsumed?.Invoke(this, message);
+
+            context.MutationConsumed += (sender, action) =>
+                MutationConsumed?.Invoke(this, action);
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


Libplanet.Net/Consensus/ConsensusContext.Event.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <inheritdoc cref="Context{T}.StateChanged"/>
+        internal event EventHandler<(int MessageLogSize, int Round, Step Step)>? StateChanged;
+
+        /// <inheritdoc cref="Context{T}.MessageConsumed"/>
+        internal event EventHandler<ConsensusMessage>? MessageConsumed;
+
+        /// <inheritdoc cref="Context{T}.MutationConsumed"/>
+        internal event EventHandler<System.Action>? MutationConsumed;
+
```
greymistcube, line 27: This loses context, i.e. height information. I was thinking more of having height info added here.

Something along the lines of

```csharp
internal event EventHandler<(long Height, int MessageLogSize, int Round, Step Step)>? StateChanged;
```

and

```csharp
context.StateChanged += (sender, size, round, step) => StateChanged?.Invoke(sender, height, size, round, step);
```

colibrishin: Amended. Seems there is no good way to concatenate the tuple, so let it handle two named tuples.
https://github.com/planetarium/libplanet/pull/2383/commits/22a6ea572d7217f318f78c5c8ff403a6e579f822#diff-d5fcd6416d32a07c2ab500ee36b81596dfd297f498c1c4d691040141b5c1ebd3

Libplanet.Net/Consensus/ConsensusContext.Event.cs
```diff
@@ -0,0 +1,86 @@
...
+    public partial class ConsensusContext<T>
+        where T : IAction, new()
+    {
+        /// <inheritdoc cref="Context{T}.ExceptionOccurred"/>
+        internal event EventHandler<(long Height, int Round, Exception)>? ExceptionOccurred;
+
+        /// <inheritdoc cref="Context{T}.TimeoutOccurred"/>
+        internal event EventHandler<
+                ((long Height, int Round) Context, (Step Step, TimeSpan TimeSpan) Event)>?
+            TimeoutOccurred;
```
greymistcube, line 16: The only additional "variable context" from `ConsensusContext` is `Height`. I think it'd be better to have

```
(long Height, (int Round, Step Step, TimeSpan TimeSpan) Event)>?
```

and additionally add `Round` for `Context` event handlers for overall consistency.

colibrishin: https://github.com/planetarium/libplanet/blob/53cc604384598000490e77f65e036e5165e1f420/Libplanet.Net/Consensus/ConsensusContext.Event.cs#L13-L16
https://github.com/planetarium/libplanet/blob/53cc604384598000490e77f65e036e5165e1f420/Libplanet.Net/Consensus/ConsensusContext.Event.cs#L61-L69

Amended. I just realized that unwrapping tuple is a one way to do so...

https://github.com/planetarium/libplanet/pull/2383/files#diff-d5fcd6416d32a07c2ab500ee36b81596dfd297f498c1c4d691040141b5c1ebd3

greymistcube: I think either way is fine as long as there is some semblance of consistency 😗

Libplanet.Net/Consensus/ConsensusContext.Event.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            context.StateChanged += (sender, tuple) =>
+            {
+                if (sender is Context<T> { } ctx)
+                {
+                    StateChanged?.Invoke(
+                        sender,
+                        (ctx.Height, tuple.MessageLogSize, tuple.Round, tuple.Step));
+                }
+            };
```
greymistcube, line 69:
```suggestion
            context.StateChanged += (sender, tuple) =>
                StateChanged?.Invoke(
                    this,
                    (context.Height, tuple.MessageLogSize, tuple.Round, tuple.Step));
```

`StateChanged`'s owner is `ConsensusContext`, not `Context`. As "wrapped" `EventHandler`s are operating at a different layer, I think `this` as a sender would be more appropriate.

colibrishin: Moved `Context` data responsibility to `Context.Event`s EventHandler, and `ConsensusContext` takes it and propagates.
fb2509329705ee26db7aaaded24fd9908602a788
0ac9fdc43be4ea3914f762752420b8b51232a6d2

limebell (approved):

PR #2390, opened 2022-10-14
Remove old contexts

limebell:
There were cases where `ConsensusContext` did not dispose `Context` instances correctly, and this PR fixed it by iterating the dictionary and removing it correctly.

Skips the changelog because PBFT has not been released yet.

PR #2411, opened 2022-10-17
Access collision in `Context` dictionary

colibrishin:
The access collision was observed in linux-unity-test from `ConsensusContext<T>._contexts` dictionary after https://github.com/planetarium/libplanet/pull/2390 patch.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -21,6 +21,7 @@ namespace Libplanet.Net.Consensus
     public partial class ConsensusContext<T> : IDisposable
         where T : IAction, new()
     {
+        private readonly object _contextLock;
         private readonly ContextTimeoutOption _contextTimeoutOption;
 
         private readonly BlockChain<T> _blockChain;
@@ -84,6 +85,8 @@ namespace Libplanet.Net.Consensus
                 .ForContext("SubTag", "ConsensusContext")
                 .ForContext<ConsensusContext<T>>()
                 .ForContext("Source", nameof(ConsensusContext<T>));
+
+            _contextLock = new object();
         }
 
         /// <summary>
@@ -131,9 +134,12 @@ namespace Libplanet.Net.Consensus
         public void Dispose()
         {
             _newHeightCts?.Cancel();
-            foreach (Context<T> context in _contexts.Values)
+            lock (_contextLock)
             {
-                context.Dispose();
+                foreach (Context<T> context in _contexts.Values)
+                {
+                    context.Dispose();
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


greymistcube:
Just my two cents, but as `ConsensusContext` is accessed asynchronously and observing how `_context` is used, I don't think simply using a `ConcurrentDictionary` is a proper solution. Either all dictionary access method should be rewritten with concurrency in mind if we are to use `ConcurrentDictionary` **or** use a `lock` on `_context`. 🙄

limebell:
How about using a lock instead of `ConcurrentDictionary`? The logic of `ConsensusContext` is not considered to be used concurrently IMO.

colibrishin:
Changed from `ConcurrentDictionary` to `lock` and set it to where race conditions occurred.

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -324,7 +333,7 @@ private void ClearOldLastCommitCache(long maxSize = 30)
         }
 
         /// <summary>
-        /// Discard and remove all contexts that has lower or equal height with
+        /// Discard and remove all contexts that has lower height with
```
limebell, line 336: 👍

limebell (approved):

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -199,14 +202,17 @@ public void NewHeight(long height)
...
             {
-                _contexts[height] = new Context<T>(
-                    this,
-                    _blockChain,
-                    height,
-                    _privateKey,
-                    _getValidators(height).ToList(),
-                    Step.Default,
-                    contextTimeoutOptions: _contextTimeoutOption);
+                lock (_contextLock)
```
greymistcube, line 205: This should be moved one layer up I think?

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -251,13 +257,16 @@ public void HandleMessage(ConsensusMessage consensusMessage)
...
             if (!_contexts.ContainsKey(height))
             {
-                _contexts[height] = new Context<T>(
-                    this,
-                    _blockChain,
-                    height,
-                    _privateKey,
-                    _getValidators(height).ToList(),
-                    _contextTimeoutOption);
+                lock (_contextLock)
```
greymistcube, line 260: Same as above

Libplanet.Net/Consensus/ConsensusContext.cs
colibrishin: https://github.com/planetarium/libplanet/blob/ab82ef9d1219224f7d877fb27b45004a6e7e1e36/Libplanet.Net/Consensus/ConsensusContext.cs#L206-L223

Amended.

Libplanet.Net/Consensus/ConsensusContext.cs
colibrishin: https://github.com/planetarium/libplanet/blob/ab82ef9d1219224f7d877fb27b45004a6e7e1e36/Libplanet.Net/Consensus/ConsensusContext.cs#L261-L275

Amended.

PR #2415, opened 2022-10-19
Unexpected vote created with arbitrary `VoteFlag`

colibrishin:
There is a case in which a forged vote can be created with an arbitrary `VoteFlag` for `ConsensusVote/Commit` (e.g., `VoteFlag.PreVote` + `ConsensusCommit`.) These could lead to an unexpected state change of `Context<T>`.

## `Context`
The current `GetVotes()` and `GetCommits()` behaviour could count different `VoteFlag`. For example, If a `ConsensusVote` with a vote contains `VoteFlag.PreCommit` can be counted as a PreVote message.

https://github.com/planetarium/libplanet/blob/b580d379a2f092a9054504990e2b0812194144fc/Libplanet.Net/Consensus/MessageLog.cs#L52-L80

The methods are used in state change condition methods, `HasTwoThirdsPreVote()` and `HasTwoThirdsPreCommit()`, which do not also check `VoteFlag`.

https://github.com/planetarium/libplanet/blob/b580d379a2f092a9054504990e2b0812194144fc/Libplanet.Net/Consensus/Context.cs#L402-L438

This behavior also has an impact on `VoteSet` because `GetVotes()` and `GetCommits()` are used.

https://github.com/planetarium/libplanet/blob/b580d379a2f092a9054504990e2b0812194144fc/Libplanet.Net/Consensus/Context.cs#L224-L236

## Changes
In the constructor of `ConsensusVote` and `ConsensusCommit`, the sanity of `VoteFlag` is checked.

### DIFF
--- a/Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
@@ -334,14 +334,14 @@ namespace Libplanet.Net.Tests.Consensus.Context
             context.ProduceMessage(
                 new ConsensusCommit(
                     TestUtils.CreateVote(
-                        TestUtils.Peer2Priv, 1, 0, hash: null, flag: VoteFlag.PreVote))
+                        TestUtils.Peer2Priv, 1, 0, hash: null, flag: VoteFlag.PreCommit))
                 {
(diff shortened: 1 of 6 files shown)


limebell (commented):
Can we have a regression test for this?

limebell (approved):

PR #2419, opened 2022-10-19
Added constraint on vote metadata

greymistcube:
~~`Null` and `Unknown` type `Vote`s are no longer allowed to have `BlockHash`.~~
`Null` and `Unknown` type `Vote`s are no longer allowed to have `BlockHash` as `null`.

### DIFF
--- a/Libplanet/Consensus/VoteSet.cs
+++ b/Libplanet/Consensus/VoteSet.cs
@@ -29,7 +29,7 @@ namespace Libplanet.Consensus
         public VoteSet(
             long height,
             int round,
-            BlockHash? blockHash,
+            BlockHash blockHash,
             IEnumerable<PublicKey> validatorSet)
         {
             Height = height;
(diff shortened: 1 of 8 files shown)


Libplanet/Consensus/VoteSet.cs
```diff
@@ -41,7 +41,7 @@ public class VoteSet
                 .Select(x => new VoteMetadata(
                     height,
                     round,
-                    blockHash,
+                    null,
```
colibrishin, line 44: Could you explain why did you make this change?

greymistcube: Initially, I thought it'd save space. On second thought, I think the requirement might have to be other way around?

greymistcube: Did a complete U-turn and now requires `BlockHash` for `Vote`s with `Null` or `Unknown` flag. 😅

colibrishin (approved):

PR #2421, opened 2022-10-19
Thread-safe NewHeight method in `ConsensusContext`

limebell:
The `ConsensusContext.NewHeight()` method was not thread-safe and could be invoked multiple times for the same height. No regression test for this because it's a timing problem and hard to reproduce.


greymistcube:
More tests seem to fail than usual with this. 😥

colibrishin (approved):

PR #2422, opened 2022-10-20
Mismatch test fixture of `ConsensusContext` with actual implementation

colibrishin:
The `ConsensusReactor` initializes `ConsensusContext` with `BlockChain.Tip.Index`. This results in the initial height of `ConsensusContext` to same as the current Blockchain tip index (e.g., If the genesis block is the tip of Blockchain, then the initial height would be 0.) However, In the test fixture, the `ConsensusContext` is initialized with `BlockChain.Tip.Index + 1`, which is different from the actual implementation.

https://github.com/planetarium/libplanet/blob/c19e5f71195934fdefa5fee12903650c919bf868/Libplanet.Net/Consensus/ConsensusReactor.cs#L75-L83

https://github.com/planetarium/libplanet/blob/c19e5f71195934fdefa5fee12903650c919bf868/Libplanet.Net/Consensus/ConsensusReactor.cs#L112-L116

https://github.com/planetarium/libplanet/blob/c19e5f71195934fdefa5fee12903650c919bf868/Libplanet.Net.Tests/TestUtils.cs#L207-L254

### DIFF
--- a/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextTest.cs
+++ b/Libplanet.Net.Tests/Consensus/ConsensusContext/ConsensusContextTest.cs
@@ -142,9 +142,9 @@ namespace Libplanet.Net.Tests.Consensus.ConsensusContext
                 TestUtils.Peer1Priv,
                 validators);
 
-            Assert.Equal(1, consensusContext.Height);
+            Assert.Equal(0, consensusContext.Height);
             blockChain.Append(blockChain.ProposeBlock(new PrivateKey()));
-            Assert.Equal(1, consensusContext.Height);
+            Assert.Equal(0, consensusContext.Height);
             await Task.Delay(newHeightDelay + TimeSpan.FromSeconds(1));
             Assert.Equal(2, consensusContext.Height);
         }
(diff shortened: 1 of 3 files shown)


greymistcube:
https://github.com/planetarium/libplanet/blob/c19e5f71195934fdefa5fee12903650c919bf868/Libplanet.Net/Consensus/ConsensusContext.cs#L45-L47

This is how it is written in the docs tho. I think the description should change. 🙄

colibrishin:
> https://github.com/planetarium/libplanet/blob/c19e5f71195934fdefa5fee12903650c919bf868/Libplanet.Net/Consensus/ConsensusContext.cs#L45-L47
> 
> This is how it is written in the docs tho. I think the description should change. 🙄

Tried to add extra detail to this.
https://github.com/planetarium/libplanet/blob/6b14c01ddc2318974c98e081efa8478cac9e8f60/Libplanet.Net/Consensus/ConsensusContext.cs#L101-L108

colibrishin:
> I think it'd be better if we remove the parameter from the constructor altogether and set `Height` initially to something like `-1` to indicate that there is no consensus under consideration. Otherwise, `Height` property does not have any real meaning without larger runtime context.

That's a good point. It seems PR #2421 needs this to be fixed, so perhaps, we could decide this later on.

greymistcube:
Sure, opened a new issue #2423.

greymistcube (approved):

PR #2431, opened 2022-10-20
Change removal `LastCommit` cache target

colibrishin:
## Context
`ClearOldLastCommitCache` removes every `LastCommit` caches if stored count is higher than `maxSize` (`lastCommitClearThreshold`.) This also removes the previous height `LastCommit`.

The problem is, for example, if a node goes down after the `NewHeight()` is called and then restarts, the `LastCommit` for the previous height does not exist in memory and even in the store.

## Changes
- Removes `LastCommit` cache in FIFO manner and it does not remove all `LastCommit` cache now. It keeps the `maxSize` count of `lastCommit` cache.
- Some chores (Add documentation for method, simplifying test.)

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -334,6 +334,12 @@ namespace Libplanet.Net.Consensus
                 _newHeightCts.Token);
         }
 
+        /// <summary>
+        /// Removes old last commit (<see cref="BlockCommit"/>) cache in store, if the cache count
+        /// is over <paramref name="maxSize"/>. The removal starts from lowest height cache and
+        /// keep the last commit cache count in <paramref name="maxSize"/>.
+        /// </summary>
+        /// <param name="maxSize">A maximum count value of <see cref="BlockCommit"/> cache.</param>
         private void ClearOldLastCommitCache(long maxSize = 30)
         {
             IEnumerable<long> indices = _blockChain.Store.GetLastCommitIndices().ToArray();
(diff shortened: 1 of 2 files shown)


Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -333,7 +333,7 @@ private void OnBlockChainTipChanged(object? sender, (Block<T> OldTip, Block<T> N
                 _newHeightCts.Token);
         }
 
-        private void ClearOldLastCommitCache(long maxSize = 30)
+        private void ClearOldLastCommitCache(long currentHeight, long maxSize = 30)
```
greymistcube, line 336: Is this kind of fine control really necessary? I think referencing `_blockChain.Tip.Index` directly might be better. 😶

colibrishin: I thought there would be a case if a block has appended while in `ClearOldLastCommitCache`, then exclude target can be changed if `_blockChain.Tip.Index - 1` is used. On second thought, It has no problem doing so due to the lack of `LastCommit` of the synced block. Still, I don't see any advantage of using `_blockChain.Tip.Index`.

greymistcube: In general, additional parameter means the caller needs to be aware of the wider context in which the method should be called. Specifically, the caller needs to know not only "when" but also "how" this method should be called. I don't think that kind of logical burdening is necessary to the outside caller.

limebell (approved):

PR #2432, opened 2022-10-21
`Context<T>.VoteSet` only returns same `BlockHash` with proposed block

colibrishin:
This PR aligns `Context<T>.VoteSet()` return value with `BlockCommit` specification. `BlockCommit` specification has changed to take `Vote`s only if its `BlockHash` is the same as given `BlockHash`.
https://github.com/planetarium/libplanet/blob/253e997cbf34d3bb7231416826d3f778f477cec5/Libplanet/Blocks/BlockCommit.cs#L85-L96

## Changes
- `Context<T>.VoteSet()` will now return `PreCommit` votes only if a vote has the same `BlockHash` with the known propose `BlockHash`.

### DIFF
--- a/Libplanet.Net/Consensus/Context.cs
+++ b/Libplanet.Net/Consensus/Context.cs
@@ -233,7 +233,14 @@ namespace Libplanet.Net.Consensus
                 ? new VoteSet(Height, round, b.Hash, _validators)
                 : throw new NullReferenceException(
                     $"Cannot create a {nameof(Libplanet.Consensus.VoteSet)} for a null block");
-            _messageLog.GetPreCommits(round).ForEach(commit => voteSet.Add(commit.PreCommit));
+            _messageLog.GetPreCommits(round)
+                .ForEach(commit =>
+                {
+                    if (commit.PreCommit.BlockHash.Equals(block.Hash))
+                    {
+                        voteSet.Add(commit.PreCommit);
+                    }
+                });
             return voteSet;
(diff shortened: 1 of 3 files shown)


Libplanet.Net/Consensus/Context.cs
```diff
@@ -233,7 +233,14 @@ public VoteSet VoteSet(int round)
...
                     $"Cannot create a {nameof(Libplanet.Consensus.VoteSet)} for a null block");
-            _messageLog.GetPreCommits(round).ForEach(commit => voteSet.Add(commit.PreCommit));
+            _messageLog.GetPreCommits(round)
+                .ForEach(commit =>
+                {
+                    if (commit.PreCommit.BlockHash.Equals(block.Hash))
+                    {
+                        voteSet.Add(commit.PreCommit);
+                    }
+                });
```
greymistcube, line 243:
```csharp
            _messageLog.GetPreCommits(round)
                .Where(commit => commit.PreCommit.BlockHash.Equals(block.Hash))
                .ForEach(commit => voteSet.Add(commit.PreCommit));
```

I'd go further with `LINQ`. 😗

colibrishin: I have avoided suggesting approach because `GetPreCommits().Where()` results have to casted in `List<T>`, `.ToList()`, again.

greymistcube (approved):

PR #2438, opened 2022-10-21
Validate lastcommit only when lastblock is PBFT based block

limebell:
It matters at the edge of PoW based block and PBFT based block.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -183,18 +183,27 @@ namespace Libplanet.Net.Consensus
             {
                 _newHeightCts?.Cancel();
 
+                _logger.Debug(
+                    "Invoke {FName}() for height #{Height} (was {WHeight}).",
+                    nameof(NewHeight),
+                    height,
+                    Height);
+
                 if (height == Height)
                 {
-                    throw new InvalidHeightIncreasingException(
-                        $"{nameof(NewHeight)}: Context of height {height} is already running.");
+                    var msg =
+                        $"{nameof(NewHeight)}: Context of height #{height} is already running.";
+                    _logger.Error(msg);
+                    throw new InvalidHeightIncreasingException(msg);
                 }
 
                 if (height != _blockChain.Tip.Index + 1)
                 {
-                    throw new InvalidHeightIncreasingException(
-                        $"{nameof(NewHeight)}: Given new height is not increasing " +
-                        $"monotonically by 1. " +
-                        $"(expected: {_blockChain.Tip.Index + 1}, actual: {height})");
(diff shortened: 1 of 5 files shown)


colibrishin (approved):

Libplanet.Net/Consensus/Context.cs
```diff
@@ -367,6 +385,14 @@ private bool IsValid(Block<T> block)
             {
                 var exception = _blockChain.ValidateNextBlock(block);
                 bool isValid = exception is null;
+                _logger.Debug(
+                    exception,
+                    "Block #{Index} {Block} is valid? {Bool}. {@E}",
```
greymistcube, line 390: I don't think `{@E}` is necessary. Also, why `{@E}` instead of `{E}`?

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -311,7 +325,26 @@ private void OnBlockChainTipChanged(object? sender, (Block<T> OldTip, Block<T> N
...
+                                "Unexpected exception occurred during {FName}(): {E}",
+                                nameof(NewHeight),
+                                exc);
+                        }
+                    }
+                    else
+                    {
+                        _logger.Error(
+                            "Did not invoke {FName}() for height " +
+                            "#{Height} because cancellation is requested.",
```
greymistcube, line 345:
```suggestion
                            "#{Height} because cancellation is requested",
```

Also see #2439.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1520,7 +1520,9 @@ public InvalidBlockException ValidateNextBlock(Block<T> block)
                     $"the block #{index - 1}'s ({prevTimestamp}).");
             }
 
-            if (block.ProtocolVersion > BlockMetadata.PoWProtocolVersion)
+            // Since the version cannot be downgraded,
+            // the target block is PBFT based block if the last block is PBFT based block.
+            if (lastBlock?.ProtocolVersion > BlockMetadata.PoWProtocolVersion)
```
greymistcube, line 1525: I think this logic needs to be checked again. `Block<T>.LastCommit` **must** be `null` on cases where we don't check for the validity of `LastCommit`.

limebell: I tried to make policy to determine it, but there was a problem making totally new field to policy and make it pass. You can find this argument on discord https://discord.com/channels/928926944937013338/998498344416579636/1032935816944701540

greymistcube: My bad. I forgot about that. Still looks a bit icky at the moment. 😥
As long as we fix this later down the line, I guess it's fine.

greymistcube:
I've made some adjustments to `LastCommit`'s nullability logic. 😶

PR #2444, opened 2022-10-21
added description tag to each element tag in libplanet/assests/curren…

yashsarnaik22:
added a description tag inside the all the items tags for the bullets on the libplanet/assets/currency page

### DIFF
--- a/Libplanet/Assets/Currency.cs
+++ b/Libplanet/Assets/Currency.cs
@@ -38,12 +38,12 @@ namespace Libplanet.Assets
     /// do not have the said limit. To define a <see cref="Currency"/> you may call either of the
     /// following.
     /// <list type="bullet">
-    /// <item><see
+    /// <item><description><see
     /// cref="Capped(string,byte,ValueTuple{BigInteger, BigInteger},IImmutableSet{Address}?)"/>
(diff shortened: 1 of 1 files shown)


Libplanet/Assets/Currency.cs
```diff
@@ -38,12 +38,12 @@ namespace Libplanet.Assets
...
     /// <list type="bullet">
-    /// <item><see
+    /// <item><description><see
     /// cref="Capped(string,byte,ValueTuple{BigInteger, BigInteger},IImmutableSet{Address}?)"/>
-    /// </item>
-    /// <item><see cref="Capped(string,byte,ValueTuple{BigInteger, BigInteger},Address)"/></item>
-    /// <item><see cref="Uncapped(string,byte,IImmutableSet{Address}?)"/></item>
-    /// <item><see cref="Uncapped(string,byte,Address)"/></item>
+    /// </description></item>
+    /// <item><description><see cref="Capped(string,byte,ValueTuple{BigInteger, BigInteger},Address)"/></description></item>
```
greymistcube, line 44: We have a linting rule that enforces 100 character limit per line, otherwise the project won't build. 😥

PR #2452, opened 2022-10-22
feat: update tags

### LINKED ISSUE #2435: Change docs to use `<see langword>` tags instead of `<c>` tags when applicable.
Currently in the codebase, both `<see langword>` and `<c>` tags are used in a mixed manner for [C# keywords]. As `<see langword>` is more appropriate for tagging [C# keywords], for consistency, I suggest changing `<c>` tags to `<see langword>` tags. Often found examples include

```xml
<c>null</c>
<c>true</c>
<c>false</c>
```

and these should be changed to

```xml
<see langword="null"/>
<see langword="true"/>
<see langword="false"/>
```

[C# keywords]: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/

khalatevarun:
Fixes #2435

### DIFF
--- a/Libplanet.Net/Protocols/KBucket.cs
+++ b/Libplanet.Net/Protocols/KBucket.cs
@@ -36,13 +36,13 @@ namespace Libplanet.Net.Protocols
 
         /// <summary>
         /// The <see cref="PeerState"/> used most recently. If the bucket is empty,
-        /// this is <c>null</c>.
+        /// this is <see langword="null"/>.
         /// </summary>
         public PeerState? Head => _peerStates.Head;
 
         /// <summary>
         /// The <see cref="PeerState"/> used longest time ago. If the bucket is empty,
-        /// this is <c>null</c>.
+        /// this is <see langword="null"/>.
         /// </summary>
         public PeerState? Tail => _peerStates.Tail;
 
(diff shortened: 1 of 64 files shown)


khalatevarun:
@greymistcube please review if the PR contains expected changes.

Libplanet.Net/Protocols/KBucket.cs
```diff
@@ -86,7 +86,7 @@ public void Clear()
         /// </summary>
         /// <param name="peer">The <see cref="BoundPeer"/> to remove.</param>
         /// <returns>
-        /// <c>true</c> if <paramref name="peer"/> was removed, <c>false</c> otherwise.
+        /// <see langword="true"/> if <paramref name="peer"/> was removed, <see langword="false"/> otherwise.
```
greymistcube, line 89: Character per line limit is set to `100`. Otherwise it won't build. You'd have to change the lines manually for this. Please try building on your local machine to see if there are any offending lines. 😗

khalatevarun: sure

khalatevarun:
@greymistcube can you check once, should be fine now

greymistcube (approved):

greymistcube:
Hope you don't mind but I've made a bit more changes so that the code can be built. 😗
Thanks for your contribution!

PR #2454, opened 2022-10-24
Open RocksDB in read-only mode when gets block or tx

### LINKED ISSUE #2443: Open RocksDB store with ReadOnly mode
To enhance development (in this case, especially debug) experience, It'll be great to open local Headless node without tons of snapshot files.
To achieve this, I'm planning to share historical block and tx data using NFS and symlink to save our 100+GB disk space.
The remaining hurdle in doing this is for Headless to open the RocksDB to validate the genesis block and create a lock.
Opening RocksDB in read-only mode can solve this problem, for this I want to add read-only mode to Libplanet.
After this job is done, Libplanet will take an optional argument so that it can open RocksDB in read-only mode if necessary. It is up to the reader which one to make read-only.

U-lis:
Resolves #2443 

To open RocksDB store in read-only mode, pass `readOnly` param.
This change allows multiple clients to access the same store at the same time.

### DIFF
--- a/Libplanet.RocksDBStore/RocksDBStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBStore.cs
@@ -633,16 +633,25 @@ namespace Libplanet.RocksDBStore
                 return null;
             }
 
-            string txDbName = RocksDBStoreBitConverter.GetString(txDbNameBytes);
             _rwTxLock.EnterReadLock();
             try
             {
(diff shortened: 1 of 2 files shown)


U-lis:
I found this change breaks attaching new block.

riemannulus:
Could you write the changes on the `CHANGE.md`?

U-lis:
Close due to broken CI

PR #2461, opened 2022-10-25
Add signable consensus Proposal

colibrishin:
This PR introduces the new classes, `Proposal` and `ProposalMetaData`. These are for replacing sending raw `ConsensusProposeMsg` without a signature of its propose.

## Role
The `Proposal` is a signable class that contains the proposal information of consensus in a height and a round. The previous method, sending a raw proposal in `ConsensusProposeMsg`, can be manipulated easily by replacing the payload of its message. 

By adding a signature to the proposal, the attacker cannot manipulate the proposal until the attacker has no access to the target victim's private key.

### DIFF
--- a/Libplanet.Net/Consensus/ProposalMetaData.cs
+++ b/Libplanet.Net/Consensus/ProposalMetaData.cs
@@ -1,0 +1,126 @@
+using System;
+using System.Collections.Immutable;
+using System.Globalization;
+using System.Text.Json.Serialization;
+using Bencodex;
+using Bencodex.Types;
+using Libplanet.Crypto;
+
+namespace Libplanet.Net.Consensus
+{
+    public class ProposalMetaData
+    {
+        private const string TimestampFormat = "yyyy-MM-ddTHH:mm:ss.ffffffZ";
+        private const string HeightKey = "height";
+        private const string RoundKey = "round";
+        private const string BlockKey = "block";
+        private const string TimestampKey = "timestamp";
+        private const string ValidatorKey = "validator";
+        private const string ValidRoundKey = "validRound";
+
+        private static Codec _codec = new Codec();
+
+        public ProposalMetaData(
+            long height,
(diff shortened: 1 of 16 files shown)


Libplanet.Net/Consensus/ProposalMetaData.cs
```diff
@@ -0,0 +1,126 @@
...
+                    nameof(validRound),
+                    "ValidRound must be greater than or equal to -1.");
+            }
+
+            Height = height;
+            Round = round;
+            BlockMarshaled = blockMarshaled;
+            Timestamp = timestamp;
+            Validator = validator;
+            ValidRound = validRound;
```
greymistcube, line 56: Why do we need valid round?

Libplanet.Net/Consensus/ProposalMetaData.cs
```diff
@@ -0,0 +1,126 @@
...
+                    .Add(ValidRoundKey, ValidRound);
+
+                return encoded;
+            }
+        }
+
+        [JsonIgnore]
+        public byte[] ByteArray => _codec.Encode(Encoded);
+
+        public Proposal Sign(PrivateKey? signer)
```
greymistcube, line 109: Unlike `VoteMetadata`, I don't think there any use for unsigned `Proposal`.

Libplanet.Net/Consensus/ProposalMetaData.cs
greymistcube: Nevermind, found it in the paper. 🙃

Libplanet.Net/Consensus/ProposalMetaData.cs
colibrishin: Amended. fec4039ea3f71effb48e0f621e24c7ce009f936d

greymistcube (approved):

PR #2464, opened 2022-10-26
🔨 Prevent invalid block proposal

greymistcube:
🔨

### DIFF
--- a/Libplanet.Net.Tests/Consensus/Context/ContextProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/Context/ContextProposerTest.cs
@@ -1,10 +1,8 @@
-using System;
 using System.Threading.Tasks;
 using Libplanet.Blocks;
 using Libplanet.Consensus;
 using Libplanet.Net.Consensus;
 using Libplanet.Net.Messages;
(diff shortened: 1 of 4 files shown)


Libplanet.Net.Tests/Consensus/Context/ContextProposerTest.cs
```diff
@@ -281,29 +279,10 @@ void CheckVote(object? observer, ConsensusMsg? message)
...
-                    publicKey: fx.Miner.PublicKey,
-                    previousHash: blockChain.Tip.Hash,
-                    txHash: null,
-                    lastCommit: null))
-                .Propose().Evaluate(fx.Miner, blockChain);
-
-            context.ProduceMessage(
-                TestUtils.CreateConsensusPropose(
-                    invalidBlock, TestUtils.Peer1Priv));
-
```
colibrishin, line 303: Can we have a test case where the node is not a proposer and receives an invalid proposal?

greymistcube: See 6a98b6049090c9c40425f7b1df5b49cadf14e6b0.

colibrishin (approved):

PR #2465, opened 2022-10-26
Test cleanup

greymistcube:
This mainly deals with removing custom `Context` initial state setup.

As `Context` is a rather complex structure in and of itself, it is **extremely hard** to know all the side-effects of simply setting its round and step to a desired value. Under normal operations, any state of a non-zero round and a step implies either that particular step has been reached by multiple iterations of timeouts and/or failure to form a consensus on a network where numerous messages have been received (even if the progression is made by purely timeouts, **there will be messages received** from the node itself, meaning it is never a clean slate).

This results in test codes not being clear as to whether tested methods are in fact being tested under right circumstances.

### DIFF
--- a/Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
+++ b/Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
@@ -47,23 +47,15 @@ namespace Libplanet.Net.Tests.Consensus.Context
                     stateChangedToRoundOnePreVote.Set();
                 }
             };
+
             context.Start();
-
-            context.ProduceMessage(
-                TestUtils.CreateConsensusPropose(
-                    block, TestUtils.Peer1Priv, round: 0));
-
             context.ProduceMessage(
                 TestUtils.CreateConsensusPropose(
                     block, TestUtils.Peer2Priv, round: 1));
-
             context.ProduceMessage(
                 new ConsensusPreVoteMsg(
                     TestUtils.CreateVote(
-                        TestUtils.Peer2Priv, 1, 1, hash: block.Hash, flag: VoteFlag.PreVote))
-                {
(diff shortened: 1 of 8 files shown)


Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
```diff
@@ -47,23 +47,18 @@ public async Task EnterPreVoteBlockOneThird()
                     stateChangedToRoundOnePreVote.Set();
                 }
             };
-            context.Start();
 
+            context.Start();
             context.ProduceMessage(
                 TestUtils.CreateConsensusPropose(
                     block, TestUtils.Peer1Priv, round: 0));
```
OnedgeLee, line 54: A bit confusing, do we need this line?

OnedgeLee:
Reading these lines, I thought that it would be much convenient to have access to `_messageLog` of `Context`, though I agree that it's not a good design 😢

Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
greymistcube: Just some old artifact I suppose. I don't think we really need it. 😗

Libplanet.Net.Tests/Consensus/Context/ContextNonProposerTest.cs
```diff
@@ -289,18 +258,13 @@ public async Task EnterPreVoteNilOneThird()
...
```
OnedgeLee, line 260:
```suggestion
```
Still there are some lines like here, but since they're old artifact, won't review for those (cannot add review on those lines)

OnedgeLee (approved):

PR #2470, opened 2022-10-26
Open RocksDB in read-only mode when opening past epochs

### LINKED ISSUE #2443: Open RocksDB store with ReadOnly mode

U-lis:
Resolves #2443

To open RocksDB store in read-only mode, pass readOnly param.
This change allows multiple clients to access the same store at the same time.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -53,6 +53,8 @@ Released on November 2, 2022.
 ### Behavioral changes
 
  - Optimized `ByteUtil.CalculateHash()` method.  [[#2437], [#2459]]
+ - (Libplanet.RocksDBStore) `GetBlockDigest`, `GetTranaction` methods open
+   RocksDB in read-only mode when opening past epochs.  [[#2443]]
 
 ### Bug fixes
 
--- a/Libplanet.RocksDBStore/RocksDBStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBStore.cs
@@ -633,16 +633,25 @@ namespace Libplanet.RocksDBStore
                 return null;
             }
 
-            string txDbName = RocksDBStoreBitConverter.GetString(txDbNameBytes);
             _rwTxLock.EnterReadLock();
             try
             {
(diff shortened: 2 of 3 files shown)


moreal (commented):
IMHO, it shouldn't block to put previous timestamp blocks.

I wanted to consider deep reorganization case. Of course, I know Nine Chronicles has fixed block at every 50 blocks to get finality. But Libplanet is not only, so I think this problem should be considered.

To describe the 'deep reorganization case', my chain's tip is 100 and I'm putting blocks at 2 epoch store. But someone proposed strange block. It shares same genesis block, its total difficulty is big than my chain and the chain is different since second block with mine. Then I should replace my canonical chain with it and I downloaded the 104 blocks. And the epoch store to put the blocks will be epoch 0, 1, 2. But current latest epoch store is 2 and others will be open as readonly. And it may occur error while putting blocks 💣 

(Assumed to use 50 as epoch interval and block index as dividend) (i.e. epoch = block index / 50)

Of course, it may not have any problems to consider after migrating to pbft(?)

riemannulus:
@U-lis Could you check this PR again? I'll close this issue if staled until Feb 3.

U-lis:
> @U-lis Could you check this PR again? I'll close this issue if staled until Feb 3.

I sill think read-only mode makes sense, but I also agree with @moreal 's opinion above.
Since `libplanet` is general library not 9c specific blockchain, so I'll close this PR at the moment.
We could discuss this issue when more needs are raised in the future.

PR #2480, opened 2022-10-28
Prevent failure in `NewHeight` when getting `LastCommit`

limebell:
Moved logic into `Context` (was `ConsensusContext`) and prevent uncaught exceptions that could be occurred while creating `BlockCommit` instance.


colibrishin (approved):

limebell:
Seems like this PR is not required anymore.

PR #2486, opened 2022-10-31
Possible empty peer table of `Gossip`

colibrishin:
## Context
There is a case where a `Gossip` can be initialized with an empty peer table if the peer is the first peer in the seed node.

The `Gossip` protocol starts with the `CheckValidatorsLiveness` task, which is for checking whether other nodes are alive. This task uses peer table `Gossip._table` to dial up and check liveness, however, this table was checked and updated only once before the task. 
If the task failed to check liveness due to an empty peer table, then the `CheckValidatorsLiveness` is trying to check liveness against the empty peer table indefinitely.

This PR tries to address the problem by:

- Updating a peer table in `CheckValidatorsLiveness()` for every retrying
- Add a new regular task for updating the peer table of `Gossip` from the seed node.

## Changes
- Moved `CheckValidatorsLiveness()` invocation to `ConsensusReactor.StartAsync()`
- In `CheckValidatorsLiveness()`, invoke `UpdateTable()` if a count of the peer table is lower than `MinimumPeerCount`.
- The condition of break out from `CheckValidatorsLiveness()` is changed to `countOfPong >= MinimumPeerCount`.
- `HeartbeatTask()` does not invoke `UpdateTable()` anymore.
- Added a new task `RefreshTableAsync()` which invokes `UpdateTable()` for an interval.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusReactor.cs
+++ b/Libplanet.Net/Consensus/ConsensusReactor.cs
@@ -110,8 +110,10 @@ namespace Libplanet.Net.Consensus
         /// <returns>Returns the <see cref="ITransport.StartAsync"/>.</returns>
         public async Task StartAsync(CancellationToken cancellationToken)
         {
+            Task task = _gossip.StartAsync(cancellationToken);
+            await _gossip.CheckValidatorsLiveness(cancellationToken);
             _consensusContext.NewHeight(_blockChain.Tip.Index + 1);
-            await _gossip.StartAsync(cancellationToken);
+            await task;
         }
 
         /// <summary>
--- a/Libplanet.Net/Consensus/Gossip.cs
+++ b/Libplanet.Net/Consensus/Gossip.cs
@@ -16,6 +16,7 @@ namespace Libplanet.Net.Consensus
     /// </summary>
     public class Gossip : IDisposable
     {
+        private const int MinimumPeerCount = 4;
         private const int DLazy = 6;
         private readonly TimeSpan _heartbeatInterval = TimeSpan.FromSeconds(1);
         private readonly TimeSpan _seenTtl;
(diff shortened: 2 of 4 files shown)


Libplanet.Net/Consensus/Gossip.cs
```diff
@@ -401,6 +433,14 @@ private async Task CheckValidatorsLiveness(CancellationToken ctx)
         {
             while (!ctx.IsCancellationRequested)
             {
+                if (!_table.Any())
```
limebell, line 436: IMO this condition is not very powerful. New block can be committed only when 2/3 of the validator power exists in the network, but `_table` at this moment is a list of peers given by `--consensus-peer`s in Headless argument.

Libplanet.Net/Consensus/ConsensusReactor.cs
```diff
@@ -110,8 +110,9 @@ public void Dispose()
         /// <returns>Returns the <see cref="ITransport.StartAsync"/>.</returns>
         public async Task StartAsync(CancellationToken cancellationToken)
         {
+            _ = _gossip.StartAsync(cancellationToken);
+            await _gossip.WaitForRunningAsync();
             _consensusContext.NewHeight(_blockChain.Tip.Index + 1);
```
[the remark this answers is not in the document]
colibrishin: Amended. https://github.com/planetarium/libplanet/pull/2486/commits/53a338b79e667b9bc3405269e99738d1892122d9

Libplanet.Net/Consensus/Gossip.cs
colibrishin: For the weak condition, the `MinimumPeerCount` is added. If the `countOfPong >= MinimumPeerCount`, then consensus `Gossip` would not proceed (blocking). The value of `MinimumPeerCount` is chosen 4, which is the least value of BFT (`where f = 1`).

limebell (approved):

limebell (approved):

PR #2490, opened 2022-11-01
Feat/enrich explorer query

area363:
## Changes
1. Add a `SignedTx` field in the `chainquery transaction` data.
- The `SignedTx` field is added to test chain consistency during internal testing (to test whether past txs work in new updates).
- Reference Issue: https://github.com/planetarium/devops/issues/13

2. Use `IterateIndexes` instead of the `GetNextBlock` method when querying blocks.
- This change improves querying speed especially when the `offset` value is far from either 0 or the tip value (the original implementation only worked well for getting blocks close to the genesis or the tip block).

### DIFF
--- a/Libplanet.Explorer/GraphTypes/TransactionType.cs
+++ b/Libplanet.Explorer/GraphTypes/TransactionType.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using GraphQL;
 using GraphQL.Types;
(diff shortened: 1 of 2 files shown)


Libplanet.Explorer/GraphTypes/TransactionType.cs
```diff
@@ -54,6 +55,14 @@ public TransactionType()
                 name: "Actions",
                 description: "A list of actions in this transaction."
             );
+            Field<NonNullGraphType<StringGraphType>>(
+                name: "SignedTx",
```
dahlia, line 59: How about renaming it to `SerializedPayload` to avoid confusion with `Signature`?

area363: Sure! I think that's a better name 👍

dahlia (approved):

PR #2494, opened 2022-11-01
Parameter for how many `Context<T>` can be alive

colibrishin:


colibrishin:
This is closed due to further investigation needed.

PR #2498, opened 2022-11-01
Isolate problematic tests

OnedgeLee:
### What have done
- [x] Fix flaky netcore tests
- [x] Isolate flaky unity tests
- [x] Limit unity test coverage

### Remarks
- Flaky test methods for unity tests are isolated with `[SkippableFact]` and `[SkippableTheory]`.
- If flaky part is on test class constructor, the whole class is skipped by CI(since there are no fancy way to skip the whole test class on xunit).
- All isolated flaky tests are from signature validation error on block mining.
- Flaky tests that causes timeout are still exist, but they're not identified yet.

### Next steps
1. Find tests that causes timeout
2. Isolate tests that causes timeout
3. Figure out the reason that causes signature validation error and fix it
4. Figure out the reason that causes timeout and fix it

### DIFF
--- a/Libplanet.Tests/Blocks/BlockMetadataTest.cs
+++ b/Libplanet.Tests/Blocks/BlockMetadataTest.cs
@@ -383,11 +383,6 @@ namespace Libplanet.Tests.Blocks
                 source.Cancel();
                 bool taskEnded = task.Wait(TimeSpan.FromSeconds(10));
                 Assert.True(taskEnded);
-                if (!(hash is null))
-                {
-                    Assert.True(ByteUtil.Satisfies(hash, long.MaxValue));
-                }
-
                 Assert.Null(nonce);
                 Assert.Null(hash);
                 Assert.NotNull(exception);
--- a/Libplanet.Tests/Blocks/BlockMetadataTest.cs
+++ b/Libplanet.Tests/Blocks/BlockMetadataTest.cs
@@ -360,17 +360,19 @@ namespace Libplanet.Tests.Blocks
                     txHash: Block1Metadata.TxHash);
 
                 Exception exception = null;
+                Nonce? nonce = null;
+                ImmutableArray<byte>? hash = null;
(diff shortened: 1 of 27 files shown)


Libplanet.Tests/Blocks/BlockMetadataTest.cs
```diff
@@ -360,28 +360,36 @@ public void CancelMineNonce(int? workers)
...
                 source.Cancel();
                 bool taskEnded = task.Wait(TimeSpan.FromSeconds(10));
                 Assert.True(taskEnded);
+                if (!(hash is null))
+                {
+                    Assert.True(ByteUtil.Satisfies(hash, long.MaxValue));
+                }
+
+                Assert.Null(nonce);
+                Assert.Null(hash);
```
greymistcube, line 392: I think we can get rid of `if (!(hash is null))`? If `hash` is not null, the test won't pass anyway due to `Assert.Null(hash)` statement.

OnedgeLee: What I've thought was,
"If mining has not been correctly canceled, nonce will exist, and we have to know if generated nonce matches the given difficulty."
Actually, at the first time mining wasn't canceled, and some nonce was generated.
But I couldn't know why mining with maximum difficulty successes within short time, so I've examined if given difficulty properly works, and found difficulty wasn't properly passed.

I hesitated to remove those statements because still this can give some evidence, but this idea can be biased. I'll hear your opinion.

greymistcube: Several things:
- `if` statement is a dead piece of code **when working as expected**.
- `Assert.Null()` already gives what is inside `nonce` and `hash` if one of them is not `null`.
- Testing the wrong behavior, i.e. whether returned value satisfies the difficulty, is not in the scope of the test.

limebell (approved):

dahlia (approved):

PR #2504, opened 2022-11-02
Lock synchronization on running `ConsensusReactor`

colibrishin:


colibrishin:
In hindsight, I don't see any reason why the block synchronization should be locked in the current implementation. Closing PR.

PR #2507, opened 2022-11-02
Inject ActionEvaluator and BlockChainStates explicitly

longfin:
This PR aims to extract `ActionEvaluator<T>` creation, from `BlockChain<T>` constructor to outside. to achieve that, we need to solve the dependency chain as below.

1. `ActionEvaluator<T>` depends `BlockChain<T>` as `IBlockChainStates<T>`.
2. To perform as `IBlockChainStates<T>`, `BlockChain<T>` depends `IStore`, `IStateStore`, and itself. (due to state completion).

In other words, if we can implement `IBlockChainState<T>` without the whole `BlockChain<T>`, extracting `ActionEvaluator<T>` creation from `BlockChain<T>` constructor seems possible. 

* At first, I introduced a new `BlockChainStates<T>` implementing `IBlockChainStates<T>`, instead of `BlockChain<T>`.
* Moved `GetBlances()`, `GetStates()` and `GetTotalSupply()` from `BlockChain<T>` to `BlockChainStates<T>`.
  * I think that we have a rough consensus about obsoleting the state completion feature, but it can be another long story. so I deferred it to later PR.
  * To keep current behavior using `StateCompleter<T>`s, I introduced `BlockChainStates<T>.Bind()`. it takes `BlockChain<T>` to support previous methods.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -215,9 +215,9 @@ namespace Libplanet.Blockchain
             Store = store;
             _blockChainStates = blockChainStates;
 
-            if (_blockChainStates is BlockChainStates<T> biddableImpl)
+            if (_blockChainStates is BlockChainStates<T> bindableImpl)
             {
-                biddableImpl.Bind(this);
+                bindableImpl.Bind(this);
             }
 
             // It expects store is DefaultStore or RocksDBStore.
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -142,7 +142,7 @@ namespace Libplanet.Blockchain
 
 #pragma warning disable MEN002
 #pragma warning disable CS1573
-        /// <inheritdoc cref="BlockChain{T}(IBlockPolicy{T}, IStagePolicy{T}, IStore, IStateStore, Block{T}, IEnumerable{IRenderer{T}})" />
+        /// <inheritdoc cref="BlockChain{T}(IBlockPolicy{T}, IStagePolicy{T}, IStore, IStateStore, Block{T}, IEnumerable{IRenderer{T}}, IBlockChainStates{T})" />
         /// <param name="actionEvaluator">The <see cref="ActionEvaluator{T}" /> implementation to calculate next states when append new blocks.</param>
         public BlockChain(
             IBlockPolicy<T> policy,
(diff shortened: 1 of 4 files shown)


dahlia (approved):

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -89,14 +92,81 @@ public partial class BlockChain<T> : IBlockChainStates<T>
...
+                    genesisHash: genesisBlock.Hash,
+                    nativeTokenPredicate: policy.NativeTokens.Contains
+                )
+            )
+        {
+        }
+
+#pragma warning disable MEN002
+#pragma warning disable CS1573
+        /// <inheritdoc cref="BlockChain{T}(IBlockPolicy{T}, IStagePolicy{T}, IStore, IStateStore, Block{T}, IEnumerable{IRenderer{T}})" />
```
longfin, line 145:
```suggestion
        /// <inheritdoc cref="BlockChain{T}(IBlockPolicy{T}, IStagePolicy{T}, IStore, IStateStore, Block{T}, IEnumerable{IRenderer{T}}, IBlockChainStates{T})" />
```

😓

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -131,13 +205,20 @@ IEnumerable<IRenderer<T>> renderers
...
         {
             Id = id;
             Policy = policy;
             StagePolicy = stagePolicy;
             Store = store;
+            _blockChainStates = blockChainStates;
+
+            if (_blockChainStates is BlockChainStates<T> biddableImpl)
+            {
+                biddableImpl.Bind(this);
```
tkiapril, line 220: What does biddable mean?

tkiapril (approved):
Just a little question, but what would happen if Store and StateStore objects passed to BlockChain constructor is different from what was passed to BlockChainStates constructor that is passed to the BlockChain constructor?

Libplanet/Blockchain/BlockChain.cs
longfin: It seems just a typo. 😅

tkiapril (approved):

longfin:
> Just a little question, but what would happen if Store and StateStore objects passed to BlockChain constructor is different from what was passed to BlockChainStates constructor that is passed to the BlockChain constructor?

Basically, I believe the responsibility for the chain state and its calculation in `Blockchain<T>` should be lesser than now. but at least for now, it may cause errors or unintended behaviors. therefore a kind of check or guard would be good too.

PR #2510, opened 2022-11-03
Collision between indexing and creating in `Dictionary<long, Context<T>>`

colibrishin:
This PR adds the additional locks for preventing collision in `Dictionary<long, Context<T>>`.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -111,7 +111,16 @@ namespace Libplanet.Net.Consensus
         /// <returns>If there is <see cref="Context{T}"/> for <see cref="Height"/> returns the round
         /// of current <see cref="Context{T}"/>, or otherwise returns -1.
         /// </returns>
-        public long Round => _contexts.ContainsKey(Height) ? _contexts[Height].Round : -1;
+        public long Round
+        {
+            get
+            {
+                lock (_contextLock)
+                {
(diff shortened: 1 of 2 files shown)


Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -120,13 +129,31 @@ public partial class ConsensusContext<T> : IDisposable
...
+        internal Dictionary<long, Context<T>> Contexts
+        {
+            get
+            {
+                lock (_contextLock)
+                {
+                    return _contexts;
+                }
+            }
+        }
```
greymistcube, line 156: This change isn't necessary. `lock`s are there to lock the state of `_contexts`. This is only to return the reference to `_contexts`. 😗

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -351,14 +388,15 @@ private void ClearOldLastCommitCache(long maxSize = 30)
         /// <param name="height">The upper bound of height of the contexts to be discarded.</param>
         private void RemoveOldContexts(long height)
         {
-            var values = _contexts.Values;
-            foreach (var ctx in values)
+            lock (_contextLock)
             {
-                if (ctx.Height < height)
+                var values = _contexts.Values;
+                foreach (var ctx in values)
```
greymistcube, line 394:
```csharp
                foreach (var ctx in _contexts.Values)
```

We can do away with extraneous variable `values` since `_contexts` is locked now.

PR #2516, opened 2022-11-04
chore: change minimum peer to 3

limebell:
It does not count itself.

### DIFF
--- a/Libplanet.Net/Consensus/Gossip.cs
+++ b/Libplanet.Net/Consensus/Gossip.cs
@@ -16,7 +16,7 @@ namespace Libplanet.Net.Consensus
     /// </summary>
     public class Gossip : IDisposable
     {
-        private const int MinimumPeerCount = 4;
+        private const int MinimumPeerCount = 3;
         private const int DLazy = 6;
         private readonly TimeSpan _refreshTableInterval = TimeSpan.FromSeconds(1);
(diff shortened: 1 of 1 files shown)


colibrishin (approved):

colibrishin:
This is out of scope of this PR, but I think it is related in the aspect of a specification. Doesn't node has to add itself in a  validator set?

`Gossip` doesn't affected by this but a validator set seems to have itself. And from what I recall, our current implementation should have itself in the validator set for broadcasting (In other word, for re-feed into `ConsensusContext<T>`) own vote or proposal by a transport (Gossip).

limebell:
> This is out of scope of this PR, but I think it is related within the aspect of a specification. Doesn't node has to add itself in a validator set?
> 
> `Gossip` doesn't affected by this but a validator set seems to have itself. And from what I recall, our current implementation should have itself in the validator set for broadcasting (In other word, for re-feed into `ConsensusContext<T>`) own vote or proposal by a transport (Gossip).

Currently, `RoutingTable` does not allow it to add itself to the table to prevent sending messages to self, thus validator set for broadcasting does not include itself. I also prefer to add self to its table so that a node can send messages to itself for convenience.

limebell:
I'll close this PR because the variable is removed in #2515

PR #2517, opened 2022-11-04
Move `LastCommit` saving to `Commit`

colibrishin:
Saving LastCommit in `Context<T>.Dispose()` was hard to debug. So the LastCommit saving point is moved from `Context<T>.Dispose()` to `Commit`.

### DIFF
--- a/Libplanet.Net/Consensus/Context.cs
+++ b/Libplanet.Net/Consensus/Context.cs
@@ -213,20 +213,6 @@ namespace Libplanet.Net.Consensus
         /// <inheritdoc cref="IDisposable.Dispose()"/>
         public void Dispose()
         {
-            // Save the current height LastCommit before disposing.
-            // XXX: The reasons why it is seperated from ConsensusContext.NewHeight() is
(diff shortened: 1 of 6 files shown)


Libplanet.Net/Consensus/Context.cs
```diff
@@ -468,5 +454,19 @@ private bool HasOneThirdValidators(int round)
...
+            {
+                return;
+            }
+
+            _blockChain.Store.PutLastCommit(currentLastCommit);
+            _logger.Debug(
+                "Saving current height #{Height} and round {Round} LastCommit...",
+                currentLastCommit.Height,
+                currentLastCommit.Round);
+        }
```
greymistcube, line 470: Now that I look at it, I think direct access to `IStore` from `Context` might not be that great. Since `ConsensusContext` already does the loading, how about moving this to `ConsnesusContext.Commit()`?

colibrishin: Sure, and I also forgot to remove the "Save LastCommit with `Dispose()`" unit test. 🤧

limebell (approved):

Issue #2424, opened 2022-10-20
Failed to process `Libplanet.Net.Messages.PingMsg` during Preload


area363:
Error Message:
```
[06:38:20 ERR] [NetMQTransport] Failed to process Libplanet.Net.Messages.PingMsg fc2cb51b-4f8f-4cfc-a045-b11eb3f8347a; discarding it.											
System.Net.Internals.SocketExceptionFactory+ExtendedSocketException (00000001	 11): Resource temporarily unavailable												
   at System.Net.Dns.GetHostEntryOrAddressesCore(String hostName	 Boolean justAddresses	 AddressFamily addressFamily	 ValueStopwatch stopwatch)							
   at System.Net.Dns.GetHostEntry(String hostNameOrAddress	 AddressFamily family)												
   at System.Net.Dns.GetHostEntry(String hostNameOrAddress)													
   at NetMQ.Core.Transports.Tcp.TcpAddress.Resolve(String name	 Boolean ip4Only)												
   at NetMQ.Core.SocketBase.Connect(String addr)													
   at NetMQ.NetMQSocket..ctor(ZmqSocketType socketType	 String connectionString	 DefaultAction defaultAction)											
   at NetMQ.Sockets.DealerSocket..ctor(String connectionString)													
   at Libplanet.Net.Transports.NetMQTransport.GetRequestDealerSocket(MessageRequest request) in /app/Lib9c/.Libplanet/Libplanet.Net/Transports/NetMQTransport.cs:line 656													
   at Libplanet.Net.Transports.NetMQTransport.ProcessRequest(MessageRequest req	 CancellationToken cancellationToken) in /app/Lib9c/.Libplanet/Libplanet.Net/Transports/NetMQTransport.cs:line 695												
   at Libplanet.Net.Transports.NetMQTransport.ProcessRuntime(CancellationToken cancellationToken) in /app/Lib9c/.Libplanet/Libplanet.Net/Transports/NetMQTransport.cs:line 628
```

Please refer to https://planetariumhq.slack.com/archives/CCBRB031P/p1666166947180729 for specific context.

PR #2521, opened 2022-11-05
Fix hangs on .SendMessageAsync() & Release v0.43.3

longfin:
This PR fixes freezing on `NetMQTransport.SendMessageAsync()`. (had been reported as #2424)

### DIFF
--- a/Libplanet.Net.Tests/Transports/TransportTest.cs
+++ b/Libplanet.Net.Tests/Transports/TransportTest.cs
@@ -1,5 +1,6 @@
 #nullable disable
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


Libplanet.Net.Tests/Transports/TransportTest.cs
```diff
@@ -140,45 +141,6 @@ public void AsPeer()
         }
 
         // This also tests ITransport.ReplyMessageAsync at the same time.
-        [SkippableFact(Timeout = Timeout)]
-        public async Task SendMessageAsync()
```
greymistcube, line 144: Why is this test removed? Do we not want to test "normal" successful behavior?

longfin: It wasn't intended deletion. I'll revert.

longfin: Amended. (https://github.com/planetarium/libplanet/pull/2521/commits/bf7bba0bf2e0746e89885d2767cb0dff993bbb3b)

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -893,6 +890,15 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
...
             }
+            catch (Exception ae)
+            {
+                var se = new SendMessageFailException(
+                    $"Unexpected exception occurred during {nameof(ProcessRequest)}().",
+                    req.Peer,
+                    ae
+                );
+                tcs.TrySetException(se);
+            }
```
greymistcube, line 901: Now that we know more about `GetRequestDealerSocket()`, I think it is no longer an "unexpected exception". Perhaps wrapping `NetMQException` inside `SendMessageFailException` should also be done inside `GetRequestDealerSocket()` (instead of throwing `MetMQException` directly) and be caught by `catch when` statement above.

longfin: Well... I felt that we might have to wrap all exceptions as `SendMessageFailExceptionn` (at least on the current structure) since it will hang again if there is any unexpected exception. (actually, we've faced `System.Net.Internals.SocketExceptionFactory+ExtendedSocketException ` even on #2424).

Of course, this may be also a signal that refactorings are needed. but considering that Libplanet is a library that is already used in production, it is also necessary to take a conservative approach to errors of known patterns (i.e., hangs when unexpected errors occur).

greymistcube: Very well then, we can postpone it I suppose.

greymistcube (approved):

PR #2522, opened 2022-11-05
Move timeout related codes to `SendMessageAync()`

longfin:
(I intended to limit the amount of #2521 since it's a hotfix. 😓 , so this PR would be merged into `main` after #2521 and 0.43.3 release)

This PR is the following work of #2521. it moves checks for timeout and firing codes, from `ProcessRequest()` to `SendMessageAsync()`.

Although the semantic has changed depending on whether or not to put the time to stay in the request queue(`_requests`) in the timeout, I was worried about whether it would actually be meaningful to put this semantic, so I omitted changelog now. but feel free to comment if you think it is necessary.

### DIFF
--- a/Libplanet.Net/Transports/NetMQTransport.cs
+++ b/Libplanet.Net/Transports/NetMQTransport.cs
@@ -858,6 +858,8 @@ namespace Libplanet.Net.Transports
 
                     await channel.Writer.WriteAsync(reply, cancellationToken);
                 }
+
+                channel.Writer.Complete();
             }
             catch (Exception e)
             {
--- a/Libplanet.Net/Transports/NetMQTransport.cs
+++ b/Libplanet.Net/Transports/NetMQTransport.cs
@@ -385,10 +385,20 @@ namespace Libplanet.Net.Transports
                 throw new ObjectDisposedException(nameof(NetMQTransport));
             }
 
+            using var timerCts = new CancellationTokenSource();
+            if (timeout is { } timeoutNotNull)
+            {
+                timerCts.CancelAfter(timeoutNotNull);
+            }
+
             using CancellationTokenSource cts =
                 CancellationTokenSource.CreateLinkedTokenSource(
                     _runtimeCancellationTokenSource.Token,
-                    cancellationToken);
+                    cancellationToken,
+                    timerCts.Token
+                );
+            CancellationToken linkedToken = cts.Token;
+
             Guid reqId = Guid.NewGuid();
             try
             {
@@ -400,23 +410,20 @@ namespace Libplanet.Net.Transports
                     message
                 );
                 var tcs = new TaskCompletionSource<IEnumerable<Message>>();
+                using CancellationTokenRegistration cancelRegistration =
+                    linkedToken.Register(() => tcs.TrySetCanceled());
                 Interlocked.Increment(ref _requestCount);
-
-                // FIXME should we also cancel tcs sender side too?
-                using CancellationTokenRegistration ctr =
-                    cts.Token.Register(() => tcs.TrySetCanceled());
-                MessageRequest req = new MessageRequest(
+                var req = new MessageRequest(
                     reqId,
                     message,
                     peer,
                     now,
-                    timeout,
                     expectedResponses,
-                    returnWhenTimeout,
-                    tcs);
+                    tcs
+                );
                 await _requests.Writer.WriteAsync(
                     req,
-                    cts.Token
+                    linkedToken
                 );
                 _logger.Verbose(
                     "Enqueued a request {RequestId} to the peer {Peer}: {@Message}; " +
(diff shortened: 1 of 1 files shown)


Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -910,18 +901,16 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
...
 
                 Interlocked.Decrement(ref _socketCount);
-                timerCts.Dispose();
 
                 _logger
                     .ForContext("Tag", "Metric")
                     .ForContext("Subtag", "OutboundMessageReport")
                     .Debug(
-                        "Request {Message} {RequestId} with timeout {TimeoutMs:F0}ms " +
+                        "Request {Message} {RequestId}" +
```
greymistcube, line 909:
```suggestion
                        "Request {Message} {RequestId} " +
```

longfin:
I added https://github.com/planetarium/libplanet/pull/2522/commits/7db5426d0c04f8ac49c9f19221528450fbf94750 to address two topics.

- By adding `CancellationToken` to `MessageRequest`, manual termination of `TaskCompletionSource<T>` (i.e., `linkedToken.Register(() => tcs.TrySetCanceled()`) on the caller side can be removed.
- Removing branches for exceptions on runtime also can be achieved.
  - All exceptions on `ProcessRequest()` will be forwarded to the caller via `tcs`, without wrapping. 
  - To keep the previous behavior, it will not set exceptions but set the result when `dealerSocket` has received one or more messages from the remote.
    - It felt a little counter-intuitive, but I think it's effective to separate the refactoring and the change in the behavior.

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -439,9 +446,26 @@ CancellationToken cancellationToken
...
                 {
-                    return new Message[0];
+                    return Array.Empty<Message>();
                 }
             }
+            catch (TaskCanceledException toe) when (timerCts.IsCancellationRequested)
+            {
+                if (returnWhenTimeout)
+                {
+                    return Array.Empty<Message>();
```
limebell, line 456: `returnWhenTimeout` was intended to return received messages by the timeout. For example, if 10 replies were expected and only 5 messages were received when the timeout was fired, 5 messages will be returned when `returnWhenTimeout` is set to true. If `returnWhenTimeout` is false, then throw a timeout exception.

longfin: Thanks for pointing. then as my understanding, this behavior will only be needed when `returnWhenTimeout` is true. is it right?

https://github.com/planetarium/libplanet/pull/2522/files#diff-87d2edd35fa01273fa9229c93431561886cf57fae9e11593c90c4165a89c2ce7R878-R881

limebell: Yes. it can look little weird but required for the situation when broadcasting big amount of staged transactions and the timeout occurs while receiving some of them.

longfin: 🆗 I'm going to try to unify those considerations... 🤔

longfin:
I modified the code to leverage `Channel<Message>` instead of `TaskCompletionSource`, due to a kind of streaming for previous behavior compatibility as @limebell mentioned. cc @greymistcube

Libplanet.Net/Transports/NetMQTransport.cs
longfin: @limebell Added https://github.com/planetarium/libplanet/pull/2522/commits/a7efbe1d0fead17b8d8e7e2c2d1f7efdb52e9dd8 to address. PTAL

limebell (approved):

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -808,98 +804,64 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
...
-            }
-            catch (Exception ae)
-            {
-                var se = new SendMessageFailException(
-                    $"Unexpected exception occurred during {nameof(ProcessRequest)}().",
-                    req.Peer,
-                    ae
-                );
-                tcs.TrySetException(se);
+                channel.Writer.Complete(e);
```
greymistcube, line 864: Perhaps may not be strictly necessary, but maybe we should write `Complete()` on successful operation of `for` loop also? 😶

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -808,98 +804,64 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
...
+                            reply.Remote,
+                            message,
+                            req.Id);
                         throw;
                     }
-                }
 
-                tcs.TrySetResult(result);
+                    await channel.Writer.WriteAsync(reply, cancellationToken);
+                }
```
longfin, line 860:
```suggestion
                }

                channel.Writer.Complete();
```

Libplanet.Net/Transports/NetMQTransport.cs
longfin: Like https://github.com/planetarium/libplanet/pull/2522/files#r1018739172 ?

longfin: Added https://github.com/planetarium/libplanet/pull/2522/commits/5a41085db23b04e8ead8abb0b3d66259c38e4f28.

greymistcube (approved):

PR #2527, opened 2022-11-08
Added additional checks when proposing a block

greymistcube:
Just an additional guardrail to prevent proposing an invalid block.
As a proposer, it was possible to add an invalid `ConsensusProposalMsg` to `MessageLog`.

### DIFF
--- a/Libplanet.Net/Consensus/Context.cs
+++ b/Libplanet.Net/Consensus/Context.cs
@@ -307,14 +307,27 @@ namespace Libplanet.Net.Consensus
         private Block<T>? GetValue()
         {
             Block<T> block = _blockChain.ProposeBlock(_privateKey, lastCommit: _lastCommit);
-            if (block.Index == Height)
+            if (_blockChain.ValidateNextBlock(block) is InvalidBlockException e)
             {
-                _blockChain.Store.PutBlock(block);
-                return block;
+                _logger.Error(
+                    e, "Could not propose a valid block");
+                ExceptionOccurred?.Invoke(this, e);
+                return null;
+            }
+            else if (block.Index != Height)
+            {
+                InvalidBlockIndexException ibie = new InvalidBlockIndexException(
+                    $"Proposed block's index {block.Index} must be the same " +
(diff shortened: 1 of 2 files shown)


limebell (commented):
How about make `BlockChain<T>.ProposeBlock()` more secure so that it cannot make invalid block?

greymistcube:
> How about make BlockChain<T>.ProposeBlock() more secure so that it cannot make invalid block?

Thought about removing `lastcommit` parameter for a bit, but we have too many options open for `BlockChain<T>.ProposeBlock()`. Felt weird to **only** automate collecting the `LastCommit` part. 😐

Libplanet.Net/Consensus/Context.cs
```diff
@@ -307,15 +307,28 @@ private TimeSpan TimeoutPropose(long round)
         private Block<T>? GetValue()
         {
             Block<T> block = _blockChain.ProposeBlock(_privateKey, lastCommit: _lastCommit);
-            if (block.Index == Height)
+            if (_blockChain.ValidateNextBlock(block) is InvalidBlockException e)
             {
-                _blockChain.Store.PutBlock(block);
-                return block;
+                _logger.Error(
+                    e, "Could not propose a valid block");
```
colibrishin, line 313:
```suggestion
                    e, "Could not propose a invalid block");
```

Libplanet.Net/Consensus/Context.cs
```diff
@@ -307,15 +307,28 @@ private TimeSpan TimeoutPropose(long round)
...
+                return null;
             }
-            else
+            else if (block.Index != Height)
             {
+                InvalidBlockIndexException ibie = new InvalidBlockIndexException(
+                    $"Proposed block's index {block.Index} must be the same " +
+                    $"as context's height {Height}.");
+                _logger.Error(
+                    ibie, "Could not propose a valid block");
```
colibrishin, line 323:
```suggestion
                    ibie, "Could not propose a invalid block");
```

colibrishin (approved):

PR #2531, opened 2022-11-09
Unit test of double proposal test scenario

colibrishin:
This PR addresses a set of unit tests about the double proposal. 

`DoubleProposalTwoThridNotLockedBlock` is skipped because currently locking to multiple proposals is not implemented.


colibrishin:
I'd like to close this PR for now. 

- These tests are valid, but too weak proof of double proposal.
- There is no actual implementation to be tested by this as a unit test. Probably this should be done as an integration test. 
- In my opinion, however, making tests like posplant is hard to be done (libplanet standalone, building byzantine node) and cannot be done in parallel with other jobs for the current or future workflow.

PR #2532, opened 2022-11-09
chore: fix build failure in linux mono  [skip changelog]

limebell:
`Zip` method has different overloads in .NET Framework and .NET Core.

### DIFF
--- a/Libplanet.Tests/TestUtils.cs
+++ b/Libplanet.Tests/TestUtils.cs
@@ -300,12 +300,21 @@ Actual (C# array lit):   new byte[{actual.LongLength}] {{ {actualRepr} }}";
             );
         }
 
-        public static void AssertSorted<T>(IEnumerable<T> list)
+        public static void AssertSorted<T>(IEnumerable<T> expected)
             where T : IComparable<T>
         {
-            Assert.True(list
-                .Zip(list.Skip(1))
-                .All(pair => pair.First.CompareTo(pair.Second) <= 0));
+            // Was implemented by `.Zip()`, but the method has different
+            // overloads in .NET Framework, so cannot be used.
+            T[] arr = expected.ToArray();
+            foreach ((int i, (T first, T second)) in Enumerable
+                         .Range(0, arr.Length - 1)
+                         .Select(i => (i, (arr[i], arr[i + 1]))))
+            {
+                string errorMessage =
+                    $"Given list is not sorted since item {first} at {i} " +
+                    $"is greater than item {second} at {i + 1}";
+                Assert.True(first.CompareTo(second) <= 0, errorMessage);
+            }
         }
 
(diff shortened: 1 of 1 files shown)


greymistcube (approved):

Libplanet.Tests/TestUtils.cs
```diff
@@ -300,12 +300,17 @@ public static void AssertBencodexEqual(IValue expected, IValue actual)
...
         }
 
-        public static void AssertSorted<T>(IEnumerable<T> list)
+        public static void AssertSorted<T>(IEnumerable<T> expected)
             where T : IComparable<T>
         {
-            Assert.True(list
-                .Zip(list.Skip(1))
-                .All(pair => pair.First.CompareTo(pair.Second) <= 0));
+            List<T> list = expected.ToList();
```
longfin, line 306: It would be better if there is a mention about why we don't use `.Zip()` here.

longfin (approved):

Libplanet.Tests/TestUtils.cs
```diff
@@ -300,12 +300,17 @@ public static void AssertBencodexEqual(IValue expected, IValue actual)
...
+            var ret = true;
+            for (int i = 0; ret && i < list.Count - 1; i++)
+            {
+                ret &= list[i].CompareTo(list[i + 1]) <= 0;
+            }
+
+            Assert.True(ret, "Given list is not sorted.");
```
greymistcube, line 313: I know I originally implemented this, but if possible, it'd be also better to tell which indices are not in order.

colibrishin (approved):

Libplanet.Tests/TestUtils.cs
greymistcube:
```suggestion
            foreach ((int Index, (var First, var Second)) in Enumerable
                .Range(0, list.Count() - 1)
                .Select(index => (index, (list[index], list[index + 1]))))
            {
                string errorMessage =
                    $"Given list is not sorted since item {First} at {Index} " +
                    $"is greater than item {Second} at {Index + 1}";
                Assert.True(First.CompareTo(Second) <= 0, errorMessage);
            }
```

Maybe something like this? 😗

limebell: Amended at https://github.com/planetarium/libplanet/pull/2532/commits/664941f54ed04cdcb87adf7144da1643c93f0c78

Libplanet.Tests/TestUtils.cs
limebell: Amended at https://github.com/planetarium/libplanet/pull/2532/commits/664941f54ed04cdcb87adf7144da1643c93f0c78

greymistcube (approved):

PR #2534, opened 2022-11-09
Try set exception when an exception occur in process request

boscohyun:
SSIA


boscohyun:
This change does not necessary.

PR #2546, opened 2022-11-14
Changed `BlocksMsg` layout

greymistcube:
In preparation to requiring a `BlockCommit` when appending a PBFT `Block<T>`.

### DIFF
--- a/Libplanet.Net/Messages/BlocksMsg.cs
+++ b/Libplanet.Net/Messages/BlocksMsg.cs
@@ -9,11 +9,13 @@ namespace Libplanet.Net.Messages
     {
         /// <summary>
         /// Creates an instance of <see cref="BlocksMsg"/> with given <paramref name="payloads"/>.
-        /// The length of <paramref name="payloads"/> is twice the length of
-        /// <see cref="Block{T}"/>s to send.  Even indices are encoded <see cref="Block{T}"/>s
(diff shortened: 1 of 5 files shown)


Libplanet.Net/Messages/BlocksMsg.cs
```diff
@@ -1,22 +1,51 @@
...
 {
     internal class BlocksMsg : Message
     {
+        /// <summary>
+        /// Creates an instance of <see cref="BlocksMsg"/> with given <paramref name="payloads"/>.
+        /// The length of <paramref name="payloads"/> is twice the length of
+        /// <see cref="Block{T}"/>s to send.  Even indices are encoded <see cref="Block{T}"/>s
+        /// and odd indices are encoded <see cref="BlockCommit"/>s.
+        /// </summary>
+        /// <param name="payloads">The payload for this <see cref="Message"/>.</param>
```
limebell, line 16: How about adding a description written in [`Payload`](https://github.com/planetarium/libplanet/pull/2546/files#diff-980c0d1b8703be0df9e101478a4dd704a81c9a3790fe8bb8ee020a9a89d89e25R41-R44) also in here?

greymistcube: Like [this](3ad5eae46c63a95e9fa1ca51302f9f3374aaebca)?

limebell (approved):

PR #2558, opened 2022-11-17
Send `BlockCommit` in `BlocksMsg`

### LINKED ISSUE #2551: Send `BlockCommit`s with `BlocksMsg`
Although not strictly necessary, probably better to be worked on after #2550.

Currently, an empty `byte` array is being sent as a placeholder in `BlocksMsg.Payloads`. When sending `Block<T>`s through `BlocksMsg`, we should also retrieve the result of `GetBlockCommit()` and attach to `BlocksMsg`.

greymistcube:
Closes #2551.

### DIFF
--- a/Libplanet.Net/Swarm.BlockSync.cs
+++ b/Libplanet.Net/Swarm.BlockSync.cs
@@ -327,12 +327,13 @@ namespace Libplanet.Net
                     window: InitialBlockDownloadWindow
                 );
 
-                Block<T> initialTip = BlockChain.Tip;
+                (Block<T> Block, BlockCommit Commit) initialTip =
+                    (BlockChain.Tip, BlockChain.GetBlockCommit(BlockChain.Tip.Hash));
                 long totalBlocksToDownload = 0L;
                 long receivedBlockCount = 0L;
-                Block<T> tipCandidate = initialTip;
+                (Block<T> Block, BlockCommit Commit) tipCandidate = initialTip;
 
-                Block<T> tempTip = tipCandidate;
+                (Block<T> Block, BlockCommit Commit) tempTip = tipCandidate;
                 Block<T> branchpoint = null;
 
                 var demandBlockHashes = GetDemandBlockHashes(
@@ -441,11 +442,11 @@ namespace Libplanet.Net
                     );
                     block.ValidateTimestamp();
                     workspace.Store.PutBlock(block);
-                    if (tempTip is null ||
+                    if (tempTip.Block is null ||
                         BlockChain.PerceiveBlock(block).Index >
-                            BlockChain.PerceiveBlock(tempTip).Index)
+                            BlockChain.PerceiveBlock(tempTip.Block).Index)
                     {
-                        tempTip = block;
+                        tempTip = (block, commit);
                     }
 
                     receivedBlockCount++;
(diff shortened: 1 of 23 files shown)


colibrishin (changes requested):
https://github.com/planetarium/libplanet/blob/075a864b968fedfefac976efebdba8c1c9d2a5f1/Libplanet.Net/Swarm.BlockSync.cs#L397-L399

https://github.com/planetarium/libplanet/blob/075a864b968fedfefac976efebdba8c1c9d2a5f1/Libplanet.Net/Swarm.BlockSync.cs#L542-L548

`Append()` with `BlockCommit` in `CompleteBlockAsync()` is not implemented.

limebell (commented):
`byte[]`-typed `.ToByteArray()` method and `ImmutableArray<byte>`-typed `.ByteArray` method looks confusing to me by its name. How about rename method that returning `ImmutableArray` to `ImmutableByteArray` or something?

colibrishin:
Seems I misused the word `implemented`. What I wanted to say was, the received `BlockCommit` should be passed to `Append()`.

greymistcube:
@limebell 
It's really not by my choice. We do have `ImmutableArray<byte> ByteArray` and `byte[] ToByteArray()` sprinkled here and there throughout the codebase, such as `Address`, `HashDigest<T>`, `Nonce`, etc.; I'm just trying to be consistent. 🙄

Libplanet.Net/Swarm.BlockSync.cs
```diff
@@ -466,34 +470,35 @@ in completedBlocks.WithCancellation(cancellationToken))
...
+                    (Block<T> Block, BlockCommit Commit) blockToAdd;
+                    if (deltaBlocks.First is LinkedListNode<(Block<T>, BlockCommit)> node)
                     {
-                        Block<T> b = node.Value;
+                        (Block<T> b, BlockCommit c) = node.Value;
                         if (b.PreviousHash is { } p && !workspace.ContainsBlock(p))
                         {
-                            blockToAdd = workspace.Store.GetBlock<T>(p);
+                            blockToAdd =
+                                (workspace.Store.GetBlock<T>(p), workspace.GetBlockCommit(p));
```
colibrishin, line 493: This fails in some tests. It seems to sort the fetched block, so the previous `BlockCommit` is needed. 

Problem was throwing an exception in `GetBlockCommit()`. `GetBlockCommit()` works if the target `Block<T>` has been appended into the chain before, however, in this case, we have not yet appended a block.

```csharp
(Block<T> b, BlockCommit _) = node.Value; // <-- We don't BlockCommit of the current block.
if (b.PreviousHash is { } p && !workspace.ContainsBlock(p))
{
  Block<T> block = workspace.Store.GetBlock<T>(p);
  blockToAdd = (block, workspace.Store.GetBlockCommit(block.Index)); // <-- Take BlockCommit from store.
}
```

I think my simplest (not sure performance-wise) workaround is storing all `BlockCommit` and taking it from the store while doing it.

```csharp
_logger.Verbose(
    "Add a block #{BlockIndex} {BlockHash}...",
    block.Index,
    block.Hash
);
block.ValidateTimestamp();
workspace.Store.PutBlock(block);
workspace.Store.PutBlockCommit(commit); // <-- Store BlockCommit
if (tempTip.Block is null ||
    BlockChain.PerceiveBlock(block).Index >
        BlockChain.PerceiveBlock(tempTip.Block).Index)
{
    tempTip = (block, commit);
}
```

colibrishin (approved):

PR #2561, opened 2022-11-18
Validate `BlockCommit` while in `BlockChain<T>.Append()`

### LINKED ISSUE #2541: Add additional `BlockCommit` validation for `Append()`
Related to #2504.

## Premise

I think we were thinking about the issue in a wrong way. Although we have implemented `Block<T>.LastCommit` validation, we are missing the entire other half for validating the `BlockCommit` as a proof that a consensus has been reached **for the `Block<T>`** we are trying to append. Said `BlockCommit` is different from `Block<T>.LastCommit` as it will reference the `Block<T>` itself, not the `Block<T>`'s previous `Block<T>`.

## Explanation

**Any** PBFT `Block<T>` that is not a genesis `Block<T>` should require a `BlockCommit` to be present as an evidence that a consensus has been reached to include the said `Block<T>` (albeit not included in the `Block<T>` as data) in order to be appended to `BlockChain<T>`. It is important to differentiate the purpose of this `BlockCommit` from `Block<T>.LastCommit`. If `Block<T>` has an index of `n`, then `Block<T>.LastCommit` with height of `n - 1` "signals" that whoever proposed the `Block<T>` **had the right to propose `Block<T>`** in the sense that the validator has observed all the evidence to actually append the previous `Block<T>` of index `n - 1`. This, however, **does not mean** the `Block<T>` in question, of index `n`, is in any way valid, i.e. that a consensus has been reached, for it to be appended to a `BlockChain<T>`. It just so happens `Swarm<T>` propagates any `Block<T>`'s in its own `BlockChain<T>`.

Any node observing `Block<T>`, either through `Swarm<T>` or `Context`, must do its due diligence by whether it can observe/gather a `BlockCommit` of **height `n`** that is valid. If it is through `Swarm<T>`, this can be observed/hijacked from a `Block<T>` of **index `n + 1`** which would have a `LastCommit` field of height `n`. This means, when syncing `Block<T>`s through a `Swarm<T>`, it should never be possible to completely sync up with a peer's `BlockChain<T>.Tip`, but only up to the penultimate `Block<T>`.

Additionally, this is possible since for `Block<T>` of index `n`, the set of validators can be derived for height `n` as all the necessary state transitions must have taken place by having a `BlockChain<T>` of height `n - 1`. This allows to verify and validate the `LastCommit` of height `n`, i.e. the one attached to a `Block<T>` of index `n + 1`.

## Remedy

A starting point would be to add an additional parameter for `BlockChain<T>.Append()` that is of type `BlockCommit` and prevent appending a `Block<T>` if a valid `BlockCommit` is not provided (except for the genesis `Block<T>`[^1]).

## Notes

By some unforeseen fault, either by missing `ConsensusMsg`s due to some network failure or by receiving invalid `ConsensusMsg`s due to an activity of a malicious node, a node's `Context` might become "out-of-sync" with the rest of the network. As long as it is below tolerance, i.e. 1/3 of the total power of the network, it seems the network should recover on its own. Even when there was a race condition between the `Swarm<T>` and a `Context`, it seems this wasn't really much of a problem (other than there being an obvious security hole where a malicious actor could've forced a fork on an honest node since `Swarm<T>` currently bypasses the consensus check). With the updated approach provided here, `Context` would generally take precedence to `Swarm<T>`. Consider the following thought experiment.

As-is scenario:

- A local node is running a `Context` of height `n` (i.e. its `BlockChain<T>.Tip.Index` is `n - 1`).
- There is some fault and the node's `Context` is stuck at `n`.
- `Swarm<T>` forcibly syncs its `BlockChain<T>.Tip` to a known peer of height **`n`** (although it shouldn't as explained above).
- Running `Context` is discarded and a new `Context` of height `n + 1` gets started.
  - At this point, the network's `Context` is on height **`n + 1`**.
- Since there is no `BlockCommit` for height `n` to provide as `LastCommit` of `Block<T>` of index `n + 1`, the node is incapable of proposing a `Block<T>` for a `Context` of height `n + 1`[^2].
- Once a consensus is reached for `Context` of height `n + 1`, `Context` of height `n + 2` should start normally.

To-be scenario:

- A local node is running a `Context` of height `n` (i.e. its `BlockChain<T>.Tip.Index` is `n - 1`).
- There is some fault and the node's `Context` is stuck at `n`.
- `Swarm<T>` forcibly syncs its `BlockChain<T>.Tip` to a known peer of height **`n + 1`**.
- Running `Context` is discarded and a new `Context` of height `n + 1` gets started.
  - At this point, the network's `Context`[^3] is on height **`n + 2`**.
- Although there is a `BlockCommit` for height `n` to provide as `LastCommit` of `Block<T>` of index `n + 1`, and the node is free to propose a `Block<T>` of index `n + 1`, it will be ignored.
  - As other honest nodes are on height `n + 2`, the node will not get the votes for its proposed `Block<T>`.
- As the node has cached `ConsensusMsg`s (assuming there were no fault during network's `Context` of height `n + 1` phase), the node will quickly reach the same consensus as other nodes (even if it hasn't participated in it) for height `n + 1`.
- Depending on when the node join's the network's `Context` for height `n + 2`, although the node might've missed several rounds, the rest should progress normally onwards. This means, the node `Context` will become normal by the time `Context` for height `n + 3` starts.

[^1]: Note that this requirement is slightly different from whether `LastCommit` should or should not be `null` for a PBFT `Block<T>`. For example, a PBFT `Block<T>` of index `1` should be provided with a `BlockCommit` during the `BlockChain<T>.Append()` call, but the `Block<T>` itself would be missing a `LastCommit`, i.e. it'll be `null`.

[^2]: This is not entirely true, as the node can propose under certain conditions by "copying" the `Block<T>` proposed by someone else.

[^3]: I am using this term very loosely. 😗

colibrishin:
Resolves #2541
This PR adds validation of `BlockCommit` in `BlockChain<T>.Append()`.

There are some cases we should consider: 
- `PreloadAfterReorg` is failing but there will be no reorganization in PBFT, so skipped the test (Reorganization code should be removed.)
- When the validator set is dynamically changing, in my opinion, we should revisit this and think about how would block synchronization be done.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,7 @@ Version PBFT
  -  Added `BlockMarshaler.UnmarshalBlockHash()` method. [[#PBFT]]
  -  Added `BlockChain<T>.GetBlockCommit()` method.  [[#PBFT]]
  -  Added `BlockChain<T>.CleanupBlockCommitStore()` method.  [[#PBFT]]
+ -  Added `InvalidBlockCommitException` class.  [[#PBFT]]
  -  Added `BlockChain<T>.ValidateBlockCommit()` method  [[#PBFT]]
  -  (Libplanet.Net) Added `IReactor` interface.  [[#PBFT]]
  -  (Libplanet.Net) Added `ConsensusReactor` class which inherits
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,7 @@ Version PBFT
  -  Added `BlockMarshaler.UnmarshalBlockHash()` method. [[#PBFT]]
  -  Added `BlockChain<T>.GetBlockCommit()` method.  [[#PBFT]]
  -  Added `BlockChain<T>.CleanupBlockCommitStore()` method.  [[#PBFT]]
+ -  Added `BlockChain<T>.ValidateBlockCommit()` method  [[#PBFT]]
  -  (Libplanet.Net) Added `IReactor` interface.  [[#PBFT]]
  -  (Libplanet.Net) Added `ConsensusReactor` class which inherits
     `IReactor` interface.  [[#PBFT]]
@@ -88,6 +89,12 @@ Version PBFT
  -  `PreEvaluationBlockHeader()` constructor became to throw
     `InvalidBlockLastCommitException` when its metadata's `LastCommit` is
     invalid.  [[#PBFT]]
+ -  `BlockChain<T>.Append()` has new parameter `BlockCommit blockCommit`, which
+    is a set of commits for given block. `BlockCommit` is used for checks
+    whether a block is committed in consensus.  [[#PBFT]]
+ -  `BlockChain<T>.Append()` method became to throw
+    `InvalidBlockCommitException` when the given `BlockCommit` is invalid with
+    given block.  [[#PBFT]]
 
 ### Bug fixes
 
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1578,7 +1578,7 @@ namespace Libplanet.Blockchain
         }
 
 #pragma warning disable SA1202
-        public InvalidBlockCommitException ValidateBlockCommit(
+        internal InvalidBlockCommitException ValidateBlockCommit(
             Block<T> block,
             BlockCommit blockCommit)
 #pragma warning restore SA1202
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1551,6 +1551,32 @@ namespace Libplanet.Blockchain
             }
         }
 
+        /// <summary>
+        /// Clean up <see cref="BlockCommit"/>s in the store. The <paramref name="except"/> height
+        /// of <see cref="BlockCommit"/> will not be removed. If the stored
+        /// <see cref="BlockCommit"/> count is not over <paramref name="maxCacheSize"/>, the removal
+        /// is skipped.
+        /// </summary>
+        /// <param name="except">A exceptional index that is not to be removed.</param>
+        /// <param name="maxCacheSize">A maximum count value of <see cref="BlockCommit"/> cache.
+        /// </param>
+        internal void CleanupBlockCommitStore(long except, long maxCacheSize = 30)
+        {
+            IEnumerable<long> indices = Store.GetBlockCommitIndices().ToArray();
+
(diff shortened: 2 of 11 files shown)


limebell (commented):
Can you please add one line in "Behavioral changes" section that `BlockChain<T>.Append()` became to throw an `InvalidBlockCommitException`?

Libplanet.Tests/TestUtils.cs
```diff
@@ -355,7 +355,7 @@ public static byte[] GetRandomBytes(int size)
         public static BlockCommit CreateBlockCommit<T>(Block<T> block)
             where T : IAction, new() =>
                 block.Index > 0 &&
-                block.ProtocolVersion > BlockMetadata.PoWProtocolVersion
+                block.ProtocolVersion >= BlockMetadata.PoWProtocolVersion
```
greymistcube, line 358: Shouldn't it be `>`?

CHANGES.md
```diff
@@ -59,6 +59,7 @@ Version PBFT
  -  Added `BlockMarshaler.UnmarshalBlockHash()` method. [[#PBFT]]
  -  Added `BlockChain<T>.GetBlockCommit()` method.  [[#PBFT]]
  -  Added `BlockChain<T>.CleanupBlockCommitStore()` method.  [[#PBFT]]
+ -  Added `BlockChain<T>.ValidateBlockCommit()` method  [[#PBFT]]
```
greymistcube, line 62: Is there a need for this to be `public`?

colibrishin: Amended. I've changed it to `internal` for testing.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1540,6 +1549,77 @@ internal IEnumerable<BlockHash> IterateBlockHashes(int offset = 0, int? limit =
...
+            Block<T> block,
+            BlockCommit blockCommit)
+#pragma warning restore SA1202
+        {
+            // To allow the PoW block to be appended, we skips the validation.
+            if ((block.Index == 0 && blockCommit == null) ||
+                block.ProtocolVersion <= BlockMetadata.PoWProtocolVersion)
+            {
+                return null;
+            }
```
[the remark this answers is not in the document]
colibrishin: Amended. https://github.com/planetarium/libplanet/pull/2561/commits/8c825681e41c20f6c78c805c808754eee22cb7c1

limebell (approved):

PR #2566, opened 2022-11-23
Introduce `IFeeCalculator`

longfin:
# Context
NineChronicles has a Tx spamming problem. And it is caused by the non-fee Transaction model. 
In this PR. introduce `IFeeCalculator`. It is an interface for calculating Transaction fees and collecting from signers.

# Description
Get `IFeeCalculator` as an `IBlockPolicy` and check the fee in `ActionEvaluator.EvaluateActions()`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -75,6 +75,8 @@ deployment if possible.
      -  Removed `Vote(byte[])`, `Vote.ByteArray`, and `Vote.ToByteArray()`.
      -  Removed `VoteMetadata(byte[])`, `VoteMetadata.ByteArray`,
         and `VoteMetadata.ToByteArray()`.
+ -  `ActionEvaluator()` constructor now explicitly requires a new
+    `feeCalculator` parameter.  [[#2566]]
 
 ### Backward-incompatible network protocol changes
 
@@ -95,6 +97,7 @@ deployment if possible.
      -  Added `ImportableKeyStore` interface.
      -  Added `KeyStore` interface.
      -  Added `MutableKeyStore` interface.
+ -  Added `IFeeCalculator` interface. [[#2566]]
 
 ### Behavioral changes
 
@@ -107,6 +110,7 @@ deployment if possible.
 ### CLI tools
 
 [Libplanet 0.52.0]: https://www.nuget.org/packages/Libplanet/0.52.0
+[#2566]: https://github.com/planetarium/libplanet/pull/2566
 [#2703]: https://github.com/planetarium/libplanet/pull/2703
 [#2915]: https://github.com/planetarium/libplanet/pull/2915
 [#2954]: https://github.com/planetarium/libplanet/pull/2954
--- a/Libplanet/Action/ActionEvaluator.cs
+++ b/Libplanet/Action/ActionEvaluator.cs
@@ -32,6 +32,7 @@ namespace Libplanet.Action
         private readonly Func<BlockHash, ITrie>? _trieGetter;
         private readonly Predicate<Currency> _nativeTokenPredicate;
         private readonly IActionTypeLoader _actionTypeLoader;
+        private readonly IFeeCalculator? _feeCalculator;
 
 #pragma warning disable MEN002
 #pragma warning disable CS1573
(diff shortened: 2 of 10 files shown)


Libplanet/Action/ActionEvaluator.cs
```diff
@@ -364,13 +389,40 @@ ActionContext CreateActionContext(IAccountStateDelta prevStates, int randomSeed)
...
+                    }
+                    else if (rehearsal)
+                    {
+                        logger?.Verbose("rehearsal don't be charged.");
+                    }
+                    else if (blockAction)
+                    {
+                        logger?.Verbose("blockAction don't be charged.");
+                    }
+                    else if (feeCalculator is null)
```
longfin, line 411: Considering static calculator instead of `null`

Libplanet/Action/IFeeCalculator.cs
```diff
@@ -0,0 +1,9 @@
+using Libplanet.Assets;
+
+namespace Libplanet.Action
+{
+    public interface IFeeCalculator
+    {
+        FungibleAssetValue CalculateFee(IAction action);
```
longfin, line 7: Maybe we should need additional contextual informations

longfin:
I integrated and adjusted some works from @riemannulus 😅 

- First, accessing `context.PreviousStateRootHash` has been reverted like previously due to performance reasons. (I think we need to revisit this in near future)
- And more, I turned back `ActionEvaluator._feeCalculator` to nullable due to considering the characteristics of the `FungibleAssetValue` that requires `Currency` and the API of `transferAsset` which is impossible to transfer 0.
  - but on the other hand, I also think that `null` may not be the proper choice to indicate "fee-free". (`null` vs `NoFeeCalculator.Instance` 🤔 )
-  Last, I adjusted and removed some previous changelogs to keep the convention.

longfin:
Revised for continuing from [fce695c](https://github.com/planetarium/libplanet/pull/2566/commits/fce695ca20e23cffedeb9fcbd106cc1d3e437703)

longfin:
Rebased to follow 41829777672ec9e2a34763f25ec451ac8224e9e4.

CHANGES.md
```diff
@@ -18,19 +18,24 @@ To be released.
     -  `ValidatorSetStateCompleters`
  -  Removed type parameter `T` for `IAction` from `ActionEvaluator` and
     `IBlockChainStates`.  [[#2703]]
+ -  `ActionEvaluator<T>()` constructor now explicitly requires a new
+    `feeCalculator` parameter.  [[#2566]]
```
moreal, line 22: Because this pull request opened a long time ago, this line seems staled. (ref #2703)
```suggestion
 -  `ActionEvaluator()` constructor now explicitly requires a new
    `feeCalculator` parameter.  [[#2566]]
```

Libplanet/Blockchain/Policies/BlockPolicy.cs
```diff
@@ -172,6 +172,9 @@ public class BlockPolicy<T> : IBlockPolicy<T>
         /// </summary>
         public TimeSpan BlockInterval { get; }
 
+        /// <inheritdoc/>
+        public IFeeCalculator? FeeCalculator => null;
```
moreal, line 176: Unlike other properties, is it intended to restrict using `IFeeCalculator` yet?

Libplanet/Blocks/PreEvaluationBlock.cs
```diff
@@ -252,7 +252,8 @@ IStateStore stateStore
                 trieGetter: null,
                 genesisHash: null,
                 nativeTokenPredicate: nativeTokenPredicate,
-                actionTypeLoader: StaticActionTypeLoader.Create<T>()
+                actionTypeLoader: StaticActionTypeLoader.Create<T>(),
+                feeCalculator: null
```
moreal, line 256: Can it cause the `InvalidStateRootException`? Because it passes `null` as `feeCalculator` and it doesn't restrict any actions so all of them will be executed unlike evaluation while appending blocks

moreal: Hmm, okay. I remembered there was `BlockChain<T>.ProposeGenesisBlock` (was `BlockChain<T>.MakeGenesisBlock()`). I think this `PreEvaluationBlock.DetermineStateRootHash()` method should have an `internal` access modifier instead of `public` if it will continue to provide `BlockChain<T>.ProposeGenesisBlock()` static method in the future.

CHANGES.md
```diff
@@ -45,6 +45,9 @@ defined in *[@planetarium/account]*.
      -  The return type of `encodeCurrency()` became `Value` (was `Encodable`).
      -  The return type of `encodeMint()` became `Value` (was `Encodable`).
      -  The return type of `encodeTransfer()` became `Value` (was `Encodable`).
+ -  `ActionEvaluator<T>()` constructor now explicitly requires a new
+ -  `ActionEvaluator()` constructor now explicitly requires a new
```
moreal, line 49: I guess you intended this.

```suggestion
 -  `ActionEvaluator()` constructor now explicitly requires a new
```

moreal (approved):

PR #2573, opened 2022-11-25
Refactor `BlockCommit` caching

greymistcube:
Now uses `BlockHash` based caching. There may be some performance implications.


limebell (commented):
Please write changelog since public API has been modified

PR #2577, opened 2022-11-25
Introduce Validator, Add validator actions

OnedgeLee:


OnedgeLee:
Close due to duplicated PR #2596

PR #2579, opened 2022-11-25
Fix `Swarm<T>.Preload()`

greymistcube:
See #2652.


greymistcube:
On second thought, I think we might need to revisit this. I'm not sure this has been resolved. 🙄

Issue #2583, opened 2022-11-26
Change `BlockLocator` index selection logic


greymistcube:
Given hashes `[0x00, 0x01, ..., 0x0e, 0x0f]`, we should expect the following results:

- `BlockLocator` created with `sampleAfter` value `0`: `[0x0f, 0x0e, 0x0c, 0x08, 0x00]`
- `BlockLocator` created with `sampleAfter` value `1`: `[0x0f, 0x0e, 0x0d, 0x0b, 0x07, 0x00]`
  - Current actual value: `[0x0f, 0x0e, 0x0c, 0x08, 0x00]`
- `BlockLocator` created with `sampleAfter` value `2`: `[0x0f, 0x0e, 0x0d, 0x0c, 0x0a, 0x06, 0x00]`
  - Current actual value: `[0x0f, 0x0e, 0x0d, 0x0b, 0x07, 0x00]`

In the last example above, `[0x0f, 0x0e]` would correspond to non-sampled indices and and `[0x0d, 0x0c, 0x0a, 0x06, 0x00]` would correspond to sampled indices. From my perspective, there are two problems with current implementation:

- There is no difference between `BlockLocator` created with `sampleAfter` value of `0` and `1`.
- The initial delta of indices for "sampling" starts with `2` not `1`.

Given starting index $n$, I think the cleanest way to define the act of "sampling indices" would be the following:
- $i_{0} = n$
- $i_{1} = i_{0} - 1$
- $i_{k} = i_{k-1} - 2(i_{k - 2} - i_{k - 1})$ for $k \geq 2$

Or in a more compact form:
- $i_{0} = n$
- $i_{k} = i_{k - 1} - d_{k - 1}$ where $d_{t} = 2^{t}$ for $k \geq 1$

Or in a closed form:
- $i_{k} = n - (2^{k} - 1)$

Issue #2582, opened 2022-11-26
Make `FindNextHashes()` behavior different when no branchpoint is found


greymistcube:
Related to #2581. 

When the method fails to find a branchpoint between itself and the given `BlockLocator` object, it simply "assumes" that the branchpoint should be the genesis. This means, this method does not differentiate a case between a valid `BlockLocator` object with actual genesis branchpoint and some arbitrary/invalid `BlockLocator`.

If we dig a little deeper, this is due to underlying assumptions, or lack there of, on `BlockLocator` objects and possible/allowed states for `BlockChain<T>`. I'll be making a separate issue on this shortly.

Issue #2581, opened 2022-11-26
Make `FindNexHashes()` return at most `count` number of `BlockHash`es


greymistcube:
As far as I can tell, currently, the behavior is rather odd in the following way that is not immediately apparent:

- If the branchpoint is "found" (in quotation since this part is also problematic in its current form) to be the genesis block, the method can return at most `count` number of `BlockHash`es.
- If the branchpoint is found to be a non-genesis block, the method can return at most `count + 1` number of `BlockHash`es.

This is rather confusing. 🙄

PR #2584, opened 2022-11-27
♻️ Refactor `BlockLocator`

### LINKED ISSUE #2580: Remove `Func<BlockHash, long>` parameter from `BlockLocator()`
As there is no requirement for `indexBlockHash` and `indexByBlockHash`, although presumably these are supposed to be inverses of each other, this can actually affect and mess up the internal flow of logic. For instance

```csharp
Func<long, BlockHash?> indexToBlockHash = _ => new BlockHash();
Func<BlockHash, long> blockHashToIndex = _ => long.MaxValue;
BlockLocator locator = new BlockLocator(
    indexToBlockHash, blockHashToIndex, 10);
```

results in `OverflowException` due to creating an "infinite loop" (the loop is only broken by an unexpected error, not due to `break` condition being met).

greymistcube:
Closes #2580, #2581, #2582, and #2583.

The implementation is still subject to change depending on what constraints we decide to impose on `BlockChain<T>` instances and how we want to handle variety of cases arising from constraints, or lack there of. The issue is rather far reaching and not obvious to resolve. See #2585.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,14 @@ To be released.
 
 ### Deprecated APIs
 
+ -  Removed `BlockLocator(Func<long, BlockHash?>, Func<BlockHash, long>, int)`
+    constructor.  Use `BlockLocator.Create()` static method instead.
+    [[#2580], [#2584]]
+
 ### Backward-incompatible API changes
 
+ -  Changed `BlockLocator` to throw an `ArgumentException` if an empty set of
+    `BlockHash`es are given during construction.  [[#2584]]
  -  `BlockChain<T>()` now explicitly requires both `store` and `stateStore`
     arguments to be not `null`.  [[#2609]]
  -  `BlockChain<T>.Swap()` now throws an `InvalidOperationException` if called
@@ -60,6 +66,7 @@ To be released.
     [[#2474], [#2505]]
  -  (Libplanet.Explorer) Added `actionsLogsList` field to `TxResultType`.
     [[#2474], [#2505]]
+ -  Added `BlockLocator.Create()` static method.  [[#2584]]
  -  Added `PolicyBlockActionGetter` delegator type.  [[#2646]]
  -  Added `IActionTypeLoader.LoadAllActionTypes()` method.  [[#2646]]
      -  Added `StaticActionTypeLoader.LoadAllActionTypes()` method.
@@ -77,6 +84,14 @@ To be released.
 
 ### Behavioral changes
 
+ -  Changed `BlockChain<T>.FindNextHashes()` to return at most `count`
+    number of `BlockHash`es regardless of the result.
+    `BlockChain<T>`. [[#2581], [#2584]]
+ -  Changed `BlockChain<T>.FindNextHashes()` to return zero `BlockHash`es
+    if no branch point `BlockHash` is found instead of returning
+    `BlockHash`es starting with the genesis `BlockHash`.  [[#2582], [#2584]]
+ -  Chnaged the behavior of `BlockLocator` index selection and sampling when
+    creating an instance.  [[#2583], [#2584]]
  -  Changed the default `VolatileStagePolicy<T>.Lifetime` from 3 hours
     to 10 minutes.  [[#2718]]
 
@@ -100,6 +115,11 @@ To be released.
 
 [#2474]: https://github.com/planetarium/libplanet/discussions/2474
 [#2505]: https://github.com/planetarium/libplanet/pull/2505
+[#2580]: https://github.com/planetarium/libplanet/issues/2580
+[#2581]: https://github.com/planetarium/libplanet/issues/2581
+[#2582]: https://github.com/planetarium/libplanet/issues/2582
+[#2583]: https://github.com/planetarium/libplanet/issues/2583
+[#2584]: https://github.com/planetarium/libplanet/pull/2584
 [#2609]: https://github.com/planetarium/libplanet/pull/2609
 [#2619]: https://github.com/planetarium/libplanet/pull/2619
 [#2646]: https://github.com/planetarium/libplanet/pull/2646
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,10 +8,6 @@ To be released.
 
 ### Deprecated APIs
 
- -  Removed `BlockLocator(Func<long, BlockHash?>, Func<BlockHash, long>, int)`
-    constructor.  Use `BlockLocator.Create()` static method instead.
-    [[#2580], [#2584]]
-
 ### Backward-incompatible API changes
 
  -  Changed `BlockLocator` to throw an `ArgumentException` if an empty set of
@@ -24,6 +20,9 @@ To be released.
     [[#2474], [#2505]]
  -  Added `actionsLogsList` parameter to `TxFailure` constructor.
     [[#2474], [#2505]]
+ -  Removed `BlockLocator(Func<long, BlockHash?>, Func<BlockHash, long>, int)`
+    constructor.  Use `BlockLocator.Create()` static method instead.
+    [[#2580], [#2584]]
  -  Replaced `IAction?`-typed `policyBlockAction` parameter with
     `PolicyBlockActionGetter`-typed `policyBlockActionGetter` parameter
     in `ActionEvaluator` constructor.  [[#2646]]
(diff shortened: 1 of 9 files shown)


greymistcube:
Few notes:
- Reverts most changes made in #2667 on `BlockLocator`
  - Decision to synchronize only `Tip.Index` and `Id` seem arbitrary.
  - Doing away with `_rwlock` on `IStore` access reliant upon "others will use it responsibly" is shaky.
  - If we don't want to care about `IStore` and want to explicitly synchronize only `Tip.Index` and `Id`, a separate lock should be used instead of using `_rwlock` as a proxy (which would require a lot more work).
  - I suspect the performance bottleneck was mostly from old `BlockHash` to `long` mapping which required to **load** `BlockDigest` each time. The problem with the parameter was already mentioned in #2580 and this is no longer true by getting rid of the access altogether.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1095,26 +1097,18 @@ public BlockChain<T> Fork(BlockHash point, bool inheritRenderers = true)
...
         {
-            Guid chainId;
-            BlockHash tipHash;
-
             _rwlock.EnterReadLock();
             try
             {
-                chainId = Id;
-                tipHash = Tip.Hash;
+                return BlockLocator.Create(
```
longfin, line 1103: Well...  I'm still convinced that lock separation on here wasn't a trade-off between safety and performance, but was a sort of correction against unnecessary locking. (because we keep the immutability on previous blocks for a given chain id).

CHANGES.md
```diff
@@ -55,6 +62,15 @@ To be released.
 
 ### Behavioral changes
 
+ -  Changed `BlockChain<T>.FindNextHashes()` to return at most `count`
+    number of `BlockHash`es regardless of the result.
+    `BlockChain<T>`. [[#2581], [#2584]]
```
longfin, line 67: Does it mean that `BlockChain<T>.FindNextHashes()` will return more hashes even after the branchpoint (to fit `count`)?

Libplanet/Blockchain/BlockLocator.cs
```diff
@@ -1,80 +1,109 @@
...
             {
-                hashes.Add(hash);
-                long currentBlockIndex = indexByBlockHash(hash);
+                throw new ArgumentOutOfRangeException(
+                    $"Given {nameof(startIndex)} cannot be negative: {startIndex}");
+            }
+
+            BlockHash genesisHash = indexToBlockHash(0) ??
+                throw new ArgumentException(
+                    $"Given {nameof(indexToBlockHash)} should not be null at zero index.");
```
longfin, line 90:
```suggestion
                throw new ArgumentException(
                    $"Given {nameof(indexToBlockHash)} should not be null at zero index.",
                    nameof(indexToBlockHash));
```

Libplanet/Blockchain/BlockLocator.cs
```diff
@@ -1,80 +1,109 @@
...
-            BlockHash? current = tipHash;
-            long step = 1;
-            var hashes = new List<BlockHash>();
-            while (current is { } hash)
+            if (startIndex < 0)
             {
-                hashes.Add(hash);
-                long currentBlockIndex = indexByBlockHash(hash);
+                throw new ArgumentOutOfRangeException(
+                    $"Given {nameof(startIndex)} cannot be negative: {startIndex}");
```
longfin, line 85:
```suggestion
                throw new ArgumentOutOfRangeException(
                    $"Given {nameof(startIndex)} cannot be negative: {startIndex}",
                    nameof(startIndex));
```

Libplanet/Blockchain/BlockLocator.cs
```diff
@@ -1,80 +1,109 @@
...
         /// </summary>
         /// <param name="hashes">Enumerable of <see cref="BlockHash"/>es to convert from.</param>
+        /// <exception cref="ArgumentException">Thrown when <paramref name="hashes"/> is empty.
+        /// </exception>
         public BlockLocator(IEnumerable<BlockHash> hashes)
         {
-            _impl = hashes;
+            _impl = hashes.Any()
+                ? hashes.ToList()
+                : throw new ArgumentException($"Given {nameof(hashes)} cannot be empty.");
```
longfin, line 26:
```suggestion
                : throw new ArgumentException($"Given {nameof(hashes)} cannot be empty.", nameof(hashes));
```

it may exceed the line limit 😅

CHANGES.md
```diff
@@ -8,8 +8,14 @@ To be released.
 
 ### Deprecated APIs
 
+ -  Removed `BlockLocator(Func<long, BlockHash?>, Func<BlockHash, long>, int)`
```
longfin, line 11: I think that we need to move this sentence to the Backward-incompatible API changes section.

longfin (commented):
Apart from my hope(i.e., escaping from branch-point detection), it seems to help to clarify confusing points. 

Detailed questions and opinions are left as separate comments.

Libplanet/Blockchain/BlockChain.cs
greymistcube: > because we keep the immutability on previous blocks for a given chain id

This is only implicitly guaranteed by our current practice/codebase. I do agree there is little to no danger in how we are currently using `IStore` and `BlockChain<T>`, but the `IStore` interface itself does not contractually guarantee such immutability (if we had something like `IWriteOnlyStore` with appropriate constraints, then I'd say we *really* don't need to worry about safety). The immutability assumption only comes from *knowing* the whole stack all the way up to `Swarm<T>` is not going to mess with "bad" `IStore` methods.

There is also the question of how we specify `BlockLocator`'s behavior. One of the following should apply:
- If `IStore` is assumed to be immutable on all previous `Block<T>`s for the current `BlockChain<T>.Tip` *and* `BlockChain<T>` (and in turn the `IStore`) is assumed to have *all* `Block<T>`s, then `BlockLocator` should not handle `null` for missing `Block<T>`s.
- If `IStore` is not assumed to be immutable or could be missing some `Block<T>`s, then `BlockLocator` should be designed to deal with missing `Block<T>`s.

The problem is, we have the latter as an implementation while assuming the former via usage (of `BlockChain<T>` and `IStore` *not* `BlockLocator`, which can be confusing). 🙄

longfin: - In the long term, I hope `BlockLocator` disappears and `BlockChain<T>` and `IStore` adjust together. but I don't think this is something that can be done right now.
- If the lock range can be reduced by tweaking `IStore`'s interface, I think it's worth considering for at least now.  
  - I'm not a fan of the Chain ID concept and think a universal storage interface is needed
  - but also I think it's expensive to pay for reducible locks until that happens.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -964,35 +970,31 @@ public IImmutableSet<TxId> GetStagedTransactionIds()
...
-            if (branchpointIndex > 0)
+            if (!(Store.GetBlockIndex(branchpoint) is long branchpointIndex))
             {
-                count++;
+                return new Tuple<long?, IReadOnlyList<BlockHash>>(null, new BlockHash[0]);
             }
 
             var result = new List<BlockHash>();
-            foreach (BlockHash hash in Store.IterateIndexes(Id, branchpointIndex, count))
+            foreach (BlockHash hash in Store.IterateIndexes(Id, (int)branchpointIndex, count))
```
colibrishin, line 984: It's a bit off-topic from this PR, but casting `long` block index into `int` seems odd.

greymistcube: Eh, it is weird. We need to change `IStore` interface to accept `long` values at some point. 🙄

colibrishin (approved):

greymistcube:
Welp, I don't see this getting merged anytime soon, so I'll scrap this. We still need to resolve issues #2580, #2581, #2582, and #2583 somehow. 🙄

greymistcube:
On further inspection, we really need a better spec, and this has too many issues to discard really, affecting too many things up and down the stack (down to `IStore` and all the way up to `Swarm`). I see the lack of spec making the whole syncing process pretty much unworkable in its current form. 😑

I'll do away with `_lock`s for now, as it'd be safe. Changes of things breaking are fairly low, but I still think it'll be nearly undiscoverable implicit spec adding much maintenance overhead (i.e. not being sure whether a call to `BlockLocator()` constructor is safe or not) to whoever tries to work on syncing. 🙁

PR #2589, opened 2022-11-28
Improve performance of `VolatileStagePolicy.Iterate()` and `.GetNextTxNonce()` & Release 0.44.2

longfin:
As the title suggests, it enhances `VolatilePolicy`'s performance through two approaches.

1. Change `VolatilePolicy._ignored`'s implementation, from `ConcurrentBag<T>` to `HashSet<T>`.
    - On my machine[^1], `ConcurrentBag<T>.Contains` was 20x slower than plain `HashSet<T>`'s.
    - I've considered `ConccurentDictionary<K, V>` as an alternative, but `HashSet<T>` might be okay because `VolatilePolicy`already assumes rw lock hardly[^2].
2. Adjusts lock-scopes of `.Iterate()` to avoid unintended blocking.

Also, this PR doesn't suggest a long-term solution for `._ignored` at this moment, but we need to revisit it again in the near future because there is still no consideration about its expiration. 😢 

[^1]: i9-12900k, 32GB RAM, .NET 6.0.400, arch linux
[^2]: But also I believe that it can be the next challenge since it will be the bottleneck of whole block processing again, due to holding the global lock.

### DIFF
--- a/Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
+++ b/Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
@@ -39,7 +39,7 @@ namespace Libplanet.Blockchain.Policies
         where T : IAction, new()
     {
         private readonly ConcurrentDictionary<TxId, Transaction<T>> _staged;
-        private readonly ConcurrentBag<TxId> _ignored;
+        private readonly HashSet<TxId> _ignored;
         private readonly ReaderWriterLockSlim _lock;
         private readonly ILogger _logger;
 
@@ -62,7 +62,7 @@ namespace Libplanet.Blockchain.Policies
             _logger = Log.ForContext<VolatileStagePolicy<T>>();
             Lifetime = lifetime;
             _staged = new ConcurrentDictionary<TxId, Transaction<T>>();
-            _ignored = new ConcurrentBag<TxId>();
+            _ignored = new HashSet<TxId>();
             _lock = new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
         }
 
--- a/Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
+++ b/Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
@@ -189,39 +189,14 @@ namespace Libplanet.Blockchain.Policies
         /// <inheritdoc/>
         public Transaction<T>? Get(BlockChain<T> blockChain, TxId id, bool filtered = true)
         {
-            Transaction<T>? transaction = null;
-
-            _lock.EnterWriteLock();
+            _lock.EnterReadLock();
             try
             {
-                _staged.TryGetValue(id, out transaction);
-
-                if (transaction is Transaction<T> tx)
-                {
-                    if (Expired(tx) || _ignored.Contains(tx.Id))
-                    {
-                        _staged.TryRemove(id, out _);
-                        return null;
-                    }
-                    else if (filtered)
-                    {
-                        return blockChain.Store.GetTxNonce(blockChain.Id, tx.Signer) <= tx.Nonce
-                            ? tx
-                            : null;
-                    }
-                    else
-                    {
-                        return tx;
-                    }
-                }
-                else
-                {
-                    return null;
-                }
+                return GetInner(blockChain, id, filtered);
             }
             finally
             {
-                _lock.ExitWriteLock();
+                _lock.EnterReadLock();
             }
         }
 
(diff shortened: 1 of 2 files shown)


Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
```diff
@@ -189,39 +189,14 @@ public bool Ignores(BlockChain<T> blockChain, TxId id)
         /// <inheritdoc/>
         public Transaction<T>? Get(BlockChain<T> blockChain, TxId id, bool filtered = true)
         {
-            Transaction<T>? transaction = null;
-
-            _lock.EnterWriteLock();
+            _lock.EnterReadLock();
```
greymistcube, line 192: Write lock was used since we wanted to modify `_staged` during lookup. Are we sure this won't cause a problem?

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
```diff
@@ -275,5 +250,31 @@ public long GetNextTxNonce(BlockChain<T> blockChain, Address address)
...
+                else
+                {
+                    return tx;
+                }
+            }
+            else
+            {
+                return null;
+            }
+        }
```
greymistcube, line 278: Not sure why this was separated out. Detaching this from the `lock` statement might make things more prone to mistakes. We might need to at least document it should be operating under a lock?

longfin: I didn't want unnecessarily lock for inner invocations (e.g., `.Iteratate()` invokes `.Get()` and it already has a read lock).

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
longfin: At least from what I checked, it would be okay because...

1. `_staged` is already `ConccurrentDictionary<K, V>` so it might be safe against multiple threads itself.
2. While `_staged` updating, I think that we wouldn't perform any other ops(i.e., `.TryUpdate()` and return null) that should be synchronized as a writer (not a reader because we're already in that scope).

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
greymistcube: Ah, it is shared with `Iterate()`. I missed that part. 🙃

Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
greymistcube: It almost seems okay also to my eyes but still...

- `_staged` being `ConcurrentDictionary<K, V>` isn't that relevant. It has to do with look-up and conditional check operations being separate from `TryRemove()` call.
- While `_staged` is updating, external context may change, i.e. `Expired(tx) || _ignored.Contains(tx.Id)`, which is affected by other calls such as (`Ignore()`).

It is not that clear that this would be a safe change that doesn't break the concurrency of `VolatileStagePolicy`'s internal states.

longfin: > _staged being ConcurrentDictionary<K, V> isn't that relevant. It has to do with look-up and conditional check operations being separate from TryRemove() call.

Right, but nevertheless I mentioned `ConcurrentDictionary<K, V>` since plain `Dictionary<K, V>` tends to misbehave even just one operation in a multi-threaded environment. 😅 

> While _staged is updating, external context may change, i.e. Expired(tx) || _ignored.Contains(tx.Id), which is affected by other calls such as (Ignore()).

Basically, locking mechanisms are used to prevent invalid behaviors by controlling the accessibility of an intermediate state. but in this case, there are several points to consider 🤔 

1. As this logic only mutate `_staged` and the mutation against it might be atomic (due to `ConcurrentDictionary<K, V>`). it means that other viewers have no write lock or no difference.
2. At the same time, we must protect the current execution context from being affected by changes in other threads. and I guess that it's okay because we've acquired `_lock` as reader and other threads can't enter as writer.
    - Of course,  since these read-write locks are entirely based on the programmer's code, it's also possible that assumptions can be broken by improper locking. 
    - But if we assume that it won't acquire write locks properly, then all read locks in the class, not just this logic, will be meaningless.

longfin: > At the same time, we must protect the current execution context from being affected by changes in other threads. and I guess that it's okay because we've acquired _lock as reader and other threads can't enter as writer.

But yeah.. on second thought, I may need to more consider `Get()` vs other `Get()`s... 🤔

greymistcube (dismissed):

PR #2595, opened 2022-11-30
♻️ Refactor `VolatileStagePolicy`

greymistcube:
♻️ 

Several points:
- Handling of ignored and expired `Transaction<T>`s can be completely separated. There was no reason to silently unstage ignored `Transaction<T>`s during retrieval. It isn't even necessary to reference ignored `Transaction<T>`s as long as we keep the internal states `_ignored` and `_staged` concurrent.
- Wanting to unstage on-the-fly while iterating over `_staged` seemed problematic causing frequent write lock and unlock. I made it so the removal can be done in bulk at the end.

### DIFF
--- a/Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
+++ b/Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Libplanet.Action;
 using Libplanet.Tx;
 using Serilog;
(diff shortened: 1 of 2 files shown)


Libplanet/Blockchain/Policies/VolatileStagePolicy.cs
```diff
@@ -248,37 +293,73 @@ public long GetNextTxNonce(BlockChain<T> blockChain, Address address)
...
+        /// aggregate expired <see cref="Transaction{T}"/>s and remove them in bulk.
+        /// </para>
         /// </remarks>
-        private Transaction<T>? GetInner(BlockChain<T> blockChain, TxId id, bool filtered)
+        private (Transaction<T> Transaction, bool Expired)? GetInner(
+            BlockChain<T> blockChain, TxId id, bool filtered)
         {
             if (_staged.TryGetValue(id, out Transaction<T>? tx) && tx is { })
             {
-                if (Expired(tx) || _ignored.Contains(tx.Id))
```
longfin, line 262: is it okay we don't care about `_ignored` here?

greymistcube: `Ignore()` has been changed to manage `_ignored` to be in tandem with `_staged`. The only way for a `TxId` to be marked ignored is through `Ignore()` and when marked, it is always removed from `_staged`.

greymistcube:
So... I've reworked things a bit to reduce the lock scope, but it seems a bit hacky with `Task.Run()` (see 0eb4e977e80256310a9b22510690067371faa504).

greymistcube:
Welp, I don't see a way to remove more `_lock`s safely without complete reimplementation. Probably needs to be worked on from a completely different angle. 😕

PR #2607, opened 2022-12-04
Context bootstrapping

greymistcube:
This PR allows network nodes to come online sequentially, at least under the assumption all nodes' `BlockChain<T>` are synced[^1].

[^1]: If not, there are cases where a previously committed `Block<T>` can be invalidated. Consider the following scenario. Let there be a single node with the highest tip index (for whatever reason, before shutting down the network, other nodes might have not been able to sync and/or commit the latest `Block<T>`). If such node does not come online before 2/3+ of the network and reach the consensus, effectively, the said node's tip will be ignored.

### DIFF
--- a/Libplanet.Net/Consensus/MessageLog.cs
+++ b/Libplanet.Net/Consensus/MessageLog.cs
@@ -346,5 +346,21 @@ namespace Libplanet.Net.Consensus
                 }
             }
         }
+
+        internal ConsensusMsg? GetRandomMessage()
+        {
+            lock (_lock)
+            {
+                Random random = new Random();
+                List<ConsensusMsg> pool = new List<ConsensusMsg>()
+                    .Concat(_proposals.Values)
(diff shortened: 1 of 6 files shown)


Libplanet.Net/Consensus/MessageLog.cs
```diff
@@ -346,5 +346,21 @@ internal int GetTotalCount()
...
+                    .Concat(_proposals.Values)
+                    .Concat(_preVotes.Values.SelectMany(kv => kv.Values))
+                    .Concat(_preCommits.Values.SelectMany(kv => kv.Values))
+                    .ToList();
+                return pool.Count > 0
+                    ? pool[random.Next(pool.Count)]
+                    : null;
+            }
+        }
     }
```
riemannulus, line 365: IMO select from own message only is way better for this scenario.

PR #2609, opened 2022-12-05
Removed automatic internal reassignment of `Store` to `StateStore`

greymistcube:
Issues:
- We don't really have any concrete `class` that implements both `IStore` and `IStateStore`.
- In any case, if one wishes to use the same object for both storing blocks and states, it shouldn't be up to `BlockChain<T>` to decide. The caller of `BlockChain<T>` should be responsible by passing the same object as both `IStore` and `IStateStore`.
- If we want to support the current behavior (or something similar to it), then we should have a combined `interface` such as `ISuperStore : IStore, IStateStore` to make the intention more pronounced.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -213,13 +213,11 @@ namespace Libplanet.Blockchain
         {
             if (store is null)
             {
-                throw new ArgumentNullException(
-                    nameof(store), $"Given {nameof(store)} cannot be null");
+                throw new ArgumentNullException(nameof(store));
             }
             else if (stateStore is null)
             {
-                throw new ArgumentNullException(
-                    nameof(stateStore), $"Given {nameof(stateStore)} cannot be null");
+                throw new ArgumentNullException(nameof(stateStore));
             }
 
             Id = id;
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -81,6 +81,8 @@ namespace Libplanet.Blockchain
         /// events made by unsuccessful transactions too; see also
         /// <see cref="AtomicActionRenderer{T}"/> for workaround.</param>
         /// <param name="stateStore"><see cref="IStateStore"/> to store states.</param>
+        /// <exception cref="ArgumentNullException">Thrown when either of <paramref name="store"/>
+        /// or <paramref name="stateStore"/> is <see langword="null"/>.</exception>
         /// <exception cref="InvalidGenesisBlockException">Thrown when the <paramref name="store"/>
         /// has a genesis block and it does not match to what the network expects
         /// (i.e., <paramref name="genesisBlock"/>).</exception>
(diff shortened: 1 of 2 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -209,24 +211,29 @@ ActionEvaluator<T> actionEvaluator
             IBlockChainStates<T> blockChainStates,
             ActionEvaluator<T> actionEvaluator)
         {
+            if (store is null)
+            {
+                throw new ArgumentNullException(
+                    nameof(store), $"Given {nameof(store)} cannot be null");
+            }
+            else if (stateStore is null)
```
longfin, line 219:
```suggestion

            if (stateStore is null)
```

Just nitpick.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -209,24 +211,29 @@ ActionEvaluator<T> actionEvaluator
             IBlockChainStates<T> blockChainStates,
             ActionEvaluator<T> actionEvaluator)
         {
+            if (store is null)
+            {
+                throw new ArgumentNullException(
+                    nameof(store), $"Given {nameof(store)} cannot be null");
```
longfin, line 217: Would we need this message? 🤔

longfin (approved):

Libplanet/Blockchain/BlockChain.cs
greymistcube: Eh, I suppose `ArgumentNullException` does sufficient job in this case and we don't need to elaborate (or provide a custom message). 🙄

Libplanet/Blockchain/BlockChain.cs
greymistcube: I think it's a personal preference thing. It's logically the same, and depends on how you look at it.

- Separate `if` statements indicate two conditionals are conceptually separate, which they are in the sense that there is no inherent dependency between `store` and `stateStore`.
- Use of `if` and `else if` statements is to more closely follow the actual code flow itself as early termination due to the first conditional means the second won't ever run. Additionally, I tend to prefer a chain of `if` and `else if`s, especially within a constructor, if they can be logically checked upfront to represent "if the first constraint is valid, then check the next one".

As always, this depends on what we're checking and how complicated the logic is. I think both have their own pros and cons. 😗

longfin: Well, yeah. I think that it can be a perspective difference about a sort of narrative(?) upon these validity checks as you said. 😅 

- If they have an important dependency on each other to the extent that this chaining has meaning, I think it can be seen as a problem with the situation itself where these two are input separately.
  - I also think the nullability check is too trivial(?) to reveal the semantics between `store` and `stateStore`.
- There are cases where tools related to code style suggest omissions in this case, and this is one of those cases.
  - Since we want the project to be contributed by multiple contributors (regardless of whether or not I agree with these tools), I also feel like I want to conform to that lower bar to some degree.
  - Of course, it can also be covered by project-level settings.
- It's a thought I've been thinking about for a while, but maybe turning on the C# nullable check feature and appending `?` to the codes that assume null now is also a way. (even if it's not this PR)

The comment is a bit long, but I think if - else if chains are fine too. :)

longfin (approved):

Libplanet/Blockchain/BlockChain.cs
greymistcube: Welp, whatever the "lower bar" is, I can't foresee a third party tool's complaint. 😕
It's not my hill to die on, so if we are to enforce such rule via the project settings, I won't particularly complain.
I think it is more important to have conditionals at the top, if possible, especially when they aren't side-effect dependent at runtime(?), then having them scattered around just before each assignment (again, this could also be a personal preference thing tho). 😗

PR #2614, opened 2022-12-08
Exclude linux-unity-test on workflow

OnedgeLee:
Temporarily exclude `linux-unity-test` on workflow, due to the frequent error on `xunit-unity-runner`.
This change can be reverted on the condition the problem been solved.


riemannulus:
It seems need changelog skip.

OnedgeLee:
> It seems need changelog skip.

Amended!

longfin (approved):

Issue #2174, opened 2022-07-22
Remove `bundle` Github Actions workflow


tkiapril:
Apparently all-in-one bundles (#811) are no longer used in Unity tests. To speed up CI checks, removing it may be beneficial.

riemannulus:
Resolved at #2625

PR #2625, opened 2022-12-12
test: upscale ci runner

riemannulus:
If we upscale ci runner plain, those flaky tests will be much better than now?


dahlia (dismissed):

riemannulus:
I also remove `main.yaml` workflow in this PR. see also #2174

longfin (approved):

PR #2654, opened 2022-12-20
Improve `NetMQTransport` again

longfin:
This PR fixes internal codes using NetMQ in `NetMQTransport` to avoid blocking and improve latency (like #2631 ).

- Add `.CofigureAwait(false)` to `await`s, on NetMQ context-free methods. (e.g., `PingAsync()`).
- Increased I/O thread of NetMQ from 1 to 3.
- Increased `NetMQRuntime` itself (not its task), from 1 to `Environment.ProcessorCount`.
   - I tried making this value follow `workers`, but it gave an error when there were 200+ runtimes.
   - In a regular release it would be nice to remove `workers` and replace it with a constant or add it as another option. 
- `DealerSocket`s creating for `SendMessageAsync()` will have a `DisableTimeWait` option to avoid TCP TIME_WAIT.

Also, it tweaks `KBucketDictionary` a little bit too.
- Loosened the lock in `KBucketDictionary` by introducing `ReaderWriterLockSlim` instead of a single lock object.
- Tweaked `.Head`, `.Tail` by `.Aggregate()` instead of full ordering.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@ Version 0.44.5
 
 To be released.
 
+ -  Improved overall performance of `NetMQTransport` and `Swarm<T>`
+    classes. [[#2654]]
+
+[#2654]: https://github.com/planetarium/libplanet/pull/2654
+
 
 Version 0.44.4
 --------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@ Version 0.44.5
 To be released.
 
  -  Improved overall performance of `NetMQTransport` and `Swarm<T>`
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -6,6 +6,11 @@ Version 0.44.5
 
 To be released.
 
+ -  Improved overall performance of `NetMQTransport` and `Swarm<T>`
+    classes. [[#2654]]
```
limebell, line 10:
```suggestion
    classes.  [[#2654]]
```

limebell (approved):

PR #2664, opened 2022-12-23
🧹 Removed `IStore.GetCanonicalGenesisBlock<T>()`

greymistcube:
See [the discussion](https://github.com/planetarium/libplanet/discussions/2617) for more context.

The naming `GetCanonicalGenesisBlock<T>()` suggests there can be non-canonical genesis `Block<T>`s in `IStore`. This is problematic since this can be interpreted two different ways:

- There can be other genesis `Block<T>`s in `IStore`
- There can be some other chain id with a different genesis `Block<T>` in `IStore`

If it is the former, then there isn't much of a problem , but since the term "canonical" is closely tied to chain ids, I would say it is more likely than not to be interpreted as the latter. The rest of the API for `IStore` allowing there to be a chain id with a different genesis `Block<T>` also seem to suggest that the latter may be intended.

In any case, and thankfully, `IStore.GetCanonicalGenesisBlock<T>()` doesn't seem to be really used anywhere, so I think its just better to remove it altogether to reduce overall confusion.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,9 @@ To be released.
  -  Replaced `IAction?`-typed `policyBlockAction` parameter with
     `PolicyBlockActionGetter`-typed `policyBlockActionGetter` parameter
     in `ActionEvaluator` constructor.  [[#2646]]
+ -  Removed `IStore.GetCanonicalGenesisBlock<T>()` interface method and all its
+    implementations.  [[#2664]]
+
 
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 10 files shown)


CHANGES.md
```diff
@@ -8,6 +8,9 @@ To be released.
 
 ### Deprecated APIs
 
+ -  Removed `IStore.GetCanonicalGenesisBlock<T>()` interface method and all its
+    implementations.  [[#2664]]
```
dahlia, line 12: Shouldn't it be placed under <q>Backward-incompatible API changes</q>?

greymistcube: You mean because the change is `interface` related?

greymistcube: Fixed!

dahlia (approved):

PR #2669, opened 2022-12-26
🧹 Minor refactor for `KBucketDictionary`

greymistcube:
🧹


longfin:
There is the code in the same style on `.Head` too. how about modifying it too?

greymistcube:
@longfin Fixed!

longfin (approved):

PR #2679, opened 2022-12-28
Fix benchmark problem

riemannulus:
1. use `gh-pages`'s previous bench instead cache.
2. merging multiple benchmarks
3. using a merged bench instead of running action multiple times.


.github/workflows/benchmarks-pr.yml
```diff
@@ -20,78 +20,78 @@ jobs:
     - uses: actions/checkout@v3
       with:
         ref: ${{ github.pull_request.head.sha }}
-    - uses: actions/setup-dotnet@v3
+    - uses: actions/setup-dotnet@v1
```
moreal, line 23: I wonder the reason to roll back its version.

riemannulus: just mistake 😓  I'll fix this mistake

greymistcube (approved):

PR #2681, opened 2022-12-28
🔧 Try running benchmarks to completion on other os failure

greymistcube:
If any of the operating system fails, then all benchmarks have to be run again. I suppose this could be intended depending on our policy. 🙄


riemannulus:
Should we merge this PR after https://github.com/planetarium/libplanet/pull/2682

riemannulus (approved):

PR #2684, opened 2022-12-29
Fix bugs and stabilize unit test again

longfin:
This PR fixes a bug causing instability in environments with limited available CPU resources (e.g., CI).
it also fixes that `NetMQTransport.ReplyMessageAsync()` hadn't worked properly. [^1]

[^1]: To be honest, I thought that the method would never wait, so I plan to fix the interface in the main, but I couldn't because Gossip, which is for PBFT, relies on that wait.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,13 @@ Version 0.44.7
 
 To be released.
 
+ -  (Libplanet.Net) Fixed bugs where `NetMQTransport` hadn't worked expected
+    when not many threads were available.  [[#2684]]
+ -  (Libplanet.Net) Fixed a bug where `NetMQTransport.ReplyMessageAsync()`
+    hadn't worked properly.  [[#2684]]
+
+[#2684]: https://github.com/planetarium/libplanet/pull/2684
+
 
 Version 0.44.6
 --------------
--- a/Libplanet.Net.Tests/SwarmTest.Broadcast.cs
+++ b/Libplanet.Net.Tests/SwarmTest.Broadcast.cs
@@ -529,7 +529,7 @@ namespace Libplanet.Net.Tests
                 await swarmB.AddPeersAsync(new[] { swarmC.AsPeer }, null);
 
                 swarmA.BroadcastTxs(new[] { tx3, tx4 });
-                await swarmC.TxReceived.WaitAsync();
+                await swarmB.TxReceived.WaitAsync();
 
                 // SwarmB receives tx3 and is staged, but policy filters it.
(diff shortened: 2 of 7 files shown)


limebell (dismissed):

Libplanet.Net/Swarm.cs
```diff
@@ -217,6 +217,8 @@ public partial class Swarm<T> : IDisposable
 
         internal AsyncAutoResetEvent TxReceived => TxCompletion?.TxReceived;
 
+        internal AsyncAutoResetEvent TxStaged => TxCompletion?.TxStaged;
+
```
greymistcube, line 221: This feels a bit icky. The amount of knowledge required as to why this is set up this way seems vast.

- There's a huge discrepancy between uses of `BlockCompletion` (where the instance is created for each use) and `TxCompletion` (where a single instance is used for `Swarm<T>`.
- Not sure handling of `TxReceived` and `TxStaged` in `TxCompletion` is apt. Though seemingly simple events, it is rather hard to concisely describe the conditions in which these events are raised.

Not that this would be easily fixable (I'd have no idea where to start), but I'm worried about the general maintainability of the code for this PR, for which this is just an example. 😶

limebell: I think maintenence is not a big problem because these events are not used in actual logic, but only used in tests.

longfin: its main motivation was the stabilization of tests. the previous `.TxReceived` has been used in two different situations. 

1. To check just receiving transactions (even if we except ignoring): https://github.com/planetarium/libplanet/blob/159984972bc6bc1d5629ef030b81bdf5b831c05c/Libplanet.Net.Tests/SwarmTest.cs#L1159
2. To check until staging: https://github.com/planetarium/libplanet/blob/87053edaa8bdf130604e5b4f0f414c0fe2e89d9d/Libplanet.Net.Tests/SwarmTest.Broadcast.cs#L277-L278

So I just intended to separate them via a dedicated event to fix flaky tests after #2631, but. well... on second thought, we can make [previous codes](https://github.com/planetarium/libplanet/pull/2631/files#diff-3ac16fe3c2da0cceabdfc2119fe56393dae424160f879d36ad87037166c378b5L179-R172) without new event. 

I'll amend that commit.

longfin: @greymistcube Amended. [53f1a99](https://github.com/planetarium/libplanet/pull/2684/commits/53f1a99fee173be26f75c870edc88a9e18c164c5)

greymistcube (approved):

PR #2687, opened 2023-01-03
Add unit tests related to `SetValidator` system action

limebell:
This test just including tests that checks the action actually modified the state.

### DIFF
--- a/Libplanet.Tests/Action/Sys/SetValidatorTest.cs
+++ b/Libplanet.Tests/Action/Sys/SetValidatorTest.cs
@@ -1,0 +1,131 @@
+using System;
+using System.Numerics;
+using Bencodex.Types;
+using Libplanet.Action;
+using Libplanet.Action.Sys;
+using Libplanet.Assets;
+using Libplanet.Blocks;
+using Libplanet.Consensus;
+using Libplanet.Crypto;
+using Xunit;
+using static Libplanet.Tests.TestUtils;
+using Random = System.Random;
+
+namespace Libplanet.Tests.Action.Sys
+{
+    public class SetValidatorTest
+    {
+        [Fact]
+        public void Constructor()
+        {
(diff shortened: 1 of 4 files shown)


Libplanet.Tests/Action/Sys/SetValidatorTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            var random = new Random();
+            Address signer = random.NextAddress();
+            Currency bazCurrency = Currency.Uncapped("BAZ", 0, random.NextAddress());
+            var prevStates = new AccountStateDeltaImpl(
+                accountStateGetter: addr => new IValue[addr.Count],
+                accountBalanceGetter: (addr, c) => c * 0,
+                totalSupplyGetter: c => c * 0,
+                validatorSetGetter: () => new ValidatorSet(),
+                signer: signer
+            );
```
OnedgeLee, line 70: May be some `AccountStateDeltaImpl` mockup fixture could be needed?
This `prevStates` construction seems to be repeated on `Libplanet.Tests.Action.Sys`.
Defining parent test base class for system actions seems to be better for me.

Libplanet.Tests/Action/Sys/SetValidatorTest.cs
```diff
@@ -0,0 +1,131 @@
...
+                miner: random.NextAddress(),
+                blockIndex: 123L,
+                previousStates: prevStates,
+                randomSeed: 123,
+                rehearsal: false,
+                previousBlockStatesTrie: null,
+                blockAction: false,
+                genesisHash: genesisHash,
+                nativeTokenPredicate: null
+            );
```
OnedgeLee, line 84: These seems better to be defined as some fixture, also.

OnedgeLee (approved):
I've left some opinion, but it seems like just a matter of preference.
Since test code size is small, I think it can be better not to make some base fixture.

PR #2701, opened 2023-01-10
Introduce `HostOptions`

greymistcube:
Continuation of #2693.

On a side note, I thought about making `HostOptions` as a plain collection of properties with `{ get; set; }` for each, like `AppProtocolVersionOptions` and `SwarmOptions`, but decided against it. This breaks overall consistency, but I think parsing options should be done at "options level". If anyone disagrees with the design, I can change it back. 🙄

### DIFF
--- a/Libplanet.Net/HostOptions.cs
+++ b/Libplanet.Net/HostOptions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 
@@ -11,7 +12,7 @@ namespace Libplanet.Net
     {
         public HostOptions(
             string? host,
-            ImmutableList<IceServer> iceServers,
+            IEnumerable<IceServer> iceServers,
             int port = 0)
         {
             if (port < 0)
@@ -25,7 +26,7 @@ namespace Libplanet.Net
             }
 
             Host = host;
-            IceServers = iceServers;
+            IceServers = iceServers.ToImmutableList();
             Port = port;
         }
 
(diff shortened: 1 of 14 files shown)


Libplanet.Net/HostOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+            IEnumerable<IceServer> iceServers,
+            int port = 0)
+        {
+            if (port < 0)
+            {
+                throw new ArgumentException($"Given {nameof(port)} cannot be negative: {port}");
+            }
+            else if (!(host is { } ^ iceServers.Any()))
+            {
+                throw new ArgumentException(
```
longfin, line 24: Just an idea: how about sub-classing `HostOptions` and enforcing to choose of one via type (e.g., `TurnProxyHostOptions`, `StaticHostOptions`)?

greymistcube: I've had a similar idea, but wasn't sure if it'd be better to off-load type parsing responsibility to the client. As it stands, thought it'd be easier from a user standpoint to just dump all arguments and have the library to tell early on that certain combinations of arguments can't be used together. I think this can be later further refactored to include the said functionality/implementation when SDK/client API gets more streamlined. 🙄

longfin: I guess that we can offer a parsing feature even with these types if we need it. (e.g., `HostOptions.Create()`). 

and... I also think that it isn't parsing actually because options for the relay server and its host don't share the same representation.

greymistcube: Eh, tried implementing `TrunProxyOptions` and `StaticHostOptions` using `HostOptions.Create()`, but seemed like a bit of an overkill. 😐

I have some vague idea about creating an additional layer for handling `HostOptions` where type handling of subclasses of the `HostOptions` class might be more appropriate. Imma kick this can down the road a bit more. 😗

PR #2703, opened 2023-01-10
Remove StateCompleters

longfin:
This PR removes state completer feature with related APIs. I still believe state completion concept is worth in various situations but current implementation(i.e., only recalcuation) can't help that situations due to performance. 😢 

Also, it drops type parameter `T` from `ActionEvaluator` to fit `IActionTypeLoader` semmentics.


longfin:
Related discussion: #2440

longfin:
I'll rebase this PR onto #2863 since that tweaks `BlockChain<T>`'s constructors. cc @planetarium/libplanet

longfin:
I've revised it for continuing from 4818f4b67a932f2b1caa8c782b2e34d4a153f357

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -228,18 +197,21 @@ internal static ValidatorSet NullValidatorSetGetter()
 
         /// <summary>
         /// Retrieves the set of <see cref="Address"/>es that will be updated when
-        /// a given <see cref="Transaction{T}"/> is evaluated.
+        /// a given <see cref="ITransaction"/> is evaluated.
         /// </summary>
-        /// <param name="tx">The <see cref="Transaction{T}"/> to evaluate.</param>
+        /// <param name="tx">The <see cref="ITransaction"/> to evaluate.</param>
```
moreal, line 202: The `tx` parameter seems like a `Transaction<T>` type yet. Is this okay?

moreal (commented):
It seems good to me. There is only one comment to ask.

Libplanet/Action/ActionEvaluator.cs
longfin: For `GetUpadatedAddresses()`, we still need to keep the type parameter since it needs materialization statically. but this behavior doesn't compatible with the action type loader concept and we need to remove it in near future. cc @planetarium/libplanet

in this PR, I'll fix the docstring and add comments about that.

CHANGES.md
```diff
@@ -19,6 +19,15 @@ To be released.
  -  `BlockChain<T>()` now throws an `ArgumentException` if provided `IStore`
     does not have its canonical chain id set or provided chain id is not found.
     [[#1486], [#2585], [#2889]]
+ -  Removed state completion feature with related classes as below.  [[#2703]]
+        -  `StateCompleter`
+        -  `StateCompleters`
+        -  `TotalSupplyStateCompleter`
+        -  `TotalSupplyStateCompleters`
+        -  `ValidatorSetStateCompleter`
+        -  `ValidatorSetStateCompleters`
```
dahlia, line 28: Trivial, but these lines are apparently too indented.

dahlia (approved):

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -233,14 +202,23 @@ internal static ValidatorSet NullValidatorSetGetter()
         /// <param name="tx">The <see cref="Transaction{T}"/> to evaluate.</param>
         /// <returns>An <see cref="IImmutableSet{T}"/> of updated <see cref="Address"/>es.
         /// </returns>
+        /// <typeparam name="T">An <see cref="IAction"/> type.  It should match
+        /// the <see cref="Block{T}"/>'s type parameter.</typeparam>
```
moreal, line 206: Well.. I'm not sure but it is better to use `Transaction{T}` instead of `Block{T}` in this method.

```suggestion
        /// <typeparam name="T">An <see cref="IAction"/> type.  It should match
        /// the <see cref="Transaction{T}"/>'s type parameter.</typeparam>
```

moreal (approved):

CHANGES.md
longfin: Fixed on https://github.com/planetarium/libplanet/pull/2703/commits/8881d80599f12b5347ad2442810f56dc1bf1b7b8.

Libplanet/Action/ActionEvaluator.cs
longfin: Fixed on https://github.com/planetarium/libplanet/pull/2703/commits/8ac312e3a74ae2b3fda68968bc7efb8c492dfc26.

PR #2708, opened 2023-01-13
fix/transport: socket count

colibrishin:
This PR addresses a logging bug where the count of sockets is going negative value. Before the start, I couldn't find the current socket count-related property or field from NetMQ.

### Socket is not allocated but canceled
The socket count is incremented when the `GetRequestDealerSocket()` is called. If the `CancellationToken.ThrowIfCancellationRequested()` is thrown, then the socket is not allocated (i.e., `GetRequestDealerSocket()` is not called, the count is not incremented.)

### Failed to allocate a socket
The `DealerSocket` in `GetRequestDealerSocket()` throws `NetMQException` if allocating socket is failed. In this case, the socket is not allocated (socket count remains) and breaks out from `new DealerSocket` and `GetRequestDealerSocket()`.

### Socket is allocated and canceled or exception thrown
The socket count is incremented by calling `GetRequestDealerSocket()`. `OperationCanceledException` also can be thrown at `dealer.ReceiveMultipartMessageAsync()`, as an example. In this case, the socket has been allocated (count has incremented), and cancellation has been requested.

### DIFF
--- a/Libplanet.Net/Transports/NetMQTransport.cs
+++ b/Libplanet.Net/Transports/NetMQTransport.cs
@@ -771,48 +771,53 @@ namespace Libplanet.Net.Transports
                 req.Peer
             );
             int receivedCount = 0;
+            DealerSocket dealer = null;
 
             // Normal OperationCanceledException initiated from outside should bubble up.
             try
             {
-                using var dealer = GetRequestDealerSocket(req);
+                cancellationToken.ThrowIfCancellationRequested();
 
-                if (dealer.TrySendMultipartMessage(req.Message))
+                using (dealer = GetRequestDealerSocket(req))
                 {
-                    _logger.Debug(
-                        "Request {RequestId} sent to {Peer}.",
-                        req.Id,
-                        req.Peer);
+                    if (dealer.TrySendMultipartMessage(req.Message))
+                    {
+                        _logger.Debug(
+                            "Request {RequestId} sent to {Peer}.",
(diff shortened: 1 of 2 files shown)


longfin (approved):

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -832,7 +837,10 @@ private async Task ProcessRequest(MessageRequest req, CancellationToken cancella
                     await Task.Delay(1000);
                 }
 
-                Interlocked.Decrement(ref _socketCount);
+                if (dealer is { IsDisposed: true })
+                {
+                    Interlocked.Decrement(ref _socketCount);
+                }
```
greymistcube, line 843: I'm not sure disposal check is any better than a simple `null` check. For whatever reason if sockets are not disposed properly, we might overcount the number of `DealerSocket`s instead. I think `ITransport<T>` layer should only be concerned with its own references if possible, and try not to deal with the underlying `NetMQ` layer. 😶

longfin: At least, `NetMQSocket.Close()` seems to turn off `.IsDisposed` immediately. but... yeah I also think that a simple null check is enough in this context.

longfin (approved):

greymistcube (approved):

longfin (approved):

PR #2710, opened 2023-01-13
♻️ Optimize preload

greymistcube:
This is mainly optimized for memory usage. Could be slower on tx restaging as `Transaction<T>`s are loaded again when traversing. Also possibly slightly less safe as instead of directly traversing a `BlockChain<T>` object, this assumes some business logic and directly accesses the `IStore`. 🙄

Shouldn't be much of a problem:
- Rewinding should be relatively small. Aside from trying to move to PBFT, if we have a long rewind path, we have a larger issue.
- Takes relatively smaller amount of time compared to block append and execution.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.Swap.cs
+++ b/Libplanet/Blockchain/BlockChain.Swap.cs
@@ -78,9 +78,9 @@ namespace Libplanet.Blockchain
 
                 Block<T> oldTip = Tip, newTip = other.Tip;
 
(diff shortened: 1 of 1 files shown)


longfin:
> Rewinding should be relatively small. Aside from trying to move to PBFT, if we have a long rewind path, we have a larger issue.

Agree with it deeply... and so I really hope we won't have to do the rewinding itself.

longfin (approved):

PR #2716, opened 2023-01-17
Pow dynamic validator

limebell:


Libplanet/Consensus/Validator.cs
```diff
@@ -0,0 +1,165 @@
...
+            {
+                return Equals(other);
+            }
+
+            return false;
+        }
+
+        public bool Equals(Validator? other)
+        {
+            return PublicKey.Equals(other?.PublicKey);
```
greymistcube, line 130: Is not considering power intended? If so, why?

OnedgeLee: What I've thought was, we're often use validator's identity as it's public key.
In practical use, when we compare validator `A` and validator `B`, we often don't pay attention to its powers. What we're interested in is their identity.
So, I've thought judging `Validator("A", 100)` same as `Validator("A", 1000)` would be better for practical use.
I thought, when we're interested in their power, we would call `ValidatorA.Power.Equals(ValidatorB.Power)`.
May people can confused with this notation?

OnedgeLee (dismissed):

Libplanet/Action/Sys/SetValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+        public SetValidator(PublicKey validatorPublicKey, BigInteger power)
+        {
+            ValidatorPublicKey = validatorPublicKey;
+            Power = power;
+        }
+
+        public SetValidator(Validator validator)
+            : this(validator.PublicKey, validator.Power)
+        {
+        }
```
greymistcube, line 31: Seems a bit redundant. I think `SetValidator(Validator)` should suffice?

limebell: For now, yes but in some cases such as mutation in headless, `SetValidator(PublicKey validatorPublicKey, BigInteger power)` can be used.

Libplanet/Consensus/Validator.cs
OnedgeLee: Resolved

CHANGES.md
```diff
@@ -56,9 +66,20 @@ To be released.
...
+ -  Added `ValidatorSetStateCompleter<T>` delegate.  [[#2716]]
+ -  Added `ValidatorSetStateCompleters<T>` class.  [[#2716]]
+ -  (Libplanet.Explorer) Added `ValidatorType` class.  [[#2716]]
+ -  (Libplanet.Explorer) Added `validators` query in `StateQuery`.  [[#2716]]
 
 ### Behavioral changes
 
+ -  `Swarm<T>` became to execute actions one by one when preloading blocks.
+    (used to execute actions after appending all blocks)  [[#2716]]
+
```
greymistcube, line 82: This no longer applies...? 😶

OnedgeLee (approved):

greymistcube (approved):

PR #2719, opened 2023-01-17
Fix NetMQSocket leak

longfin:
This PR fixes a leakage of `NetMQSocket` on `NetMQTransport.SendMessageAsync()` when it has failed on connecting.

I splitted commits with intention because

- We should back-port this change to 0.45-maintenance after merging with #2708.
- I borrowed 2a194d7 from my other experimental PR(#2714) and perhaps I think the current way of handling is enough. however, there may be better implementations or implementations also need test cases for confirmation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -84,6 +84,9 @@ To be released.
 
  -  (Libplanet.Net) Fixed a bug `NetMQTransport` log shows socket count wrongly.
     [[#2708]]
+ -  (Libplanet.Net) Fixed a bug where `NetMQTransport.SendMessageAsync()` method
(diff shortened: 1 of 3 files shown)


colibrishin (approved):

limebell (dismissed):

longfin:
I've updated the test case since we need to prepare a more specific hostname to mimic DNS lookup failure after #2721 .  cc @moreal 

PTAL @limebell @colibrishin

limebell (commented):
LGTM, but confict should resolved before merge

limebell (approved):

PR #2722, opened 2023-01-18
feat/kademlia: handshake before add peer to table

colibrishin:
Based on #2719 and #2721 (a case of `BoundPeer` with an unreachable hostname thrown in `DealerSocket.connect()`), a peer with an unreachable hostname can be added to `RoutingTable`. This PR suggests adding a `BoundPeer` only if one can be sent `PingMsg` to and can respond `PongMsg`.

### DIFF
--- a/Libplanet.Net.Tests/Protocols/ProtocolTest.cs
+++ b/Libplanet.Net.Tests/Protocols/ProtocolTest.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Libplanet.Crypto;
+using Libplanet.Net.Messages;
 using Libplanet.Net.Protocols;
 using Libplanet.Net.Transports;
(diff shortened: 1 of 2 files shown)


colibrishin:
Closed due to further investigation needed. 

- This could prevent existing unreachable hosts in `RoutingTable`, so a node will not try to send a request in the first place. 
  - When the host, which has the correct format of hostname, cannot be resolved from DNS, then it will immediately fail to connect, catching an exception, and ignoring the handshake.
- However, a host with no responses holds the socket for an amount of timeout, If my understanding of `NetMQ` is correct. In that perspective, solely this method doesn't prevent a message or contact from a reachable but no responding host.
- It can be somehow controlled with a "peer quality" variable abstractly, but there might be a simple solution or a more concrete idea to prevent all of the cases. And also, I think we need time and confidence that the current transport implementation is stable.

PR #2723, opened 2023-01-18
Implement dynamic validator

### LINKED ISSUE #2576: Introduce `Validator`
For now, validator is used as its public key, but consensus power is also needed.
New struct `Validator` feature will be added, and will be used on the consensus.

limebell:
This patch adds `SetValidator` action in order to manage validators as blockchain state.

Closes #2576


Libplanet.Net.Tests/TestUtils.cs
```diff
@@ -275,7 +270,8 @@ in PrivateKeys.Zip(Peers, (first, second) => (first, second)))
                 long height = 1,
                 IBlockPolicy<DumbAction>? policy = null,
                 PrivateKey? privateKey = null,
-                ContextTimeoutOption? contextTimeoutOptions = null)
+                ContextTimeoutOption? contextTimeoutOptions = null,
+                ValidatorSet? validatorSet = null)
```
OnedgeLee, line 274: I think this parameter would be better to be omitted.
`CreateDummyContext()` returns both of `BlockChain<T>` and `Context<T>`.
There were no dependency on `ValidatorSet` with `BlockChain<T>` for static validators, but on dynamic validators, dependency has been made.
So, If we inject `validatorSet`, returned `BlockChain<T>` and `Context<T>` would have different `ValidatorSet`.

Libplanet.Net.Tests/Consensus/Context/ContextProposerTest.cs
```diff
@@ -230,7 +230,8 @@ public async void EnterPreVoteNil()
             var stepChangedToPreVote = new AsyncAutoResetEvent();
             var nilPreVoteSent = new AsyncAutoResetEvent();
             var (_, context) = TestUtils.CreateDummyContext(
-                height: 5); // Peer1 should be a proposer
+                height: 5,
+                validatorSet: Libplanet.Tests.TestUtils.ValidatorSet); // Peer1 should be a proposer
```
OnedgeLee, line 234:
```suggestion
                height: 5); // Peer1 should be a proposer
```
Seems like original code would work properly.

Libplanet.Net.Tests/Consensus/Context/ContextProposerTest.cs
```diff
@@ -304,7 +305,8 @@ public async void CannotProposeWithoutLastCommitWhenRequired()
             Exception? exception = null;
             var (blockChain, context) = TestUtils.CreateDummyContext(
                 privateKey: TestUtils.PrivateKeys[2],
-                height: 2);
+                height: 2,
+                validatorSet: TestUtils.ValidatorSet);
```
OnedgeLee, line 309:
```suggestion
                height: 2);
```

Libplanet.Net.Tests/Consensus/Context/ContextTest.cs
```diff
@@ -74,7 +74,8 @@ public async void StartAsProposerWithLastCommit()
             // is the proposer for height 2.
             var (blockChain, context) = TestUtils.CreateDummyContext(
                 height: 2,
-                privateKey: TestUtils.PrivateKeys[2]);
+                privateKey: TestUtils.PrivateKeys[2],
+                validatorSet: Libplanet.Tests.TestUtils.ValidatorSet);
```
OnedgeLee, line 78:
```suggestion
                privateKey: TestUtils.PrivateKeys[2]);
```

Libplanet.Net.Tests/Consensus/Context/ContextTest.cs
```diff
@@ -122,7 +123,8 @@ public async Task CanAcceptMessagesAfterCommitFailure()
 
             var (blockChain, context) = TestUtils.CreateDummyContext(
                 height: 2,
-                privateKey: TestUtils.PrivateKeys[2]);
+                privateKey: TestUtils.PrivateKeys[2],
+                validatorSet: TestUtils.ValidatorSet);
```
OnedgeLee, line 127:
```suggestion
                privateKey: TestUtils.PrivateKeys[2]);
```

Libplanet.Net.Tests/TestUtils.cs
limebell: Ommitting this causes `ArgumentOutOfException` exception.

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2032,6 +2049,12 @@ private void ConstructWithGenesisBlock()
                 .Add(storeFixture.Address2)
                 .Add(storeFixture.Address3);
 
+            var validatorPrivKey = new PrivateKey();
+            var systemActions = new IAction[]
+            {
+                new SetValidator(new Validator(validatorPrivKey.PublicKey, BigInteger.One)),
+            };
```
OnedgeLee, line 2056: Seems better to use validator private keys on `TestUtils`

colibrishin (approved):

OnedgeLee (approved):

Libplanet.Tests/Blockchain/BlockChainTest.cs
limebell: I think it's better to create a new key every time cause it makes this test more secure. Hard coded value may leads some corner cases, not in this case though.

OnedgeLee: What I was concerned was the number of validators. We don't prevent single validator, but I think it's abnormal. Should `ValidatorSet` have to set spec for number of validators?

PR #2727, opened 2023-01-18
Elide duplicated prefix of paths in Web3KeyStore & Release 0.45.5

longfin:
This PR adjusts internal codes of `Web3KeyStore`, to remove duplicated prefixes from return values of `.ListFiles()`. ([`Directory.EnumerateFiles()`](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.enumeratefiles?view=netstandard-2.0#system-io-directory-enumeratefiles(system-string)) returns full name so we don't need to combine return values and `Path`)

I couldn't prepare a regression test case since `System.IO.Path.Combine()` seems to elide duplicated parts automatically. but it didn't work so that on IL2CPP environment.

p.s. thanks for helping debug :) @sonohoshi

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ Libplanet changelog
 Version 0.45.5
 --------------
 
-To be released.
+Released on January 19, 2023.
 
   -  Fixed a bug when `Web3KeyStore.Get()` hadn't worked properly on IL2CPP
      environment.  [[#2727]]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@ Version 0.45.5
 
 To be released.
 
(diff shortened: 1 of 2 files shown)


longfin:
> The combined paths. If one of the specified paths is a zero-length string, this method returns the other path. If path2 contains an absolute path, this method returns path2.

More correctly, [`Path.Combine()`](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netstandard-2.0#system-io-path-combine(system-string-system-string)) returns `path2` without any concatenation if `path2` is an absolute path. in other words, the return value of `Directory.EnumerateFiles()` might not be considered an absolute path, at least IL2CPP build.

dahlia (dismissed):

dahlia (approved):

PR #2733, opened 2023-01-19
Remove Validator related methods from interfaces

limebell:
`Validator`s dealt only in internal methods, so it had been removed from public interfaces.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  Removed `IAccountStateView.GetValidatorSet()` method.  [[#2733]]
+ -  Removed `IAccountStateDelta.SetValidator(Validator)` method.  [[#2733]]
  -  (Libplanet.Extensions.Cocona)  Dropped .NET Standard 2.0 and .NET Core 3.1
     target assemblies.  [[#2732]]
  -  (Libplanet.Extensions.Cocona)  Added .NET 6 target assembly.  [[#2732]]
@@ -29,6 +31,7 @@ To be released.
 ### CLI tools
 
 [[#2732]]: https://github.com/planetarium/libplanet/pull/2732
+[[#2733]]: https://github.com/planetarium/libplanet/pull/2733
 
 
 Version 0.46.0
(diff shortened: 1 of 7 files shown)


OnedgeLee (dismissed):

CHANGES.md
```diff
@@ -20,6 +22,8 @@ To be released.
 
 ### Added APIs
 
+ -  Added `IValidatorSupportStateDelta` interface.  [[#2733]]
```
longfin, line 25: If our main motivation is hiding validator-related features on `IAccountStateDelta` & `IAccountStateView` against applications, how about hiding this interface too?

Libplanet/Action/ValidatorStateExtensions.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using Libplanet.Consensus;
+
+namespace Libplanet.Action
+{
+    internal static class ValidatorStateExtensions
+    {
+        public static ValidatorSet GetValidatorSet(this IAccountStateDelta delta)
```
longfin, line 8: AFIK, placing `public` members onto `internal` class can be a little bit debated ([see also](https://ericlippert.com/2014/09/15/internal-or-public/) 

but [just for consistency](https://github.com/planetarium/libplanet/blob/0.46.0/Libplanet/Action/AccountStateDeltaExtensions.cs#L12), I support using `internal` atm. 😕

limebell: Many libplanet classes now using `internal` extension but `public` methods. (`NetMQFrameExtensions`, `BytesConvertExtensions`, `StunAddressExtensions`, `ActionEvaluationsExtensions`, `EnumerableExtensions`, `KeyValuePairExtensions` and many more..)

CHANGES.md
limebell: Applied and squashed!

OnedgeLee (approved):

Libplanet/Action/ValidatorStateExtensions.cs
longfin: I see... then, it'd be good to incorporate them in another PR or discussion.

longfin (approved):

PR #2742, opened 2023-01-20
log/transport: alternative Identity for NetMQTransport log

### LINKED ISSUE #2738: Set identity of DealerSocket as MessageRequest.Id

colibrishin:
Closes #2738.

### DIFF
--- a/Libplanet.Net/Transports/NetMQTransport.cs
+++ b/Libplanet.Net/Transports/NetMQTransport.cs
@@ -504,8 +504,10 @@ namespace Libplanet.Net.Transports
                 throw new ObjectDisposedException(nameof(NetMQTransport));
             }
 
-            string identityHex = ByteUtil.Hex(message.Identity);
-            _logger.Debug("Reply {Message} to {Identity}...", message, identityHex);
+            string reqId = !(message.Identity is null) && message.Identity.Length == 16 ?
+                new Guid(message.Identity).ToString() : "unknown";
+
+            _logger.Debug("Reply {Message} to {Identity}...", message, reqId);
 
             var ev = new AsyncManualResetEvent();
             _replyQueue.Enqueue(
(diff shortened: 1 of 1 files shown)


longfin:
- My intention was, to leverage this `GUID` for `DealerSocket`'s identifier. (it can be set through `dealerSocket.Options.Identity = req.Id.ToByteArray()`).
- In addition, specifying identifiers in a content-addressable manner, as in the current implementation, can be problematic from a strict point of view, because even if the message has exactly the same content, it is necessary to distinguish each message when it is sent twice or more.

longfin (dismissed):

longfin:
Well, maybe we need a changelog? (or need to skip explicitly via commit message).

colibrishin:
> Well, maybe we need a changelog? (or need to skip explicitly via commit message).

I think this is a minor change to be added. I'll skip the changelog.

longfin (approved):

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -600,11 +600,13 @@ private void ReceiveMessage(object sender, NetMQSocketEventArgs e)
                             try
                             {
                                 Message message = _messageCodec.Decode(copied, false);
+                                Guid reqId = new Guid(copied[0].ToByteArray());
```
longfin, line 603: Ah... we can't assume the first frame can be translated as GUID always, so need handling for that. 😢

longfin (changes requested):

Libplanet.Net/Transports/NetMQTransport.cs
colibrishin: Added a fallback exception If `Identity` cannot be parsed as GUID, create a new one.

Libplanet.Net/Transports/NetMQTransport.cs
```diff
@@ -599,14 +599,37 @@ private void ReceiveMessage(object sender, NetMQSocketEventArgs e)
                         {
                             try
                             {
+                                Guid reqId;
```
longfin, line 602: I think that `reqId` will be used only for logging purposes, so turning it to`string` and parse like as below would be enough.

```csharp
string requestId;
try
{
    requestId =  new Guid(copied[0].ToByteArray()).ToString();
}
catch (ArgumentException)
{
    requestId = "unknown";
}
```

longfin (approved):

riemannulus:
Could you rebase this PR, please?

colibrishin:
Rerouting branch to use next internal testnet.

greymistcube (approved):

PR #2743, opened 2023-01-20
Refactor/extract transport

greymistcube:
Several points:
- The ownership of `AppProtocolVersionOptions` shouldn't be `ITransport`.
- This further exposes the mangling of various options that should be for different layers. For instance, `SwarmOptions.MessageTimestampBuffer` should either be moved to a separate option explicitly for `ITransport`, or this should be moved together with `AppProtocolVersionOptions` to `Swarm<T>`. Also there are certain `TimeoutOptions` properties that should only be `ITransport`'s concern.
- At the moment, `Swarm<T>` should probably check `ITransport`'s `PublicKey` to make sure they both have the same `PrivateKey`. I'm not sure about whether they should actually be the same.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,13 @@ To be released.
  -  (Libplanet.Extensions.Cocona)  Dropped .NET Standard 2.0 and .NET Core 3.1
     target assemblies.  [[#2732]]
  -  (Libplanet.Extensions.Cocona)  Added .NET 6 target assembly.  [[#2732]]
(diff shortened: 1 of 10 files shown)


riemannulus (commented):
LGTM but `CHANGES.md` is deprecated. I'll fix that problem and approve and merge now. cc. @longfin

riemannulus (approved):

longfin (approved):

colibrishin (approved):

PR #2749, opened 2023-01-21
Introduce check-changelog action

greymistcube:


greymistcube:
Closing since the behavior is not the same as the currently used status check. Skipping changelog check is label based for this action rather than commit message based. I suppose we might first need a discussion whether label based would be better or not.

PR #2759, opened 2023-02-01
♻️ Removed nullability from `TrustedAppProtocolVersionSigners`

greymistcube:
Several points:
- I'm of the opinion that [`null`s are bad in general if they can be avoided](https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/).
- Currently, `null` option isn't really being used.
- Other than "it'd be nice to have trust any signer as an option", I still fail to see a concrete use case for this.
- Having both `null` and empty list as meaningful options are incompatible with current CLI.

If `null` is still that important, we should change the CLI to accept a formatted `string` to allow all possible cases. 🙄


Atralupus (approved):
agree

PR #2764, opened 2023-02-02
Catch exception of `AddressType`

boscohyun:
An exception occurred when the length of the `hex` less then 2.
So I catch that exception, wrap it inside of an `ArgumentException` and throw it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,9 @@ To be released.
 
 ### Behavioral changes
 
+ - If the length of the given hexadecimal string is less than 2,
+   `AddressType.ParseValue(object)` will throw `ArgumentException`. [[#2764]]
+
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


greymistcube:
> An exception occurred when the length of the hex less then 2.

I suppose the type of exception thrown would be different, but why is that a problem? 🤔

boscohyun:
> > An exception occurred when the length of the hex less then 2.
> 
> I suppose the type of exception thrown would be different, but why is that a problem? 🤔

I can expect the `AddressType.ParseValue(object)` throw an `ArgumentsException` when the argument(object) is empty string(""). But it isn't.🥲
So I changed it. And.. this change is not major stuff and I know this well.🙂
It just one of my tries to feel the warmth of the Libplanet repository.

greymistcube:
Hmm... In that case, I think it'd be better to update `Address` to give more helpful error message than try to parse it at the query level. 🙄

boscohyun:
Replaced with #2781

PR #2778, opened 2023-02-03
♻️ Introduce `IBencodable`

greymistcube:
Baby steps. 😶
- Allows to remove `[Serializable]` attribute on our own `class`es. `[Serializable]` was being used only for the sake of having objects to be serialized as a part of custom `Exception`s that we wanted to `throw`. This can reduce the exposure to the [warning] by Microsoft.
- `Address` now explicitly accepts `IValue` type not `Binary`. This is my own design decision to have overall consistency down the line.
  - Before this PR, `Address(Binary)` is probably not even necessary as `Binary` would be implicitly casted to `Address(byte[])`. The same goes with `Address(Text)` where the parameter is implicitly casted to `Address(string)`. This makes it more error prone by allowing non-standard usage of `Address` encoding.

[warning]: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/serialization/#binary-and-xml-serialization

### DIFF
--- a/Libplanet.Net.Tests/BoundPeerTest.cs
+++ b/Libplanet.Net.Tests/BoundPeerTest.cs
@@ -1,9 +1,7 @@
 #nullable disable
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Net;
-using System.Runtime.Serialization.Formatters.Binary;
 using Libplanet.Crypto;
 using Xunit;
 
@@ -29,22 +27,6 @@ namespace Libplanet.Net.Tests
                     new DnsEndPoint("0.0.0.0", 1234)),
             },
         };
-
-        [Theory]
-        [MemberData(nameof(GetBoundPeers))]
-        public void Serializable(BoundPeer peer)
-        {
-            var formatter = new BinaryFormatter();
-            using (var stream = new MemoryStream())
-            {
-                formatter.Serialize(stream, peer);
-                byte[] serialized = stream.ToArray();
-                stream.Seek(0, SeekOrigin.Begin);
-                BoundPeer deserialized = (BoundPeer)formatter.Deserialize(stream);
-                Assert.IsType(peer.GetType(), deserialized);
-                Assert.Equal(peer, deserialized);
-            }
-        }
 
         [Theory]
         [MemberData(nameof(GetBoundPeers))]
--- a/Libplanet.Net.Tests/BoundPeerTest.cs
+++ b/Libplanet.Net.Tests/BoundPeerTest.cs
@@ -48,12 +48,12 @@ namespace Libplanet.Net.Tests
 
         [Theory]
         [MemberData(nameof(GetBoundPeers))]
-        public void Serialize(BoundPeer peer)
+        public void Bencode(BoundPeer peer)
         {
-            Bencodex.Types.Dictionary serialized = peer.ToBencodex();
-            var deserialized = new BoundPeer(serialized);
+            Bencodex.Types.IValue bencoded = peer.Bencoded;
+            var decoded = new BoundPeer(bencoded);
 
-            Assert.Equal(peer, deserialized);
+            Assert.Equal(peer, decoded);
         }
 
         [Fact]
--- a/Libplanet.Net/BoundPeer.cs
+++ b/Libplanet.Net/BoundPeer.cs
@@ -31,7 +31,12 @@ namespace Libplanet.Net
         }
 
         public BoundPeer(Bencodex.Types.IValue bencoded)
-            : this((Bencodex.Types.Dictionary)bencoded)
+            : this(bencoded is Bencodex.Types.Dictionary dict
+                ? dict
+                : throw new ArgumentException(
+                    $"Given {nameof(bencoded)} must be of type " +
+                    $"{typeof(Bencodex.Types.Dictionary)}: {bencoded.GetType()}",
+                    nameof(bencoded)))
         {
         }
 
--- a/Libplanet.Net/BoundPeer.cs
+++ b/Libplanet.Net/BoundPeer.cs
@@ -2,17 +2,14 @@ using System;
 using System.Diagnostics.Contracts;
 using System.Globalization;
 using System.Net;
-using System.Runtime.Serialization;
 using Bencodex;
 using Bencodex.Types;
 using Destructurama.Attributed;
 using Libplanet.Crypto;
-using Libplanet.Serialization;
 
 namespace Libplanet.Net
 {
-    [Serializable]
-    public sealed class BoundPeer : ISerializable, IEquatable<BoundPeer>, IBencodable
+    public sealed class BoundPeer : IEquatable<BoundPeer>, IBencodable
     {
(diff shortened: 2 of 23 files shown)


Libplanet.Net/BoundPeer.cs
```diff
@@ -103,6 +90,17 @@ private BoundPeer(SerializationInfo info, StreamingContext context)
...
+        [Pure]
+        public Bencodex.Types.IValue Bencoded =>
+            Bencodex.Types.Dictionary.Empty
+                .Add(PublicKeyKey, PublicKey.Format(true))
+                .Add(EndPointHostKey, EndPoint.Host)
+                .Add(EndPointPortKey, EndPoint.Port)
+                .Add(
+                    PublicIpAddressKey,
+                    PublicIPAddress is IPAddress ip ? (IValue)(Text)ip.ToString() : Null.Value);
+
```
tkiapril, line 103: A question: we currently have objects that are encoded to bencodex values with interfaces like `IValue Serialize()` (`Dictionary ToBencodex()` too, but I see that this is being changed too.) Is it intended that objects that can be represented to bencodex values all use the `IBencodable` interface and have those different APIs changed into `Bencoded`?

Libplanet.Net/Transports/CommunicationFailException.cs
```diff
@@ -36,9 +40,7 @@ public class CommunicationFailException : Exception
         public CommunicationFailException(SerializationInfo info, StreamingContext context)
             : base(info, context)
         {
-            Peer = info.GetValue(nameof(Peer), typeof(BoundPeer)) is BoundPeer peer
-                ? peer
-                : throw new SerializationException($"{nameof(Peer)} is of an invalid type.");
+            Peer = new BoundPeer(_codec.Decode(info.GetValue<byte[]>(nameof(Peer))));
```
tkiapril, line 43: Would it be possible that `info.GetValue<byte[]>()` returns null or non-Dictionary bencodex value?

Libplanet.Net.Tests/BoundPeerTest.cs
```diff
@@ -32,28 +30,12 @@ public class BoundPeerTest
...
-        [MemberData(nameof(GetBoundPeers))]
-        public void Serialize(BoundPeer peer)
-        {
-            Bencodex.Types.Dictionary serialized = peer.ToBencodex();
-            var deserialized = new BoundPeer(serialized);
+            Bencodex.Types.IValue bencoded = peer.Bencoded;
+            var decoded = new BoundPeer(bencoded);
 
-            Assert.Equal(peer, deserialized);
+            Assert.Equal(peer, decoded);
```
tkiapril, line 38: Maybe it might be beneficial to have a `NotEqual` case.

Libplanet/Address.cs
```diff
@@ -135,24 +133,20 @@ public Address(string hex)
...
-        /// <param name="address">A Bencodex <see cref="Binary"/> of 20 <see cref="byte"/>s which
+        /// <param name="bencoded">A Bencodex <see cref="Binary"/> of 20 <see cref="byte"/>s which
         /// represents an <see cref="Address"/>.
         /// </param>
-        /// <exception cref="ArgumentException">Thrown when the given <paramref name="address"/>
-        /// did not lengthen 20 bytes.</exception>
-        public Address(Binary address)
-            : this(address.ByteArray)
+        /// <exception cref="ArgumentException">Thrown when given <paramref name="bencoded"/>
+        /// is not an encoding of a <see cref="byte"/> array of length 20.</exception>
```
tkiapril, line 142:
```suggestion
        /// is not an encoded Bencodex <see cref="Binary"/> value of length 20.</exception>
```

Might be more clear.

tkiapril (commented):
I don't see anything that seems off, but a few questions:

I was wondering why bencodex values in `SerializationInfo` were changed to contain encoded bencodex, instead of `IValue`. Is there a convention?

Also, I see that some constructors with IValue as the parameter uses casting to convert the bencodex value to the appropriate types to call the generic constructor, and that when an inappropriate type is given, they will throw an exception. I don't particularly oppose this design, as this has been used throughout libplanet in the past, but I can imagine a case when the exception may be cryptic when an undesired value is provided. I think it might be beneficial to throw an explicit exception about receiving an undesired value for debugging in the future.

greymistcube:
> I was wondering why bencodex values in SerializationInfo were changed to contain encoded bencodex, instead of IValue. Is there a convention?

This is because `IValue` does not support serialization directly. If we want to support this functionality, we'd have to update `Bencodex.IValue`. I'm not sure this is a good idea tho.

> Also, I see that some constructors with IValue as the parameter uses casting to convert the bencodex value to the appropriate types to call the generic constructor, and that when an inappropriate type is given, they will throw an exception.

I guess something like this?

```csharp
        public BoundPeer(Bencodex.Types.IValue bencoded)
            : this(bencoded is Bencodex.Types.Dictionary dict
                ? dict
                : throw new ArgumentException(
                    $"Given {nameof(bencoded)} must be of type " +
                    $"{typeof(Bencodex.Types.Dictionary)}: {bencoded.GetType()}",
                    nameof(bencoded)))
        {
        }
```

Libplanet.Net/BoundPeer.cs
greymistcube: Yup, trying to have some sort of standard for housekeeping. As you've mentioned, there are two issues:

- Method names for "serializing" to bencodex are not standardized.
- Return types are not standardized.

Perhaps the decision to not return a concrete type might cause some issues down the line, but we'll deal with it later. 😅

Libplanet.Net/Transports/CommunicationFailException.cs
greymistcube: It is. I guess its pretty sloppy. Then again, it could be argued whether `SerializationException` is all that necessary in this context (seems like a defensive practice for the sake of the practice itself). 🙄

I'll take another look.

Libplanet.Net.Tests/BoundPeerTest.cs
greymistcube: I'm not quite getting it, what would be a `NotEqual()` case?

tkiapril: I was thinking of something like creating a `BoundPeer` that has roughly the default value, and doing `Assert.NotEqual()` with it, but now that I think of it the idea of a default `PublicKey` might be not well-defined. 🤔 I take it back.

tkiapril:
Aha, I see the reason with `SerializationInfo` having `byte[]`.

The constructor exception looks good. I think it might be good if the ctor can throw when the expected value is not in the `Dictionary` as well, but again it might be an overcautious feat.

Libplanet.Net/Transports/CommunicationFailException.cs
greymistcube: Seems like this problem runs a bit deeper than it looks. I think for proper `null` handling, changes to `SerializationInfoExtensions` might be the way to go (although, I'm more of the opinion we need to get rid of `SerializationInfoExtensions`).

tkiapril (approved):

PR #2782, opened 2023-02-06
Use voting power instead of vote count when validating blockcommit

limebell:
TODO: Should add test case for newly added logic

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -5,6 +5,7 @@ using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Linq;
+using System.Numerics;
 using System.Security.Cryptography;
 using System.Threading;
 using Bencodex.Types;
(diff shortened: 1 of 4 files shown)


OnedgeLee (approved):

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1662,6 +1663,20 @@ internal void CleanupBlockCommitStore(long limit)
                         string.Empty, (s, key) => s + key.ValidatorPublicKey + ", \n"));
             }
 
+            // TODO: Add an unit case for this method.
+            BigInteger commitPower = blockCommit.Votes.Aggregate(
+                BigInteger.Zero,
+                (power, vote) => power + (vote.Flag == VoteFlag.PreCommit
+                    ? validators.GetValidator(vote.ValidatorPublicKey).Power
+                    : BigInteger.Zero));
+            if (validators.TotalPower * 2 / 3 > commitPower)
```
greymistcube, line 1672: Doesn't `ValidatorSet` have `TwoThirdsPower` property?

greymistcube: Also, I think it should be `>=`.

limebell: Amended!

OnedgeLee (approved):

greymistcube (approved):

PR #2794, opened 2023-02-09
♻️ Refactor `BlockCandidateTable`

greymistcube:
Preliminary refactoring to further specify the context in which a set of `Block<T>`s gets thrown around.
I'll probably be implementing a `Branch<T>` class shortly as the contractual guarantee from `GetCurrentRoundCandidate()` is pretty weak for the returned `List<T>` of `Block<T>`s when passing it around. For that reason, the usage of `SortedList<long, Block<T>>` has been removed as it does not sufficiently guarantee the desired properties. 😶

This adds two additional constraints to the previous implementation:
- `Block<T>`s have to have consecutive indices.
- `Block<T>`s should form a linked list in the sense that a "next" `Block<T>`'s `PreviousHash` should match that of the "previous" `Block<T>`s `Hash`.

As a side note, I think it'd be also a good idea to only allow `Block<T>`s that are at least of size 2. This would require the table to always include the branchpoint and some logic change up the stack. 🙄

### DIFF
--- a/Libplanet.Net/BlockCandidateTable.cs
+++ b/Libplanet.Net/BlockCandidateTable.cs
@@ -40,17 +40,19 @@ namespace Libplanet.Net
         /// Adds given <paramref name="blocks"/> to the table.
         /// </para>
         /// <para>
-        /// The internal table is only updated if a given pair satisfy both of the following:
+        /// The internal table is only updated if a given pair satisfy all of the following:
         /// <list type="bullet">
+        ///     <item><description>
+        ///         The internal dictionary does not already contain <paramref name="blockHeader"/>
+        ///         as its key.
+        ///     </description></item>
         ///     <item><description>
         ///         Given <paramref name="blocks"/> is non-empty.
         ///     </description></item>
         ///     <item><description>
-        ///         Given <paramref name="blocks"/> have unique indices.
-        ///     </description></item>
-        ///     <item><description>
-        ///         The internal dictionary does not already contain <paramref name="blockHeader"/>
-        ///         as its key.
+        ///         Given <paramref name="blocks"/> are consecutive in the sense that indices
+        ///         are unique consecutive and every <see cref="Block{T}.PreviousHash"/> match
+        ///         the <see cref="Block{T}.Hash"/> of the previous <see cref="Block{T}"/>.
         ///     </description></item>
         /// </list>
         /// </para>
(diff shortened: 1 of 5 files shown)


Libplanet.Net/Swarm.BlockCandidate.cs
```diff
@@ -17,20 +17,20 @@ public partial class Swarm<T>
...
             while (!cancellationToken.IsCancellationRequested)
             {
-                if (BlockCandidateTable.Any())
+                if (BlockCandidateTable.Count > 0)
                 {
                     BlockHeader tipHeader = BlockChain.Tip.Header;
-                    SortedList<long, Block<T>> blocks =
+                    List<Block<T>> blocks =
                         BlockCandidateTable.GetCurrentRoundCandidate(tipHeader);
-                    if (!(blocks is null) && blocks.Count > 0)
```
riemannulus, line 25: `object is { }` is compact version of `!(object is null)`. https://devblogs.microsoft.com/dotnet/do-more-with-patterns-in-c-8-0/

Libplanet.Net/Swarm.BlockCandidate.cs
```diff
@@ -17,20 +17,20 @@ public partial class Swarm<T>
             var checkInterval = TimeSpan.FromMilliseconds(10);
             while (!cancellationToken.IsCancellationRequested)
             {
-                if (BlockCandidateTable.Any())
+                if (BlockCandidateTable.Count > 0)
                 {
                     BlockHeader tipHeader = BlockChain.Tip.Header;
-                    SortedList<long, Block<T>> blocks =
```
riemannulus, line 23: I want to make sure this list is indeed sorted. Have we a way to prohibit adding a non-sorted list in this code? Without `SortedList<K, V>`?

Libplanet.Net/Swarm.BlockCandidate.cs
riemannulus: Aha, I misread. You changed correctly. 😓 Ignore this comment, please.

Libplanet.Net/Swarm.BlockCandidate.cs
greymistcube: Not really. However, a list being sorted is not a sufficient condition (for consumption), nor is it reflective of the actual constraints (for `BlockCandidateTable`), as I have already mentioned in the PR message above. 😶

Libplanet.Net/BlockCandidateTable.cs
```diff
@@ -10,88 +10,121 @@
...
-                }
+                Log.Debug(
+                    "Given blocks will not be added to the table as an empty set of blocks " +
+                    "is not allowed",
+                    sorted.Count);
+                return;
             }
-            catch (ArgumentException e)
+            else if (!sorted
+                .Zip(sorted.Skip(1), (prev, next) =>
```
riemannulus, line 85: Have you any reason for skipping the first item?

Libplanet.Net/Swarm.BlockCandidate.cs
riemannulus: At first, I'm not sure about just guarding in `BlockCandidateTable.Add()` and leaving the inner data structure to `List<T>`. Because if left to `List<T>`, other contributors can mistake. (e.g., Make new non-sorted append I guess.)
I'm curious about your opinion about this.

greymistcube: I agree, hence the need for a new `class`. If there isn't any logic issues, once this gets merged I'll work on it ASAP. 😗

Libplanet.Net/BlockCandidateTable.cs
greymistcube: Otherwise it would map `{ 1, 2, 3, 4 }` to `{ (1, 1), (2, 2), (3, 3), (4, 4) }`? 🙃

riemannulus (approved):
Great work.

PR #2795, opened 2023-02-09
♻️ `IBencodable` `TxId`

greymistcube:
♻️

### DIFF
--- a/Libplanet.Tests/Tx/TxIdTest.cs
+++ b/Libplanet.Tests/Tx/TxIdTest.cs
@@ -12,12 +12,6 @@ namespace Libplanet.Tests.Tx
     public class TxIdTest
     {
         [Fact]
-        public void ConstructorDoesNotTakeNullValue()
-        {
-            Assert.Throws<ArgumentNullException>(() => new TxId(null));
-        }
-
-        [Fact]
         public void TxIdMustBe32Bytes()
         {
             for (int size = 0; size < 36; size++)
@@ -31,12 +25,10 @@ namespace Libplanet.Tests.Tx
                 ImmutableArray<byte> immutableBytes = bytes.ToImmutableArray();
                 Assert.Throws<ArgumentOutOfRangeException>(
                     "txid",
-                    () => new TxId(immutableBytes)
-                );
+                    () => new TxId(immutableBytes));
                 Assert.Throws<ArgumentOutOfRangeException>(
                     "txid",
-                    () => new TxId(bytes)
-                );
+                    () => new TxId(bytes));
             }
         }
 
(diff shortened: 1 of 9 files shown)


Libplanet.Tests/Tx/TxIdTest.cs
```diff
@@ -231,6 +193,28 @@ public void Compare()
...
 
+        [Fact]
+        public void Bencoded()
+        {
+            var expected = new TxId(TestUtils.GetRandomBytes(TxId.Size));
+            var deserialized = new TxId(expected.Bencoded);
+            Assert.Equal(expected, deserialized);
+            expected = default(TxId);
+            deserialized = new TxId(expected.Bencoded);
+            Assert.Equal(expected, deserialized);
```
tkiapril, line 204:
```suggestion
            var expected2 = default(TxId);
            var deserialized2 = new TxId(expected.Bencoded);
            Assert.Equal(expected2, deserialized2);
            Assert.NotEqual(expected, deserialized2);
```

This might help if Bencoded implementation is unintentionally changed to return something constant.

Libplanet.Tests/Tx/TxIdTest.cs
```diff
@@ -231,6 +193,28 @@ public void Compare()
...
+
+        [Fact]
+        public void Serializable()
+        {
+            var expected = new TxId(TestUtils.GetRandomBytes(TxId.Size));
+            var deserialized = TestUtils.BinarySerializeDeserialize<TxId>(expected);
+            Assert.Equal(expected, deserialized);
+            expected = default(TxId);
+            deserialized = TestUtils.BinarySerializeDeserialize<TxId>(expected);
+            Assert.Equal(expected, deserialized);
```
tkiapril, line 215: Here too:
```suggestion
            var expected2 = default(TxId);
            var deserialized2 = TestUtils.BinarySerializeDeserialize<TxId>(expected);
            Assert.Equal(expected2, deserialized2);
            Assert.Equal(expected, deserialized2);
```

Libplanet/Address.cs
```diff
@@ -81,18 +83,15 @@ public Address(in ImmutableArray<byte> address)
...
+        /// represents an <see cref="Address"/>.</param>
         /// <exception cref="ArgumentException">Thrown when the given <paramref
         /// name="address"/> array did not lengthen 20 bytes.</exception>
         /// <remarks>A valid <see cref="byte"/> array which represents an
         /// <see cref="Address"/> can be gotten using <see cref="ToByteArray()"
         /// /> method.</remarks>
         /// <seealso cref="ToByteArray()"/>
         public Address(byte[] address)
-            : this(address?.ToImmutableArray() ?? throw new ArgumentNullException(nameof(address)))
+            : this(address.ToImmutableArray())
```
tkiapril, line 94: I might be wrong, but couldn't arrays be implicitly null?

Libplanet/Tx/TxId.cs
```diff
@@ -20,9 +20,10 @@ namespace Libplanet.Tx
     /// (See also <see cref="Size"/> constant.)</para>
     /// </summary>
     /// <seealso cref="Transaction{T}.Id"/>
-    [JsonConverter(typeof(TxIdJsonConverter))]
```
tkiapril, line 23: Is this removed because Bencodex can handle Json marshalling?

tkiapril (commented):
Again, nothing particularly off, but some suggestions and questions. 👍

Libplanet/Address.cs
greymistcube: `#nullable enabled` has been applied in #2778. Although this doesn't syntactically prevent `null` from being passed, if we are to perform null check every time, I don't see the point of having `null`able context. 🙄

Libplanet/Tx/TxId.cs
greymistcube: This isn't removed. It's just `git diff` acting up. 🙃

Libplanet.Tests/Tx/TxIdTest.cs
greymistcube: I understand the precaution, but what does this achieve exactly?
If `Bencoded` implementation gets broken to return something constant, the test for encoding and decoding the first random `TxId` would fail anyways. 😶

tkiapril: You're right, in the hypothetical situation it should fail on the first `Assert.Equal()`. I thought the precautionary test might flag on something, but I can't really come up with something that makes sense. 😞

tkiapril (approved):

PR #2827, opened 2023-02-17
update some log level from Debug to Information

sky1045:
SSIA

### DIFF
--- a/Libplanet/Blockchain/Renderers/LoggedActionRenderer.cs
+++ b/Libplanet/Blockchain/Renderers/LoggedActionRenderer.cs
@@ -44,7 +44,7 @@ namespace Libplanet.Blockchain.Renderers
         public LoggedActionRenderer(
             IActionRenderer<T> renderer,
             ILogger logger,
-            LogEventLevel level = LogEventLevel.Debug
+            LogEventLevel level = LogEventLevel.Information
         )
             : base(renderer, logger, level)
         {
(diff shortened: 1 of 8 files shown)


Libplanet/Blockchain/Renderers/LoggedActionRenderer.cs
```diff
@@ -44,7 +44,7 @@ public class LoggedActionRenderer<T> : LoggedRenderer<T>, IActionRenderer<T>
         public LoggedActionRenderer(
             IActionRenderer<T> renderer,
             ILogger logger,
-            LogEventLevel level = LogEventLevel.Debug
+            LogEventLevel level = LogEventLevel.Information
```
area363, line 47: Wouldn't this prevent debug messages from being logged as stdout? I might have misunderstood but I thought we were logging all the debug messages on the stdout but filter info logs to CW.

greymistcube: I suspect this breaks tests. 🤔

area363 (approved):

greymistcube:
Lemme push some additional changes before merging. 🙃

sky1045:
by the way, i'm not sure why some tests fail even if the change i made is only for log levels 🤷‍♂️

greymistcube:
I've tried to be a bit more selective on which logs get elevated to `Information` level.

greymistcube (approved):

PR #2828, opened 2023-02-17
🧹 Changed method name logging for more consistency

greymistcube:
Although a case could be made whether method names should be handled as properties, I've decided to include method names as logged property. The reason is twofold:

- Cleaner codebase
- Easier parsing through property

Also, added `()` for missing cases. These are mainly for comprehension.

In the long run, a better solution would be to automatically generate the method name and line number as context at some point. 🙄


riemannulus (approved):
LGTM

PR #2832, opened 2023-02-17
🧹 Sanitized logging

greymistcube:
Just bringing up to date from a working branch.

### DIFF
--- a/Libplanet.Net.Tests/Protocols/TestTransport.cs
+++ b/Libplanet.Net.Tests/Protocols/TestTransport.cs
@@ -132,7 +132,7 @@ namespace Libplanet.Net.Tests.Protocols
                 throw new ObjectDisposedException(nameof(TestTransport));
             }
 
-            _logger.Debug("Starting transport of {Peer}.", AsPeer);
+            _logger.Debug("Starting transport of {Peer}", AsPeer);
             _swarmCancellationTokenSource = new CancellationTokenSource();
             CancellationToken token = cancellationToken.Equals(CancellationToken.None)
                 ? _swarmCancellationTokenSource.Token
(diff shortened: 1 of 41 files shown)


dahlia (approved):

Libplanet.Net.Tests/Protocols/TestTransport.cs
```diff
@@ -244,30 +244,30 @@ async Task DoAddPeersAsync()
...
-                    _logger.Verbose("Trying to ping all {PeersNumber} peers.", tasks.Count);
+                    _logger.Verbose("Trying to ping all {PeersNumber} peers", tasks.Count);
                     await Task.WhenAll(tasks);
-                    _logger.Verbose("Update complete.");
+                    _logger.Verbose("Update complete");
                 }
                 catch (DifferentAppProtocolVersionException)
                 {
-                    _logger.Debug("Different version encountered during AddPeersAsync().");
+                    _logger.Debug("Different version encountered during AddPeersAsync()");
```
riemannulus, line 253: Do we not replace `nameof()` instead of a plain string this time?

Libplanet.Net.Tests/Protocols/TestTransport.cs
```diff
@@ -244,30 +244,30 @@ async Task DoAddPeersAsync()
...
                 }
                 catch (PingTimeoutException)
                 {
                     var msg =
-                        $"Timeout occurred during {nameof(AddPeersAsync)}() after {timeout}.";
+                        $"Timeout occurred during {nameof(AddPeersAsync)}() after {timeout}";
```
riemannulus, line 258: Do we not replace a template string instead of a plain string this time?

greymistcube: Eh, it's because this is used both for logging and an exception. Also this being part of the test code, I didn't think too much about it. 🙄

Libplanet.Net.Tests/Protocols/TestTransport.cs
greymistcube: I guess I missed it since it wasn't caught by regex. I'll fix it on the other branch to be included on a follow-up PR.

PR #2837, opened 2023-02-20
test/pbft: double vote byzantine test

colibrishin:
## Context
This is a part of the byzantine tests which is a node commits a double vote.

## Double vote
If a node commits a double vote, by the `MessageLog` specification, other validator nodes will accept one of the votes and discard one. Additionally, When the evidence is added this test should check whether a node has caught the evidence.

## Double proposal
There is another byzantine case, a double proposal, solely `Context<T>` does not guarantee the recovery but it is covered by `Swarm<T>` block synchronization. 
- Even if the 2 nodes couldn't catch up with the consensus, (byzantine and the minority proposal received node), another majority received 2 nodes will able to commit the block, so the down node will recover while the block synchronization and `TipChange`. (I will try to cover this case in follow-up PR.)
- The failing `Context<T>` test is not included for the ease of managing test code. 
- Note that the current implementation would work in a double proposal scenario, however, any situation that uses only  `Context<T>` cannot recover from the double proposal byzantine case.

```csharp
[Fact(
    Timeout = Timeout,
    Skip = "If one of validators sees the +2/3 votes, new context will be created by " +
            "block sync. If the context need to be more stable, this should be fixed.")]
public async Task DoubleProposal()
{
    const int validatorCount = 4;
    Gossip[]? gossips = null;

    // Due to giving a BroadcastMessage delegation, here Context<T> is created in
    // first and defined null. The initialization will be done in later.
    var peerList = Enumerable.Range(0, 4)
        .Select(x =>
            new BoundPeer(
                TestUtils.PrivateKeys[x].PublicKey,
                new DnsEndPoint("127.0.0.1", 6000 + x)))
        .ToArray();

    (BlockChain<DumbAction> BlockChain, Context<DumbAction> Context)[]?
        validators = null;
    var newHeightStarted = new AsyncAutoResetEvent[validatorCount];
    for (int i = 0; i < validatorCount; ++i)
    {
        newHeightStarted[i] = new AsyncAutoResetEvent();
    }

    gossips = Enumerable.Range(0, 4)
        .Select(x => CreateGossip(
            message =>
            {
                if (message is ConsensusMsg consensusMsg)
                {
                    // ReSharper disable once AccessToModifiedClosure
                    validators![x].Context.ProduceMessage(consensusMsg);
                }
            },
            TestUtils.PrivateKeys[x],
            6000 + x,
            peerList))
        .ToArray();

    validators = Enumerable.Range(0, validatorCount)
        .Select(x =>
            TestUtils.CreateDummyContext(
                1,
                privateKey: TestUtils.PrivateKeys[x],
                broadcastMessage: gossips[x].AddMessage,
                contextTimeoutOptions: new ContextTimeoutOption(
                    preVoteSecondBase: 1,
                    preCommitSecondBase: 1)))
        .ToArray();

    try
    {
        // If a proposer sends two split proposals in a round, one of the node cannot see
        // the +2/3 votes from other nodes due to locking into "canonical" proposal, which
        // is the first seen proposal. A node might need to track all proposals and known
        // votes of blocks in a round for switching to the proposal with +2/3 votes to
        // keep itself into consensus.
        // Note that this is separate case with the validity of the proposed value.

        // Setup next height listener.
        for (int i = 0; i < validatorCount; i++)
        {
            var nodeNo = i;
            validators[i].Context.StateChanged += (sender, mrs) =>
            {
                if (mrs.Step == Step.EndCommit)
                {
                    newHeightStarted[nodeNo].Set();
                }
            };
        }

        // Byzantine will propose a multiple block in a round, and split out to other nodes.
        var blockB = validators[1].BlockChain.ProposeBlock(TestUtils.PrivateKeys[1]);

        // A proposal for that actually will get the +2/3 and canonical proposal.
        var proposalB = TestUtils.CreateConsensusPropose(blockB, TestUtils.PrivateKeys[1]);

        var proposalBPreVote = TestUtils.CreateVote(
            TestUtils.PrivateKeys[1],
            1,
            0,
            blockB.Hash,
            VoteFlag.PreVote);

        var proposalBPreCommit = TestUtils.CreateVote(
            TestUtils.PrivateKeys[1],
            1,
            0,
            blockB.Hash,
            VoteFlag.PreCommit);

        // And other validator 2, 3 will see the proposalB as canonical proposal.
        // The total votes for proposalB is +2/3 (by proposer, 2, and 3), so the value will
        // be committed in the perspective of validator 2, 3.
        validators[2].Context.ProduceMessage(proposalB);
        validators[3].Context.ProduceMessage(proposalB);

        // Proposer will send the proposal, preVote, and preCommit for all split blocks.
        // This is the setup before sending the proposal.
        validators[2].Context.ProduceMessage(new ConsensusPreVoteMsg(proposalBPreVote));
        validators[2].Context.ProduceMessage(new ConsensusPreCommitMsg(proposalBPreCommit));
        validators[3].Context.ProduceMessage(new ConsensusPreVoteMsg(proposalBPreVote));
        validators[3].Context.ProduceMessage(new ConsensusPreCommitMsg(proposalBPreCommit));

        foreach (var gossip in gossips)
        {
            _ = gossip.StartAsync(default);
            await gossip.WaitForRunningAsync();
        }

        AsyncAutoResetEvent proposalACreated = new AsyncAutoResetEvent();
        validators[1].Context.MessageBroadcasted += (sender, msg) =>
        {
            if (msg is ConsensusProposalMsg)
            {
                proposalACreated.Set();
            }
        };

        // Messages will be delivered in this order:
        // Validator 2, 3 : proposalB, preVoteB, preCommitB, proposalA ...
        // Validator 0 : proposalA, proposalB, preVoteA ...
        validators[1].Context.Start();
        await proposalACreated.WaitAsync();
        validators[0].Context.Start();
        validators[2].Context.Start();
        validators[3].Context.Start();

        // Validator 2, 3 reaches the PreCommit and add the block to the chain. However, the
        // validator 0 will not commit the proposalB, because it has ignored the proposalB
        // so the validator 0 will move to the next round. With that so, The next height
        // consensus can't be started.

        // From our current implementation of Swarm, even if the validator 0 and proposer
        // failed, the consensus resumed by synchronization. Context<T> cannot recover
        // solely.
        await Task.WhenAll(newHeightStarted.Select(x => x.WaitAsync()));
        foreach (var validator in validators)
        {
            Assert.Equal(1, validator.Context.Height);
        }
    }
    finally
    {
        foreach (var gossip in gossips)
        {
            await gossip.StopAsync(TimeSpan.FromMilliseconds(100), default);
            gossip.Dispose();
        }

        foreach (var validator in validators)
        {
            validator.Context.Dispose();
        }
    }
}
```

### DIFF
--- a/Libplanet.Net.Tests/Consensus/ByzantineTest.cs
+++ b/Libplanet.Net.Tests/Consensus/ByzantineTest.cs
@@ -42,57 +42,63 @@ namespace Libplanet.Net.Tests.Consensus
             // Due to giving a BroadcastMessage delegation, here Context<T> is created in
             // first and defined null. The initialization will be done in later.
             const int validatorCount = 4;
-            Gossip[]? gossips = null;
+            var validatorKeys = TestUtils.PrivateKeys.Take(validatorCount).ToArray();
 
-            var peerList = Enumerable.Range(0, 4)
-                .Select(x =>
+            var peerList = validatorKeys
+                .Select((key, index) =>
                     new BoundPeer(
-                        TestUtils.PrivateKeys[x].PublicKey,
-                        new DnsEndPoint("127.0.0.1", 6000 + x)))
+                        key.PublicKey,
+                        new DnsEndPoint("localhost", 6000 + index)))
                 .ToArray();
 
-            (BlockChain<DumbAction> BlockChain, Context<DumbAction> Context)[]?
-                validators = null;
-            var newHeightStarted = new AsyncAutoResetEvent[validatorCount];
-            for (int i = 0; i < validatorCount; ++i)
-            {
-                newHeightStarted[i] = new AsyncAutoResetEvent();
-            }
-
-            gossips = Enumerable.Range(0, 4)
-                .Select(x => TestUtils.CreateGossip(
+            (BlockChain<DumbAction> BlockChain, Context<DumbAction> Context)[]? validators = null;
+            var gossips = validatorKeys
+                .Select((key, index) => TestUtils.CreateGossip(
                     content =>
                     {
                         if (content is ConsensusMsg consensusMsg)
                         {
                             // ReSharper disable once AccessToModifiedClosure
-                            validators![x].Context.ProduceMessage(consensusMsg);
+                            validators![index].Context.ProduceMessage(consensusMsg);
                         }
                     },
-                    TestUtils.PrivateKeys[x],
-                    6000 + x,
-                    peerList))
+                    privateKey: key,
+                    port: 6000 + index,
+                    peers: validatorKeys.Select((k, i) => new BoundPeer(
+                        k.PublicKey, new DnsEndPoint("localhost", 6000 + i))).ToArray()))
                 .ToArray();
 
-            validators = Enumerable.Range(0, validatorCount)
-                .Select(x =>
+            validators = validatorKeys
+                .Select((key, index) =>
                     TestUtils.CreateDummyContext(
                         1,
-                        privateKey: TestUtils.PrivateKeys[x],
-                        broadcastMessage: gossips[x].AddMessage))
+                        privateKey: key,
+                        broadcastMessage: gossips[index].AddMessage,
+                        newHeightDelay: TimeSpan.Zero))
                 .ToArray();
 
             try
             {
                 // Setup next height listener.
-                for (int i = 0; i < validatorCount; i++)
+                var newHeightStarted = new AsyncAutoResetEvent[validatorCount];
+                var messageExceptionOccured = new AsyncAutoResetEvent[validatorCount];
+                foreach (var item in validators.Select((validator, index) => (validator, index)))
                 {
-                    var nodeNo = i;
-                    validators[i].Context.StateChanged += (sender, mrs) =>
+                    newHeightStarted[item.index] = new AsyncAutoResetEvent();
+                    messageExceptionOccured[item.index] = new AsyncAutoResetEvent();
+                    item.validator.Context.StateChanged += (sender, mrs) =>
                     {
                         if (mrs.Step == Step.EndCommit)
                         {
-                            newHeightStarted[nodeNo].Set();
+                            newHeightStarted[item.index].Set();
+                        }
+                    };
+
+                    item.validator.Context.ExceptionOccurred += (sender, e) =>
+                    {
+                        if (e is InvalidConsensusMessageException)
+                        {
+                            messageExceptionOccured[item.index].Set();
                         }
                     };
                 }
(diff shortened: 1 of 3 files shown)


OnedgeLee:
Rebased onto main, and reset target branch to main.

Libplanet.Net.Tests/Consensus/ByzantineTest.cs
```diff
@@ -0,0 +1,160 @@
...
+                    peerList))
+                .ToArray();
+
+            validators = Enumerable.Range(0, validatorCount)
+                .Select(x =>
+                    TestUtils.CreateDummyContext(
+                        1,
+                        privateKey: TestUtils.PrivateKeys[x],
+                        broadcastMessage: gossips[x].AddMessage))
+                .ToArray();
```
OnedgeLee: Does `gossips` needed to be created in advance to `validators`' creation?

Libplanet.Net.Tests/Consensus/ByzantineTest.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+
+            gossips = Enumerable.Range(0, 4)
+                .Select(x => TestUtils.CreateGossip(
+                    message =>
+                    {
+                        if (message is ConsensusMsg consensusMsg)
+                        {
+                            // ReSharper disable once AccessToModifiedClosure
+                            validators![x].Context.ProduceMessage(consensusMsg);
```
OnedgeLee, line 69: What if count of `validators` is different from number of peers?

OnedgeLee: Do we need to separate number of peers and number of validators?

Libplanet.Net.Tests/Consensus/ByzantineTest.cs
```diff
@@ -0,0 +1,160 @@
...
+                validators[3].Context.ProduceMessage(new ConsensusPreVoteMsg(proposalAPreVoteNil));
+
+                validators[0].Context.Start();
+                validators[2].Context.Start();
+                validators[3].Context.Start();
+
+                await Task.WhenAll(newHeightStarted.Select(x => x.WaitAsync()));
+                foreach (var validator in validators)
+                {
+                    Assert.Equal(1, validator.Context.Height);
```
OnedgeLee, line 140: I'm not sure about purpose of this assertion.

Libplanet.Net.Tests/Consensus/ByzantineTest.cs
```diff
@@ -0,0 +1,160 @@
...
+                // Setup next height listener.
+                for (int i = 0; i < validatorCount; i++)
+                {
+                    var nodeNo = i;
+                    validators[i].Context.StateChanged += (sender, mrs) =>
+                    {
+                        if (mrs.Step == Step.EndCommit)
+                        {
+                            newHeightStarted[nodeNo].Set();
+                        }
```
OnedgeLee, line 96: State change to `EndCommit` doesn't assure `newHeight` has been started, due to the existence of `newHeightDelay`.

OnedgeLee:
I think this can be covered with tests https://github.com/planetarium/libplanet/blob/51a95183937a76829f5166d4169c5b9fd20f6fc2/Libplanet.Net.Tests/Consensus/Context/MessageLogTest.cs#L82-L142.

Tests on evidence will be covered after implementation.

So, discarding this PR for now.

Issue #2730, opened 2023-01-19
Fix `Libplanet.Net.Tests.TestUtils.CreateDummyContext()` to return proper `BlockChain<T>`


OnedgeLee:
https://github.com/planetarium/libplanet/blob/ea4e74e996df15e6b81c81e2c3cb8225b96643d1/Libplanet.Net.Tests/TestUtils.cs#L266-L304

`CreateDummyContext()` returns both of `BlockChain<T>` and `Context<T>`, on the returned `BlockChain<T>`, only genesis block exists. So, the height of the `Context<T>` is different from `BlockChain<T>`'s height.

This makes hard to omit `validatorSet` parameter from this method. After dynamic validator has been merged, injecting `ValidatorSet` can cause some unexpected mistakes, and it would be better to get `ValidatorSet` from the state always.

To do that, `BlockChain<T>` with full blocks is needed.

PR #2839, opened 2023-02-21
Dummy context generation to fill blockchain on test

OnedgeLee:
### Changes
- Make `CreateDummyConsensusContext` to fill returned `BlockChain`.
- Also, `CreateDummyContext` will fill returned `BlockChain`.
- Removed `validatorSet` parameter from `CreateDummyContext`
- Removed test : Remove `CannotProposeWithoutLastCommitWhenRequired`, since block proposal without required lastcommit will be prohibited by block appending, before proposal.

### Related issues
- #2730

### DIFF
--- a/Libplanet.Net.Tests/TestUtils.cs
+++ b/Libplanet.Net.Tests/TestUtils.cs
@@ -227,6 +227,7 @@ namespace Libplanet.Net.Tests
             ConsensusContext<DumbAction> ConsensusContext)
             CreateDummyConsensusContext(
                 TimeSpan newHeightDelay,
+                long height = 1L,
                 IBlockPolicy<DumbAction>? policy = null,
                 PrivateKey? privateKey = null,
                 ConsensusContext<DumbAction>.DelegateBroadcastMessage? broadcastMessage = null,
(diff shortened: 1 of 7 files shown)


Libplanet.Net.Tests/TestUtils.cs
```diff
@@ -236,6 +237,14 @@ in PrivateKeys.Zip(Peers, (first, second) => (first, second)))
...
             var fx = new MemoryStoreFixture(policy.BlockAction);
             var blockChain = CreateDummyBlockChain(fx);
+            for (int i = 1; i < height; i++)
+            {
+                var block = blockChain.ProposeBlock(
+                    Libplanet.Tests.TestUtils.GenesisProposer,
+                    lastCommit: CreateBlockCommit(blockChain[i - 1]));
+                blockChain.Append(block, CreateBlockCommit(block));
+            }
+
```
colibrishin, line 247: Could you change the `PrivateKey` field so that the blocks can have the proposer in a round-robin order, for now? 
Or, We could fix it later because we will not have a reward until the PoS, so instead flag it here as TODO.

OnedgeLee: What I've thought on that time was, unit test doesn't need to be work as real work, but now my mind has been changed.
May be this kind of mock up can be harmful especially after future sortition logic implementation.
I think it would be better to make some fixture like `ValidatorSet` that have private keys also, for tests.
I'll add some.

OnedgeLee: Instead of create new fixture, added dictionary from `publicKey` to `privateKey` on `TestUtils`. And, proposer is now selected with `ValidatorSet.GetProposer()`.

colibrishin (approved):

PR #2857, opened 2023-02-24
♻️ Refactor validate nonces

greymistcube:
This is a follow-up to #2845.

PR size is unnecessarily large due to indentation fix in `Swarm.BlockCandidate.cs`. 🙄

- Tx nonce validation has been extracted out as a separate method.
  - There could be a slight performance degradation as we are iterating over `Block<T>.Transactions` twice.
  - This is to mainly separate policy validation and structural validation more easily.
  - Future plan is to re-use this method during genesis `Block<T>` validation.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1137,7 +1137,6 @@ namespace Libplanet.Blockchain
                 sampleAfter: threshold);
         }
 
-#pragma warning disable MEN003
         internal void Append(
             Block<T> block,
             bool evaluateActions,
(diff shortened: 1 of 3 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1173,39 +1172,25 @@ public BlockLocator GetBlockLocator(int threshold = 10)
...
                     throw ibe;
                 }
 
-                var nonceDeltas = new Dictionary<Address, long>();
+                var nonceDeltas = ValidateNonces(
+                    block.Transactions
+                        .Select(tx => tx.Signer)
+                        .Distinct()
+                        .ToDictionary(signer => signer, signer => Store.GetTxNonce(Id, signer)),
+                    block);
```
riemannulus, line 1183: 👍

riemannulus (approved):

PR #2874, opened 2023-03-02
🧹 Logging sanitization

greymistcube:
🧹

### DIFF
--- a/Libplanet/Action/ActionEvaluator.cs
+++ b/Libplanet/Action/ActionEvaluator.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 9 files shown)


longfin (dismissed):

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -576,7 +576,7 @@ internal static IImmutableSet<Address> GetUpdatedAddresses(Transaction<T> tx)
                     .ForContext("Subtag", "TxEvaluationDuration");
                 logger.Information(
                     "{ActionCount} actions {ActionTypes} in transaction {TxId} " +
-                    "by {Signer} with timestamp {TxTimestamp} evaluated in {DurationMs:F0}ms",
```
colibrishin, line 579: Is this intended to use a delta `DateTimeOffset` as Duration?

greymistcube: Fixed! Must've missed it somehow. Nice catch. 🙂

colibrishin (approved):

PR #2882, opened 2023-03-03
`@planetarium/tx`: support tx signing / signed txs and use `Account` from `@planetarium/account`

tkiapril:
Added `signTx()` and `encodeSignedTx()` functions which can use an `Account` to sign either `UnsignedTxWithSystemAction` or `UnsignedTxWithCustomActions`, and return a `SignedTx`, which can then be converted to `bencodex` representation with `encodeSignedTx()` function.
Also, now `Account` is exported as a regular symbol instead of a type.


@planetarium/account/src/Account.ts
```diff
@@ -8,6 +8,13 @@ export interface Account {
   sign(message: Message): Promise<Signature>;
 }
 
+export function isAccount(account: unknown): account is Account {
+  return (
+    (account as { publicKey: unknown }).publicKey instanceof PublicKey &&
+    (account as { sign: unknown }).sign instanceof Function
+  );
+}
```
moreal, line 16: I guess this part should be recored in `CHANGES.md` too.

tkiapril: I thought it should be fine as-is as `@planetarium/account` is yet to be released. What do you think?

moreal: I forgot the version on npm is the `dev` release version yet. You're right 👍🏻

moreal (dismissed):

moreal (approved):
The bencodex part `Encodable -> Map<string | Buffer, Encodable>` should be resolved if merging after https://github.com/planetarium/libplanet/pull/2889 is merged.

moreal (dismissed):

PR #2894, opened 2023-03-07
♻️ Refactor block commit cache

### LINKED ISSUE #2878: `BlockChain<T>.Fork()` does not copy the `BlockCommit` at the branching point
With the following test code:
```csharp
using System;
using System.Diagnostics;
using System.Linq;
using Bencodex.Types;
using Libplanet.Action;
using Libplanet.Action.Sys;
using Libplanet.Blockchain;
using Libplanet.Blockchain.Policies;
using Libplanet.Blocks;
using Libplanet.Consensus;
using Libplanet.Crypto;
using Libplanet.Store;
using Libplanet.Store.Trie;
using Xunit;

namespace Libplanet.Tests
{
    public class ForkTest
    {
        [Fact]
        public void TipBlockCommitNotForked()
        {
            var pk = new PrivateKey();
            var chain = new BlockChain<Dum>(
                new NullBlockPolicy<Dum>(),
                new VolatileStagePolicy<Dum>(),
                new MemoryStore(),
                new TrieStateStore(new MemoryKeyValueStore()),
                BlockChain<Dum>.ProposeGenesisBlock(
                    privateKey: pk,
                    systemActions: new[]
                    {
                        new SetValidator(new Validator(pk.PublicKey, 1)),
                    }));
            foreach (var i in Enumerable.Range(0, 5))
            {
                var block = chain.ProposeBlock(
                    pk, lastCommit: chain.GetBlockCommit(chain.Tip.Index));
                chain.Append(
                    block,
                    new BlockCommit(
                        chain.Tip.Index + 1,
                        0,
                        block.Hash,
                        ImmutableArray<Vote>.Empty
                            .Add(
                                new VoteMetadata(
                                    chain.Tip.Index + 1,
                                    0,
                                    block.Hash,
                                    DateTimeOffset.UtcNow,
                                    pk.PublicKey,
                                    VoteFlag.PreCommit)
                                    .Sign(pk))));
            }

            Debug.Assert(chain.Tip.PreviousHash != null, "chain.Tip.PreviousHash != null");
            var forked = chain.Fork(chain.Tip.PreviousHash.Value);

            Assert.Equal(chain.Count, forked.Count + 1);
            Assert.Equal(chain.Tip.Index, forked.Tip.Index + 1);
            foreach (var i in Enumerable.Range(0, (int)forked.Count))
            {
                Assert.Equal(chain[i], forked[i]);
                Assert.Equal(chain[i].Hash, forked[i].Hash);
            }

            foreach (var i in Enumerable.Range(0, (int)forked.Count))
            {
                Assert.Equal(chain.GetBlockCommit(i), forked.GetBlockCommit(i));
            }
        }

        private class Dum : IAction
        {
            public IValue PlainValue => null;

            public void LoadPlainValue(IValue plainValue)
            {
            }

            public IAccountStateDelta Execute(IActionContext context) => context.PreviousStates;
        }
    }
}
```

Assert fails at index 4 as the following:
```
Assert.Equal() Failure
Expected: {"block_hash":"64388ae0c8ec676ac6d6ee15acc14f9c9e54cded5695403cd48a41b1a48e92e0","height":4,"round":0,"votes":[{"validator_public_key":"0334622c7cfdd93e83a0ee875bd8b8ac638837ed7d6205392e5fd946e29ccea63a","flag":2,"timestamp":"03/03/2023 03:29:03 \u002B00:00"}]}
Actual:   (null)
```

Which seems to indicate that the `BlockCommit` at the branching point is not copied to the forked chain (`null`.)

greymistcube:
Resolves #2878.

There are numerous issues pertaining to handling "globally" cached `BlockCommit`s, but this is more of a problem with how we handle `IStore` generally. In particular, having multiple chain instances with different chain ids with access to shared state (be it stored `Block<T>`s or `Transaction<T>`s, etc.) and **deletion** operations, the root of the problem is not having a scheme/spec/etc. for managing chain ids.

In any case, this PR only further specifies that any instance of `BlockChain<T>` should have `BlockCommit` for its `BlockChain<T>.Tip`, when applicable, i.e. when it is a non-genesis PBFT `Block<T>`.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1634,8 +1634,19 @@ namespace Libplanet.Blockchain
         /// <see cref="BlockCommit.Height"/> less than <paramref name="limit"/>.
         /// </summary>
         /// <param name="limit">A exceptional index that is not to be removed.</param>
+        /// <exception cref="InvalidOperationException">Thrown when <see cref="IsCanonical"/> is
+        /// <see langword="false"/>.</exception>
         internal void CleanupBlockCommitStore(long limit)
         {
+            // FIXME: This still isn't enough to prevent the canonical chain
+            // removing cached block commits that are needed by other non-canonical chains.
+            if (!IsCanonical)
+            {
+                throw new InvalidOperationException(
+                    $"Cannot perform {nameof(CleanupBlockCommitStore)}() from a " +
+                    "non canonical chain.");
+            }
+
             List<BlockHash> hashes = Store.GetBlockCommitHashes().ToList();
 
             _logger.Debug("Removing old BlockCommits with heights lower than {Limit}...", limit);
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1245,7 +1245,7 @@ namespace Libplanet.Blockchain
         /// <param name="index">A index value (height) of <see cref="Block{T}"/> to retrieve.
         /// </param>
         /// <returns>Returns a <see cref="BlockCommit"/> of given <see cref="Block{T}"/> index.
-        /// Following conditions will returns <see langword="null"/>:
+        /// Following conditions will return <see langword="null"/>:
         /// <list type="bullet">
         ///     <item>
         ///         Given <see cref="Block{T}"/> <see cref="Block{T}.ProtocolVersion"/> is
@@ -1271,7 +1271,7 @@ namespace Libplanet.Blockchain
             }
 
             return index == Tip.Index
-                ? Store.GetBlockCommit(block.Hash)
+                ? Store.GetChainBlockCommit(Id)
                 : this[index + 1].LastCommit;
         }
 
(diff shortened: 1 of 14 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1639,8 +1639,19 @@ internal IEnumerable<BlockHash> IterateBlockHashes(int offset = 0, int? limit =
         /// <see cref="BlockCommit.Height"/> less than <paramref name="limit"/>.
         /// </summary>
         /// <param name="limit">A exceptional index that is not to be removed.</param>
+        /// <exception cref="InvalidOperationException">Thrown when <see cref="IsCanonical"/> is
+        /// <see langword="false"/>.</exception>
         internal void CleanupBlockCommitStore(long limit)
```
colibrishin, line 1644: I think `DeleteBlockCommit()` became a dead code. Behavior has been changed to putting/getting a `BlockCommit` in/from `_chainDB`, but putting/getting in/from `BlockCommit` is not called, and so `DeleteBlockCommit()` will remove nothing for all time.

colibrishin (changes requested):

Libplanet/Blockchain/BlockChain.cs
greymistcube: `DeleteBlockCommit()` is intended to remove a `BlockCommit` from `_blockCommitDB`, which is chain agnostic, not `_chainDB`, which is tied to a specific `Guid`. `BlockCommit`s are cached during download during preloading phase.

colibrishin: Ah, yes, I forgot that a `BlockCommit` is stored while in synchronization.

colibrishin (approved):

PR #2902, opened 2023-03-09
ci/bench: benchmark for branches

colibrishin:
## Context
This PR addresses changing the benchmark to check the multiple branches and compare the result for each.

## Change
- Benchmark results will be stored in `gh-pages/dev/bench/<branch_name>`.
- Failed benchmark name will be shown in Github Actions log.
- PR benchmark compares the result from the target branch directory. (e.g., gh-pages/dev/bench/main/...)
- Merged benchmark is now triggered only if the PR is merged. (`on.push` with an arbitrary branch it was also triggered in PR.)
- Minor fix for `BlockChain<T>.Create()` changes.


riemannulus (approved):
👍

PR #2915, opened 2023-03-10
`@planetarium/account-web3-secret-storage`

tkiapril:
This PR provides account-web3-secret-storage, which provides `Web3KeyStore` that implements Ethereum Web3 Secret Storage and complies to the `KeyStore` interface that `@planetarium/account` provides (also included in this PR.)

### DIFF
--- a/@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
+++ b/@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
@@ -1,0 +1,227 @@
+import { KeyId } from "./KeyId.js";
+import { PassphraseEntry } from "./PassphraseEntry.js";
+import {
+  type AccountDeletion,
+  type AccountGeneration,
+  type AccountImportation,
+  type AccountMetadata,
+  type AccountRetrieval,
+  Address,
+  type ImportableKeyStore,
+  RawPrivateKey,
+} from "@planetarium/account";
+import {
+  type KeystoreAccount,
+  decryptKeystoreJson,
+  encryptKeystoreJson,
+  isKeystoreJson,
+} from "ethers";
+import { Dirent } from "node:fs";
+import * as fs from "node:fs/promises";
+import { homedir } from "node:os";
+import * as path from "node:path";
+
+export interface Web3KeyStoreOptions {
+  path?: string;
+  passphraseEntry: PassphraseEntry;
+}
+
+/**
+ * Determines the default key store path.  It depends on the platform:
+ *
+ * - Linux/macOS: `$HOME/.config/planetarium/keystore`
+ * - Windows: `%AppData%\planetarium\keystore`
+ */
+export function getDefaultWeb3KeyStorePath(): string {
+  const baseDir =
+    process.platform === "win32"
+      ? process.env.AppData || path.join(homedir(), "AppData", "Roaming")
+      : process.env.XDG_CONFIG_HOME || path.join(homedir(), ".config");
+  // Note that it's not necessary to explicitly choose one of `path.win32` or
+  // `path.posix` here, but it makes unit tests less dependent on mocks:
+  return (process.platform === "win32" ? path.win32 : path.posix).join(
+    baseDir,
+    "planetarium",
+    "keystore",
+  );
+}
+
+const pattern =
+  /^(?:UTC--([0-9]{4}-[0-9]{2}-[0-9]{2})T([0-9]{2}-[0-9]{2}-[0-9]{2})Z--)?([0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12})(?:.json)?$$/i;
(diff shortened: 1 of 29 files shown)


@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
```diff
@@ -0,0 +1,218 @@
...
+          return { result: "error", keyId, message: `${e}` };
+        }
+        firstAttempt = false;
+        continue;
+      }
+      break;
+    }
+    return {
+      result: "success",
+      // TODO: Declare Web3Account
```
Akamig, line 155: Can you elaborate on this one? what kind of methods are on the table? I'm just curious.

Akamig (commented):
This is the cleanest and most elegant JS/TS usage I haven't seen for a long time. period. I get learned thing or two from here.

@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export interface Web3KeyStoreOptions {
+  path?: string;
+  passphraseEntry: PassphraseEntry;
+}
+
+/**
+ * Determines the default key store path.  It depends on the platform:
+ *
+ * - Linux/macOS: `$HOME/.config/planetarium/keystore`
```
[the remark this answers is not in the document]
tkiapril: As a conclusion of our discussion (@Akamig, @dahlia, and @tkiapril,) we tentatively agreed that; `Application Support` is a variable directory where the files might be moved if the applications requiring the directory is not present. However, a key being removed without user's explicit action may have adverse effect, so it might be better to have a directory that is not intended for automated deletion without user's discretion to be used in the macOS launcher.

@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
tkiapril: For the record, we discussed the possibility of carrying only the path of the keystore for the implementation, and different implementations of `PassphaseEntry` might be used for policy on what to do when key is intended to be used (so that passphrase can be off memory / can be requested once in a few minutes / or have a confirmation window, like the unix `sudo` command (or rather the Github sudo mode.))

Akamig (approved):

@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
```diff
@@ -0,0 +1,224 @@
...
+      if (
+        typeof e === "object" &&
+        e != null &&
+        "code" in e &&
+        e.code === "ENOENT"
+      ) {
+        return;
+      }
+      throw e;
+    }
```
longfin, line 89: I guess that this EAFP("It's Easier to Ask Forgiveness than Permission") doesn't seem to help enough, since there are only two cases we concerning; one is ENOENT, another one is just throwing. 🤔

longfin (approved):
I left a minor comment.

moreal (approved):

moreal (dismissed):

@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
tkiapril: The purpose of this code is to return an empty list if the directory containing web3 keystore files is nonexistent(ENOENT), but we wanted to throw when there is another issue preventing from opening the directory.

dahlia:
@maxswjeon @longfin @Akamig @moreal We've added the complete tests and rebased on the latest main.  Please take a look again!

@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
dahlia: It's just for considering the case where the keystore directory does not exist like it's empty (rather than considering it an exception case).  We left a comment to explain the intention.

longfin: I understood that intention and thought that using LBYL style(i.e., adding the check for `ENOENT`) without `catch` might be more expressive. 😅

longfin (approved):

@planetarium/account-web3-secret-storage/src/Web3KeyStore.ts
dahlia: Expressiveness-wise, you're right.  However, it's always tricky when we deal with timings… you know.  What if the directory was there a microsecond ago, and then it's suddenly gone when it's starting to read the directory entries?

longfin: > it's always tricky when we deal with timings… 

🆗  it's also fair enough.

moreal (approved):

dahlia (approved):

PR #2956, opened 2023-03-17
Fix `ResolveNetMQAddressAsync` to return IPv4

OnedgeLee:
For now, `ResolveNetMQAddressAsync` returns IPv6 addresses after [this commit](https://github.com/planetarium/libplanet/commit/977a79462cfe47b88606d6ef4e69b989b297a544).
`DealerSocket` cannot connect with IPv6 address for now, so casted to IPv4.

trivials
- Changed `DealerSocket.Option.IPv4Only` to `false`, but it couldn't receive IPv6 addr.
- `GetHostAddressesAsync()` has overloading with `AddressFamily`, but supported on `.Net 6, 7, 8`.


Libplanet.Net/Transports/BoundPeerExtensions.cs
```diff
@@ -76,8 +76,10 @@ internal static async Task<string> ResolveNetMQAddressAsync(this BoundPeer peer)
             // FIXME Dns.GetHostAddressesAsync() seems to always return at least one ip or throw
             // `SocketException`. if we can be sure, also should fix it to always assume there
             // is at least one instead of throwing additional `TransportException`.
-            string ip = addresses.FirstOrDefault()?.ToString()
+            // And, for now IPs are restricted with IPv4, and have to be fixed later on.
+            string ip = addresses.FirstOrDefault()?.MapToIPv4().ToString()
                 ?? throw new TransportException($"Failed to resolve for {addr}");
+            ip = (ip == "0.0.0.1") ? "127.0.0.1" : ip;
```
riemannulus, line 82:
```suggestion
            string ip = addresses.FirstOrDefault()?.MapToIPv4().ToString()
                ?? throw new TransportException($"Failed to resolve for {addr}");
            string ipv4 = (ip == "0.0.0.1") ? "127.0.0.1" : ip;
```
I don't have much good idea for using ipv4 only when localhost but IMO `ipv4` is the correct variable name in this case.

riemannulus: Oh, I'm misreading it. 😓  just ignore this comment.

riemannulus (approved):

greymistcube:
Just my two cents. We could maybe use something like below? 🙄

```csharp
            IPAddress addresses = (await Dns.GetHostAddressesAsync(addr).ConfigureAwait(false))
                .FirstOrDefault(addr => addr.AddressFamily is AddressFamily.InterNetwork);
```

OnedgeLee:
> Just my two cents. We could maybe use something like below? 🙄
> 
> ```cs
>             IPAddress addresses = (await Dns.GetHostAddressesAsync(addr).ConfigureAwait(false))
>                 .FirstOrDefault(addr => addr.AddressFamily is AddressFamily.InterNetwork);
> ```

What I've thought was, this solution will filter just IPv4 format results, and ignore IPv6 formats, but I'm not sure so I'll make a test.

Libplanet.Net/Transports/BoundPeerExtensions.cs
```diff
@@ -76,11 +77,13 @@ internal static async Task<string> ResolveNetMQAddressAsync(this BoundPeer peer)
             // FIXME Dns.GetHostAddressesAsync() seems to always return at least one ip or throw
             // `SocketException`. if we can be sure, also should fix it to always assume there
             // is at least one instead of throwing additional `TransportException`.
-            string ip = addresses.FirstOrDefault()?.ToString()
+            // And, for now IPs are restricted with IPv4, and have to be fixed later on.
```
greymistcube, line 80: I think `XML` documentation would make this method more usable (that this method only checks whether `BoundPeer.EndPoint.Host` resolves to an IPv4 address).

Also looking at (although this is for `GetHostEntry()` instead of `GetHostAddressesAsync()`)

https://stackoverflow.com/questions/156585/can-dns-gethostentry-ever-return-an-iphostentry-with-an-empty-addresslist

seems like it is technically possible to get an empty list. Maybe we can also change the `FIXME` comment accordingly. 😗

greymistcube (approved):

PR #2971, opened 2023-03-21
Get rid of Libplanet.Node

dahlia:
*Context on Discord (Korean): <https://discord.com/channels/928926944937013338/929910863123591199/1087633931697934386>.*

This patch gets rid of *Libplanet.Node* package, which is not used anymore<del datetime="2023-03-22T15:26:23+09:00">, and flattens `ITxMetadata` and `ITxExcerpt` interfaces into `ITransaction`</del>. <ins datetime="2023-03-22T15:26:23+09:00">I left only a commit to get rid of *Libplanet.Node* in this pull request.  Instead, I created another pull request <https://github.com/planetarium/libplanet/pull/2977> that dedicates flattening transaction interfaces.</ins>


riemannulus (dismissed):

riemannulus (dismissed):
👋

greymistcube:
Any particular reason why `ITxMetadata` was removed? 🤔

dahlia:
Actually, *Libplanet.Node* was removed to get rid of `ITxMetadata`.  @tkiapril and I are going to add more sensible interfaces to represent various aspects/forms of transactions.

greymistcube:
I thought it might be good to have "everything ready except signing" `interface`. I can see how `ITxMetadata` might get in the way of complicating things when we try to add more "sensible" `interface`s (whether to include `Nonce` etc.). Although I imagine anything short of `ITxMetadata` should be something `ITxMetadata` would ineherit. 🙃

tkiapril:
We actually are thinking of something similar to that for a follow-up PR, where you would only have to provide the nonce and sign with the private key. The signer and the public key would be derived from the pk.

greymistcube:
> We actually are thinking of something similar to that for a follow-up PR, where you would only have to provide the nonce and sign with the private key. The signer and the public key would be derived from the pk.

I understand that this might be useful from an end-user perspective. I was suggesting from a code-contractual standpoint. Of course, making *everything* into an `interface` is not the best way to go, but I still think `ITxMetadata` itself might be useful as an `interface` for an `ITransaction` with only signature and txid missing. 😶

dahlia:
> I understand that this might be useful from an end-user perspective. I was suggesting from a code-contractual standpoint. Of course, making _everything_ into an `interface` is not the best way to go, but I still think `ITxMetadata` itself might be useful as an `interface` for an `ITransaction` with only signature and txid missing. 😶

First of all, we're going to define interfaces to represent smaller pieces of transactions than `ITxMetadata`.  If we can achieve such abstract layers we could define a new type built on top of them to replace the existing `ITxMetadata`.

dahlia:
I left only a commit to get rid of *Libplanet.Node* in this pull request.  Instead, I created another pull request <https://github.com/planetarium/libplanet/pull/2977> that dedicates flattening transaction interfaces.

greymistcube (dismissed):

dahlia:
@greymistcube  Rebased.  Please take a look again!

greymistcube (approved):

PR #2977, opened 2023-03-22
Flatten `ITxMetadata`/`ITxExcerpt` into `ITransaction`

### LINKED ISSUE #1997: Give `ITxExcerpt` a better name
> Can we clarify it's use a bit more?
>
> As I understand it, seems like this is to "group" `UntypedTransaction` and `Transaction<T>` for validating nonces. In any case, the relation between `ITxExcerpt` and `ITxMetadata` should mirror that of `IBlockExcerpt` and `IBlockMetadata` if possible. 😕

_Originally posted by @greymistcube in https://github.com/planetarium/libplanet/pull/1978#discussion_r881171057_

> I defined the term *excerpt* as a set of metadata functioning as a proof without complete data (usually lacking its body like actions and transactions).  That's why `IBlockExcerpt` has `Hash`, and `ITxExcerpt` has `Id`.

_Originally posted by @dahlia in https://github.com/planetarium/libplanet/pull/1978#discussion_r881190462_

> > I defined the term excerpt as a set of metadata
> 
> I see _excerpts_ are some sort of metadata. On the other hand, `ITxMetadata` and `IBlockMetadata` have analogous specific uses. I'm still confused as to what this is "proving" really and if `ITxExcerpt : ITxMetadata` is necessary.
> 
>     * If the term _excerpt_ is to _extend_ metadata interfaces with additional proof data, than inheritance is granted.
> 
>     * If the term _excerpt_ is to be used for naming an auxiliary `interface` with minimal amount of information than there shouldn't be an inheritance relation.
> 
> 
> I'm suggesting that either we have `IBlockExcerpt : IBlockMetadata` and `ITxExcerpt : ITxMetadata` (this is the former of the above) or not have inheritance relations between them (the latter of the above). The split design seems confusing.

_Originally posted by @greymistcube in https://github.com/planetarium/libplanet/pull/1978#discussion_r881229427_

> I tried to make `ITxMetadata` to inherit nothing, but it wasn't that easy as `TxExcerptExtensions.ValidateTxNonces()` used several properties of `ITxMetadata`.  Instead, I'm going to rename its name later.

_Originally posted by @dahlia in https://github.com/planetarium/libplanet/pull/1978#discussion_r881441594_

dahlia:
*Closes #1997.*


Libplanet/Tx/ITransaction.cs
```diff
@@ -1,15 +1,65 @@
...
+
+        /// <summary>
+        /// The number of previous <see cref="Transaction{T}"/>s committed by
+        /// the <see cref="Signer"/> of this transaction.  This nonce is used for preventing
+        /// replay attack.
+        /// </summary>
+        long Nonce { get; }
+
+        /// <summary>
+        /// A <see cref="PublicKey"/> of the account who signs this transaction.
```
moreal, line 32: Isn't it the `Address` of the account instead of `PublicKey`?

Libplanet/Tx/TxMetadata.cs
```diff
@@ -81,24 +80,45 @@ public TxMetadata(Bencodex.Types.Dictionary dictionary)
...
+        /// <summary>
+        /// The number of previous <see cref="Transaction{T}"/>s committed by
+        /// the <see cref="Signer"/> of this transaction.  This nonce is used for preventing
+        /// replay attack.
+        /// </summary>
         public long Nonce { get; set; }
 
-        /// <inheritdoc cref="ITxMetadata.Signer"/>
+        /// <summary>
+        /// A <see cref="PublicKey"/> of the account who signs this transaction.
```
moreal, line 91: Isn't it the `Address` of the account instead of `PublicKey`?

Libplanet/Tx/ITransaction.cs
dahlia: Amended.

Libplanet/Tx/TxMetadata.cs
dahlia: Amended.

moreal (approved):

longfin (approved):

PR #2996, opened 2023-03-27
fix: validate block content in PreVote step

### LINKED ISSUE #2973: Validate proposed `Block<T>` before making a `Vote`
Currently, it is possible to vote on and to reach a consensus on a `Block<T>` with an invalid state root hash. If this happens, the entire network can be faulted by a single malicious validator.

colibrishin:
Closes #2973 

## Context
Current block validation in PBFT consensus guarantees the sanity of a block header but the block content (e.g., Tx nonces, Policy defined validation rules or state root hash.) This could lead `Context<T>` to being locked in an invalid block and committed by consensus, even though the block is invalid.

## Changes
- Rename `BlockChain<T>.ValidateNextBlock()` to `ValidateNextBlockHeader()`
- Refactor Tx nonce/policy validation and calculate and commit the state root hash to the `BlockChain<T>.ValidateNextBlockContent()` method.
- `Context<T>` validates the block with `ValidateNextBlockContext()` and if a node is a proposer, bypass the actions evaluation.
- Added `Context<T>.IsCurrentRoundProposer()` method due to SA1113.

## Further work
This fix will resolve the problem, however, lengthen the consensus time by evaluating the block twice, in the PreVote and Commit step, and the PreVote step timeout might not be triggered accurately while validating block content.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1201,7 +1201,7 @@ namespace Libplanet.Blockchain
             Block<T> prevTip = Tip;
             try
             {
-                if (ValidateNextBlock(block) is { } ibe)
+                if (ValidateNextBlockHeader(block) is { } ibe)
                 {
                     throw ibe;
                 }
@@ -1608,7 +1608,7 @@ namespace Libplanet.Blockchain
             return null;
         }
 
-        internal InvalidBlockException ValidateNextBlock(Block<T> block)
+        internal InvalidBlockException ValidateNextBlockHeader(Block<T> block)
         {
             if (block.Index == 0)
             {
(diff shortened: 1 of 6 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1792,6 +1781,69 @@ internal InvalidBlockException ValidateNextBlock(Block<T> block)
...
+                    block.Index,
+                    block.Hash);
+                var actionEvaluations = ExecuteActions(block, writeTxExecution);
+                _logger.Information(
+                    "Executed actions in block #{BlockIndex} {BlockHash}",
+                    block.Index,
+                    block.Hash);
+
+                return (actionEvaluations, nonceDeltas);
+            }
```
greymistcube, line 1842: I have general qualms with evaluating the content inside `ValidateNextBlockContent()`. 🙄

- Evaluation result isn't part of `BlockContent<T>`'s scope.
- I think usage `writeTxExecution` for the sake of having an ephemeral execution is bad. Actually `UpdateTxExecutions()` isn't where it should be (that is, inside `EvaluateActions()` instead of `Append()`).
- Combining the two, the signature for this should be
```csharp
Dictionary<Address, long> ValidateNextBlockContent(IBlockContent<T>)
```

greymistcube (approved):

limebell (approved):

PR #2999, opened 2023-03-28
Add `GetState` measurement for single action execution

limebell:


longfin (approved):

longfin:
Just for note) we may need `ActionContext.Current` to access the current `ActionContext` handy, from nested codes. 🤔

PR #3022, opened 2023-03-29
🔀 Port 0.50.7 to 0.53.2

riemannulus:
🔀

### DIFF
--- a/Libplanet/Action/ActionEvaluator.cs
+++ b/Libplanet/Action/ActionEvaluator.cs
@@ -613,8 +613,6 @@ namespace Libplanet.Action
             bool rehearsal = false,
             ITrie? previousBlockStatesTrie = null)
         {
-            Stopwatch stopwatch = new Stopwatch();
-            stopwatch.Start();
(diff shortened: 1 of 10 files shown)


greymistcube (approved):

Libplanet/Action/ActionEvaluator.cs
```diff
@@ -605,7 +605,7 @@ internal static IImmutableSet<Address> GetUpdatedAddresses<T>(Transaction<T> tx)
                     tx.Id,
                     tx.Signer,
                     block.Index,
-                    ByteUtil.Hex(block.PreEvaluationHash));
+                    ByteUtil.Hex(block.PreEvaluationHash.ByteArray));
```
limebell, line 608: Maybe just use `block.PreEvaluationHash` is more clean? (It implements `.ToString()` methods)

riemannulus: That is `ByteUtil.Hex`, not `ByteUtil.ToHex`. Parameter type is `byte[]`.

PR #3028, opened 2023-03-30
feat: cache evaluated actions in `Context<T>`

### LINKED ISSUE #3045: Multiple block evaluations in consensus
A proposed block by PBFT consensus will be evaluated multiple times (while in `PreVote` and `Commit`). This causes lengthening the consensus time.

colibrishin:
Following PR of #2996 and partially resolves #3045

## Context
Currently, block evaluation is done in proposing a block, which is required to get a final block from `PreEvaluationBlock`, `PreVote`, which is a stage of consensus for validating a given block, and `Commit`, which is a stage of consensus of finalizing the given block. This PR will cache an evaluated action from a block in the `PreVote` stage, and use it in the `Commit` stage.

## Changes
With this, an evaluation count will be reduced from two to one, if a node is a non-proposer, or from three to two, if a node is a proposer. From a network perspective, the evaluation will be done two times, which are from the proposer in `Propose`, and from all nodes in `PreVote`.

### Considerations
A proposer could bypass the `PreVote` evaluation by caching an evaluation created from proposing stage. This was ruled out due to complicating the logic in `Context<T>` without any positive effect or trade-off, optimization-wise and safety-wise. For example, even if a proposer bypasses the validation, the proposer has to wait for `PreVote` from +2/3 validators, therefore, the consensus time of the network stays remain. `BlockChain<T>` could have different statuses with `Context<T>`, or `BlockChain<T>` proposing an invalid block, but can be caught on validation.

## Further works
If a block synchronizes before the `Context<T>` committing, a cached action is not used because of that, `BlockChain<T>` will re-evaluate the block again.

### DIFF
--- a/Libplanet.Net/Consensus/Context.Mutate.cs
+++ b/Libplanet.Net/Consensus/Context.Mutate.cs
@@ -25,7 +25,7 @@ namespace Libplanet.Net.Consensus
                 ToString());
             Round = round;
             Step = Step.Propose;
-            if (IsCurrentRoundProposer())
+            if (_validatorSet.GetProposer(Height, Round).PublicKey == _privateKey.PublicKey)
             {
                 _logger.Information(
                     "Starting round {NewRound} and is a proposer.",
--- a/Libplanet.Net/Consensus/Context.Mutate.cs
+++ b/Libplanet.Net/Consensus/Context.Mutate.cs
@@ -124,7 +124,7 @@ namespace Libplanet.Net.Consensus
                     ToString());
                 Step = Step.PreVote;
 
-                if (IsValid(p1.Block) && (_lockedRound == -1 || _lockedValue == p1.Block))
+                if (IsValid(p1.Block, out _) && (_lockedRound == -1 || _lockedValue == p1.Block))
                 {
                     BroadcastMessage(
                         new ConsensusPreVoteMsg(MakeVote(Round, p1.Block.Hash, VoteFlag.PreVote)));
@@ -150,7 +150,7 @@ namespace Libplanet.Net.Consensus
                     ToString());
                 Step = Step.PreVote;
 
-                if (IsValid(p2.Block) &&
+                if (IsValid(p2.Block, out _) &&
                     (_lockedRound <= p2.ValidRound || _lockedValue == p2.Block))
                 {
                     BroadcastMessage(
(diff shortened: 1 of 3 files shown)


Libplanet.Net/Consensus/Context.Mutate.cs
```diff
@@ -266,7 +266,24 @@ private void ProcessHeightOrRoundUponRules(ConsensusMsg message)
                         block4.Index,
                         block4.Hash,
                         ToString());
-                    _blockChain.Append(block4, GetBlockCommit());
+
+                    bool found = _blockValidationCache.TryGet(
+                        (block4.Hash, block4.StateRootHash), out var cached);
```
greymistcube, line 271: I think new instances of anon pairs will have different `GetHashCode()` result. I don't think this would work without a new `struct`. Besides, I think `BlockHash` would suffice as a key. 🙄

Libplanet.Net/Consensus/Context.cs
```diff
@@ -96,7 +97,10 @@ public partial class Context<T> : IDisposable
         private readonly CancellationTokenSource _cancellationTokenSource;
 
         private readonly ILogger _logger;
-        private readonly LRUCache<BlockHash, bool> _blockHashCache;
+        private readonly
+            LRUCache<(BlockHash BlockHash, HashDigest<SHA256> EvaluatedActions),
+                (bool IsValid, IReadOnlyList<ActionEvaluation>? EvaluatedActions)>
+            _blockValidationCache;
```
greymistcube, line 103: I think we can get rid of `HashDigest<SHA256>`? 🤔

Libplanet.Net/Consensus/Context.Mutate.cs
greymistcube: Also instead of trying to grab a cache here again, I think `bool IsValid(Block<T>, out IReadOnlyList<ActionEvaluation>)` might work better. 😗

colibrishin: Oh, good point. It does not pair the block with actions but it only increases the input value for the cyclic hash.

_After some searches, found that those are boxed into `ValueTuple` and the hash code is calculated from the `GetHashCode()` of items. I'm not sure but probably `ValueTuple` will be compared with the hash code. (Because in this case, `LRUCache` uses `Dictionary.ContainsKey()` internally and it uses the hash code if `IEqualityComparer<T>` is not defined.)_

https://github.com/dotnet/roslyn/blob/f5a0fc0b9ddfdd58ee2609b36076059c6a0069e1/src/Compilers/Test/Resources/Core/NetFX/ValueTuple/ValueTuple.cs#L637-L641

greymistcube: I guess I was mistaken. 🙄

colibrishin: But, your statement seems to be right, [[stackoverflow link goes here]](https://stackoverflow.com/questions/2240231/how-well-does-net-dictionary-resolve-collisions). It seems .net `Dictionary` is handling collision internally. If we want this to be properly handled, `Equal()` should be implemented in `struct` or so.

greymistcube (approved):

PR #3042, opened 2023-04-03
Adjust consensus loggins

limebell:
This changes `MessageLog.Add()`'s return type to `void` (which was `bool`), and throws error instead if failed. Since the method was `internal`, skip changelog.

### DIFF
--- a/Libplanet.Net/Consensus/Gossip.cs
+++ b/Libplanet.Net/Consensus/Gossip.cs
@@ -378,10 +378,10 @@ namespace Libplanet.Net.Consensus
             MessageId[] ids = contents.Select(c => c.Id).ToArray();
 
             _logger.Debug(
-                "WantMessage: Requests are: {Idr}, Ids are: {Id}, Messages are: {@Messages}",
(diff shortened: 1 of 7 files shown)


Libplanet.Net/Consensus/Gossip.cs
```diff
@@ -378,10 +378,10 @@ private async Task HandleWantAsync(Message msg, CancellationToken ctx)
             MessageId[] ids = contents.Select(c => c.Id).ToArray();
 
             _logger.Debug(
-                "WantMessage: Requests are: {Idr}, Ids are: {Id}, Messages are: {@Messages}",
+                "WantMessage: Requests are: {Idr}, Ids are: {Id}, Messages are: {Messages}",
                 wantMessage.Ids,
                 ids,
-                contents);
+                contents.Select(content => content.Type + "(" + content.Id + ")"));
```
greymistcube, line 384: I think `Serilog` is smart enough to handle the following:

```
 contents.Select(content => (content.Type, content.Id)));
```

😗

Issue #2709, opened 2023-01-13
`PublicKey(string hex)` constructor (or `PublicKey.FromString()` method)


dahlia:
There is `PublicKey.ToString()` which returns a hexadecimal string, but there is no way to parse it into `PublicKey` instance.  It would be useful if `PublicKey` has a constructor (or a static method) taking `string`.

PR #3044, opened 2023-04-03
Implement TypeConverter for commonly used types 

### LINKED ISSUE #2711: Implement `TypeConverter` for commonly used types
[`TypeConverter`][1] is a de facto standard in .NET for serializing/deserializing a .NET object into/from a string.  It is used by [Microsoft.Extensions.Configuration][2] and [Cocona][3].  Currently, we need to take values like public keys, addrsses, and block hashes in string types first, and then manually convert them into an appropriate type.  Similar conversion code is spread here and there.  We could avoid such duplication by letting commonly used Libplanet types support `TypeConverter`.

[1]: https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter
[2]: https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration#binding
[3]: https://github.com/mayuki/Cocona/blob/v2.1.0/src/Cocona.Core/Command/Binder/CoconaValueConverter.cs#L29

dahlia:
*Closes #2711 & #2709.*

### DIFF
--- a/Libplanet.Tests/AddressTest.cs
+++ b/Libplanet.Tests/AddressTest.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Immutable;
+using System.ComponentModel;
 using System.Linq;
 using Bencodex.Types;
 using Libplanet.Crypto;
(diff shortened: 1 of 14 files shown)


Libplanet.Tests/AddressTest.cs
```diff
@@ -311,6 +312,24 @@ public void Serializable()
...
+            Assert.Equal(
+                address,
+                converter.ConvertFrom("0x0123456789ABcdefABcdEfABcdEFabcDEFabCDEF"));
+            Assert.Equal(
+                address,
+                converter.ConvertFrom("0123456789ABcdefABcdEfABcdEFabcDEFabCDEF"));
+            Assert.True(converter.CanConvertTo(typeof(string)));
+            Assert.Equal(
+                "0123456789ABcdefABcdEfABcdEFabcDEFabCDEF",
+                converter.ConvertTo(address, typeof(string)));
```
tkiapril, line 330: I'm not sure about how type conversion works here so this is just a question, but is it okay to not implement some kind of validation for the string to be converted to an Address instance, or a fail test case? (PublicKeyTest too)

tkiapril (approved):

Libplanet.Tests/AddressTest.cs
dahlia: They will throw exceptions such as `ArgumentOutOfRangeException` or `FormatException`, and Cocona treat it as a wrong option value if whatever they throw.

dahlia: On second thoughts, however, it would be better to have tests for exceptional cases as well.  I'm going to add them.

dahlia: @tkiapril Amended.  Please take a look again!

tkiapril (approved):

PR #3057, opened 2023-04-05
🔀 Port 0.53.3 to main

greymistcube:
🔀


Libplanet/Action/Sys/Registry.cs
```diff
@@ -72,7 +72,7 @@ action switch
                 Mint _ => TypeId.Mint,
                 Transfer _ => TypeId.Transfer,
                 Initialize _ => TypeId.Initialize,
-                _ => null,
+                _ => throw new ArgumentException("Unknown system action type.", nameof(action)),
```
tkiapril, line 75: We intentionally made this null.

greymistcube: Yup just saw a test fail. 😥

tkiapril (approved):

PR #3060, opened 2023-04-05
Implement `Gossip.PublishMessage()`

### LINKED ISSUE #3054: Implement `Publish` method for `Gossip`
`Gossip` depends only on `HeartbeatTask` to broadcast message, thus messages are delayed. In Libp2p gossipsub, defining `Publish` method to broadcast message immediately.

limebell:
`Gossip` used to broadcast messages via `HeartbeatTask`, which is called every second. Message can be delayed at most 1 second and it may hurt consensus health.

This patch implements `Gossip.PublishMessage()` method that broadcasts message immediately, and gossips it after it sent.

closes #3054

### DIFF
--- a/Libplanet.Net.Tests/Consensus/GossipTest.cs
+++ b/Libplanet.Net.Tests/Consensus/GossipTest.cs
@@ -50,9 +50,9 @@ namespace Libplanet.Net.Tests.Consensus
             var key2 = new PrivateKey();
             var receivedEvent = new AsyncAutoResetEvent();
             var gossip1 = CreateGossip(
-                message =>
+                content =>
                 {
-                    if (message.Content is ConsensusProposalMsg)
+                    if (content is ConsensusProposalMsg)
                     {
                         received1 = true;
                     }
@@ -61,9 +61,9 @@ namespace Libplanet.Net.Tests.Consensus
                 6001,
                 new[] { new BoundPeer(key2.PublicKey, new DnsEndPoint("127.0.0.1", 6002)) });
             var gossip2 = CreateGossip(
-                message =>
+                content =>
                 {
-                    if (message.Content is ConsensusProposalMsg)
+                    if (content is ConsensusProposalMsg)
                     {
                         received2 = true;
                         receivedEvent.Set();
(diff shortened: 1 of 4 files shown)


colibrishin (dismissed):

Libplanet.Net.Tests/Consensus/GossipTest.cs
```diff
@@ -40,19 +40,74 @@ public void Dispose()
...
+            {
+                _ = gossip1.StartAsync(default);
+                _ = gossip2.StartAsync(default);
+                await gossip1.WaitForRunningAsync();
+                await gossip2.WaitForRunningAsync();
+                gossip1.PublishMessage(
+                    TestUtils.CreateConsensusPropose(fx.Block1, new PrivateKey(), 0));
+                await receivedEvent.WaitAsync();
+                Assert.True(received1);
+                Assert.True(received2);
```
greymistcube, line 85: I think this should test somehow how fast the message is received. Maybe put a cap on using a timer?

limebell: I agree, but it is not easy to implement because broadcast also has its own delay and it is also random.

IMO, to check whether the message is received via `HaveMsg` is most accurate, but it needs some "holes"...

greymistcube: I was thinking something like the following:

```csharp
                Stopwatch stopwatch = new Stopwatch();
                var proposeMsg = TestUtils.CreateConsensusPropose(fx.Block1, new PrivateKey(), 0);
                _ = gossip1.StartAsync(default);
                _ = gossip2.StartAsync(default);
                await gossip1.WaitForRunningAsync();
                await gossip2.WaitForRunningAsync();
                stopwatch.Start();
                gossip1.PublishMessage(proposeMsg);
                await receivedEvent.WaitAsync();
                stopwatch.Stop();
                Assert.True(stopwatch.ElapsedMilliseconds < 100);
                Assert.True(received1);
                Assert.True(received2);
```

This is rather hacky tho. 🙄

greymistcube (dismissed):

PR #3061, opened 2023-04-05
@planetarium/account-web3-secret-storage: `Web3Account` (deferred unlock) & asyncify `getPublicKey()`

tkiapril:
Now, *@planetarium/account-web3-secret-storage* uses `Web3Account` instead of `RawPrivateKey`. This change enables the users to interact with an instance of `Account` before unlocking the account, and defers the authentication (`PassphraseEntry.authenticate()`) to when the operation that requires unlocked account is requested.

Additionally, since some backends (web3) require accounts to be unlocked before publickey can be shown, Account.publicKey property is now changed to async Account.getPublicKey() method.

### DIFF
--- a/@planetarium/account-web3-secret-storage/src/Web3Account.ts
+++ b/@planetarium/account-web3-secret-storage/src/Web3Account.ts
@@ -1,0 +1,92 @@
+import { KeyId } from "./KeyId.js";
+import { PassphraseEntry } from "./PassphraseEntry.js";
+import {
+  ExportableAccount,
+  PublicKey,
+  RawPrivateKey,
+  Signature,
+} from "@planetarium/account";
+import { type KeystoreAccount, decryptKeystoreJson } from "ethers";
+
+export type Web3KeyObjectKdf =
+  | {
+      kdf: "pbkdf2";
+      kdfparams: {
+        c: number;
+        dklen: number;
+        prf: "hmac-sha256";
+        salt: string;
+      };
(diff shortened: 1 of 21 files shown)


@planetarium/account-web3-secret-storage/src/Web3Account.ts
```diff
@@ -0,0 +1,92 @@
...
+    let account: KeystoreAccount;
+    while (true) {
+      const passphrase = await this.#passphraseEntry.authenticate(
+        this.#keyObject.id,
+        firstAttempt,
+      );
+      let currentProgress = 0;
+      const json = JSON.stringify(this.#keyObject);
+      try {
+        // TODO: ethers does not support scrypt yet, however Libplanet.KeyStore.Web3KeyStore does.
```
Akamig, line 66: ethers does not support scrypt?!?!?!?!?!?!?!?!?

OH...............

Akamig (approved):
Getting address thorough `Web3Account` seems appropriate. Um, but if I remember correctly, address field sometimes does not exists in some implementation of Web3 SS generators (i.e. ethereumjs-wallet). is that covered as well?

tkiapril:
@Akamig 🤔 As the spec requires that field, we might as well as consider generated wallets that lack the address malformed. (Though we should validate the JSON format, we shall leave it for next PR.)

dahlia (approved):

PR #3063, opened 2023-04-05
Replace newHeightDelay with contextMinInterval

OnedgeLee:
### Problem to solve
Context interval is expected to be constant, but it varies.

### Changes
`newHeightDelay`, that suspends `NewHeight()` with fixed time after `OnTipChanged` will be removed, and replaced with `contextMinInterval` that suspends `NewHeight()` with dynamic time, till interval between previous `NewHeight()` call and current `NewHeight()` call reaches to it.

### Limitation
Validator that doesn't have context of new tip appends block without delay, since it doesn't know the time that context has been made.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -26,6 +26,7 @@ namespace Libplanet.Net.Consensus
 
         private readonly BlockChain<T> _blockChain;
         private readonly PrivateKey _privateKey;
+        private readonly TimeSpan _contextMinInterval;
         private readonly ILogger _logger;
         private readonly Dictionary<long, Context<T>> _contexts;
 
@@ -58,8 +59,8 @@ namespace Libplanet.Net.Consensus
             BroadcastMessage = broadcastMessage;
             _blockChain = blockChain;
             _privateKey = privateKey;
+            _contextMinInterval = contextMinInterval;
             Height = -1;
-            ContextMinInterval = contextMinInterval;
 
             _contextTimeoutOption = contextTimeoutOption;
 
@@ -135,11 +136,6 @@ namespace Libplanet.Net.Consensus
         /// height of value, and value is the <see cref="Context{T}"/>.
         /// </summary>
         internal Dictionary<long, Context<T>> Contexts => _contexts;
-
-        /// <summary>
-        /// A minimum time interval between each <see cref="Context{T}"/> starts.
-        /// </summary>
-        internal TimeSpan ContextMinInterval { get; private set; }
 
         /// <inheritdoc cref="IDisposable.Dispose"/>
         public void Dispose()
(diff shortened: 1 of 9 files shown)


Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -137,6 +136,11 @@ public Step Step
         /// </summary>
         internal Dictionary<long, Context<T>> Contexts => _contexts;
 
+        /// <summary>
+        /// A minimum time interval between each <see cref="Context{T}"/> starts.
+        /// </summary>
+        internal TimeSpan ContextMinInterval { get; private set; }
```
limebell, line 142: I cannot find usage of this value, how about using `private` instead?

limebell (approved):
LGTM

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -327,7 +327,24 @@ private void OnTipChanged(object? sender, (Block<T> OldTip, Block<T> NewTip) e)
             Task.Run(
                 async () =>
                 {
-                    await Task.Delay(_newHeightDelay, _newHeightCts.Token);
+                    var startTime = _contexts.ContainsKey(e.NewTip.Index)
+                        ? _contexts[Height].StartTime
+                        : DateTimeOffset.MinValue;
+                    var contextIntervalLacked = _contextMinInterval
+                        - (DateTime.UtcNow - startTime);
+                    if (contextIntervalLacked.Ticks > 0)
```
greymistcube, line 335:
```suggestion
                    DateTimeOffset startTime = _contexts.ContainsKey(e.NewTip.Index)
                        ? _contexts[Height].StartTime
                        : DateTimeOffset.MinValue;
                    TimeSpan contextIntervalLacked = _contextMinInterval
                        - (DateTime.UtcNow - startTime);
                    if (contextIntervalLacked > TimeSpan.Zero)
```

greymistcube: Eh, its been merged already. 🙃

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -327,7 +327,24 @@ private void OnTipChanged(object? sender, (Block<T> OldTip, Block<T> NewTip) e)
             Task.Run(
                 async () =>
                 {
-                    await Task.Delay(_newHeightDelay, _newHeightCts.Token);
+                    var startTime = _contexts.ContainsKey(e.NewTip.Index)
+                        ? _contexts[Height].StartTime
+                        : DateTimeOffset.MinValue;
```
greymistcube, line 332: Highly unlikely, but this could cause an unexpected exception due to (lack of) concurrency.
Using `_contexts.TryGetValue()` would be more apt here.

greymistcube: Or we could just use context `lock` like everywhere else. 😗

OnedgeLee: Seems better to use `TryGetValue` here, to minimize `lock`. Thanks.

PR #3067, opened 2023-04-06
🧹 Remove mining

greymistcube:
Mining code was left in place just in case we wanted to test PoW backward compatibility.
However, realistically speaking, there seems to be no easy way to make a valid PoW chain with current codebase.
Probably an easier approach is to use a prior version to make a PoW chain fixture and import it if we really want to test the backward compatibility feature. 🙄


riemannulus (approved):
🚀

PR #3074, opened 2023-04-10
Preserve order of `Transaction<T>.UpdatedAddresses`

dahlia:
Previously, `Transaction<T>.UpdatedAddresses` in serialized form should have be ordered in a certain way, which depends on `ImmutableHashSet<T>`'s internals.  If a transaction payload had multiple updated addresses and they were ordered in another way, it failed to verify the signature.

This patch fixes it by letting `Transaction<T>.UpdatedAddresses` preserve the order regardless of the way to order it.  Note that it still does not allow duplicates though.  A new internal data structure named `AddressSet` is also introduced.

### DIFF
--- a/Libplanet/Tx/AddressSet.cs
+++ b/Libplanet/Tx/AddressSet.cs
@@ -1,0 +1,205 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics.Contracts;
+using System.Linq;
+
+namespace Libplanet.Tx
+{
+    /// <summary>
+    /// Order-preserving set of <see cref="Address"/>es.
+    /// </summary>
+    internal class AddressSet : IImmutableSet<Address>, IEquatable<AddressSet>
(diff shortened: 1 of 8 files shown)


limebell (dismissed):

Libplanet/Tx/AddressSet.cs
```diff
@@ -0,0 +1,212 @@
...
+                ? o
+                : other.ToImmutableHashSet();
+            return operand.IsSupersetOf(_addresses);
+        }
+
+        /// <inheritdoc cref="IImmutableSet{T}.IsSupersetOf(IEnumerable{T})"/>
+        [Pure]
+        public bool IsSupersetOf(IEnumerable<Address> other)
+        {
+            // TODO: test
```
tkiapril, line 179: I think you forgot to remove the TODO: test commands. 😋

dahlia: Amended!

tkiapril (approved):

PR #3077, opened 2023-04-11
♻️ `BlockChain<T>.ProposeBlock()` overhaul

greymistcube:
Follow-up on #3072.

Notable changes:
- Removed `Timestamp` parameter from `ProposeBlock()`. It might be nice to have, but isn't being used anywhere. Besides, there is no easy path to assign custom timestamps for a running node.
- ~~Added `ProposeBlock()` that takes a list of `Transaction<T>`s to propose a somewhat custom `Block<T>`.~~
  - On second thought, made such method `internal` only for testing.
  
Addendum:
We can also probably remove `BlockCommit` and `IComparer<Transaction<T>>`. 🙄


OnedgeLee (approved):
LGTM

PR #3083, opened 2023-04-13
♻️ Refactor `TxActionList` encoding

greymistcube:
My own general design opinions:
- Owning class generally should not care about *how* its member class is encoded.
- Similarly, a member class should not care about what its property name (dictionary key?) should be from the perspective of an owning class.

I have no idea whether `JsonConverters` still works with this change. 🙄

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,12 @@ To be released.
 
 ### Backward-incompatible API changes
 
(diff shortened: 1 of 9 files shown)


longfin (dismissed):

greymistcube:
Merged with a stale approval after a rebase. 😗

PR #3084, opened 2023-04-13
@planetarium/account: `Account.getAddress()`

tkiapril:
Now, `Account` has `getAddress()` async method that can get the address of the given account directly. Also, `Web3KeyStore` now uses `Web3KeyMetadata` which contains `address` field.


riemannulus (approved):

dahlia (approved):

Akamig (approved):
Now 'Account' seems taking it's shape!

PR #3094, opened 2023-04-17
Fixed interval

OnedgeLee:
On https://github.com/planetarium/libplanet/pull/3066, https://github.com/planetarium/libplanet/pull/3063,  https://github.com/planetarium/libplanet/pull/3064 has been reverted, since those would be permanently break time sync between validators.

Using the timing on tip changed, this harmful aspect can be resolved.

Still it's imperfect due to the state transition time, but to resolve it, block appending on consensus is needed.

I think those direction is OK, but it's not easy to decide.
So, to minimize difference, decided to use this way.


OnedgeLee:
Has some problem on logic, so discarding.

Issue #3103, opened 2023-04-19
Make `ConsensusContext` not to asynchronous


OnedgeLee:
Seems like asynchronous part on `ConsensusContext` can be moved to `Context.Async`.

PR #3104, opened 2023-04-20
Move delay to `EndCommit` from `OnTipChanged`

OnedgeLee:
### Rationale
1. `ConsensusContext` doesn't have to be asynchronous, but it's asynchronous only due to delay before `NewHeight`.
2. For now, consensus waits after block has been committed, so `BlockCommit` only have `Vote`s before waiting period while appending.
3. Timing will be synced when validators comes to `EndCommit` step, and block appending will be executed immediately without state transition, so it doesn't matter if we move waiting period to before appending. (Timing of validators' `OnTipChanged` still will be synced)
4. If we can sync timing of `OnTipChanged`, it seems more natural not to wait after block has been committed.

### Changes
- Move `newHeightDelay` to `EndCommit` step, to be wait before block has been committed.

### Trivials
- This can be done since action evaluation caching(#3028) has been merged already.

### Related PRs
- https://github.com/planetarium/libplanet/pull/3028

### Related issues
- https://github.com/planetarium/libplanet/issues/3103

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,10 @@ To be released.
  -  (Libplanet.Net) Removed optional `render` parameter from
     all `Swarm<T>.PreloadAsync()` overload methods.  No rendering is done
     during the preloading phase.  [[#3108]]
+ -  (Libplanet.Net) Removed `newHeightDelay` parameter from constructor of
+    `ConsensusReactor<T>`, `ConsensusContext<T>`.  [[#3104]]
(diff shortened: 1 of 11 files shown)


OnedgeLee:
Closing this PR, since there are some vulnerability.
If some validator want to manage commit, they can append block before delay has been passed, and we cannot prevent this kind of action from affecting to other nodes.
So, purpose of this PR, especially `Rationale 2` cannot be achieved under trustless condition.

PR #3111, opened 2023-04-25
More generic action type attribute

greymistcube:
This is mainly in preparation to get rid of a distinction between system actions and custom actions. 😗

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,12 @@ To be released.
         `IImmutableList<IValue>`.
  -  `TxActionList`'s JSON representation has changed.  It no longer has
     `"type"` field.  [[#3110]]
+ -  Changed the type for `ActionTypeAttribute.TypeIdentifier`
+    from `string` to `IValue`.  [[#3111]]
+ -  Changed the return type for `ActionTypeAttribute.ValueOf()` from
+    `string?` to `IValue?`.  [[#3111]]
+ -  Changed the return type for `IActionTypeLoader()` from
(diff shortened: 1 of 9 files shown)


Libplanet/Action/PolymorphicAction.cs
```diff
@@ -243,7 +243,7 @@ public string GetInnerActionTypeName()
 
         public void LoadPlainValue(Dictionary plainValue)
         {
-            var typeStr = (Text)plainValue["type_id"];
+            var typeStr = plainValue["type_id"];
```
moreal, line 246: It doesn't have to be updated in this pull request, the variable name can be renamed into another like `typeId` or.

CHANGES.md
```diff
@@ -34,6 +34,12 @@ To be released.
         `IImmutableList<IValue>`.
  -  `TxActionList`'s JSON representation has changed.  It no longer has
     `"type"` field.  [[#3110]]
+ -  Changed the type for `ActionTypeAttribute.TypeIdentifier`
+    from `string` to `IValue`.  [[#3111]]
+ -  Changed the return type for `ActionTypeAttribute.ValueOf()` from
+    `string?` to `IValue?`.  [[#3111]]
+ -  Changed the return type for `IActionTypeLoader()` from
```
moreal, line 41: You may intend:

```suggestion
 -  Changed the return type for `IActionTypeLoader.Load(IActionContext)` from
```

or

```suggestion
 -  Changed the return type for `IActionTypeLoader.Load()` from
```

greymistcube: Suggestions applied! 😊

moreal (approved):

PR #3116, opened 2023-04-26
♻️ 🧹 Remove distinction between system action and custom action

greymistcube:
Notable changes:
- So called "system action" types are no longer treated as special.
  - `Transaction<T>` now only has `Actions` which can hold any number and any combination of "system action"s and "custom action"s
  - The rationale behind this is to make `Transaction<T>` know less about `IAction` in general. Having a separate API for "system action"s and "custom action"s meant `Transaction<T>` had to know about the distinction between the two.
  - `Transaction<T>.Create(IEnumerable<IAction>)` simply calls `IAction.PlainValue` on each `IAction`.
- Now `Transaction<T>.Create(IEnumerable<IAction>)` is now purely for convenience as there is no way to get `IAction` back solely from `Transaction<T>` due to removal of `Transaction<T>.SystemAciton` and `Transaction<T>.CustomActions`. You can only retrieve `IValue`s once `Transaction<T>` is created.
- JSON representation of actions field has been simplified accordingly:

Before:
```json
{
    "actions": {
         "customActions" : [
             {
                 "type_id": "foo",
                 "values": ["bar", 0]
             },
             [1, "baz"]
         ]
    }
}
```

After:
```json
{
     "actions": [
         {
             "type_id": "foo",
             "values": ["bar", 0]
         },
         [1, "baz"]
    ]
}
```

After this, I think we can easily remove `<T>`, which will be handled in a follow-up PR.

Todo:
- As `Transaction<T>` no longer instantiates `IAction`s on deserialization from the data retrieved from the network, this can cause a problem when a malformed serialized `Transaction<T>` is received. In particular, `ActionEvaluator` can break when `Transaction<T>.Actions` are actually instantiated for evaluation. This must be handled within the same release.


dahlia (changes requested):
TypeScript tests are failing due to changes on JSON representation of actions.  These tests should be updated together.

longfin (dismissed):

greymistcube:
@dahlia 
Fixed the tests. Probably need to change the underlying structure for TypeScript to reflect the changes in libplanet at some point tho.

dahlia (approved):
We should postpone the next release until *@planetarium/tx* is reshaped in the same way to this change.  /cc @tkiapril

PR #3125, opened 2023-05-02
Update txexecution

limebell:
Fixed a bug where `BlockChain<T>.Append()` hadn't update tx executions even `evaluateActions` set to `true` when `actionEvaluations` are given.

### DIFF
--- a/Libplanet/Blockchain/BlockChain.cs
+++ b/Libplanet/Blockchain/BlockChain.cs
@@ -1158,21 +1158,24 @@ namespace Libplanet.Blockchain
                 _rwlock.EnterWriteLock();
                 try
                 {
-                    if (evaluateActions && actionEvaluations is null)
+                    if (evaluateActions)
                     {
-                        _logger.Information(
-                            "Executing actions in block #{BlockIndex} {BlockHash}...",
-                            block.Index,
-                            block.Hash);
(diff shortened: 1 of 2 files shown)


Libplanet/Blockchain/BlockChain.cs
```diff
@@ -1158,13 +1158,17 @@ public BlockCommit GetBlockCommit(long index)
...
-                        ValidateBlockStateRootHash(block, out actionEvaluations);
+                        if (actionEvaluations is null)
+                        {
+                            _logger.Information(
+                                "Executing actions in block #{BlockIndex} {BlockHash}...",
+                                block.Index,
+                                block.Hash);
+                            ValidateBlockStateRootHash(block, out actionEvaluations);
+                        }
+
```
OnedgeLee, line 1171: Won't logging `Executed actions in block...` on L1176 better to be moved?
Seems a bit weired on cached situation to see those log, without `Executiong actions...` log on L1165.

OnedgeLee (approved):
LGTM

Libplanet/Blockchain/BlockChain.cs
limebell: Ok, seems reasonable.

PR #3135, opened 2023-05-03
♻️ Changed `IActionTypeLoader` to `IActionLoader`

greymistcube:
This also moves the responsibility of loading `IAction`s from `IActionEvaluator` to `IActionLoader`. 😗


moreal (approved):

CHANGES.md
```diff
@@ -74,6 +74,10 @@ To be released.
      -  Changed `PreEvaluationBlock<T>` to `PreEvaluationBlock`.
          -  `PreEvaluationBlock.Evaluate<T>()` now requires type parameter `T`.
  -  Removed `PreEvaluationBlock.Evaluate<T>()` method.  [[#3127]]
+ -  Renamed `IActionTypeLoader` to `IActionLoader`.  [[#3135]]
+     -  Added `IActionLoader.Load(long, IValue)` interface method.
+     -  Removed `ActionTypeLoaderContext` class.  Use `long` instead.
```
moreal, line 79: Originally, the `ActionTypeLoaderContext` class has internal access modifier. And I guess you wanted to notice that argument type was changed. If then, it will be better to say in other way 🧐

greymistcube: Hmm... It is odd then in the first place for `public` interface methods for `IActionTypeLoader` to have had `ActionTypeLoaderContext` as parameter. 🙄

Issue #3134, opened 2023-05-03
Lazy validator that has been skipped `PreCommit` step cannot commit to `Block`


OnedgeLee:
Since [action evaluation validation](https://github.com/planetarium/libplanet/blob/51a95183937a76829f5166d4169c5b9fd20f6fc2/Libplanet.Net/Consensus/Context.Mutate.cs#L127) takes time, `PreCommit` step can be skipped, since +2/3 `PreCommit` votes can be collected during this validation time and moved directly to `EndCommit` step. 
On that condition, proposer cannot broadcast `PreCommit` on that round, since it will be blocked with [`EndCommit` no-op](https://github.com/planetarium/libplanet/blob/51a95183937a76829f5166d4169c5b9fd20f6fc2/Libplanet.Net/Consensus/Context.Mutate.cs#L110).

PR #3137, opened 2023-05-04
Fix mutation to be consumed sequentially

OnedgeLee:
### Changes
- Mutation does not produce subsequent mutation any more.
- Instead, mutation will handle subsequent mutation with iteration.
- So, mutation produced from later messages will be consumed after every subsequent mutations has been consumed.
- This will secure `upon rules` to be performed as expected order.

### Related issue
- Resolves https://github.com/planetarium/libplanet/issues/3134

### Restriction
- `TimeoutPreVote` and `TimeoutPreCommit` can be called on inaccurate time, due to `IsValid` operation.
- But, I don't think it's so harmful, and negligible.
- According to PBFT spec, `upon rules` have to be executed `atomically`, so this seems to be a part of spec rather than restriction.
- This restriction exists from 1.0.0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -98,6 +98,9 @@ To be released.
     `IRenderer<T>.RenderReorgEnd()`, `IActionRenderer<T>.UnrenderAction()`,
     and `IActionRenderer<T>.UnrenderActionError()`, i.e., these interface
     methods are no longer invoked by a `BlockChain<T>`.  [[#3087]]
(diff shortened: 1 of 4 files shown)


greymistcube (dismissed):
No changelog? 😶

limebell (dismissed):

OnedgeLee:
> No changelog? 😶

Forgot to add. I'll add some before merging it.

PR #3136, opened 2023-05-04
🔧 Filter malformed `Transaction`s when proposing `Block`s

greymistcube:
There should be an accompanying PR to filter malformed `Block`s containing malformed `Transaction`s received from the network.

PR #3140, opened 2023-05-08
🔧 Filter malformed `Block`s from `BlockChain<T>.Append()` and `Context<T>.IsValid()`

greymistcube:
A companion PR to #3136.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -79,6 +79,9 @@ To be released.
      -  Removed `ActionTypeLoaderContext` class.  Use `long` instead.
      -  Renamed `StaticActionTypeLoader` to `StaticActionLoader`.
  -  Added `IActionEvaluator.IActionLoader` property.  [[#3136]]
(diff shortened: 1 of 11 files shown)


CHANGES.md
```diff
@@ -79,6 +79,8 @@ To be released.
      -  Removed `ActionTypeLoaderContext` class.  Use `long` instead.
      -  Renamed `StaticActionTypeLoader` to `StaticActionLoader`.
  -  Added `IActionEvaluator.IActionLoader` property.  [[#3136]]
+ -  Changed `IActionLoader.LoadAction()` to throw `InvalidActionException`
+    instead of `ArgumentException`.  [[#3140]]
```
limebell, line 83:
```suggestion
    instead of `ArgumentException` when the action cannot be instantiated.  [[#3140]]
```
How about describe detail of the situation when this exception occurs.

limebell (approved):

OnedgeLee (approved):

PR #3148, opened 2023-05-10
Refactor `IActionLoader`

greymistcube:
Mostly in preparation for removing `T` and `PolymorphicAction<T>`.


Libplanet/Action/Loader/TypedActionLoader.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <param name="assembly">The <see cref="Assembly"/> to search <see cref="IAction"/> types.
+        /// </param>
+        /// <param name="baseType">The type of <see cref="IAction"/> to filter.</param>
+        /// <returns>A <see cref="TypedActionLoader"/> created from provided
+        /// <paramref name="assembly"/> and <paramref name="baseType"/>.</returns>
+        /// <exception cref="ArgumentException">Thrown when both <paramref name="assembly"/> and
+        /// <paramref name="baseType"/> are <see langword="null"/>.
+        /// </exception>
+        /// <seealso cref="TypedActionLoader(IDictionary{IValue, Type})"/>
+        public static TypedActionLoader Create(Assembly? assembly = null, Type? baseType = null)
```
longfin, line 80: Even in `TypedActionLoader`, should we care about `baseType`?

Libplanet/Action/Loader/SimpleActionLoader.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <exception cref="ArgumentException">Thrown when given type <paramref name="type"/>
+        /// is not a <see langword="class"/> or cannot be assigned to <see cref="IAction"/>.
+        /// </exception>
+        public SimpleActionLoader(Type type)
+        {
+            if (!type.IsClass)
+            {
+                throw new ArgumentException(
+                    $"Given {nameof(type)} must be a class: {type}", nameof(type));
+            }
```
longfin, line 32: Despite of below check(`typeof(IAction).IsAssignableFrom(type)`), does this check needed?

Libplanet.Tests/Action/StaticActionLoaderTest.cs
```diff
@@ -0,0 +1,38 @@
+using Bencodex.Types;
+using Libplanet.Action;
+using Libplanet.Action.Loader;
+using Libplanet.Tests.Common.Action;
+using Xunit;
+
+namespace Libplanet.Tests.Action
+{
+    public class StaticActionLoaderTest
```
longfin, line 9:
```suggestion
    public class SimpleActionLoader
```

`SingleActionLoader` also may be intuitive.

Libplanet/Action/Loader/TypedActionLoader.cs
greymistcube: Here `baseType` is optional. This serves several purposes.

- `PolymorphicAction<T>` had to have **something** for `T`, and this allows to mimic the previous `PolymorphicAction<T>`'s behavior (before `StaticActionLoader` was introduced).
- Such behavior can still act as a defensive mechanism to filter types (just in case there are some unwanted attributed `IAction`s lying around in an assembly?).
- If unwanted, it can simply be ignored. 😗

Libplanet/Action/Loader/SimpleActionLoader.cs
greymistcube: Welp, for one thing, `typeof(IAction)` is assignable to `typeof(IAction)` (i.e. getting rid of it allows something like `new SingleActionLoader(typeof(IAction))`), which is I think is non-sensical. 😅

Only allowing `class` but not `struct` might be a little arbitrary tho. 🤔

Libplanet.Tests/Action/StaticActionLoaderTest.cs
greymistcube: Actually, it was originally named `SingleActionLoader`. 😅
I've changed it back to `SingleActionLoader`.

riemannulus (dismissed):

riemannulus:
https://app.circleci.com/pipelines/github/planetarium/libplanet/8197/workflows/18ff94d5-3452-4a82-9da1-bd03717e377f/jobs/41838 Seems this is not a just CI bug. Could you check this problem? @greymistcube

greymistcube:
@riemannulus 
Currently fixing it 😥

PR #3145, opened 2023-05-09
[Experimental, WIP] Remove `T` from `BlockChain`

longfin:
- Remove `StaticActionLoader.Create<T>()` call in `BlockChain<T>.Create()`. (#3149)
- Remove `T` from `IRenderer` and its implementations. (#3147 )
- [x] Remove `StaticActionLoader.Create<T>()` call in `BlockChain<T>.EvaluateGenesis()`. 
- [ ] Remove `T` from `IBlockPolicy`, `IStagePolicy`, and `BlockChain<T>` itself.

PR #3149, opened 2023-05-10
Remove `StaticActionLoader.Create<T>()` from `BlockChain<T>`

longfin:
This PR is a subset of #3145 (like #3147). it removes the `StaticActionLoader.Create<T>()` call in `BlockChain<T>.Create()` by requiring `ActionEvaluator` explicitly.

And also, it removes `trieGetter` parameter and adds `GetTrie()` interface method to `IBlockChainStates`  since `trieGetter` has been introduced to fetch state trie by block offset, and it seemed to in  `IBlockChainStates`'s interest. (confirmation needed)

Related PRs

- https://github.com/planetarium/lib9c/pull/1893
- https://github.com/planetarium/NineChronicles.Headless/pull/2022
- https://github.com/planetarium/NineChronicles/pull/2633


longfin:
`BlockChain<T>.EvaluateGenesis()` also can matter... 🤔

longfin:
> `BlockChain<T>.EvaluateGenesis()` also can matter... 🤔

it will be addressed different PR for the sake of length.

moreal:
> And also, it removes `trieGetter` parameter and adds `GetTrie()` interface method to `IBlockChainStates` since `trieGetter` has been introduced to fetch state trie by block offset, and it seemed to in `IBlockChainStates`'s interest. (confirmation needed)

@longfin 

To summarize, I don't think `GetTrie()` should be in `IBlockChainStates`. I think `ITrie` is a detailed implementation for storing state, and `IBlockChainStates` is a higher layer that abstracts the raw state well with Balance, State, ValidatorSet, TotalSupply, etc. So I don't think it's right to expose `ITrie` in `IBlockChainStates`.

Now, in this PR, I understand that the reason we need `IBlockChainStates.GetTrie()` is to get rid of the `trieGetter` argument in `ActionEvaluator.ctor(...)`. So, while I didn't think much about it before this PR, I would like to talk about the usefulness of `trieGetter` (i.e., `IActionContext.PreviousStateRootHash`). The `trieGetter` is used inside the `ActionEvaluator` to get an `ITrie` object called `previousBlockStatesTrie`, and the `previousBlockStatesTrie` is used to create an `ActionContext` object during the action evaluation process, and the `ActionContext` is used to get the `IActionContext. PreviousStateRootHash`, which is then utilized to provide the `IActionContext.PreviousStateRootHash`, which is then utilized to log errors in the execution of the action. Alternatively, the `IAction` implementation could also utilize it, but this is not yet the case in the Lib9c example. The `ActionEvaluator` shows the `IActionContext.PreviousStateRootHash` when the `IAction` throws an `Exception` during execution. While having an `IActionContext.PreviousStateRootHash` would make redoing easier, throwing an `Exception` is not a situation where you need to debug state computation through redoing.

How do you think about it? Can `trieGetter` be just removed?

longfin:
> > And also, it removes `trieGetter` parameter and adds `GetTrie()` interface method to `IBlockChainStates` since `trieGetter` has been introduced to fetch state trie by block offset, and it seemed to in `IBlockChainStates`'s interest. (confirmation needed)
> 
> @longfin
> 
> To summarize, I don't think `GetTrie()` should be in `IBlockChainStates`. I think `ITrie` is a detailed implementation for storing state, and `IBlockChainStates` is a higher layer that abstracts the raw state well with Balance, State, ValidatorSet, TotalSupply, etc. So I don't think it's right to expose `ITrie` in `IBlockChainStates`.
> 
> Now, in this PR, I understand that the reason we need `IBlockChainStates.GetTrie()` is to get rid of the `trieGetter` argument in `ActionEvaluator.ctor(...)`. So, while I didn't think much about it before this PR, I would like to talk about the usefulness of `trieGetter` (i.e., `IActionContext.PreviousStateRootHash`). The `trieGetter` is used inside the `ActionEvaluator` to get an `ITrie` object called `previousBlockStatesTrie`, and the `previousBlockStatesTrie` is used to create an `ActionContext` object during the action evaluation process, and the `ActionContext` is used to get the `IActionContext. PreviousStateRootHash`, which is then utilized to provide the `IActionContext.PreviousStateRootHash`, which is then utilized to log errors in the execution of the action. Alternatively, the `IAction` implementation could also utilize it, but this is not yet the case in the Lib9c example. The `ActionEvaluator` shows the `IActionContext.PreviousStateRootHash` when the `IAction` throws an `Exception` during execution. While having an `IActionContext.PreviousStateRootHash` would make redoing easier, throwing an `Exception` is not a situation where you need to debug state computation through redoing.
> 
> How do you think about it? Can `trieGetter` be just removed?

@moreal 
Thanks for the comment. it seems that there are two topics.

1. About `IBlockChainStates.GetTire()`

As you said, `ITrie` is too specific to handle from `IBlockChainStates`. but the root hash of the previous states (as the digest of whole states) might be still affordable, from both the abstraction and usage perspectives. so how about `IBlockChainStates.GetStateRootHash()` (or `.GetStateDigest()`)?

2. About `IActionContext.PreviousStateRootHash`

I guess the previous motivation of #1030 seems still valid for debugging purposes. but it doesn't fit `IActionContext`'s intention (i.e., provide execution context to action code), as you said. 
so I agree to remove `PreviousStateRootHash` from `IActionContext`, but I don't have strong confidence much that about against `ActionEvaulator`.

moreal:
> @moreal Thanks for the comment. it seems that there are two topics.
> 
>     1. About `IBlockChainStates.GetTire()`
> 
> 
> As you said, `ITrie` is too specific to handle from `IBlockChainStates`. but the root hash of the previous states (as the digest of whole states) might be still affordable, from both the abstraction and usage perspectives. so how about `IBlockChainStates.GetStateRootHash()` (or `.GetStateDigest()`)?
> 
>     2. About `IActionContext.PreviousStateRootHash`
> 
> 
> I guess the previous motivation of #1030 seems still valid for debugging purposes. but it doesn't fit `IActionContext`'s intention (i.e., provide execution context to action code), as you said. so I agree to remove `PreviousStateRootHash` from `IActionContext`, but I don't have strong confidence much that about against `ActionEvaulator`.

1. I'm not sure what `.GetStateDigest()` will return. Could you write a pseudo-code for the return type?
2. For the `.GetStateRootHash()`, I imagine it can be used like `GetStateRootHash(IPreEvaluationBlock.PreviousHash)` but it cannot make `IActionContext.PreviousStateRootHash` with the state root hash. (because `IActionContext.PreviousStateRootHash` is `MerkleRootHash(PreviousBlockStates + IActionContext.PreviousStates)`)

longfin:
> I'm not sure what .GetStateDigest() will return. Could you write a pseudo-code for the return type?

Just thought like as below

```csharp
interface IBlockChainStates 
{
    HashDigest<SHA256>? GetStateRootHash(BlockHash offset);
}

class BlockChainStates
{
    public HashDigest<SHA256>? GetStateRootHash(BlockHash offset) => _store.GetBlockDigest(offset)?.StateRootHash;
}
```

it seems just the wrapper of `IStore.GetBlockDigest()` and `BlockDigest.StateRootHash`, but can be a useful abstraction for such as `RemoteBlockChainStates`.

moreal:
> I don't have strong confidence much that about against ActionEvaulator.

If the signature of `GetStateDigest()` (or `GetStateRootHash()`)  is the same as the code, the `PreviousStateRootHash` cannot be calculated as I said.

And, about the `IBlockChainStates.GetStateRootHash()`, I think it can be introduced but not a strong opinion. (because Libpalnet's block has `StateRootHash` field and it assumes the chain will data structures like merkle trie.)

longfin:
> because Libpalnet's block has StateRootHash field and it assumes the chain will data structures like merkle trie

Right, since `ActionEvalautor` already can access `Block` during evaluation. 😅

longfin:
> but it cannot make IActionContext.PreviousStateRootHash with the state root hash. (because IActionContext.PreviousStateRootHash is MerkleRootHash(PreviousBlockStates + IActionContext.PreviousStates))

OK, I understood as below

1. We needed the intermediate state root hash, not just on the previous block.
2. To calculate 1, we need `ITrie`, not `HashDigest<SHA256>`.
3. To get`ITrie` in `ActionEvaluator`, we need `IStateStore`, or adding `IBlockStates.GetTrie()`.

Hence there're 4 choices
 
1. Add `IBlockStates.GetTrie()`. (like PR)
2. Drop the features requiring `ITrie` in `ActionEvaluator`. (suggested by @moreal)
3. Inject `IStateStore` instead of `IBlockStates` to `ActionEvaluator`.
4. Keep `trieGetter`delegate as previous.

I'm considering about 3 or 4. but 2 become more attractive to me. 😓

greymistcube (approved):
Eh, better to have this merged and then remove `ITrie` access if we really want to.

greymistcube (approved):

longfin:
Regarding `IBlockChainStates.GetTrie()`, it seems that there is already a direction in which some consensus has been reached, but there are follow-up PRs scheduled, so we will cover it there. cc @moreal @greymistcube

longfin:
Merged without squashing since we may need to refer separate commits in this PR. (like https://github.com/planetarium/libplanet/pull/3149/commits/a855d0069466bb1943ed3cf635f74fea93def86b)

PR #3152, opened 2023-05-11
Periodic `MessageId` handling in `Gossip`

limebell:
This PR delays reply of `HaveMessage`, to remove total amount of message in the network.

You may concern that this patch may delay the network, but gossip broadcasts messages via `Publish()` method at first, and than gossips through `HaveMessage` if any peer fails to receive published message. If a peer failed to receive published message, it means the message is already delayed, so this PR does not significantly affect delay.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,11 +18,18 @@ To be released.
 
 ### Behavioral changes
 
+ -  `Gossip` became to store the `MessageId`s received through the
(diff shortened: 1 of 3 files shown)


riemannulus (commented):
IMO we add these changes on `Behavioral changes`

riemannulus (approved):

PR #3156, opened 2023-05-12
Introduce: Fee and Gas field on `ITxInvoice`

riemannulus:
## Rationale
https://github.com/planetarium/libplanet/discussions/3119

## Abstract
Introduce `GasLimit` and `MaxGasFee` on `ITxInvoice` for use in the `IActionEvaluator`


CHANGES.md
```diff
@@ -43,6 +43,10 @@ To be released.
     [[#3144]]
  -  Added `IActionContext.GasLimit()` method.
     [[#3144]]
+ -  Added `ITransaction.MaxGasPrice` property.
+    [[#3156]]
+ -  Added `ITransaction.GasLimit` property.
+    [[#3156]]
```
longfin, line 49: Basically, adding interface methods seems to the API breaking changes because its implementation should be updated.

Libplanet/Assets/FungibleAssetValue.cs
```diff
@@ -637,6 +657,10 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
         [Pure]
         public override string ToString() =>
             $"{GetQuantityString()} {Currency.Ticker}";
+
+        public IValue Serialize() => Dictionary.Empty
```
longfin, line 661: If there is no serious backward compatibility concern, I strongly suggest a simple `Bencodex.Types.List` representation, instead of `Bencodex.Types.Dictionary`

Libplanet/Tx/ITxInvoice.cs
```diff
@@ -30,7 +31,7 @@ public interface ITxInvoice : IEquatable<ITxInvoice>
 
         /// <summary>
         /// A <see cref="HashDigest{SHA256}"/> value of the genesis which this transaction is made
-        /// from.  This can be <see langword="null"/> iff the transaction is contained in
+        /// from.  This can be <see langword="null"/> if the transaction is contained in
```
longfin, line 34: I guess that iff isn't a typo. 😅  ([see also](https://en.wikipedia.org/wiki/If_and_only_if))

CHANGES.md
riemannulus: This is all nullable so do not directly affect implementation I guess. but should I mark `Breaking change` on those comments?

Libplanet/Assets/FungibleAssetValue.cs
riemannulus: Does not have a strong opinion about this. I was mocking like other `Serialize()` methods. I'll change it to `Bencodex.Types.List`.

riemannulus:
amended. PTAL 🙏  @longfin @OnedgeLee

PR #3165, opened 2023-05-15
Improve commit() perf on trie related APIs

longfin:
As the title suggests, this PR improves `StateStoreExtensions.Commit()` and `MerkleTrie.Commit()` methods by limiting the size of batch writes.

And more, this PR fixes `ReadOnlySpan<T>` usages in `HashDigest<T>`, to improve the performance when .NET Standard 2.1+.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,11 +18,18 @@ To be released.
 
 ### Behavioral changes
 
+ -  Improved performance of `StateStoreExtensions.Commit()` extension method
+    and `MerkleTrie.Commit()` method.  [[#3165]]
+ -  Improved performance of `HashDigest<T>.DeriveFrom()` static method on
+    .NET Standard 2.1+.  [[#3165]]
+
 ### Bug fixes
 
 ### Dependencies
 
 ### CLI tools
+
+[#3165]: https://github.com/planetarium/libplanet/pull/3165
 
 
 Version 1.4.0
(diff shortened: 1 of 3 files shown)


moreal (approved):

longfin:
I'll reopen this PR after the 1.4 release (hence, it'll be included in 1.5+)

riemannulus (approved):

greymistcube:
If I'm not mistaken, understanding that the scope of `WriteBatch` being limited to `ITrie.Commit()`, seems like this would work as intended, but `WriteBatch` seems a bit hacky as there is already `WriteBatch` for `RocksDB` already. 🙄

Also aside from the format change, could you explain how this improves the performance?
I think it could be that performance improvement might simply be coming from using `Dictionary` inside `WriteBatch` instead of `ConcurrentDictionary` for original `values`. 🤔

I guess from certain perspective this could be thought of as improving code readability? 😶

longfin:
> If I'm not mistaken, understanding that the scope of `WriteBatch` being limited to `ITrie.Commit()`, seems like this would work as intended, but `WriteBatch` seems a bit hacky as there is already `WriteBatch` for `RocksDB` already. 🙄

Agree, could you suggest a better naming?

> Also aside from the format change, could you explain how this improves the performance? I think it could be that performance improvement might simply be coming from using `Dictionary` inside `WriteBatch` instead of `ConcurrentDictionary` for original `values`. 🤔

This PR's main idea is, to limit `ConcurrentDictionary<K, V>`'s size through flushing per threshold. if we maintain values as a vanilla dictionary (either `ConcurrentDictionary<K, V>` and `Dictionary<K, V>`), it means that we need to keep all values on memory while traversing. (of course, in that case, the root cause is long traversal time. but we can't address it right away, this approach can help memory pressures)

greymistcube (approved):

PR #3176, opened 2023-05-17
introduce: IAccount

riemannulus:
## Rationale
We need to own the state by each address for the next step about PoS. (I'll describe this background on the issue later.)

## Describe
This PR is introducing `IAccount`. It will make the use `MPT` more effective.

### DIFF
--- a/Libplanet/State/Account.cs
+++ b/Libplanet/State/Account.cs
@@ -1,0 +1,50 @@
+using System.Security.Cryptography;
+using Bencodex.Types;
+
+namespace Libplanet.State
+{
+    /// <summary>
+    /// An implementation of <see cref="IAccount"/> interface.
+    /// </summary>
+    public class Account : IAccount
+    {
+        public Account(
+            Address id,
+            string? memo,
+            HashDigest<SHA256> stateRootHash)
+        {
+            Id = id;
+            Memo = memo;
+            StateRootHash = stateRootHash;
+        }
+
+        public Account(Bencodex.Types.List serialized)
+        {
+            Nonce = (Integer)serialized[0];
+            Id = new Address(serialized[1]);
+            StateRootHash = new HashDigest<SHA256>((Binary)serialized[2]);
+            Memo = serialized.Count > 3 ? (Text?)serialized[3] : null;
+        }
+
+        public long Nonce { get; }
+
+        public Address Id { get; }
+
+        public string? Memo { get; }
+
+        public HashDigest<SHA256> StateRootHash { get; }
+
+        public IValue Serialize()
+        {
+             var list = Bencodex.Types.List.Empty
+                .Add(Id.Bencoded)
+                .Add(StateRootHash.ByteArray);
+             if (Memo is { } memo)
+             {
+                 list.Add(memo);
+             }
+
+             return list;
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


Libplanet/State/AccountDeltaImpl.cs
```diff
@@ -0,0 +1,41 @@
...
+        public IAccountDelta SetState(Address key, IValue value) =>
+            new AccountDeltaImpl(Account, _trie.Set(ToStateKey(key), value));
+
+        public IAccountDelta SetAccountMemo(string memo) =>
+            new AccountDeltaImpl(new Account(Account.Id, memo, Account.StateRootHash), _trie);
+
+        ITrie IAccountDelta.Commit() => _trie.Commit();
+
+        private KeyBytes ToStateKey(Address address) =>
+            new KeyBytes(ByteUtil.Hex(address.ByteArray), Encoding.UTF8);
```
longfin, line 39:
```suggestion
            new KeyBytes(address.ByteArray);
```

Or, is there any special reason about `ByteUtil.Hex()`?

Libplanet/State/Account.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public string? Memo { get; }
+
+        public HashDigest<SHA256> StateRootHash { get; }
+
+        public IValue Serialize()
+        {
+             var list = Bencodex.Types.List.Empty
+                .Add(Id.Bencoded)
+                .Add(StateRootHash.ByteArray);
```
longfin, line 41:
```suggestion
             var list = new Bencodex.Types.List.List(Id.Bencoded, StateRootHash.ByteArray);
```

Also, I think `Bencodex.Types.List` might be a little bit verbose, despite its intention (i.e., differ from `List<T>`). 🤔  (we may need to discuss as other issue than this PR).

Libplanet/Store/StateStoreExtensions.cs
```diff
@@ -105,5 +106,34 @@ IReadOnlyList<string> rawStateKeys
             HashDigest<SHA256> stateRootHash
         ) =>
             stateStore.GetStateRoot(stateRootHash).Recorded;
+
+        public static IAccount GetAccount(
+            this IStateStore stateStore,
+            HashDigest<SHA256>? stateRootHash,
+            string key
```
longfin, line 113: I think that it's okay to require `Address` other than `string`.

Libplanet/Store/StateStoreExtensions.cs
```diff
@@ -105,5 +106,34 @@ IReadOnlyList<string> rawStateKeys
...
+        )
+        {
+            ITrie trie = stateStore.GetStateRoot(stateRootHash);
+            if (trie.Get(new[] { EncodeKey(key) })[0] is Bencodex.Types.List rawAddress)
```
longfin, line 117: I guess It may throw NRE when there is no entity for the given `key`.

longfin (commented):
it would be better that we have test cases for this, to check usage.

Libplanet/Store/StateStoreExtensions.cs
riemannulus: So, should we support `CreateAccount` in `StateStoreExtension`?

Libplanet/State/Account.cs
riemannulus: I agree `Bencodex.Types.List` is too verbose but most of our code base is already written like that. Could we talk about this code convention on the other issue?

PR #3201, opened 2023-06-07
introduce: fee collector

riemannulus:
SSIA


Libplanet/Assets/FungibleAssetValue.cs
```diff
@@ -637,6 +652,10 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
         [Pure]
         public override string ToString() =>
             $"{GetQuantityString()} {Currency.Ticker}";
+
+        public IValue Serialize() => Bencodex.Types.List.Empty
+                .Add(Currency.Serialize())
+                .Add((Bencodex.Types.Integer)RawValue);
```
longfin, line 658:
```suggestion
        public IValue Serialize() => new List(Currency.Serialize(), (Bencodex.Types.Integer)RawValue);
```

longfin (dismissed):

Libplanet/Tx/ITxInvoice.cs
```diff
@@ -39,5 +40,20 @@ public interface ITxInvoice : IEquatable<ITxInvoice>
         /// A list of actions in this transaction.
         /// </summary>
         TxActionList Actions { get; }
+
+        /// <summary>
+        /// The maximum amount of <see cref="FungibleAssetValue"/> that the transaction author is
+        /// willing to pay for the transaction.
+        /// This can be <see langword="null"/> if align the handling of that transaction
+        /// with the <see cref="IBlockMetadata.ProtocolVersion"/>.
```
greymistcube, line 48: Wait, does this mean metering gas is dependent on `IBlockMetadata.ProtocolVersion`? 🤔

Libplanet.Tests/Tx/TxInvoiceTest.cs
```diff
@@ -123,14 +127,20 @@ public void Equality()
...
+                        i == 1 ? null : updatedAddresses,
+                        i == 2 ? (DateTimeOffset?)DateTimeOffset.MinValue : timestamp,
+                        i == 3 ? null : actions,
+                        i == 4 ? (FungibleAssetValue?)null : new FungibleAssetValue(
+                            Currency.Uncapped(
+                                "FOO",
+                                18,
+                                new PrivateKey().ToAddress()),
+                            100),
+                        i == 5 ? (long?)null : 10);
```
greymistcube, line 143: One too many tabs? 😗

Libplanet/Assets/Currency.cs
```diff
@@ -671,9 +671,53 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
         /// <summary>
         /// Serializes the currency into a Bencodex value.
         /// </summary>
+        /// <param name="hash">Whether to serialize the currency for hash calculation.</param>
         /// <returns>The serialized Bencodex value.</returns>
         [Pure]
-        public IValue Serialize()
+        public IValue Serialize(bool hash = false) =>
```
longfin, line 677: - Adding a parameter with a default value also will break binary compatibility (not source compatibility), so we need to mention it.
- default value of `hash` is `false` and it will be redirected `SerializeForState()`, but the previous behavior seems `SerializeForHash()`. is it intended change?

riemannulus: There's a little bit of complexity here. `NineChronicles` was not using `Currency.Serialize` from `libplanet`. Instead, I was creating my own `CurrencyExtensions` to serialize/deserialize. Therefore, in order to read the Currency used by `NineChronicles` from `libplanet`, we had no choice but to bring the method of `NineChronicles` to `libplanet`.

This is a problem caused by `libplanet` not providing `Currency.Serialize()`, and rather, the existing `Currency.Serialize()` is a version that breaks backward compatibility of `NineChronicles`.

riemannulus: So I chose to import the `Serialize` method of `NineChronicles`, and I wanted to delete the existing unused `Currency.Serialize()`, but this required a lot of modification at the code level, so I added a boolean parameter on `Serialize()` for much easier refactoring.

Libplanet/Assets/Currency.cs
```diff
@@ -671,9 +671,53 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
...
+            hash ? SerializeForHash() : SerializeForState();
```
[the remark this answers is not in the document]
riemannulus:
```suggestion
        public IValue Serialize() => SerializeForState();
        public HashDigest<SHA256> Hash => code.Encode(SerializeForHash());
```
Did you suggest code like this?

Libplanet/Assets/Currency.cs
longfin: Well... OK. if we can convince `Currency.Serialize()` hadn't been used outside of anywhere and we want to change its behavior like as `SerlizeForState()`, hiding `SerializeForHash()` and changing its behavior directly without new parameter introducing (i.e., https://github.com/planetarium/libplanet/pull/3201#discussion_r1227843653)  would be better than the current implementation.

riemannulus: > Hiding SerializeForHash() and changing its behavior directly without a new parameter introducing 

I strongly agree with your opinion. I'll fix it.

riemannulus: https://github.com/planetarium/libplanet/pull/3201/commits/2ad90d87437c3fd5d266da40052ef27a41fe3426 amended.

Libplanet/Tx/ITxInvoice.cs
riemannulus: amended. https://github.com/planetarium/libplanet/pull/3201/commits/2ad90d87437c3fd5d266da40052ef27a41fe3426

longfin (approved):

longfin (approved):

PR #3206, opened 2023-06-09
Replace proposal when +2/3 valid votes are collected

limebell:
## Introduction
This PR introduces two new classes `Maj23`, `VoteSetBits` and its related classes to solve following problem.

## Problem

1. There exists 4 validator, where one of them (validator D) is attacker.

2. Validator D is the proposer for the round and height, and validator D sends block X's proposal to validator A, and send Y's proposal to validator B and C.

| |ValidatorA|ValidatorB|ValidatorC|ValidatorD|
|-|-|-|-|-|
|Proposal|X|Y|Y|XY|

3. Each validator broadcasts prevote vote according to received proposal. But validator D sends X's prevote to validator A and B, and sends Y's prevote to validaotr C.

| |ValidatorA|ValidatorB|ValidatorC|ValidatorD|
|-|-|-|-|-|
|Proposal|X|Y|Y|XY|
|PreVote|XYYX|XYYX|XYYY|XYYY|

4. Validator A and B failed to collect +2/3 prevote for block X, and validator C and D collected +2/3 prevote for block Y thus send precommit for block Y.

| |ValidatorA|ValidatorB|ValidatorC|ValidatorD|
|-|-|-|-|-|
|Proposal|X|Y|Y|XY|
|PreVote|XYYX|XYYX|XYYY|XYYY|
|lock|-|-|Y|Y|
|PreCommit|--YY|--YY|--YY|--YY|

5. No any block received +2/3 precommit, round increased.
6. Validator C and D cannot vote to other block that Validator A and B proposed (block X' != X and X' != Y) because block Y is already locked, so block X' cannot be committed.
7. When validator C is proposer, validator C will propose block Y with its valid round. Validator A and B's prevote for such round is `XYYX`, thus will not vote for block Y.
8. Round infinetly increased, block never created.

## Solution
When a validator has block X as the proposal for the round, and somehow detects other peer success to collect vote for block Y that is not same with proposal has, the validator replace proposal from block X's to block Y's.

When a validator success to collect +2/3 votes for a proposal (prevote or precommit), broadcasts `Maj23` message to other validators. If the receiver's proposal does not match with received `Maj23`'s block information, try to collect missing votes for the block using `VoteSetBits` message. Proposal will be replaced if the votes are collected successfully.

## Additional Tasks
- Add a unit test for `Maj23` auto broadcasting.
- Add a unit test for consensus-level scenario test.

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusReactor.cs
+++ b/Libplanet.Net/Consensus/ConsensusReactor.cs
@@ -6,6 +6,7 @@ using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using Libplanet.Blockchain;
+using Libplanet.Consensus;
 using Libplanet.Crypto;
 using Libplanet.Net.Messages;
 using Libplanet.Net.Transports;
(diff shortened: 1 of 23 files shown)


OnedgeLee (commented):
A bit confused for its logic. Could you explain me?
The point confusing is,
"Whether `MessageLog` can receive republished `Votes` from `ConsensusVoteSetBitsMsg`, and verify `Maj23` properly, if duplicated `Votes` are published previously."
For above example, Validator A has already has `vote on X` on its message log, and public key of `vote on X` and `vote on Y` is same. On that condition, `vote on Y` will be rejected, IMHO.

Libplanet.Net/Consensus/ConsensusReactor.cs
```diff
@@ -155,6 +156,80 @@ private void ProcessMessage(MessageContent content)
...
+                                .GetVoteSetBits(
+                                    maj23Msg.Round,
+                                    maj23Msg.Maj23.BlockHash,
+                                    maj23Msg.Maj23.Flag);
+                            var sender = _gossip.Peers.First(
+                                peer => peer.PublicKey.Equals(maj23Msg.ValidatorPublicKey));
+
+                            _gossip.PublishMessage(
+                                new ConsensusVoteSetBitsMsg(voteSetBits),
+                                new[] { sender });
```
OnedgeLee, line 196: If context already has sufficient messages, or on the higher step, we don't need to send `ConesneusVoteSetBitsMsg`, IMHO.

Libplanet.Net/Consensus/Context.Mutate.cs
```diff
@@ -286,6 +343,11 @@ private void ProcessGenericUponRules()
                     PublishMessage(
                         new ConsensusPreCommitMsg(
                             MakeVote(Round, p3.Block.Hash, VoteFlag.PreCommit)));
+
+                    // Maybe need to broadcast periodically?
+                    PublishMessage(
+                        new ConsensusMaj23Msg(
+                            MakeMaj23(Round, p3.Block.Hash, VoteFlag.PreVote)));
```
OnedgeLee, line 350: Do we have to publish `ConesnsusMaj23Msg` only on condition that we have proper proposal of +2/3 votes?
Just for asking, since other implementation doesn't seem to check proposal condition.
I'm not sure which one would be better, and think this implementation is also good, since it can reduce a number of messages.

Not sure whether `Maj23` should be the flag of `Single correct process`, or `Single +2/3 votes`.

OnedgeLee (approved):

PR #3207, opened 2023-06-09
♻️ Streamline various getters for `ActionEvaluator`

greymistcube:
♻️ First step towards a long and arduous journey for overhauling state, context, and delta.

- Strictly speaking, `IBlockChainStates.GetStateRoot()` doesn't have to be exposed at this very moment as it is only used internally by `BlockChainStates`. It is merely a preparation in case we want to retrieve and inject state roots in `ActionEvaluator` and `AccountStateDeltaImpl`.
- `ActionEvaluator` only builds state getters from `IBlockChainStates` for the initial delta. Delta objects takes its previous delta directly to "inherit" previous getters.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,11 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  `IBlockChainStates` interface overhauled.  [[#3207]]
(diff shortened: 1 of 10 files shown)


greymistcube:
Closing temporarily. I'll be splitting this up to two separate PRs.

PR #3214, opened 2023-06-14
♻️ Expand `IBlockChainStates`

greymistcube:
♻️

- This mainly allows `IBlockChainStates` to query with `null` offset returning somewhat sensible default values.
- Removed `IncompleteBlockStatesException` as this no longer well-defined.
  - Previously, this would be thrown if either given `offset` is not in `IStore` **or** the state root hash associated with `offset` is not in `IStateStore`.
    - This was somewhat problematic as "incompleteness" doesn't actually refer to the incompleteness of data (as a well managed `BlockChain` storage), but rather given `offset` is simply not found.
  - Now, as `null` is allowed for `offset`, an `offset` of `null` is now considered to be "in" `IStore` in a sense, and the same goes for the state root associated with `null` state root hash.

This is preparatory work to get rid of special null-getters inside `ActionEvaluator`.


Libplanet/Blockchain/BlockChainStates.cs
```diff
@@ -56,85 +62,112 @@ public BlockChainStates(IStore store, IStateStore stateStore)
...
+            else if (_store.GetStateRootHash(hash) is { } stateRootHash)
+            {
+                if (_stateStore.ContainsStateRoot(stateRootHash))
+                {
+                    return _stateStore.GetStateRoot(stateRootHash);
+                }
+                else
+                {
+                    throw new ArgumentException(
+                        $"Could not find state root {stateRootHash} in {nameof(IStateStore)}.");
```
longfin, line 163: adding `offset` to the exception might be good for investigation.

longfin (dismissed):

Libplanet/Blockchain/BlockChainStates.cs
greymistcube: Suggestion applied. 🙂

longfin (approved):

riemannulus (approved):

PR #3215, opened 2023-06-14
♻️ Remove null getters for `ActionEvaluator`

greymistcube:
Null getters are no longer necessary as expanded API of `IBlockChainStates` in #3214 now return "sensible" values for empty states.

### DIFF
--- a/Libplanet.Tests/Action/ActionEvaluatorTest.cs
+++ b/Libplanet.Tests/Action/ActionEvaluatorTest.cs
@@ -231,14 +231,12 @@ namespace Libplanet.Tests.Action
                     txHash: BlockContent.DeriveTxHash(txs),
                     lastCommit: null),
                 transactions: txs).Propose();
-            IAccountStateDelta previousStates = AccountStateDeltaImpl.ChooseVersion(
-                genesis.ProtocolVersion,
-                ActionEvaluator.NullAccountStateGetter,
-                ActionEvaluator.NullAccountBalanceGetter,
-                ActionEvaluator.NullTotalSupplyGetter,
-                ActionEvaluator.NullValidatorSetGetter,
-                genesis.Miner,
-                ImmutableDictionary<(Address, Currency), BigInteger>.Empty);
+            IAccountStateDelta previousStates =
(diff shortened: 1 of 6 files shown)


Libplanet/State/AccountStateDeltaImpl.cs
```diff
@@ -339,53 +340,118 @@ public IAccountStateDelta SetValidator(Validator validator)
...
+                return new AccountStateDeltaImpl(
+                    delta.GetStates,
+                    delta.GetBalance,
+                    delta.GetTotalSupply,
+                    delta.GetValidatorSet,
                     signer)
                     {
-                        TotalUpdatedFungibles = previousTotalUpdatedFungibles,
+                        TotalUpdatedFungibles = impl.TotalUpdatedFungibles,
                     };
```
riemannulus, line 447: IMO using a `switch statement` is much better than an `if statement` in this usage.

riemannulus (dismissed):

Libplanet.Tests/Action/ActionEvaluatorTest.cs
```diff
@@ -469,21 +440,12 @@ DumbAction MakeAction(Address address, char identifier, Address? transferTo = nu
...
-                block2.ProtocolVersion,
-                accountStateGetter,
-                accountBalanceGetter,
-                totalSupplyGetter,
-                ActionEvaluator.NullValidatorSetGetter,
-                block2.Miner,
-                ImmutableDictionary<(Address, Currency), BigInteger>.Empty);
+                evals1.Last().OutputStates, block2.ProtocolVersion);
+            previousStates = AccountStateDeltaImpl.ChooseSigner(
+                previousStates, block2.Miner);
```
riemannulus, line 448: Plus, is it ok to backward compatibility?

Libplanet/State/AccountStateDeltaImpl.cs
greymistcube: Fixed. 🙂

riemannulus (approved):

PR #3218, opened 2023-06-16
Eager TotalUpdatedFungibles update

longfin:
This PR tweaks the codes maintaining `AccountStateDeltaImpl.TotalUpdatedFungibles`, to care performance even in the extreme case of updating many addresses in one action.

### DIFF
--- a/Libplanet/State/AccountStateDeltaImpl.cs
+++ b/Libplanet/State/AccountStateDeltaImpl.cs
@@ -206,6 +206,8 @@ namespace Libplanet.State
             }
 
             FungibleAssetValue balance = GetBalance(recipient, currency);
+            (Address, Currency) assetKey = (recipient, currency);
+            BigInteger rawBalance = (balance + value).RawValue;
 
             if (currency.TotalSupplyTrackable)
             {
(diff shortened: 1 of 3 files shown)


Libplanet/State/AccountStateDeltaImpl.cs
```diff
@@ -238,6 +242,8 @@ public virtual IAccountStateDelta MintAsset(Address recipient, FungibleAssetValu
             bool allowNegativeBalance = false
         )
         {
+            var sw = new System.Diagnostics.Stopwatch();
```
riemannulus, line 245: Do you think it's necessary?

greymistcube (approved):

Libplanet/State/AccountStateDeltaImpl.cs
longfin: My bad, I'll amend it.

Libplanet/State/AccountStateDeltaImpl.cs
```diff
@@ -238,6 +242,8 @@ public virtual IAccountStateDelta MintAsset(Address recipient, FungibleAssetValu
...
+            sw.Start();
```
longfin, line 246:
```suggestion
```

greymistcube (approved):

PR #3219, opened 2023-06-16
♻️ Expose actual delta for `IAccountStateDelta`

greymistcube:
♻️ As per usual, this is mostly a preparatory work to separate state and delta. Some names have been changed for better consistency, but might change further in a following PR. 🙄


greymistcube:
Dealt in #3245.

PR #3220, opened 2023-06-17
✨ Introduce `IDelta`

greymistcube:
This introduces `IDelta` that actually represents the delta. `IDelta` can be thought more of as a vector where it is completely ignorant of its base state. 😗


greymistcube:
Dealt in #3245

PR #3221, opened 2023-06-19
introduce: `ActionEvaluatorEventHandler`

riemannulus:
## Rationale
We need to provide a network-wide action execution interface that will be executed when a block is executed or ended. For example, F1 Fee Distribution stores the fees collected during Transaction execution in a specific address and then distributes them at the next block execution. Introducing `ActionEvaluatorEventHandler` to provide this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,10 @@ To be released.
 
 ### Added APIs
 
+ -  Added `ActionEvaluatorEventHander` class.  [[#3221]]
+ -  Added `beginBlock` parameter from `ActionEvaluator`.  [[#3221]]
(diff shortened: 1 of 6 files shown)


greymistcube:
Looks pretty nifty, but I'm not a big fan of using an entirely new stack to resolve the issue, as I'm more of the opinion that these pseudo-`IAction`s should actually be `IAction`s. 😥
I understand the current interface does not provide sufficient functionality for what this PR is trying to do so I'll let it slide. 😕

greymistcube:
Also, not that critical at the moment, but the name `ActionEvaluatorEventHandler` sounds confusing to me. 🤔

PR #3222, opened 2023-06-19
♻️ Remove context related data from `IAccountStateDelta` implementation

greymistcube:
♻️ 
- In particular, this removes `AccountStateDeltaImpl.Signer`. 
- Access control for `Mint()` and `Burn()` are still way too awkward.
- `Mint()` and `Burn()` now accepts `IActionContext` as an argument.
  - The rationale being that these methods are meant to be called **inside** `IAction.Execute()` with `IActionContext` as a parameter.
- This is the most substantial API change yet. I expect this to break things quite a bit. 😥


greymistcube:
Moved to another PR.

PR #3223, opened 2023-06-20
♻️ Combine `AccountStateDeltaImpl` and `AccountStateDeltaImplV0`

greymistcube:
This is a companion PR to #3222. This completely remove's `IAccountStateDelta`'s **implicit** dependency on its running context and require `IActionContext` as an argument for those methods that actually require its context. Hence it is no longer necessary to maintain two separate `class`es. I think it was a bad idea to type differentiate `AccountStateDeltaImpl` and `AccountStateDeltaImplV0` in the first place, and even if we needed it, the original design of inheritance made `IAccountStateDelta` very clunky to use inside `IActionEvaluator` and made it very **prone to mistakes**. 😕


greymistcube:
Moved to another PR.

PR #3229, opened 2023-06-22
♻️ Explicit `IActionContext` dependency for `TransferAsset()`

greymistcube:
Among numerous problems with the design of `IAccountStateDelta` (and its implementation `AccountStateDeltaImpl`) is having implicit context dependency. Not only that, dependency implicitness isn't even streamlined. In particular,

- `Block`'s `ProtocolVersion` dependency is handled by `type` differentiation between `AccountStateDeltaImpl` and `AccountStateDeltaImplV0`.
- `Signer` (whatever that is actually supposed to mean) is handled by **hidden** internal property of `AccountStateDeltaImpl` and `AccountStateDeltaImplV0`.

This resolves the former, thus getting rid of the necessity to have two different implementations for `IAccountStateDelta`.


Libplanet/State/IAccountStateDelta.cs
```diff
@@ -136,8 +139,15 @@ public interface IAccountStateDelta : IAccountStateView
...
         /// the <paramref name="allowNegativeBalance"/> option is turned off.</exception>
+        /// <remarks>
+        /// The behavior is different depending on <paramref name="context"/>'s
+        /// <see cref="IActionContext.BlockProtocolVersion"/>.  There is a bug for version 0
+        /// where this may not act as intended.  Such behavior is left intact for backward
+        /// compatibility.
+        /// </remarks>
         [Pure]
         IAccountStateDelta TransferAsset(
+            IActionContext context,
```
riemannulus, line 150: If `IActionContext` need on `TransaferAsset` for ACL, How about `MintAsset` or `BurnAsset`? They are needless `IActionContext`?

greymistcube: That will be dealt with in a separate PR, which is almost ready. 🙃
Coming right up after this gets merged. 😗

greymistcube: Also, although both `TransferAsset` and the `MintAsset` and `BurnAsset` pair depend on `IActionContext`, the way these depend on `IActionContext` are different. As for the former, the dependency is baked in to the implementation `type`, and for the latter, the dependency is baked in to a hidden variable named `Signer`. 😑

riemannulus (approved):

PR #3248, opened 2023-06-26
♻️ API streamlining related to `IAccountStateDelta`

greymistcube:
Mostly housekeeping stuff. 😗

### DIFF
--- a/Libplanet/Tx/TxSuccess.cs
+++ b/Libplanet/Tx/TxSuccess.cs
@@ -41,7 +41,7 @@ namespace Libplanet.Tx
             BlockHash blockHash,
             TxId txId,
             List<List<string>>? actionsLogsList,
-            IImmutableDictionary<Address, IValue?> updatedStates,
+            IImmutableDictionary<Address, IValue> updatedStates,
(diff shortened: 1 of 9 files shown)


Libplanet/Tx/TxSuccess.cs
```diff
@@ -41,7 +41,7 @@ public sealed class TxSuccess : TxExecution, ISerializable
             BlockHash blockHash,
             TxId txId,
             List<List<string>>? actionsLogsList,
-            IImmutableDictionary<Address, IValue?> updatedStates,
+            IImmutableDictionary<Address, IValue> updatedStates,
```
riemannulus, line 44: Is it ok to code-level backward compatibility?

greymistcube: This depends on whether `IAccountStateDelta.SetState()` accepted `null` or not.
- Although we are more stringent with `null`ables these days, if the current `interface` is to be believed, then that should have never been `null`.
- If we dig a little deeper, `ITrie.Set()` throws a `ArgumentNullException` when trying to set `null` as a value.

So, I'd say **probably**, but I'm not entirely sure. 😣

limebell (approved):

PR #3249, opened 2023-06-27
Feature: Introduce VoteSet

limebell:
This PR removes `MessageLog` and introduces `HeightVoteSet` and `VoteSet` class.

Both classes manage votes received in context, but `VoteSet` manages only inner votes, not proposal.
`VoteSet` manages votes and classify by its `PublicKey` and `BlockHash`, whereas `MessageLog` can only keep single message from a certain `PublicKey`.

Since `VoteSet` no longer manage `Proposal`, `Context` will handle it.


CHANGES.md
```diff
@@ -51,6 +55,23 @@ To be released.
...
+ -  (Libplanet.Net) Added `HeightVoteSet` class.  [[#3249]]
+ -  (Libplanet.Net) Added `ConsensusVoteMsg` abstract class.  [[#3249]]
+ -  (Libplanet.Net) Added `InvalidProposalException` class.  [[#3249]]
+ -  (Libplanet.Net) Added `InvalidVoteException` class.  [[#3249]]
+ -  (Libplanet.Net) Added `InvalidMaj23Exception` class.  [[#3249]]
+ -  (Libplanet.Net) Added
+    `Gossip.PublishMessage(MessageContent, IEnumerable<BoundPeer>)` method.
+    [[#3249]]
 
 ### Behavioral changes
```
OnedgeLee, line 77: I think change on `nil` vote expression have to be expressed. (`BlockHash = null` => `BlockHash = default`)

Libplanet.Net/Consensus/Context.Mutate.cs
```diff
@@ -208,20 +287,35 @@ private void ProcessGenericUponRules()
...
+                    _logger.Debug(
+                        "PreCommit nil for the round {Round} because 2/3+ PreVotes " +
+                        "were collected. (context: {Context})",
+                        Round,
+                        ToString());
+                    Step = ConsensusStep.PreCommit;
+                    BroadcastMessage(
+                        new ConsensusPreCommitMsg(MakeVote(Round, default, VoteFlag.PreCommit)));
+                }
+                else if (!(Proposal is null) && !hash3.Equals(Proposal.BlockHash))
```
OnedgeLee, line 304:
```suggestion
                else if (Proposal is { } proposal && !proposal.BlockHash.Equals(hash3))
```

Libplanet.Net/Consensus/HeightVoteSet.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when given <paramref name="voteFlag"/>
+        /// is not either <see cref="VoteFlag.PreVote"/> or <see cref="VoteFlag.PreCommit"/>.
+        /// </exception>
+        public VoteSet GetVoteSet(int round, VoteFlag voteFlag)
+        {
+            RoundVoteSet roundVoteSet;
+
+            // TODO: Check if try-catch is needed for KeyNotFoundException.
+            roundVoteSet = _roundVoteSets[round];
```
OnedgeLee, line 238: Still I think it's better to replace `KeyNotFoundException` with explicit custom exception.

Libplanet.Net/Consensus/VoteSet.cs
```diff
@@ -0,0 +1,402 @@
...
+            {
+                _logger.Information("Failed to create block commit since no +2/3 votes collected");
+                return null;
+            }
+
+            return new BlockCommit(
+                _height, _round, _maj23!.Value, MappedList().ToImmutableArray());
+        }
+
+        internal void AddVerifiedVote(Vote vote)
```
OnedgeLee, line 263: Would it be okay not to rename this method?

Libplanet.Net/Consensus/VoteSet.cs
```diff
@@ -0,0 +1,402 @@
...
+                _votesByBlock[blockHash] = new BlockVotes(blockHash);
+
+                // We'll add the vote in a bit.
+            }
+
+            BlockVotes votesByBlock = _votesByBlock[blockHash];
+
+            // Before adding to votesByBlock, see if we'll exceed quorum
+            BigInteger origSum = votesByBlock.Sum;
+            BigInteger quorum = _validatorSet.TwoThirdsPower + 1;
```
OnedgeLee, line 341: Seems nice, but don't we have to apply same approach to other +2/3 majority?
May be it would be better to `+1` on `ValidatorSet.TwoThirdsPower` etc.
For now, we check `ValidatorSet.TwoThirdsPower < Power to examine`,
but when some decimals were truncated, `ValidatorSet.TwoThirdsPower + 1 <= Power to examine` would be better like this.

OnedgeLee (commented):
Added some comments!

Libplanet.Net/Consensus/HeightVoteSet.cs
limebell: IMO its okay, but I'll add `<exception>` XML docs to the method for clearity.

Libplanet.Net/Consensus/VoteSet.cs
limebell: Do you mean change
```csharp
        public bool HasTwoThirdsAny()
        {
            lock (_lock)
            {
                return Sum > _validatorSet.TwoThirdsPower;
            }
        }
```
to
```csharp
        public bool HasTwoThirdsAny()
        {
            lock (_lock)
            {
                return Sum >= _validatorSet.TwoThirdsPower + 1;
            }
        }
```

OnedgeLee: I think it would be better to modify `ValidatorSet.TwoThirdsPower` to add 1.
I think it's okay to let alone here and change with another PR.

OnedgeLee (dismissed):

OnedgeLee (dismissed):

OnedgeLee (approved):

PR #3258, opened 2023-06-30
Reimplement bootstrapping

OnedgeLee:
Since `GetRandomMsg` has been removed, alternative method is needed for bootstrapping.
New bootstrapping acts like below.

1. When a validator participates in consensus and starts a new round, it broadcasts `ConsensusBootstrapMsg`.
2. Online peers who receives `ConsensusBootstrapMsg` collects proper votes from its `HeightVoteSet`, and replies with `ConsensusVotesRecallMsg`, which contains recalled votes.
3. If a validator receives `ConsensusVotesRecallMSg`, it checks the content of it and add those votes to its `HeightVoteSet`.

- Round of returned `ConsensusVotesRecallMsg` will be `Bootstrap.Round` + 1, if given round is smaller than replyer's round, otherwise, `Bootstrap.Round`.
- `ConsensusVotesRecallMsg` will be ignored, if its round is bigger than receiver's round + 1, to be fitted with spamming prevention logic.

### DIFF
--- a/Libplanet.Net/Consensus/Context.cs
+++ b/Libplanet.Net/Consensus/Context.cs
@@ -291,7 +291,7 @@ namespace Libplanet.Net.Consensus
         public void CatchupWithVotesRecall(VotesRecall votesRecall)
         {
             _logger.Debug(
-                "Catch up consensus with VotesRecall contains {count} votes : {votesRecall}",
+                "Catch up consensus with VotesRecall contains {count} votes: {votesRecall}",
                 votesRecall.Votes.Count,
                 votesRecall);
             foreach (Vote vote in votesRecall.Votes)
--- a/Libplanet.Net/Consensus/Context.cs
+++ b/Libplanet.Net/Consensus/Context.cs
@@ -105,6 +105,7 @@ namespace Libplanet.Net.Consensus
         private Block? _decision;
         private int _committedRound;
         private BlockCommit? _lastCommit;
+        private bool _bootstrapping;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Context"/> class.
@@ -178,6 +179,7 @@ namespace Libplanet.Net.Consensus
             _validRound = -1;
             _decision = null;
             _committedRound = -1;
+            _bootstrapping = false;
             _blockChain = blockChain;
             _codec = new Codec();
             _messageRequests = Channel.CreateUnbounded<ConsensusMsg>();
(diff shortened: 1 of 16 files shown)


Libplanet.Net/Swarm.cs
```diff
@@ -858,7 +858,7 @@ CancellationToken cancellationToken
                     List<byte[]> payloads = blockMessage.Payloads;
                     _logger.Information(
                         "Received {Count} blocks from {Peer}",
-                        payloads.Count,
+                        payloads.Count / 2,
```
limebell, line 861: Is this intended?

Libplanet.Net/Consensus/Context.cs
```diff
@@ -251,6 +253,70 @@ public void Dispose()
...
+        /// It only accepts <see cref="Vote"/>s those rounds are less than <see cref="Round"/> + 1.
+        /// </summary>
+        /// <param name="votesRecall">Received <see cref="VotesRecall"/> from peer
+        /// <see cref="Validator"/>.</param>
+        public void CatchupWithVotesRecall(VotesRecall votesRecall)
+        {
+            _logger.Debug(
+                "Catch up consensus with VotesRecall contains {count} votes : {votesRecall}",
+                votesRecall.Votes.Count,
+                votesRecall);
```
limebell, line 296:
```suggestion
            _logger.Debug(
                "Catch up consensus with VotesRecall contains {count} votes: {votesRecall}",
                votesRecall.Votes.Count,
                votesRecall);
```
Just for convension

Libplanet.Net/Consensus/Context.cs
```diff
@@ -251,6 +253,70 @@ public void Dispose()
...
+            foreach (Vote vote in votesRecall.Votes)
+            {
+                if (vote.Round > Round + 1)
```
limebell, line 299: What if consensus is being done in round higher than `Round + 1` ?

Libplanet.Net/Swarm.cs
OnedgeLee: Since payloads contains both of `Block` and `BlockCommit`, for now it represents two times of actual `Block` count. Saparating payload can be considered also, but for now applied this to fix error on logging.

Libplanet.Net/Consensus/Context.cs
OnedgeLee: `VotesRecall` is sent as a reply of `Bootstrap`, and `Votes` in `VotesRecall` are allowed to have only rounds which is same or +1 of requested `Bootstrap.Round`.
So, even though consensus is being done in round higher than `Round + 1`, bootstrapping peer will send lower round of `Bootstrap`, and corresponding `votesRecall` will be responded, and no `Vote`s are expected to be have round higher than `Round + 1`.
If boostrapping peer gives higher round of `Bootstrap`, we don't need to give him votes.

So, rejecting `Vote`s that are higher than `Round + 1` is intended. those votes will be added via subsequent `Bootstrap` and `VotesRecall`.

Libplanet.Net/Swarm.cs
OnedgeLee: Actually, it's not related with this PR. just minor log fix.

OnedgeLee:
Close PR, since this structure is not suitable for newly introduced `HeightVoteSet`.

PR #3267, opened 2023-07-04
🧹 Remove `ActionTypeAttribute.ValueOf()`

greymistcube:
~~Not sure we even need `ValueOf()` as it can be a simple one liner. In any case, I think there would be an infinitesimal amount of performance improvement. 😗~~

Came back after working on [lib9c](https://github.com/planetarium/lib9c) a little. I think having `ValueOf()` just adds more confusion. 😐

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  Removed `ActionTypeAttribute.ValueOf()` method.  [[#3267]]
+
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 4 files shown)


riemannulus:
> Came back after working on [lib9c](https://github.com/planetarium/lib9c) a little. I think having ValueOf() just adds more confusion. 😐

Then this PR is currently in draft?

greymistcube:
@riemannulus I think we can get rid of it. Already removed all references to `ValueOf()` from lib9c.

riemannulus (approved):

PR #3270, opened 2023-07-05
🔧 Remove early termination for tx evaluation

greymistcube:
🔧


greymistcube:
Eh, seems like I made a mistake with the branch. Will reopen.

PR #3291, opened 2023-07-12
🧹 Removed ActionsLogsList from TxExecution

greymistcube:
Reasons for removal:
- Not in real use.
- Its source, `IActionContext.Logs`, is poorly implemented anyway. 🙄

Other than "it might be nice to have logs data flow back to libplanet and be useful somehow", I don't see a strong rationale for keeping it, at least for now.


moreal (approved):
👋🏻 🧹

PR #3304, opened 2023-07-14
🧹 Removed `ExtractableException`

greymistcube:
Rationale:
- Not really used anywhere.
- Somewhat wrongly used for recording `TxFailure`. Current implementation of `TxFailure` kinda assumes that `Tx` will contain only one `IAction`. There is no aggregation of `Exception`s thrown when generating `ActionEvaluation`.
- I'd say it probably would've been better to encourage client developers to use a custom `Exception` provided by libplanet. This is very unintuitive.


moreal (approved):
😥 👋🏻

PR #3328, opened 2023-07-21
🧹 Refactor to use `KeyBytes` instead of `string`s in general

greymistcube:
Removing encoding responsibility from `StateStoreExtensions`.

- It isn't at all clear which entity should actually be responsible for the encoding scheme.
- The general codebase is already severely infected with hardcoded encoding scheme (UTF8).
  - Having `KeyBytes` to be encoding agnostic is nice and all, but in reality, there is no real way to use the library with a different encoding.
  - Technically, `IStateStore` can be responsible for tracking how `string`s are encoded into bytes, but then `IStateStore` should track the encoding scheme not an extension class.
  - On another note, allowing to mix `byte[]` and `string` to be mixed seem like a bad idea, as there is no way to determine whether a `KeyBytes` is derived from a byte string or a text string. 🙄
    - To be usable at a higher level, `KeyBytes` should've retained the encoding information (i.e. whether it was created from `byte[]` or `string`).
    - The way we are handling keys in `KeyConverters` is already a mess. 😕
  - I don't think its worth the effort to allow different encoding schemes.

Anyhow, with how assumptions have already seeped in to the codebase, might well as make it more explicit. 😶

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,13 @@ To be released.
     `IImmutableDictionary<string, IValue>`.  [[#3321]]
  -  Removed `EnumerableMeasurement` class.  [[#3325]]
  -  Removed `KeyValueExtensions` class.  [[#3325]]
+ -  Removed `StateStoreExtensions.EncodeKey()` and
+    `StateStoreExtensions.DecodeKey()` methods.  [[#3328]]
+ -  Removed `StateStoreExtensions.GetStates(IStateStore, HashDigest<SHA256>?,
+    IReadOnlyList<string>)` method.  [[#3328]]
+ -  Removed `TrieExtensions.Set(ITrie,
+    IEnumerable<KeyValuePair<string, IValue?>)` method.  [[#3328]]
+ -  Removed `KeyBytes(string, Encoding)` constructor.  [[#3328]]
 
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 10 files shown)


CHANGES.md
```diff
@@ -21,6 +21,15 @@ To be released.
     `IImmutableDictionary<string, IValue>`.  [[#3321]]
  -  Removed `EnumerableMeasurement` class.  [[#3325]]
  -  Removed `KeyValueExtensions` class.  [[#3325]]
+ -  Removed `StateStoreExtensions.EncodeKey()` and
+    `StateStoreExtensions.DecodeKey()` methods.  [[#3328]]
+ -  Removed `StateStoreExtensions.GetStates(IStateStore, HashDigest<SHA256>?,
+    IReadOnlyList<string>)` method.  [[#3328]]
+ -  Removed `TrieExtensions.Set(ITrie,
+    IEnumerable<KeyValuePair<string, IValue?>)` method.  [[#3328]]
+ -  Added `KeyBytes.Encoding` static property.  [[#3328]]
```
moreal, line 30: This line should move to the `Added APIs` section.

CHANGES.md
```diff
@@ -21,6 +21,15 @@ To be released.
...
+ -  Removed `KeyBytes(string, Encoding)` constructor.  [[#3328]]
+ -  Added `KeyBytes(string)` constructor.  [[#3328]]
```
moreal, line 32: This line should move to the `Added APIs` section. 🤔 Or, since it is in the same context as the sentence above, I want to change it into one sentence and change it to something like `Replaced ...`

moreal (commented):
👍🏻

OnedgeLee (approved):
No other opinion, except above changelog comment from @moreal .

PR #3344, opened 2023-07-31
Add Static Factory Methods to TxResult Class

Atralupus:
This PR introduces `Success`, `Failure`, and `FromStatus` static factory methods to the `TxResult` class. These can be particularly beneficial in parts of the codebase such as [TransactionQuery.cs](https://github.com/planetarium/libplanet/blob/e74fb5a87684d8831b8ca42f80318bfced1494ed/Libplanet.Explorer/Queries/TransactionQuery.cs#L247C40-L247C40).

### DIFF
--- a/Libplanet.Explorer/GraphTypes/TxResult.cs
+++ b/Libplanet.Explorer/GraphTypes/TxResult.cs
@@ -47,5 +47,59 @@ namespace Libplanet.Explorer.GraphTypes
 
         public IImmutableDictionary<Address, IImmutableDictionary<Currency, FungibleAssetValue>>?
             UpdatedFungibleAssets { get; }
+
+        public static TxResult Success(
+            long blockIndex,
+            string blockHash,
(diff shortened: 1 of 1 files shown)


riemannulus:
@greymistcube Could you review it?

Libplanet.Explorer/GraphTypes/TxResult.cs
```diff
@@ -47,5 +47,59 @@ public class TxResult
...
+                blockHash,
+                exceptionName,
+                exceptionMetadata,
+                null,
+                null,
+                null
+            );
+        }
+
+        public static TxResult FromStatus(TxStatus status)
```
moreal, line 91: Well.. `Invalid` and `Staging` can be static read-only properties 🤔 .

moreal (approved):

greymistcube:
Eh, why is this needed in the first place? 🤔

PR #3359, opened 2023-08-08
♻️ Added single `Get()` to `ITrie`

greymistcube:
Long overdue change to `ITrie`. Performance-wise, new `ITrie.Get(KeyBytes)` is not that significantly better as the overhead was minimal.

However, `MerkleTrie.Get()` can be more than twice as fast, depending on the environment and the number of `KeyBytes` given.

In order to leverage this properly, `IAccount.GetStates()` and `IAccount.GetState()` should refer to different `ITrie.Get()`. Depending on where the parallelism should reside, it might even be a good idea to get rid of `ITrie.Get(IReadOnlyList<KeyBytes>)`. 🙄


riemannulus (approved):
👍

riemannulus (approved):

PR #3373, opened 2023-08-10
Introduce IWorld

OnedgeLee:
### Changes
- Introduce `IWorld` as super account that manages `Account`s.
- Former interfaces that receives or returns `IAccount` on `IActionContext`, `IActionEvaluation`, `ActionEvaluator`, `IFeeCollector`, `IAction`, `IActionRenderer` now handles `IWorld`, instead.
- `IBlockChainStates` now receives new parameter for account address.


riemannulus:
Could you rebase it?

Plus, I change the base branch to `4x-candidate` because this PR does not integrate.

PR #3377, opened 2023-08-18
Fix node equality for `INode`

greymistcube:
Preparatory work before implementing `ITrie` "diff".

### DIFF
--- a/Libplanet.Store/Trie/Nodes/FullNode.cs
+++ b/Libplanet.Store/Trie/Nodes/FullNode.cs
@@ -34,48 +34,21 @@ namespace Libplanet.Store.Trie.Nodes
             return new FullNode(Children.SetItem(index, childNode));
         }
 
-        bool IEquatable<FullNode>.Equals(FullNode? other)
+        /// <inheritdoc cref="IEquatable{T}.Equals"/>
+        public bool Equals(FullNode? other)
         {
-            if (other is null)
-            {
-                return false;
-            }
-
             if (ReferenceEquals(this, other))
             {
-                return false;
+                return true;
             }
 
-            for (var i = 0; i < ChildrenCount; ++i)
-            {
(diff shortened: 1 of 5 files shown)


Libplanet.Store/Trie/Nodes/FullNode.cs
```diff
@@ -34,48 +34,21 @@ public FullNode SetChild(int index, INode childNode)
...
-            if (other is null)
-            {
-                return false;
-            }
-
             if (ReferenceEquals(this, other))
             {
-                return false;
+                return true;
             }
```
moreal, line 43: 🙀 Sorry... 🙏🏻

Libplanet.Store/Trie/Nodes/FullNode.cs
```diff
@@ -34,48 +34,21 @@ public FullNode SetChild(int index, INode childNode)
...
-                if (other.Children[i]?.GetHashCode() != Children[i]?.GetHashCode())
-                {
-                    return false;
-                }
-            }
-
-            return true;
+            return other is { } node &&
+                Children.Where((n, i) => n is { })
+                    .SequenceEqual(node.Children.Where((n, i) => n is { }));
```
moreal, line 47: It can return `true` even if they are different.

```csharp
using System;
using System.Linq;
					
public class Program
{
	public static void Main()
	{
		var arrA = new int?[3] { null, 1, null };
		var arrB = new int?[3] { 1, null, null };
		Console.WriteLine(arrA.Where((n, i) => n is { }).SequenceEqual(arrB.Where((n, i) => n is { }))); // True
	}
}

```

image

greymistcube: My bad, I meant to compare *with* index. I'll fix it. 😱

moreal (approved):

moreal (approved):

PR #3378, opened 2023-08-18
🔧  refactor: `Account` with `ITrie`

riemannulus:
# Context
We want to use `ITrie` instead of `AccountDelta()`.

# Rationale
For using `ITrie` instead of `AccountDelta()`. First, we support backward compatibility before removing `AccountDelta()`. 
But we don't have any comparison difference for `ITrie`.
So in this PR,  I left `AccountDelta()` for backward compatibility and changed recursive `Account` to `ITrie base Account`.


Libplanet.Action/State/Account.cs
```diff
@@ -15,25 +17,44 @@ namespace Libplanet.Action.State
...
+        {
+        }
 
-        private Account(IAccountState baseState)
-            : this(baseState, new AccountDelta())
+        private Account(IAccount previousAccount, IAccountDelta delta)
+            : this(
+                previousAccount.BlockHash,
+                previousAccount.Trie.Set(delta.ToRawDelta()),
+                delta)
```
greymistcube, line 39: This seems problematic. Trying to set the whole delta every time would be very bad. 😱

greymistcube: I'll leave it as a separate issue.

PR #3391, opened 2023-08-25
Refactor/path cursor

greymistcube:
Adds `Nibbles` `struct`ure. Changed `PathCursor` to use `Nibbles` internally.
I'd say performance degradation is still acceptable. This mainly comes from converting `KeyBytes` into `Nibbles`.

This can be somewhat optimized as `KeyBytes` is used for two completely different purposes:

- To store in `IKeyValueStore` for `HashNode`s.
- To traverse `ITrie` with `PathCursor`.

The problem is `PathCursor` is never really directly used as a key to store in `IKeyValueStore`. It only represents a traversing path for `ITrie`. `KeyBytes` is actually only explicitly used to reference `HashNode`'s hash value. 🙄

That is, `KeyConverters` should be converting into `Nibbles`, not `KeyBytes`.

In any case, this would require larger `interface` overhaul. 😥


greymistcube:
Closing due to PR numbering. Will be reopened. 🙃

PR #3392, opened 2023-08-25
introduce: `IRecordableTrie` 

riemannulus:
Sliced PR for review https://github.com/planetarium/libplanet/pull/3378.

PR #3398, opened 2023-08-29
♻️ Refactor `ITrie` related interface

greymistcube:
For safety reasons, we might want to have `IReadOnlyKeyValueStore` at some point.
In any case, entirely removed `Commit()` and all its related methods from `ITrie`/`MerkleTrie` so `ITrie` can no longer write to storage. As mentioned in several places, this makes `Commit()` very code path dependent and rather more dangerous as `IStateStore.Commit()` now requires an `ITrie` with the same reference to `IKeyValueStore` that `IStateStore` has while this isn't actually being checked. 🙄

This was the reason why I closed #3350 without merging, but after seeing #3392, and with some hindsight and cost benefit analysis, I figured this might be better for the following reasons:
- `IKeyValueStore` write access was being controlled at three different layers for the same purpose: From `IKeyValueStore`, creating a `VolatileKeyValueStore` wrapper, and having `IRecordableTrie`/`Trie` interfaces.
- This makes too many points of failure, each being critical.
- Also this makes the codebase a lot more susceptible to failure due to type juggling.


Libplanet.Store/TrieStateStore.Commit.cs
```diff
@@ -0,0 +1,161 @@
...
+using Libplanet.Store.Trie.Nodes;
+
+namespace Libplanet.Store
+{
+    public partial class TrieStateStore
+    {
+        private static readonly Codec _codec = new Codec();
+
+        /// <inheritdoc cref="IStateStore.Commit"/>
+        public ITrie Commit(ITrie trie)
```
riemannulus, line 19: IMO we need some tests for this method.

greymistcube: It's mostly handled by old `ITrie.Commit()`. I suppose we need to move them. 😕

greymistcube: Just added some tests.

riemannulus (approved):

PR #3401, opened 2023-08-30
✨ Introduce `ITrie.GetNode()`

greymistcube:
Preparatory PR to introduce trie diff. Also fixes a small bug.

### DIFF
--- a/Libplanet.Store/Trie/ITrie.cs
+++ b/Libplanet.Store/Trie/ITrie.cs
@@ -89,7 +89,7 @@ namespace Libplanet.Store.Trie
         /// </para>
         /// <para>
         /// As such, for two equivalent <see cref="ITrie"/>s, <see cref="Trie"/>s that
-        /// <em>would have the same committed <see cref="Hash"/>es</em>, this may retrun different
+        /// <em>would have the same committed <see cref="Hash"/>es</em>, this may return different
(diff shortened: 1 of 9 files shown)


Libplanet.Store/Trie/ITrie.cs
```diff
@@ -59,5 +70,31 @@ public interface ITrie
         /// values are ordered in the same way to the corresponding <paramref name="keys"/>.  Absent
         /// values are represented as <see langword="null"/>.</returns>
         IReadOnlyList<IValue?> Get(IReadOnlyList<KeyBytes> keys);
+
+        /// <summary>
+        /// Gets the first node encountered at <paramref name="nibbles"/> when traversing down
+        /// from <see cref="Root"/>.
+        /// </summary>
+        /// <param name="nibbles">The path to check.  This must be a secure path.</param>
```
moreal, line 78: > This must be a secure path.

I wonder that this `secure` word has same meaning with `TrieStateStore.Secure` (hash path and use the hash as path)

moreal (approved):
It seems good to merge but I left a comment that I want to check. 🙏🏻

Libplanet.Store/Trie/ITrie.cs
greymistcube: Yes, it is the same `secure`. Probably could be worded better. 🙄

PR #3410, opened 2023-09-04
✨ Feature `ITrie.Diff()`

greymistcube:


CHANGES.md
```diff
@@ -26,6 +26,9 @@ To be released.
      -  Removed `ITrie.Secure` property.
      -  Removed `secure` parameter from `MerkleTrie()`.
      -  Removed `secure` parameter from `PathCursor()`.
+ -  (Libplanet.Store) Added `ITrie.Diff()` method.  [[#3410]]
+ -  (Libplanet.Store) Removed `MerkleTrieExtensions.DifferentNodes()`
+    static method.  [[#3410]]
```
moreal, line 31: There is no link for #3410 in `CHANGES.md`. You should add the below line:

```
[#3410]: https://github.com/planetarium/libplanet/pull/3410
```

Libplanet.Store/Trie/ITrie.cs
```diff
@@ -118,5 +118,31 @@ public interface ITrie
...
+        /// This operation has the following properties:
+        /// <list type="bullet">
+        ///     <item><description>
+        ///         This operation is non-symmetric.  That is, in general,
+        ///         <c>trieA.Diff(trieB)</c> and <c>trieB.Diff(trieA)</c> are not the same.
+        ///     </description></item>
+        ///     <item><description>
+        ///         Values existing in <paramref name="other"/> but not in the source instance,
+        ///         considered as <see langword="null"/> in the source, are not included in the
+        ///         result.
```
moreal, line 142: Then, does it mean that:

- if sourceTrie is states at block A and otherTrie is states at block A + 1
- At block A + 1, new state S was set.
- Since it was null in sourceTrie, the diff operation doesn't catch new states. right?

Libplanet.Store/Trie/MerkleTrie.Diff.cs
```diff
@@ -0,0 +1,160 @@
...
+                yield break;
+            }
+
+            var queue = new Queue<(INode, Nibbles)>();
+            queue.Enqueue((Root, new Nibbles(ImmutableArray<byte>.Empty)));
+
+            while (queue.Count > 0)
+            {
+                (INode node, Nibbles path) = queue.Dequeue();
+                INode? target = other.GetNode(path);
```
moreal, line 26: If there is a performance issue, I guess this part will be such a point.

Libplanet.Store/Trie/ITrie.cs
greymistcube: Correct. As the method only traverses the source trie to look for differences, as there can't be a `null` node (at least in current implementation) in source trie, it doesn't even get compared in the first place. 🙃

Libplanet.Store/Trie/MerkleTrie.Diff.cs
greymistcube: I do have some vague idea on how to optimize this using a lexicographically ordered trie traversal method. 🙄

moreal: Maybe I think it will be a similar idea. I left this comment for note 😀 I'll approve.

moreal (approved):

PR #3411, opened 2023-09-04
🧹 Fix `BlockChain.DetermineStateRootHash`

### LINKED ISSUE #3394: Fix `IStore.GetStateRootHash()`
See https://github.com/planetarium/libplanet/pull/3392#discussion_r1308120460.

greymistcube:
Resolves #3394.

Not the best looking solution. 🙄
This covers an edge case where a genesis `Block` is completely empty.

### DIFF
--- a/Libplanet.Store/StoreExtensions.cs
+++ b/Libplanet.Store/StoreExtensions.cs
@@ -61,12 +61,18 @@ namespace Libplanet.Store
         /// If <see langword="null"/> is present, <see langword="null"/> is returned.</param>
         /// <returns>The state root hash of the block, or <see langword="null"/> if the block is
         /// not found or <paramref name="blockHash"/> is <see langword="null"/>.</returns>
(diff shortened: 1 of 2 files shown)


Libplanet.Store/StoreExtensions.cs
```diff
@@ -61,12 +61,18 @@ public static void Copy(this IStore from, IStore to)
...
             this IStore store,
             BlockHash? blockHash
         ) =>
-            blockHash is { } hash && store.GetBlockDigest(hash) is BlockDigest digest
-                ? digest.StateRootHash
+            blockHash is { } hash
+                ? store.GetBlockDigest(hash) is BlockDigest digest
+                    ? digest.StateRootHash
+                    : throw new ArgumentException(
+                        $"Given {nameof(blockHash)} was not found in storage: {hash}")
```
riemannulus, line 75: Is it safe when a newly thrown exception on here?

PR #3416, opened 2023-09-07
Remove trie mutation from Account

greymistcube:


riemannulus (dismissed):
It seems just a rollback, LGTM.

PR #3424, opened 2023-09-14
Initial implementation of AccountDiff

greymistcube:


Libplanet.Action/State/AccountDiff.cs
```diff
@@ -0,0 +1,351 @@
...
+    ///     <item><description>
+    ///         As <see cref="Currency"/> information is in the domain of an application using
+    ///         this library, only the hash of a <see cref="Currency"/> is directly stored in
+    ///         the underlying storage.  As such, each <see cref="Currency"/> and
+    ///         <see cref="FungibleAssetValue"/> are handled as raw values, that is, as hash and
+    ///         <see cref="Integer"/>, for an <see cref="AccountDiff"/>.
+    ///     </description></item>
+    /// </list>
+    /// </summary>
+    public class AccountDiff
```
riemannulus, line 52: We might be notice this class on CHANGES.md...

greymistcube: Done. 😗

PR #3429, opened 2023-09-21
♻️ `TxExecution` overhaul

greymistcube:
Complete overhaul of `TxExecution`.

Notes:
- `TxSuccess` and `TxFailure` have been removed.
- Two are now differentiated by `TxExecution.Fail` boolean property.
- Only `TxExecution.Fail` property is compatible with old saved `TxExecution`. This is to support `TxResult`'s `TxStatus`.
  - New `TxExecution` can decode old saved `TxExecution`, but all other information is lost except `TxExecution.Fail`. The rest is set to `null`.

Todo:
- Add `InputState` and `OutputState` of type `HashDigest<SHA256>` to `TxResult`.
- As of this PR, there are now **three** commits happening. Once to validate `Block`, once to transform `IActionEvaluation`s to call `IActionRenderers`, and once to transform `IActionEvaluation`s to generate `TxExecutions`. These three should be combined to one.


Libplanet/Blockchain/BlockChain.TxExecution.cs
```diff
@@ -24,100 +22,101 @@ public partial class BlockChain
...
+                    ITrie nextTrie = trie;
+                    foreach (var eval in group.Item2)
+                    {
+                        foreach (var kv in eval.OutputState.Delta.ToRawDelta())
+                        {
+                            nextTrie = nextTrie.Set(kv.Key, kv.Value);
+                        }
+                    }
+
+                    nextTrie = StateStore.Commit(nextTrie);
```
riemannulus, line 64: 😢

riemannulus (approved):

PR #3445, opened 2023-10-05
♻️ Commit from `IActionEvaluator`

greymistcube:
Finally moves commit to `IActionEvaluator`.
API-wise, there isn't much change, but a shift in responsibility could be considered rather big. 🙄

### DIFF
--- a/Libplanet.Action/ActionEvaluator.cs
+++ b/Libplanet.Action/ActionEvaluator.cs
@@ -11,6 +11,7 @@ using Libplanet.Action.Loader;
 using Libplanet.Action.State;
 using Libplanet.Common;
 using Libplanet.Store;
+using Libplanet.Store.Trie;
 using Libplanet.Types.Blocks;
 using Libplanet.Types.Tx;
 using Serilog;
@@ -76,7 +77,7 @@ namespace Libplanet.Action
 
         /// <inheritdoc cref="IActionEvaluator.Evaluate"/>
         [Pure]
-        public IReadOnlyList<IActionEvaluation> Evaluate(
+        public IReadOnlyList<ICommittedActionEvaluation> Evaluate(
             IPreEvaluationBlock block,
             HashDigest<SHA256>? baseStateRootHash)
         {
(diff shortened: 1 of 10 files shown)


Libplanet.Action/ActionEvaluator.cs
```diff
@@ -479,6 +476,64 @@ internal IAccount PrepareInitialDelta(HashDigest<SHA256>? stateRootHash)
...
+            {
+                ITrie nextTrie = trie;
+                foreach (var kv in evaluation.OutputState.Delta.ToRawDelta())
+                {
+                    nextTrie = nextTrie.Set(kv.Key, kv.Value);
+                    setCount++;
+                }
+
+                nextTrie = _stateStore.Commit(nextTrie);
+                var committedEvaluation = new CommittedActionEvaluation(
```
riemannulus, line 502: Originally, `ActionEvaluation.Execute()` was a yield-return method. I am worried about the performance issues with these changes. How about you?

Libplanet.Action/IActionEvaluator.cs
```diff
@@ -24,16 +24,15 @@ public interface IActionEvaluator
         /// <paramref name="block"/>.</param>
         /// <returns> The result of evaluating every <see cref="IAction"/> related to
         /// <paramref name="block"/> as an <see cref="IReadOnlyList{T}"/> of
-        /// <see cref="IActionEvaluation"/>s.</returns>
+        /// <see cref="ICommittedActionEvaluation"/>s.</returns>
         /// <remarks>
-        /// <para>Publicly exposed for benchmarking.</para>
-        /// <para>First evaluates all <see cref="IAction"/>s in
```
riemannulus, line 30: IMO we need to notice `Evaluate` may cause side effects to `IStateStore`.

Libplanet.Action/ActionEvaluator.cs
greymistcube: Eh, I don't think performance isn't much to worry about. Also, double iteration is necessary down the line as `TxExecution` and `IActionRenderer` have different scopes, so I'd say `IReadOnlyList` is more appropriate here. 🙄

riemannulus: Then, can we commit parallelly? I think this makes sense... 

- Locking `IStateStore` is when only writing node.
- Expensive task when committing and calculating `StateRootHash` is organizing tree. 
- And it can be computed in parallel because each `ITrie` does not affect each other.

For sure, I do not think double iteration has an effect on performance. It is already working like that.

greymistcube: > Locking IStateStore is when only writing node.

This might be a good idea, but I think locking should be handled from `IStateStore` side not `IActionEvaluator` side.

> Then, can we commit parallelly?

- I don't quite understand this point. `ITrie`'s are independent in the sense that they are all immutable, but they are created sequentially **and** share references to nodes internally.
- Committing/computing state root hash in parallel means having write access to `KeyValueStore` in parallel. Again, I think this should be handled from `IStateStore` side not `IActionEvaluator` side.
- I don't see how committing in parallel would save time. This might be a viable option with `IWorld` later down the line, but even in that case, due to how "separate" `ITrie`'s might be tangled with each other, committing in parallel might not be as easy as you might think.

riemannulus: Ok, I might be something overthinking about this. I mean parallelism committing when migrating to `IWorld`. Don't mind what I'm saying. 😓

riemannulus (approved):

PR #3450, opened 2023-10-06
🧹 Remove `StateStoreExtensions`

### LINKED ISSUE #3323: Cleanup `StateStoreExtension`
Certain methods are purely used only for testing. 🙄

greymistcube:
Resolves #3323.


riemannulus (commented):
Does `StateStoreExteions` no longer exist use case? Have you been sure about `NineChronicles.Snapshot` too?

greymistcube:
@riemannulus 
Eh, maybe you are right. I'll have to check. From the looks of it however:
- I believe whatever `NineChronicles.Snapshot` requires, it should be achievable using public APIs present.
- I don't see it as good enough reason to keep these extension methods around.

riemannulus:
> I believe whatever NineChronicles.Snapshot requires, it should be achievable using public APIs present.

I agree with your opinion, but we need to move this extension(or function) to `NineChronicles.Snapshot` when it is used there. Just that much.

greymistcube:
@riemannulus 
I Just checked, there is only a single instance of using `ContainsStateRoot()` which can be easily replaced. 😗

riemannulus (approved):

PR #3459, opened 2023-10-25
fix import mjs to require cjs

XxshiftxX:
Previous update added mjs supports to `"import"`. It should be cjs supports for `"require"` because default module system is esm.


moreal (approved):

riemannulus:
@tkiapril Could you check this PR?

tkiapril (approved):

PR #3462, opened 2023-10-27
Introduce `IWorld`

limebell:
This PR introduces `IWorld` in order to introduce "substate" concept.

The state root hash of the block leads to `World`, and the `World` will have `Account`s' state root hashes as its value.

For more detail, check `CHANGES.md` file.


Libplanet.Action/State/ReservedAddresses.cs
```diff
@@ -0,0 +1,16 @@
...
+    public static class ReservedAddresses
+    {
+        public static readonly Address LegacyAccount
+            = new Address("1000000000000000000000000000000000000000");
+
+        public static readonly Address FungibleAssetsAccount
+            = new Address("1000000000000000000000000000000000000001");
+
+        public static readonly Address ValidatorSetAddress
+            = new Address("1000000000000000000000000000000000000002");
```
greymistcube, line 14: I'm not so sure about these. Seems like unused at the moment. We might want to have a more concrete design first. 🙄

Libplanet.Action/State/World.cs
```diff
@@ -0,0 +1,96 @@
...
+            : this(baseState, new WorldDelta())
+        {
+        }
+
+        public World(IWorldState baseState, IWorldDelta delta)
+        {
+            _baseState = baseState;
+            Delta = delta;
+            Legacy = true;
+        }
```
greymistcube, line 27: Why not just inherit `Legacy` from `IWorldState`?

Libplanet.Action/State/KeyConverters.cs
```diff
@@ -4,10 +4,10 @@
 
 namespace Libplanet.Action.State
 {
-    internal static class KeyConverters
+    public static class KeyConverters
```
greymistcube, line 7: I believe `KeyConverters` was made `internal` intentionally. Why has it changed to `public`?

Libplanet.Action.Tests/Mocks/MockWorldState.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public MockWorldState(IImmutableDictionary<Address, IAccount> accounts)
+            : this(new TrieStateStore(new MemoryKeyValueStore()).GetStateRoot(null), accounts)
+        {
+        }
+
+        public MockWorldState(ITrie trie, IImmutableDictionary<Address, IAccount> accounts)
+        {
+            Trie = trie;
+            Legacy = true;
```
greymistcube, line 26: Same with `World` in regards to its API. Seems like this is fixed to `true`. This should be able to emulate "legacy" and "modern". 😶

Libplanet.Action/Sys/Initialize.cs
```diff
@@ -47,7 +51,12 @@ public IValue PlainValue
                         s.Select(
                             kv => new KeyValuePair<IKey, IValue>(
                                 (Binary)kv.Key.Bencoded,
-                                kv.Value
+                                new Dictionary(kv.Value.Select(
+                                    kv2 => new KeyValuePair<IKey, IValue>(
+                                        (Binary)kv2.Key.Bencoded,
+                                        kv2.Value
+                                    )
+                                ))
```
greymistcube, line 59: Thought about this a bit, and I'm of the opinion that this should stay intact. Unless we need it in the near future, I don't think this should change without a consensus that this is a necessary design change. We need a proper discussion on what constitutes a "system action". 😐

Libplanet.Action/State/KeyConverters.cs
OnedgeLee: At first, it has been changed into public temporarily to be used on `BlockChain.Evaluate`.
Since these are moved into `ActionEvaluator`, they have to be reverted as internal as you mentioned.

Libplanet.Action/State/WorldExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        internal static IAccount GetValidatorSetAccount(this IWorldState world) =>
+            world.GetAccount(ReservedAddresses.LegacyAccount);
+
+        internal static IWorld SetFungibleAssetsAccount(
+            this IWorld world,
+            IAccount account) =>
+            world.SetAccount(ReservedAddresses.LegacyAccount, account);
+    }
+}
```
greymistcube, line 20: This clashes with `Addresses` by fixing FAV handling scheme.

Libplanet.Tests/Blockchain/BlockChainTest.Append.cs
```diff
@@ -256,6 +277,44 @@ public void Append(bool getTxExecutionViaStore)
...
+                miner,
+                new[] { tx2 }.ToImmutableList(),
+                commit1);
+            _blockChain.Append(block2, TestUtils.CreateBlockCommit(block2));
+            Assert.Equal(
+                (Text)"bar",
+                _blockChain.GetWorldState()
+                    .GetAccount(DumbModernAction.DumbModernAddress)
+                    .GetState(address2));
+        }
```
greymistcube, line 316: Maybe have a check that a retrieved `IWorld` is "modern"?

Libplanet.Action.Tests/Mocks/MockWorldState.cs
OnedgeLee: Hmm, Do we really need `MockWorldState` class? May be we can use `WorldBaseState` on tests too for now.

greymistcube: Maybe, its purpose is to easily mock an `IWorldState`. Mock accounts were introduced to bypass rather complicated logic implemented internally for `Account`s. I guess `World` is simple enough at this stage that we can simply use the actual implementation instead of a mock. 🙄

OnedgeLee: Actually, I thought replacing `MockAccountState` too. Recent implementation of `IAccountState` holds `Trie`, but for Mock, it's a bit hard to make use of it. (To use it, this will be copy of `AcccountBaseState` I think)

greymistcube: I agree, not enough thought has been put into `MockAccountState` and it isn't being used as widely as I originally wanted it to be. 😥

However, there is no easy way to mock a base state without actually manipulating `ITrie`, which then requires knowing how data is internally stored, which in turn requires to know how `KeyConverters` is implemented.

For practical application, when testing an `IAction.Execute()` for a client, there is no easy way to create a custom base state (mainly due to `Account` logic being overly complicated).

Libplanet.Action/State/WorldBaseState.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// Gets the <see cref="IAccount"/>s of the given <paramref name="addresses"/>.
+        /// </summary>
+        /// <param name="addresses">The <see cref="Address"/>es referring
+        /// the <see cref="IAccount"/>s to get its states.</param>
+        /// <returns>The list of <see cref="IAccountState"/>s of the given
+        /// <paramref name="addresses"/>.</returns>
+        public IReadOnlyList<IAccount> GetAccounts(IReadOnlyList<Address> addresses) =>
+            GetAccountStateRoot(addresses)
+                .Zip(addresses, (trie, address) => CreateAccount(trie))
+                .ToList();
```
greymistcube, line 52: `Zip` seems unnecessary. 😶

Libplanet.Action/State/WorldExtensions.cs
OnedgeLee: I'll remove this extension, and make reserved account addresses to be set from block policy (from app)

Libplanet.Action/State/ReservedAddresses.cs
OnedgeLee: These will be handled on https://github.com/planetarium/libplanet/pull/3483.

Libplanet.Action/State/WorldExtensions.cs
OnedgeLee: Related to https://github.com/planetarium/libplanet/pull/3483

OnedgeLee (approved):

PR #3480, opened 2023-11-03
♻️ Disallow creation of new `Transaction`s with a non-empty `UpdatedAddresses`

greymistcube:
This partially deprecates uses of `Transaction.UpdatedAddresses` property. It is no longer allowed to create a `Transaction` with non-empty `UpdatedAddresses` property through normal means.

As it is still possible to create a `Transaction` with a non-empty `UpdatedAddresses` (in particular, when unmarshalling a marshalled `Transaction`) most `Transaction`s used in tests are left intact (especially those using fixtures) to keep the old behavior intact.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,10 @@ To be released.
 
 ### Backward-incompatible API changes
 
+ -  Removed `updatedAddresses` parameter from `BlockChain.MakeTransaction()`
+    [[#3480]]
+ -  Removed `updatedAddresses` parameter from `Transaction.Create()`.  [[#3480]]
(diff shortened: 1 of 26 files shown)


riemannulus:
Could we talk about this deprecated schedule and update way? Can you make a discussion document about this?

greymistcube:
@riemannulus 
Figured it'd be better to deprecate/remove its use from the client side first.
If all goes well, this will not be used after v200110 is released (see [here](https://github.com/planetarium/NineChronicles/pull/3396)).

riemannulus (approved):

PR #3483, opened 2023-11-13
WIP : Introduce `ISystemAccounts`

OnedgeLee:
### Background
- Predefined `System Accounts` are needed.
- For better structure, they have to be found from `IActionContext`.
- Some of them (ex : fee account) have to be defined on application.
  - For future, especially for PoS, needs to specify accounts for PoS is needed, and since they're related to FAVs, that specification may be happen on application.

### Note
- Predefined account address `LegacyAccount` has been renamed to `DefaultAccount`, to cover single account application case. What I'm thinking is,
  - Provide both of initialization action : single-account initialization and multi-account initialization.
  - Provide migration path of single-to-multi account initialization
  - With this scheme, word `Legacy` seems to be better to be called `Default` I think.
  - When use word `Legacy`, seems to be legacy one is deprecated, or obsoleted, but I think it's better to provide both.

### Changes
- [x] Introduce `ISystemAccounts`
- [x] Add `IActionContext.SystemAccounts`
- [x] Introduce `ISystemAccountsGetter`
- [x] Add `IBlockPolicy.SystemAccountsGetter`
- [x] Remove `WorldExtensions`, and replace them with `SystemAccounts` 
- [ ] Migrate `ValidatorSet` from `DefaultAccount` to `ValidatorSetAccount`
  - Since `ValidatorSetOperate` is on application layer, most of work have to be done on app
  - [ ] Block improper access
  - [ ] Fix IBlockChainState, Explorer API
- [ ] Replace `ReservedAddress.Addresses` to `Context.SystemAccounts` on tests
- [ ] Make `System Accounts` cannot be set from outside of Libplanet? (ex : `IAction.Execute()` on app) -> a bit confused, does fee account needs to be mint from app actions?


OnedgeLee:
Will be handled on https://github.com/planetarium/libplanet/pull/3494, Closing

PR #3485, opened 2023-11-13
♻️ Remove `Rehearsal` from `IActionContext`

greymistcube:
♻️ In preparation for version 4.0.0. `DumbAction` needs to be refactored further to remove `RecordRehearsal` property at some point. 🙄


riemannulus:
The build seems failed.

greymistcube:
Eh, for whatever reason, builds are being flaky now. Manual retries seem to fix the issues. 🙄

riemannulus:
I'm worried that there will be too many API changes... but this PR is nothing to the problem.

riemannulus (approved):

riemannulus:
Check when you bump to lib9c and more. They have their own `IActionContext` serializer. (lib9c, headless, unity)

greymistcube:
@riemannulus
Hence I've preemptively worked on removing references to `Rehearsal` from [lib9c](https://github.com/planetarium/lib9c) as seen [here](https://github.com/planetarium/lib9c/pull/2222) and [here](https://github.com/planetarium/lib9c/pull/2224) to make the transition as smooth as possible.

riemannulus:
Ah, yes we use `ICommitedActionContext` when calling render event. I missed. 😆

PR #3487, opened 2023-11-14
replace: use `RocksDb` instead `Planetarium.RocksDbSharp`

riemannulus:
# Context
Our [Planetarium.RocksDbSharp](https://github.com/planetarium/rocksdb-sharp) is no longer a follow-up original [RocksDB](https://github.com/facebook/rocksdb) version.
But we need a new version of `RocksDB` for our new feature. So we decided to change to [RocksDb](https://github.com/curiosity-ai/rocksdb-sharp) by Curiosity-AI for the backend.

# Relative
close https://github.com/planetarium/libplanet/issues/1848

PR #3488, opened 2023-11-15
➕  introduce: `RocksDb` instance type

riemannulus:
**Caution: Merge https://github.com/planetarium/libplanet/pull/3487 first**

# Context
`RocksDb` has 3 instance types: `Primary`, `ReadOnly`, and `Secondary`. [(see also)](https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances) But we support only `Primary` and `ReadOnly` right now.

For more scalable rocksdb, we introduce the `Secondary` instance type on `Libplanet.RocksDBStore`.

### DIFF
--- a/Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
@@ -75,14 +75,19 @@ namespace Libplanet.RocksDBStore
         /// Creates a new <see cref="RocksDBKeyValueStore"/>.
         /// </summary>
         /// <param name="path">The path of the storage file will be saved.</param>
-        /// <param name="readonly">If it is true, it will open rocksdb in read-only mode.</param>
-        public RocksDBKeyValueStore(string path, bool @readonly = false)
+        /// <param name="type">Choose type of <see cref="RocksDbInstanceType"/>.</param>
+        public RocksDBKeyValueStore(
+            string path,
+            RocksDbInstanceType type = RocksDbInstanceType.Primary)
         {
             var options = new DbOptions()
                 .SetCreateIfMissing();
 
-            _keyValueDb = RocksDBUtils.OpenRocksDb(options, path, @readonly: @readonly);
+            _keyValueDb = RocksDBUtils.OpenRocksDb(options, path, type: type);
+            Type = type;
         }
+
+        public RocksDbInstanceType Type { get; }
 
         /// <inheritdoc/>
         public byte[] Get(in KeyBytes key) => _keyValueDb.Get(key.ToByteArray())
@@ -144,6 +149,11 @@ namespace Libplanet.RocksDBStore
             {
                 yield return new KeyBytes(it.Key());
             }
+        }
+
+        public void TryCatchUpWithPrimary()
+        {
+            _keyValueDb.TryCatchUpWithPrimary();
         }
     }
 }
--- a/Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
+++ b/Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
@@ -75,10 +75,10 @@ namespace Libplanet.RocksDBStore
         /// Creates a new <see cref="RocksDBKeyValueStore"/>.
         /// </summary>
         /// <param name="path">The path of the storage file will be saved.</param>
-        /// <param name="type">Choose type of <see cref="RocksDbInstanceType"/>.</param>
+        /// <param name="type">Choose type of <see cref="RocksDBInstanceType"/>.</param>
         public RocksDBKeyValueStore(
             string path,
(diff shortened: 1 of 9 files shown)


Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -1373,7 +1373,7 @@ private static (IStore Store, IStateStore StateStore) Loader(Uri storeUri)
             int txEpochUnitSeconds = query.GetInt32("tx-epoch-unit-secs", 86400);
             int blockEpochUnitSeconds = query.GetInt32("block-epoch-unit-secs", 86400);
             int dbConnectionCacheSize = query.GetInt32("connection-cache", 100);
-            bool @readonly = query.GetBoolean("readonly", false);
+            var parse = Enum.TryParse(query.Get("instance-type"), out RocksDbInstanceType type);
```
moreal, line 1376: How about using like this?

```suggestion
            var instanceType = Enum.TryParse(query.Get("instance-type"), out RocksDbInstanceType type) ? type : RocksDbInstanceType.Primary;
```

Libplanet.RocksDBStore/RocksDBUtils.cs
```diff
@@ -9,23 +10,40 @@ internal static class RocksDBUtils
...
                     ? RocksDb.OpenReadOnly(options, dbPath, false)
-                    : RocksDb.OpenReadOnly(options, dbPath, columnFamilies, false);
+                    : RocksDb.OpenReadOnly(options, dbPath, columnFamilies, false),
+                RocksDbInstanceType.Secondary => columnFamilies is null
+                    ? RocksDb.OpenAsSecondary(options, dbPath, CreateSecondaryPath(dbPath))
+                    : RocksDb.OpenAsSecondary(
+                        options,
+                        dbPath,
+                        CreateSecondaryPath(dbPath),
+                        columnFamilies),
```
moreal, line 33: Then, should we manually remove the secondary db after every disposing of `RocksDB*Store`?

Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
```diff
@@ -145,5 +150,10 @@ public IEnumerable<KeyBytes> ListKeys()
                 yield return new KeyBytes(it.Key());
             }
         }
+
+        public void TryCatchUpWithPrimary()
+        {
+            _keyValueDb.TryCatchUpWithPrimary();
+        }
```
moreal, line 157: Is it needed for `RocksDBStore` case too?

moreal (dismissed):
I don't know, but did Libplanet decide not to write CHANGELOG these days? (because there is no `changelog` pr check bot)

Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
greymistcube: Although it might not be strictly needed, I also think having this option for `RocksDBStore` might be a good idea. 🤔

Libplanet.RocksDBStore/RocksDBUtils.cs
```diff
@@ -9,23 +10,40 @@ internal static class RocksDBUtils
...
+            string secondaryPath = Path.Combine(dbPath, $"secondary-{Guid.NewGuid()}");
+            if (!Directory.Exists(secondaryPath))
+            {
+                Directory.CreateDirectory(secondaryPath);
             }
 
-            return columnFamilies is null
-                ? RocksDb.Open(options, dbPath)
-                : RocksDb.Open(options, dbPath, columnFamilies);
+            return secondaryPath;
```
greymistcube, line 46: I'm just curious. Does the secondary path get cleaned up automatically if `RocksDb` opened with `Secondary` is disposed? 😶

Libplanet.RocksDBStore/RocksDBUtils.cs
greymistcube: Yeah, this is highly problematic as `Secondary` instances leave artifacts. Implementing automatic cleanup when disposing could also be problematic as it removes log files. Besides, under current implementation, such feature might require massive refactoring. 😐

Libplanet.RocksDBStore/RocksDBKeyValueStore.cs
greymistcube: Problem is lot more complicated than it seems. `RocksDBStore` handles many different `RocksDb` instances internally. It isn't obvious at all how catch-up should work. On the other hand, other APIs such as `StoreLoaderAttribute` clashes with having separate modes open for `RocksDBStore` and `RocksDBStateStore`. 🙄

greymistcube:
@moreal 
> I don't know, but did Libplanet decide not to write CHANGELOG these days? (because there is no changelog pr check bot)

I'm still working on it. 😥

greymistcube (approved):

PR #3489, opened 2023-11-15
✨ ♻️ Cached `TrieStateStore`

greymistcube:
This overhauls existing caching strategy. Shows about 20% ~ 25% improvement for `Block` execution time with 1 million cache capacity. ~~There is a slight change in API, but should be compatible as added parameters are all optional. This should be refactored before 3.8 release.~~

Edit: I'm looking into whether `MerkleTrie()` constructors should be made `internal`. 😶


greymistcube:
Merged in #3495

PR #3495, opened 2023-11-20
✨ ♻️ `HashNode` caching

greymistcube:
Same as #3489 with changed target base branch for a faster release. 😐
#3489 will be closed if this gets included.


Libplanet.Store/Trie/MerkleTrie.cs
```diff
@@ -36,10 +38,12 @@ static MerkleTrie()
         /// nodes.</param>
         /// <param name="rootHash">The root <see cref="ITrie.Hash"/> of
         /// <see cref="MerkleTrie"/>.</param>
+        /// <param name="cache">The <see cref="HashNodeCache"/> to use as cache.</param>
         public MerkleTrie(
             IKeyValueStore keyValueStore,
-            HashDigest<SHA256> rootHash)
-            : this(keyValueStore, new HashNode(rootHash))
+            HashDigest<SHA256> rootHash,
+            HashNodeCache? cache = null)
```
riemannulus, line 45: Do you think this optional parameter is necessary? For the `HashNodeCache` config?

greymistcube: `HashNode` must be injected from `TrieStateStore`. This is to avoid API change as much as possible. 🙄

riemannulus: Aha, my mistake. nevermind.

riemannulus (approved):

PR #3503, opened 2023-11-20
Use pnpm nodeLinker (request for comments)

tkiapril:
This PR makes yarn use pnpm nodeLinker, which may fix some problems rooting from using pnp.


tkiapril:
I'm not sure how this might affect packaging/deploy etc. so I ask for comments. Any thoughts or possible problems, side effects you might think of? @Akamig @moreal cc @eseiker (thinking you might know more about nodejs packaging than me)

Issue #3518, opened 2023-11-24
Bug on gossip denial


OnedgeLee:
Once node receives messages from more than two different rounds higher than current context, It denies peer repeatedly just with single round-gapped message.

Since expected behavior is, above happens just within a height, it's a bug and have to be handled properly.

PR #3519, opened 2023-11-24
Fix bug on gossip denial

OnedgeLee:
There's a bug on gossip denial, and this PR fixes it.

Though adding `_peerCatchupRounds.Clear()` on `GossipConsensusMessageCommunicator.cs L69` is enough to solve the problem, added some height validation not to be confused.
(This will be needless after sync between `Gossip` and `HeightVoteSet`)

This PR Resolves
- https://github.com/planetarium/libplanet/issues/3518

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,14 @@ Version 3.8.1
 
 To be released.
 
+ -  (Libplanet.Net) Fixed a bug where `GossipConsensusMessageCommunicator`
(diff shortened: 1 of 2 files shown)


riemannulus (approved):

greymistcube:
Are we sure we want this to be 3.6 based? 🤔

greymistcube (approved):

PR #3520, opened 2023-11-24
♻️ `ITrie` based `IAccount`

greymistcube:
Pretty crude implementation to use `ITrie` only. It's been tested on 9c-mainnet however. 🙄


Libplanet.Action/ActionEvaluator.cs
```diff
@@ -148,39 +148,18 @@ public class ActionEvaluator : IActionEvaluator
...
-        ///     has <see cref="ActionEvaluation.OutputState"/> with
-        ///     <see cref="IAccount.Delta"/> that is a "superset"
-        ///     of the previous one, if any (possibly except for
-        ///     <see cref="IAccountDelta.ValidatorSet"/>).
-        /// </description></item>
-        /// </list>
-        /// </para>
-        /// </remarks>
         [Pure]
         internal static IEnumerable<ActionEvaluation> EvaluateActions(
```
riemannulus, line 157: As we said, is `static` necessary...?

Libplanet.Action/ActionEvaluator.cs
```diff
@@ -233,9 +213,18 @@ public class ActionEvaluator : IActionEvaluator
...
             IAction action,
+            IStateStore stateStore,
             ILogger? logger = null)
         {
+            if (!context.PreviousState.Trie.Recorded)
+            {
+                throw new InvalidOperationException(
+                    $"Given {nameof(context)} must have its previous state's " +
+                    $"{nameof(ITrie)} recorded.");
+            }
```
riemannulus, line 224: Is this check needed? we already check it on line `316...327`.

Libplanet.Action/ActionEvaluator.cs
```diff
@@ -324,6 +313,18 @@ IActionContext CreateActionContext(IAccount newPrevState)
...
+                    new AccountState(stateStore.Commit(a.Trie)),
+                    a.TotalUpdatedFungibles)
+                : throw new InvalidOperationException(
+                    $"Internal {nameof(IAccount)} is not of valid type: {state.GetType()}");
+
+            if (!state.Trie.Recorded)
+            {
+                throw new InvalidOperationException(
+                    $"Failed to record {nameof(IAccount)}'s {nameof(ITrie)}.");
+            }
```
riemannulus, line 327: Does `stateStore.Commit(a.Trie)` not throw anything if the operation fails?

Libplanet.Action/ActionEvaluator.cs
greymistcube: It's probably better to make it non-`static`. Changing requires changes to test code. Being lazy is all. 🙄
I'll deal with it in a separate PR if I get the time.

riemannulus: > Being lazy is all. 🙄

Agree.

Libplanet.Action/ActionEvaluator.cs
greymistcube: It's a precaution for `EvaluateAction` to check both its input `ITrie` and output `ITrie`. I agree it seems a bit excessive. If we are to remove one, probably the latter one could be considered unnecessary.

riemannulus: > latter one could be considered unnecessary.

I do not have a strong opinion about which one is more unnecessary. Feel free to choose. 😄

Libplanet.Action/ActionEvaluator.cs
greymistcube: Probably not, but I'm not so sure about all possible edge cases. 😥
Although, there is at least one I can come up with.
If `trie.Recorded` was `false` in the first place and it hasn't changed, `stateStore.Commit(trie)` will do nothing
and return an unrecorded `ITrie` instance. 😶

greymistcube: Then again, the more I think about it, as `ActionEvaluator` has no control over what happens inside `IAction.Execute()`, I think this might be necessary. 😑

greymistcube: There are also other issues, such as `ITrie` having to be internally coupled with `IStateStore` instance (more specifically, `IKeyValueStore` instance held internally by `IStateStore`) and the method being `static`, etc. etc. I'd say we leave it as it is for now...?

riemannulus: Ok, I understood.

Libplanet.Action/ActionEvaluator.cs
riemannulus: https://github.com/planetarium/libplanet/pull/3520#discussion_r1414787260

Then we left both of them?

riemannulus (approved):

PR #3523, opened 2023-11-27
♻️ Refactor random seed generation

greymistcube:
Mainly to mitigate unnecessary code rigidity by removing premature optimization and/or coupling.


riemannulus (commented):
Does it make the same output as before?

greymistcube:
@riemannulus 

> Does it make the same output as before?

Yes.

limebell (approved):

PR #3524, opened 2023-11-27
Adjust timing of the state migration

limebell:
Migrate legacy states to `World-Account`-based structure when evaluating block with protocol version `5`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ To be released.
 ### Backward-incompatible API changes
 
  -  Removed `BlockChainStates` class.  [[#3462]]
+ -  Bumped `BlockMetadata.CurrentProtocolVersion` to 5.  [[#3524]]
  -  (Libplanet.Action) Changed `ActionEvaluator` to accept `IWorld`
     instead of `IAccount`.  [[#3462]]
  -  (Libplanet.Action) `IActionEvaluator.OutputState` became `IWorld`.
@@ -76,6 +77,7 @@ To be released.
 
 ### Added APIs
 
+ -  Added `BlockMetadaLegacyStateVersion` constant.  [[#3524]]
  -  (Libplanet.Action) Added `IWorld` interface and its implementation.
     [[#3462]]
      -  Added `World` class.
@@ -117,6 +119,7 @@ To be released.
 ### CLI tools
 
 [#3462]: https://github.com/planetarium/libplanet/pull/3462
+[#3524]: https://github.com/planetarium/libplanet/pull/3524
 
 
 Version 3.7.0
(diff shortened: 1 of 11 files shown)


CHANGES.md
```diff
@@ -76,6 +77,7 @@ To be released.
 
 ### Added APIs
 
+ -  Added `BlockMetadaLegacyStateVersion` constant.  [[#3524]]
```
OnedgeLee, line 80:
```suggestion
 -  Added `BlockMetadata.LegacyStateVersion` constant.  [[#3524]]
```

Libplanet.Net.Tests/Consensus/ContextNonProposerTest.cs
```diff
@@ -228,7 +228,7 @@ public async Task EnterPreVoteNilOnInvalidBlockHeader()
             var invalidBlock = blockChain.EvaluateAndSign(
                 new BlockContent(
                     new BlockMetadata(
-                        protocolVersion: BlockMetadata.CurrentProtocolVersion - 1,
+                        protocolVersion: BlockMetadata.CurrentProtocolVersion,
```
OnedgeLee, line 231: I think this test is a bit weird...
Generating multiple invalid blocks with different condition would be proper instead of generating single block here.
Though, `BlockMetadata.CurrentProtocolVersion - 1` intend to generate invalid block, so I think we don't need to change protocol version here for now.

Libplanet.Tests/Blockchain/BlockChainTest.Append.cs
```diff
@@ -728,5 +731,75 @@ public void CannotAppendBlockWithInvalidActions()
...
+                new BlockMetadata(
+                    protocolVersion: BlockMetadata.LegacyStateVersion,
+                    index: 0L,
+                    timestamp: DateTimeOffset.UtcNow,
+                    miner: fx.Proposer.ToAddress(),
+                    publicKey: fx.Proposer.PublicKey,
+                    previousHash: null,
+                    txHash: BlockContent.DeriveTxHash(txs),
+                    lastCommit: null),
+                transactions: txs);
```
OnedgeLee, line 796: Seems like this variable doen't used at all.
May be, test appending `Block` with `LegacyStateVersion` after genesis is absent?

OnedgeLee (changes requested):

Libplanet.Net.Tests/Consensus/ContextNonProposerTest.cs
limebell: I've changed this since the block became to fail on evaluation process if the protocol version mismatches. Was failed on append process before.

OnedgeLee: 👌

OnedgeLee (approved):

PR #3529, opened 2023-11-28
♻️ Update `TxInvoice` specification

greymistcube:
Allowing all possible combinations of `null` and non-`null` leads to some non-sensical scenarios concerning fee collection.
This strictly prohibits such possibility by restricting possible cases.

- Both `MaxGasPrice` and `GasLimit` are `null`: This is allowed for backward compatibility.
- Both `MaxGasPrice` and `GasLimit` are non-`null`: This is used for "normal" fee collection process.

As [Lib9c](https://github.com/planetarium/lib9c) started to filter those `Transaction`s with either `MaxGasPrice` and `GasLimit` being `null` at the same time as `MaxGasPrice` and `GasLimit` were introduced, we can safely assume all `Transaction`s included in 9c chain are among the types listed above.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,12 +6,20 @@ Version 3.9.0
 
 To be released.
 
+Due to changes in [#3529], a network ran with a prior version may not
+be compatible with this version,  specifically, those that ran with
+[Libplanet 2.0.0] and onwards prior to this release that have included
+`Transaction`s that aren't compatible with the updated specification in [#3529].
+
 ### Deprecated APIs
 
 ### Backward-incompatible API changes
 
  -  (Libplanet.Action) Removed `hashedSignature` of type `byte[]` parameter
     from `ActionEvaluator.GenerateRandomSeed()`.  [[#3523]]
+ -  Changed `TxInvoice` to no longer allow having the null-ness of
+    `MaxGasPrice` and `GasLimit` to be different, i.e. either both should be
+    null or both should not be null at the same time.  [[#3529]]
 
 ### Backward-incompatible network protocol changes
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -6,12 +6,20 @@ Version 3.9.0
 
 To be released.
 
+Due to changes in [#3529], a network ran with a prior version may not
+be compatible with this version,  specifically, those that ran with
+[Libplanet 2.0.0] and onwards prior to this release that have included
+`Transaction`s that aren't compatible with the updated specification in [#3529].
```
riemannulus, line 12: Should be placed at Backward incompatible changes...?

greymistcube: Eh, thought it was important enough to be put in the header section. Not sure where exactly to put it.

Libplanet.Types/Tx/TxInvoice.cs
```diff
@@ -109,6 +115,13 @@ public TxInvoice(ITxInvoice invoice)
             {
                 throw new ArgumentNullException(nameof(updatedAddresses));
             }
+            else if (maxGasPrice is null ^ gasLimit is null)
```
longfin, line 118:
```suggestion
           if (maxGasPrice is null ^ gasLimit is null)
```

`else if` in such clause might make unnecessary diff when we need to remove above `if` block.

longfin (approved):

Libplanet.Types/Tx/TxInvoice.cs
greymistcube: Done.

riemannulus (approved):

PR #3534, opened 2023-12-04
Replace <T> with IActionLoader param

OnedgeLee:
Remove remaining `<T>`s from tests.


riemannulus (approved):

riemannulus:
Hmm... build and tests seem all failed. Could you check or re-run all?

OnedgeLee:
> Hmm... build and tests seem all failed. Could you check or re-run all?

May be messed up with automatic rebase I think...I'll check.

PR #3540, opened 2023-12-05
Add metadata for tries

limebell:
This PR adds metadata to trie, which to address of length 0, cannot be created through provided APIs.

Afterwards, `World` will be treated as `non-legacy` if given trie has metadata. If not, the trie will be treated as legacy.

### DIFF
--- a/Libplanet.Action/State/TrieMetadata.cs
+++ b/Libplanet.Action/State/TrieMetadata.cs
@@ -1,0 +1,35 @@
+using System;
+using System.Numerics;
+using Bencodex.Types;
+
+namespace Libplanet.Action.State
+{
+    public class TrieMetadata
+    {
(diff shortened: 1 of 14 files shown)


Libplanet.Action/State/TrieMetadata.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Numerics;
+using Bencodex.Types;
+
+namespace Libplanet.Action.State
+{
+    public class TrieMetadata
+    {
+        public readonly BigInteger Version;
+
+        public readonly TrieType Type;
```
OnedgeLee, line 11: Seems better to declare them as properties. (`SA1401` warning on StyleCop)

OnedgeLee (commented):
LGTM except below comment.

OnedgeLee (approved):

PR #3551, opened 2023-12-14
♻️ Allow to choose output format of `BencodexValueType`

greymistcube:
See [this discussion](https://github.com/planetarium/libplanet/discussions/3548). In particular, this breaks `stateQuery`, but it isn't yet being used widely yet (and I have plans to entirely overhaul the whole thing). 🙄

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,8 +9,13 @@ To be released.
  -  (Libplanet.Explorer) Added `BlockHashType` and `TxIdType`.  [[#3549]]
  -  (Libplanet.Explorer) Changed `HashDigestSHA256Type` to `HashDigestType<T>`.
     [[#3549]]
+ -  (Libplanet.Explorer) Changed `BencodexValueType` to inherit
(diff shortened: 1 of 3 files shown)


riemannulus (commented):
Is there any case of using this on an external service? Is it just the change of the display purpose that I mentioned last time?

greymistcube:
@riemannulus 
As far as I know, it is being used by replay command in 9c-headless.

There is much discussion to be had, but after reading much on GraphQL, I'm under the impression the existing implementation of `BencodexValueType` was ill advised (in many different ways, such as `StringGraphType` returning `byte[]` for `Serialize()`, etc.). There is much discussion on trying to transport `byte[]` over GraphQL, but most methods resort to different kind of hacks to allow such feature.

Yes, there is some downside, as this would require converting `byte[]` to `string` and back to `byte[]`, but I don't think efficiency of transferring `byte[]` is should be the main concern for using GraphQL.

Besides, whole `stateQuery` syntax **will** break when we move to 4.0.

riemannulus (approved):

PR #3553, opened 2023-12-14
feat: new argument for `RocksDBKeyValueStore.ctor`

riemannulus:
# Context
Since we updated `RocksDb`, it was newly introduced`PendingCompactionLimit`. But its default value is too low to use `NineChronicles`.

# Rationale
We introduce the `DbOptions` argument for the `RocksDBKeyValueStore.ctor` and optimize the default `PendingCompactionLimit` value.


greymistcube (approved):

greymistcube:
Seems to be working fine on mainnet.

PR #3562, opened 2023-12-18
♻️ Removed updated addresses from `IRichStore`

greymistcube:
`UpdatedAddresses` has pretty much lost its meaning, and is no longer used. There really isn't any reason to index this property anymore. 😗

Unless there is a plan to use `IRichStore`, we should remove `IRichStore` entirely if we get a chance.

- `TxId` to `BlockHash` mapping is already being handled by `IStore` (although, `IStore`'s version has its own problems).
- The only remaining functionality not supported by `IStore` directly is `Address` (signer) to `TxId` mapping.

My personal thought on this matter is that there should be a clear distinction between what should be the "core" functionalities of `IStore` (necessary for `BlockChain`'s operation) and "auxiliary" functionalities.

For instance, `TxId` -> `BlockHash` indexing, `Address` -> `TxId` indexing, and `TxResult` all could be considered non-essential. That is, [Libplanet](https://github.com/planetarium/libplanet)'s `BlockChain` would have no problem running without these. 😐

If we had a clearly defined separate interface for these functionalities, then we could **safely assume** that these indexing services are purely auxiliary and **may not be complete**.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,10 +15,15 @@ To be released.
     now one can choose its representation format.  [[#3560]]
  -  (Libplanet.Explorer) Added `HelperQuery`, a collection of utility like
     queries.  [[#3561]]
+ -  (Libplanet.Explorer) Removed `IRichStore.StoreUpdatedAddressReferences()`
+    and `IterateUpdatedAddressReferences()` interface methods.  [[#3562]]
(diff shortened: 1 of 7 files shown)


riemannulus (approved):
I'm not sure we use `IRichSthore` rn... 🤔 Might we need to re-write Libplanet.Explorer.

PR #3569, opened 2023-12-26
♻️ Refactor `KeyConverters`

greymistcube:
Current implementation drags `Currency` context to an unnecessarily low level. Ideally, `KeyConverters` should only deal with `byte[]` types. 🙄


greymistcube:
Eh... decided not to include in 3.x. On further inspection, I think we need a larger overhaul.

PR #3576, opened 2024-01-01
✨ Add `Trie.Remove()`

greymistcube:
Finally adds "value removal" to `ITrie`.
Added `ITrie.SetNull()` and `FullNode.RemoveChild()` as separate methods in order to keep backwards compatibility as much as possible.

I'll add `IAccount.RemoveState()` in a separate PR.

### DIFF
--- a/Libplanet.Tests/Store/Trie/MerkleTrieTest.cs
+++ b/Libplanet.Tests/Store/Trie/MerkleTrieTest.cs
@@ -367,7 +367,7 @@ namespace Libplanet.Tests.Store.Trie
             trie = trie.Set(key, value);
             trie = stateStore.Commit(trie);
             Assert.NotEqual(nullTrieHash, trie.Hash);
-            trie = trie.SetNull(key);
+            trie = trie.Remove(key);
             trie = stateStore.Commit(trie);
             Assert.Null(trie.Root);
             Assert.Equal(nullTrieHash, trie.Hash);
@@ -380,7 +380,7 @@ namespace Libplanet.Tests.Store.Trie
             trie = trie.Set(key00, value00);
             trie = stateStore.Commit(trie);
             Assert.NotEqual(nullTrieHash, trie.Hash);
-            trie = trie.SetNull(key00);
+            trie = trie.Remove(key00);
             trie = stateStore.Commit(trie);
             Assert.Null(trie.Root);
             Assert.Equal(nullTrieHash, trie.Hash);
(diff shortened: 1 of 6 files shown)


Libplanet.Tests/Store/Trie/MerkleTrieTest.cs
```diff
@@ -345,5 +346,137 @@ public void SetValueToFullNode()
             Assert.Equal(value0000, trie.Get(key0000));
             Assert.Equal(value0010, trie.Get(key0010));
         }
+
+        [Fact]
+        public void RemoveValue()
```
riemannulus, line 351: IMO we were supposed to check `ITrie.Remove` is removed from `IKeyValueStore`. And I might we can check it with [`MemoryKeyValueStore`](https://github.com/planetarium/libplanet/blob/81408bf2f1fdce9f3674009438e2a038322ec2b6/Libplanet.Store/Trie/MemoryKeyValueStore.cs#L63). Can you add this test for sure?

greymistcube: No, `ITrie.Remove()` should not remove the key value pair associated with the `KeyByte` being removed from `ITrie`. `IKeyValueStore.Delete()` is there for pruning. I suppose I can check that the key value pairs are **not** removed from the underlying `IKeyValueStore`. 😶

greymistcube: See 4a3e409427c33b1fc7fce2180b16c607752645af.

riemannulus: Hmmm... But you said when it commited, node is really remove from storage. right? https://github.com/planetarium/libplanet/pull/3576/files#diff-5050f0558f9c6e24f3d78bd6b017cb2bd65599e54280fa605d396f99cfcfbfc1R67-R68

greymistcube: I guess its really the problem of wording I suppose. It isn't removed from the storage in the sense that there is no `ITrie` reference recorded in the storage with a value removed at given path. That is, it is "removed" only in the sense that you get an `ITrie` reference (i.e. `HashDigest<SHA256>`) that is "recorded" in the storage. I'll try to rephrase that bit. 🙄

greymistcube: Tried to clarify a bit. See d1069af3da8589663bf448f6026ed0118613ec7a.

riemannulus: Ok, then have you any idea for removing a dangling node from storage?

greymistcube: It has to be manually removed? Depends on what we mean by dangling nodes. Technically, there shouldn't be any dangling nodes since all commits are used for `TxExecution` now. 🙄

riemannulus: Oh, ok. I think I'm sure about that rn... LGTM.

riemannulus (approved):

PR #3579, opened 2024-01-02
♻️ ⏪ Revert changes for old queries

greymistcube:
This should resolve the backwards compatibility issue with services already using the `states` query.
It needs to be phased out anyway, so might as well keep it. 🙄

P.S. From what I gather, `LegacyBencodexValueType` was implemented wrongly in the first place (serialization and parse mismatch, implicit base64 representation in playground, etc.). Hence, resulting in such an awkward way to force backwards compatibility in this PR. 😐

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,10 @@ be compatible with this version,  specifically, those that ran with
  -  (Libplanet.Store) Added `FullNode.RemoveChild()` method.  [[#3576]]
  -  (Libplanet.Action) Added `IAccount.RemoveState()` interface method.
     [[#3577]]
+ -  (Libplanet.Explorer) Added `LegacyBencodexValueType` class that is
+    a copy of an old `BencodexValueType` with its name changed
+    for backwards compatibility.  Changed old `states` query
+    to use `LegacyBencodexValueType` instead of `BencodexValueType`.  [[#3579]]
 
 [#3559]: https://github.com/planetarium/libplanet/pull/3559
 [#3560]: https://github.com/planetarium/libplanet/pull/3560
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -42,6 +42,10 @@ be compatible with this version,  specifically, those that ran with
  -  (Libplanet.Store) Added `FullNode.RemoveChild()` method.  [[#3576]]
  -  (Libplanet.Action) Added `IAccount.RemoveState()` interface method.
     [[#3577]]
+ -  (Libplanet.Explorer) Added `LegacyBencodexValueType` class that is
+    a copy of an old `BencodexValueType` with its name changed
+    for backwards compatibility.  Changed old `states` query
+    to use `LegacyBencodexValueType` instead of `BencodexValueType`.  [[#3579]]
```
riemannulus, line 48: Should we add issue or PR numbers about previous feature

Libplanet.Explorer/GraphTypes/LegacyBencodexValueType.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using Bencodex;
+using GraphQL.Language.AST;
+using GraphQL.Types;
+using Libplanet.Common;
+
+namespace Libplanet.Explorer.GraphTypes
+{
+    public class LegacyBencodexValueType : StringGraphType
```
riemannulus, line 9: Does we haven't more good naming...? I guess if you have plan about more refactoring, maybe you need LegacyBencodexValueV2Type later. 🙄

greymistcube: Well, it's just more work then. To be more sensible and be managable, we need an entirely new set of "V2" scheme which doesn't share the root. Then we'd have to create an entirely new copies of query related codes, management of V1 and V2 graph types, etc. 🙄

CHANGES.md
greymistcube: How? I guess we can refer to #3560 maybe, but I don't think it'd help much. Also the whole point of this PR is to **not fix** the issue since we don't want to deal with the consequences of breaking the API. 😑

limebell (commented):
LGTM

PR #3590, opened 2024-01-07
🧹 Leaner `IBlockChainStates`

greymistcube:
Also with more consistent API. 😶

For example, most methods use either `BlockHash` or `HashDigest<SHA256>` as a reference point in time, but a method pair such as `GetState(BlockHash, Address)` and `GetState(HashDigest<SHA256>, Address, Address)` is quite **confusing** as to what it does and/or how `IBlockChainStates` was designed. 🙄

### DIFF
--- a/Libplanet.Action/State/IBlockChainStates.cs
+++ b/Libplanet.Action/State/IBlockChainStates.cs
@@ -57,49 +57,9 @@ namespace Libplanet.Action.State
         IWorldState GetWorldState(HashDigest<SHA256>? stateRootHash);
 
         /// <summary>
-        /// Returns the <see cref="IAccountState"/> in the <see cref="BlockChain"/>
-        /// at <paramref name="offset"/>.
-        /// </summary>
-        /// <param name="offset">The <see cref="BlockHash"/> of the <see cref="Block"/> to create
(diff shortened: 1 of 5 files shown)


Libplanet.Action/State/IBlockChainStates.cs
```diff
@@ -56,51 +56,11 @@ public interface IBlockChainStates
...
-        /// </returns>
-        /// <seealso cref="IAccountState"/>
-        IAccountState GetAccountState(HashDigest<SHA256>? stateRootHash);
-
         /// <summary>
         /// Gets a state associated to specified <paramref name="address"/>.
         /// </summary>
         /// <param name="offset">The <see cref="BlockHash"/> of the <see cref="Block"/> to fetch
         /// the state from.</param>
-        /// <param name="accountAddress">The <see cref="Address"/> of the account to fetch
```
riemannulus, line 102: Maybe we should mentions about legacyState I guess...

greymistcube: Specs haven't been finalized. We might get rid of access to "lower level" values altogether. I'm making minor tweaks for now to see if it is viable. 😗

PR #3612, opened 2024-01-18
♻️ Refactored `PruneStates()`

greymistcube:
Since `Bencodex`'s fingerprinting was removed, [libplanet](https://github.com/planetarium/libplanet) has lost its support for backward compatibility with `IStateStore` that has used fingerprinting at some point (sadly which is the case for 9c). Preserving `IStateStore` compatibility was never properly implemented in the first place, so this haphazardly implemented incomplete feature has been removed. 😑

I don't think `PruneStates()` can be implemented properly in any useful manner. As `IStateStore` grows, `PruneStates()` will always be too slow to use unless the `IStateStore` in question is pruned frequently and continuously. Using `CopyStates()` and replacing the original `IStateStore` with a copied `IStateStore` will always be a more practical solution for this purpose.

### DIFF
--- a/Libplanet.Store/IStateStore.cs
+++ b/Libplanet.Store/IStateStore.cs
@@ -22,7 +22,8 @@ namespace Libplanet.Store
         ITrie GetStateRoot(HashDigest<SHA256>? stateRootHash);
 
         /// <summary>
-        /// Prunes the states no more used from the state store.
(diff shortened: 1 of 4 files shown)


greymistcube:
Decided to remove `PruneStates()` altogether instead of trying to maintain it. 🙄

PR #3620, opened 2024-01-19
Removed '#nullable disable' from the Libplanet.Strun project

s2quake:
Removed the ``#nullable enable`` keyword defined at the top of some files in the ``Stun project`` and fixed a build error.

### DIFF
--- a/Libplanet.Stun/Stun/Messages/StunMessage.cs
+++ b/Libplanet.Stun/Stun/Messages/StunMessage.cs
@@ -1,4 +1,3 @@
-#nullable disable
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -76,7 +75,7 @@ namespace Libplanet.Stun.Messages
         /// <summary>
         /// A list of <see cref="Attribute"/> of STUN packet.
         /// </summary>
-        protected IEnumerable<Attribute> Attributes { get; set; }
+        protected IEnumerable<Attribute> Attributes { get; set; } = System.Array.Empty<Attribute>();
 
         /// <summary>
         /// Parses <see cref="StunMessage"/> from <paramref name="stream"/>.
@@ -112,8 +111,7 @@ namespace Libplanet.Stun.Messages
                 transactionId
             );
 
-            StunMessage rv = null;
-            rv = @class switch
+            StunMessage rv = @class switch
             {
                 MessageClass.SuccessResponse => method switch
                 {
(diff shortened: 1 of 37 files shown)


Libplanet.Stun/Stun/Messages/StunMessage.cs
```diff
@@ -123,28 +121,23 @@ public enum MessageMethod : ushort
                     MessageMethod.Binding => new BindingSuccessResponse(),
                     MessageMethod.CreatePermission => new CreatePermissionSuccessResponse(),
                     MessageMethod.Refresh => new RefreshSuccessResponse(),
-                    _ => rv,
+                    _ => throw new TurnClientException("Parsed result is null."),
```
riemannulus, line 124: Wouldn't this change cause a change in behavior in other projects using Libplanet if they were written with the expectation that no exception would occur?

Libplanet.Stun/Stun/Messages/StunMessage.cs
```diff
@@ -170,17 +163,17 @@ public byte[] Encode(IStunContext ctx)
...
             List<Attribute> attrs = Attributes.ToList();
 
-            if (!string.IsNullOrEmpty(ctx?.Username))
+            if (ctx.Username != string.Empty)
             {
                 attrs.Add(new Username(ctx.Username));
             }
 
-            if (ctx?.Nonce != null)
+            if (ctx.Nonce.Length > 0)
```
riemannulus, line 171: `ctx?.Nonce != null` and `ctx.Nonce.Length > 0` is a different logic if `Nonce`'s length is 0. We need to check this change is ok for NineChronicles.

Libplanet.Stun/Stun/TurnClient.cs
```diff
@@ -472,7 +505,7 @@ private async Task ProcessMessage(CancellationToken cancellationToken)
                 }
             }
 
-            _logger.Debug($"{nameof(ProcessMessage)} is ended. Connected: {_control.Connected}");
+            _logger.Debug($"{nameof(ProcessMessage)} is ended. Connected: {_control!.Connected}");
```
riemannulus, line 508: `!` is ok?

Libplanet.Stun/Stun/Messages/StunMessage.cs
s2quake: Looking at the code below, if 'rv' is null, it raises an exception, so you can consider it the same code.

Libplanet.Stun/Stun/TurnClient.cs
s2quake: This method is invoked after the _control field is created, so the `!` is safe

Libplanet.Stun/Stun/Messages/StunMessage.cs
s2quake: Because the nonce type is byte[], the nonce property cannot be null.

Libplanet.Stun/Stun/Messages/StunMessage.cs
```diff
@@ -236,18 +229,23 @@ public byte[] Encode(IStunContext ctx)
             return ms.ToArray();
         }
 
+        internal static IEnumerable<Attribute> ParseAttributes(
+            IEnumerable<byte> bytes)
+        {
+            return ParseAttributes(bytes, System.Array.Empty<byte>());
+        }
+
```
greymistcube, line 237: This is unnecessary in my opinion. Attributes are parsed only on incoming messages and they are always expected to have non-null `transactionId`. This is purely for the convenience of making a test code and should not be part of the API.

s2quake:
I will create a new PR prioritizing compatibility.

PR #3622, opened 2024-01-22
Remove the '#nullable disable' keyword from 3 projects(Action, Common, Explorer)

s2quake:
This PR removes the '#nullable disable' keyword and fixes parts that cause errors during build.
Three projects have been applied, as follows.

1. Libplanet.Action/Libplanet.Action.csproj
2. Libplanet.Common/Libplanet.Common.csproj
3. Libplanet.Explorer/Libplanet.Explorer.csproj


Libplanet.Explorer/Queries/BlockQuery.cs
```diff
@@ -1,4 +1,3 @@
-#nullable disable
```
greymistcube, line 1: Also making `desc`, `offset`, and `excludeEmptyTxs` could be `NonNullGraphType` here as well.

Libplanet.Explorer/Queries/ExplorerQuery.cs
```diff
@@ -17,9 +16,11 @@ namespace Libplanet.Explorer.Queries
 {
     public class ExplorerQuery : ObjectGraphType
     {
+        private static IBlockChainContext? chainContext;
```
greymistcube, line 19: Maybe use `_chainContext` to avoid name collision in `ExplorerQuery()`? We generally use Pascal case or `_` prefixed camel case for `private static`.

Libplanet.Common/EnumerableExtensions.cs
```diff
@@ -1,4 +1,3 @@
-#nullable disable
```
greymistcube, line 1: I think we can get rid of `EnumerableExtensions` entirely. It's only used in one place, and it's probably better rephrase the LINQ expression in `Swarm`. 🙄

riemannulus:
https://github.com/planetarium/libplanet/pull/3622#issuecomment-1902867962

Please sign CLA for your code will be `LGPL-2.1` if merged.

s2quake:
I did

greymistcube (approved):

PR #3626, opened 2024-01-22
♻️ 🧹 Removed `EnumerableExtensions`

### LINKED ISSUE #3625: Remove `EnumerableExtensions`
Similar to #3623

greymistcube:
Closes #3625


s2quake (approved):
I tested it with separate code and was able to obtain the same value. There is a method called `MaxBy` in .NET 6, 7, 8. However, it seems it cannot be used in Unity.

greymistcube:
@s2quake 
Yup, which is why I had to use `Aggregate` 😥

PR #3644, opened 2024-01-30
Removal of '#nullable disable' from the Libplanet.Store project

s2quake:
🔪

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,10 @@ To be released.
 
 ### Backward-incompatible API changes
 
+- Remove the '#nullable disable' keyword from 3 projects
+  (Action, Common, Explorer). [[#3622]]
+- Removal of '#nullable disable' from the Libplanet.Store project. [[#3644]]
+
 ### Backward-incompatible network protocol changes
 
 ### Backward-incompatible storage format changes
@@ -23,6 +27,9 @@ To be released.
 ### Dependencies
 
 ### CLI tools
+
+[#3622]: https://github.com/planetarium/libplanet/pull/3622
+[#3644]: https://github.com/planetarium/libplanet/pull/3644
 
 
 Version 4.0.1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,10 +10,9 @@ To be released.
 
 ### Backward-incompatible API changes
 
-- Remove the '#nullable disable' keyword from 3 projects
-  (Action, Common, Explorer). [[#3622]]
-- Removal of '#nullable disable' from the Libplanet.Store project. [[#3644]]
-
+ -  Remove the '#nullable disable' keyword from 3 projects
+    (Action, Common, Explorer). [[#3622]]
+ -  Removal of '#nullable disable' from the Libplanet.Store project. [[#3644]]
 ### Backward-incompatible network protocol changes
 
 ### Backward-incompatible storage format changes
--- a/Libplanet.Explorer/Indexing/BlockChainIndexBase.cs
+++ b/Libplanet.Explorer/Indexing/BlockChainIndexBase.cs
@@ -298,8 +298,11 @@ public abstract class BlockChainIndexBase : IBlockChainIndex
             }
 
             var blockDigest = store.GetBlockDigest(indexEnumerator.Current)!.Value;
-            var txs = blockDigest.TxIds.Select(
-                txId => (ITransaction)store.GetTransaction(new TxId(txId))).ToArray();
+            ITransaction[] txs = blockDigest.TxIds
+                .Select(txId => store.GetTransaction(new TxId(txId))
+                    ?? throw new InvalidOperationException(
+                        $"Could not find transaction with txid {txId} in store."))
+                .ToArray();
             try
             {
                 await IndexAsyncImpl(blockDigest, txs, addBlockContext, stoppingToken)
--- a/Libplanet.Explorer/Queries/ExplorerQuery.cs
+++ b/Libplanet.Explorer/Queries/ExplorerQuery.cs
@@ -75,7 +75,7 @@ namespace Libplanet.Explorer.Queries
                     Chain.Id,
                     (int)offset,
                     limit == null ? null : (int)limit)
-                .Select((value, i) => new { i, value } );
+                .Select((value, i) => new { i, value });
 
             if (desc)
             {
@@ -84,7 +84,10 @@ namespace Libplanet.Explorer.Queries
 
             foreach (var index in indexList)
             {
-                var block = store.GetBlock(index.value);
+                Block block = store.GetBlock(index.value)
+                    ?? throw new InvalidOperationException(
+                        $"Could not find block with block hash {index.value} in store.");
+
                 bool isMinerValid = miner is null || miner == block.Miner;
                 bool isTxValid = !excludeEmptyTxs || block.Transactions.Any();
                 if (!isMinerValid || !isTxValid)
@@ -154,7 +157,7 @@ namespace Libplanet.Explorer.Queries
             return stagedTxs;
         }
 
-        internal static Block GetBlockByHash(BlockHash hash) => Store.GetBlock(hash);
+        internal static Block? GetBlockByHash(BlockHash hash) => Store.GetBlock(hash);
 
         internal static Block GetBlockByIndex(long index) => Chain[index];
 
--- a/Libplanet.Store/BaseIndex.cs
+++ b/Libplanet.Store/BaseIndex.cs
@@ -1,4 +1,3 @@
-#nullable disable
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -93,7 +92,7 @@ namespace Libplanet.Store
             }
             catch (KeyNotFoundException)
             {
-                value = default(TVal);
+                value = default!;
                 return false;
             }
         }
--- a/Libplanet.Store/BaseStore.cs
+++ b/Libplanet.Store/BaseStore.cs
@@ -1,4 +1,3 @@
-#nullable disable
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
@@ -46,7 +45,7 @@ namespace Libplanet.Store
             BlockHash branchpoint
         );
 
-        public abstract Transaction GetTransaction(TxId txid);
+        public abstract Transaction? GetTransaction(TxId txid);
 
         public abstract void PutTransaction(Transaction tx);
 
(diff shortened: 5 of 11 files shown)


riemannulus:
Could you bump on the lib9c? IMO this is easy to check API changes.

Libplanet.Explorer/Indexing/BlockChainIndexBase.cs
```diff
@@ -299,7 +299,7 @@ CancellationToken stoppingToken
 
             var blockDigest = store.GetBlockDigest(indexEnumerator.Current)!.Value;
             var txs = blockDigest.TxIds.Select(
-                txId => (ITransaction)store.GetTransaction(new TxId(txId))).ToArray();
+                txId => (ITransaction)store.GetTransaction(new TxId(txId))!).ToArray();
```
greymistcube, line 302: From the looks of it, this should throw an `Exception` instead of trying to cast to make it clear that the query expects `store.GetTransaction()` to be non-`null`.

Libplanet.Explorer/Queries/ExplorerQuery.cs
```diff
@@ -84,7 +84,7 @@ public ExplorerQuery(IBlockChainContext chainContext)
 
             foreach (var index in indexList)
             {
-                var block = store.GetBlock(index.value);
+                var block = store.GetBlock(index.value)!;
```
greymistcube, line 87: We don't need `!` here. Fix lines `74` to `78` instead.

Libplanet.Store/BaseIndex.cs
```diff
@@ -1,11 +1,11 @@
-#nullable disable
 using System;
 using System.Collections;
 using System.Collections.Generic;
 
 namespace Libplanet.Store
 {
     public abstract class BaseIndex<TKey, TVal> : IDictionary<TKey, TVal>
+        where TKey : notnull
```
greymistcube, line 8: See #3645. You can leave it as it is for now.

Libplanet.Store/StoreExtensions.cs
```diff
@@ -36,7 +35,7 @@ public static void Copy(this IStore from, IStore to)
             {
                 foreach (BlockHash blockHash in from.IterateIndexes(chainId))
                 {
-                    Block block = from.GetBlock(blockHash);
+                    Block block = from.GetBlock(blockHash)!;
```
greymistcube, line 38: Should also throw an `Exception` with an explanation. We might want to add that we expect the source `IStore` to be "complete" in the sense that it contains all `Block`s.

Libplanet.Store/MemoryStore.cs
```diff
@@ -78,11 +77,11 @@ void IStore.DeleteChainId(Guid chainId)
             _canonicalChainId = chainId;
 
         long IStore.CountIndex(Guid chainId) =>
-            _indices.TryGetValue(chainId, out ImmutableTrieList<BlockHash> index) ? index.Count : 0;
+            _indices.TryGetValue(chainId, out var index) ? index.Count : 0;
```
greymistcube, line 80: We tend to use explicit type declaration unless it is clear from the immediate context what type we'd be getting.

Libplanet.Store/MemoryStore.cs
```diff
@@ -134,8 +133,8 @@ BlockHash branchpoint
             }
         }
 
-        Transaction IStore.GetTransaction(TxId txid) =>
-            _txs.TryGetValue(txid, out object untyped) && untyped is Transaction tx
+        Transaction? IStore.GetTransaction(TxId txid) =>
+            _txs.TryGetValue(txid, out object? untyped) && untyped is Transaction tx
```
greymistcube, line 137: Might as well change the type for `_txs` to `ConcurrentDictionary<TxId, Transaction>` or is there some other reason why this can't be done?

Libplanet.Store/DefaultStore.cs
```diff
@@ -105,7 +104,7 @@ public class DefaultStore : BaseStore
         private readonly LruCache<TxId, object> _txCache;
         private readonly LruCache<BlockHash, BlockDigest> _blockCache;
 
-        private readonly MemoryStream _memoryStream;
+        private readonly MemoryStream? _memoryStream;
```
greymistcube, line 107: Looks like we can get rid of this altogether.

Libplanet.Store/BaseIndex.cs
s2quake: I have reverted that code.

Libplanet.Store/BaseStore.cs
```diff
@@ -46,23 +45,23 @@ public abstract class BaseStore : IStore
...
+        public Block? GetBlock(BlockHash blockHash)
         {
             if (GetBlockDigest(blockHash) is BlockDigest blockDigest)
             {
                 BlockHeader header = blockDigest.GetHeader();
                 (Types.Tx.TxId TxId, Transaction Tx)[] txs = blockDigest.TxIds
                     .Select(bytes => new Types.Tx.TxId(bytes.ToArray()))
                     .OrderBy(txid => txid)
-                    .Select(txid => (txid, GetTransaction(txid)))
+                    .Select(txid => (txid, GetTransaction(txid)!))
```
greymistcube, line 64: Should change to `(TxId, Transaction?)` instead of using `!`. See line 68.

Libplanet.Explorer/Queries/ExplorerQuery.cs
greymistcube: Sorry, I misread as `store.GetBlock(index.value!)` instead of `store.GetBlock(index.value)!`. In that case, you can leave the lines 74-78 intact and throw an exception here I suppose.

Libplanet.Store/BaseStore.cs
s2quake: You're right.
But '!' still needs to be used on the line 72.

Libplanet.Store/MemoryStore.cs
s2quake: yes

Libplanet.Store/StoreExtensions.cs
s2quake: How about the code below?
```csharp
Block? block = from.GetBlock(blockHash);
if (block is null)
{
    var message = $"Cannot found block: '{blockHash}' in store.";
    throw new InvalidOperationException(message);
}
```

Libplanet.Store/MemoryStore.cs
s2quake: It seems to be used only as a transaction type.

s2quake:
I've tested it with `lib9c` and `headless`.
There are some minor build errors (error 1, warning 63)
All tests passed with no failures.

Libplanet.Store/BaseStore.cs
greymistcube: Perhaps I'm not getting it. I don't see why `!` should be used in line 72? You mean in the `return` statement? In that case you can probably restructure the code to have the exact same behavior. 😗

One solution I can think of is to use `(TxId, Transaction)` with only non-`null`s and compare it to the length of `blockDigest` then throw an `Exception` if they don't match. Since an `Exception` shouldn't really be thrown, we can construct a list of missing `TxId`s when creating an `Exception`. This would more likely be more efficient as well. 🙃

Libplanet.Store/StoreExtensions.cs
greymistcube: I'd suggest

```csharp
Block block = from.GetBlock(blockHash) ??;
    throw new InvalidOperationException($"Cannot found block: '{blockHash}' in store.");
```

instead.

- It is shorter and easier to comprehend (for me at least)
- `block` isn't really ever supposed to be `null`.

riemannulus (commented):
Could you leave changes into `CHANGES.md`?

Libplanet.Store/BaseStore.cs
s2quake: I tried to solve the problem with minimal code changes. Since the code is designed to throw exceptions explicitly, I will try to restructure it without using the '!' in a way that fits the context.

s2quake:
> Could you leave changes into `CHANGES.md`?
Sure

Libplanet.Store/DefaultStore.cs
s2quake: How about this code below?
```csharp
_db = new LiteDatabase(new MemoryStream(), disposeStream: true);
```

Libplanet.Explorer/Indexing/BlockChainIndexBase.cs
```diff
@@ -298,8 +298,16 @@ CancellationToken stoppingToken
...
+            var txs = blockDigest.TxIds.Select(txId =>
+            {
+                if (store.GetTransaction(new TxId(txId)) is { } tx)
+                {
+                    return tx;
+                }
+
+                var message = $"Cannot found transaction: '{txId}' in store.";
+                throw new InvalidOperationException(message);
+            }).ToArray();
```
greymistcube, line 310: Seems unnecessarily verbose.

```suggestion
            ITransaction[] txs = blockDigest.TxIds
                .Select(txId => store.GetTransaction(new TxId(txId))
                    ?? throw new InvalidOperationException(
                        $"Could not find transaction with txid {txId} in store."))
                .ToArray();
```

Libplanet.Explorer/Queries/ExplorerQuery.cs
```diff
@@ -85,6 +85,12 @@ public ExplorerQuery(IBlockChainContext chainContext)
             foreach (var index in indexList)
             {
                 var block = store.GetBlock(index.value);
+                if (block is null)
+                {
+                    var message = $"Cannot found block: '{index.value}' in store.";
+                    throw new InvalidOperationException(message);
+                }
```
greymistcube, line 92: The main purpose of variable `block` is to be assigned `Block`. Nullable declaration is unnecessary.

```suggestion
                Block block = store.GetBlock(index.value)
                    ?? throw new InvalidOperationException(
                        $"Could not find block with block hash {index.value} in store.");
```

Libplanet.Store/StoreExtensions.cs
```diff
@@ -36,7 +35,13 @@ public static void Copy(this IStore from, IStore to)
...
                 foreach (BlockHash blockHash in from.IterateIndexes(chainId))
                 {
-                    Block block = from.GetBlock(blockHash);
+                    Block? block = from.GetBlock(blockHash);
+                    if (block is null)
+                    {
+                        var message = $"Cannot found block: '{blockHash}' in store.";
+                        throw new InvalidOperationException(message);
+                    }
+
```
greymistcube, line 44: Same as above.

```suggestion
                    Block block = store.GetBlock(index.value)
                        ?? throw new InvalidOperationException(
                            $"Could not find block with block hash {index.value} in store.");
```

Libplanet.Store/BaseStore.cs
```diff
@@ -46,35 +45,37 @@ public abstract class BaseStore : IStore
...
-                    txs.Where(pair => pair.Tx is null).Select(pair => pair.TxId).ToArray();
-                if (missingTxIds.Any())
+                TxId[] txids = blockDigest.TxIds
+                                .Select(bytes => new TxId(bytes.ToArray()))
+                                .OrderBy(txid => txid)
+                                .ToArray();
+                Transaction[] txs = txids.Select(txid => GetTransaction(txid))
+                                         .Where(tx => tx != null)
+                                         .OfType<Transaction>()
+                                         .ToArray();
```
greymistcube, line 68: Probably one last thing. I'm pretty sure `Where()` is unnecessary. 

```suggestion
                Transaction[] txs = txids.Select(txid => GetTransaction(txid))
                                         .OfType<Transaction>()
                                         .ToArray();
```

CHANGES.md
```diff
@@ -10,6 +10,10 @@ To be released.
 
 ### Backward-incompatible API changes
 
+- Remove the '#nullable disable' keyword from 3 projects
+  (Action, Common, Explorer). [[#3622]]
+- Removal of '#nullable disable' from the Libplanet.Store project. [[#3644]]
+
```
greymistcube, line 16:
```suggestion
 -  Remove the '#nullable disable' keyword from 3 projects
    (Action, Common, Explorer). [[#3622]]
 -  Removal of '#nullable disable' from the Libplanet.Store project. [[#3644]]
```

Maybe it's about time we change the formatting. 🙄

riemannulus (approved):

PR #3649, opened 2024-02-05
📝 Updated graphql schema

greymistcube:
📝


riemannulus:
This is not a PATCH version... IMO this is should be merge into main.

greymistcube:
@riemannulus It isn't in the sense that it changes the schema, but it is a patch in the sense that the schema is broken. 🙄

riemannulus (approved):

PR #3651, opened 2024-02-06
Removal of '#nullable disable' from the Libplanet.RocksDBStore project

s2quake:
🔪

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,9 +10,11 @@ To be released.
 
 ### Backward-incompatible API changes
 
- -  Remove the '#nullable disable' keyword from 3 projects
+ -  Removed the '#nullable disable' from 3 projects
     (Action, Common, Explorer). [[#3622]]
- -  Removal of '#nullable disable' from the Libplanet.Store project. [[#3644]]
+ -  Removed the '#nullable disable' from the Libplanet.Store project. [[#3644]]
+ -  Removed the '#nullable disable' from the Libplanet.RocksDBStore project.
+    [[#3651]]
 
 ### Backward-incompatible network protocol changes
 
@@ -30,6 +32,7 @@ To be released.
 
 [#3622]: https://github.com/planetarium/libplanet/pull/3622
 [#3644]: https://github.com/planetarium/libplanet/pull/3644
+[#3651]: https://github.com/planetarium/libplanet/pull/3651
 
 
 Version 4.0.3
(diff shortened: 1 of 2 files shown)


Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -645,7 +644,11 @@ BlockHash branchpoint
 
             _chainDb.Write(batch);
 
-            long bpIndex = GetBlockIndex(branchpoint).Value;
+            long bpIndex = GetBlockIndex(branchpoint) is { } index
+                ? index
+                : throw new ArgumentException(
+                    $"Could not find block index with branchpoint '{branchpoint}' in store.",
+                    nameof(branchpoint));
```
greymistcube, line 651: In general, `IStore`'s current overall design is not to throw an `Exception`. I think it'd be better to silently do nothing and terminate early.

s2quake: So is it right to return rather than throw an exception?
like below
```csharp
if (!(GetBlockIndex(branchpoint) is { } bpIndex))
{
    return;
}
```

greymistcube: Whether it is "right" is a rather hard question to answer (although, I personally think that `IStore` have too much logic baked into it to be `Exception`-free design to be the right design; probably we need some kind of redesign for `IStore` at some point). Although the change would result in a behavioral change, we don't even really have the behavior of the method well-defined. 😑

Also, moving the check/early termination to an earlier point would be better, i.e. before `_chainDb.Write()` gets called for any modification.

greymistcube:
@s2quake Sorry for the delay, but this is missing a change to `CHANGES.md`. 🙃

CHANGES.md
```diff
@@ -10,9 +10,11 @@ To be released.
 
 ### Backward-incompatible API changes
 
- -  Remove the '#nullable disable' keyword from 3 projects
-    (Action, Common, Explorer). [[#3622]]
+ -  Removal of '#nullable disable' from the Libplanet.RocksDBStore project.
+    [[#3651]]
```
greymistcube, line 14: PR numbers should be generally in order.

s2quake: I thought it should be sorted in descending order.

s2quake: I could have known right away if I had referred to the content below. However, it's not that I didn't refer to it. I wonder why I produced such results. Ultimately, it's just an excuse. Anyway, thank you for pointing it out. :)

greymistcube (approved):

PR #3662, opened 2024-02-07
Fixed an issue that caused an error when testing dotnet

s2quake:
This PR resolves problems when running command `dotnet test --filter <option>`

* Fix an issue where the collected test list differs from the expected one.
  - https://xunit.net/releases/visualstudio/2.5.1
* Fixed an issue that caused an error when running `dotnet test` command.

Additionally, I've set the package versions to the same value across all test projects.

```xml
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.1" />
<PackageReference Include="xunit.runner.visualstudio" Version="2.5.6" />
```

### DIFF
--- a/Libplanet.Action.Tests/Libplanet.Action.Tests.csproj
+++ b/Libplanet.Action.Tests/Libplanet.Action.Tests.csproj
@@ -26,7 +26,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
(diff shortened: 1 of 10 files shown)


riemannulus:
You can skip changelog check with add `[skip changelog]` in the commit message.

s2quake:
Instead, I added the 'no changelog' label.

s2quake:
> You can skip changelog check with `[skip changelog]` in the commit message.

Instead, I added the 'no changelog' label.

greymistcube (approved):

PR #3680, opened 2024-02-22
Fix Mortgage not to clean TotalUpdatedFungibleAssets

OnedgeLee:
`TotalUpdatedFungibleAssets` has been omitted since replacement of `Account` has been happend on `Mortgage`.
This makes `TotalUpdatedFungibleAssets` of transactions untractable, except last transaction of block.

This was an unintended error, so fix on this PR.

### DIFF
--- a/Libplanet.Tests/Action/ActionEvaluatorTest.cs
+++ b/Libplanet.Tests/Action/ActionEvaluatorTest.cs
@@ -968,25 +968,46 @@ namespace Libplanet.Tests.Action
         [Fact]
         public void TotalUpdatedFungibleAssets()
         {
+            var privateKeys = Enumerable.Range(0, 3).Select(_ => new PrivateKey()).ToList();
+            var gas = Currency.Uncapped("GAS", 0, null);
+
+            var gasFillActions = privateKeys.Select(pk => new UseGasAction()
+            {
+                GasUsage = 0,
+                Memo = "Refill",
+                MintValue = gas * 100,
+                Receiver = pk.Address,
+            });
+
             var (chain, actionEvaluator) = MakeBlockChainAndActionEvaluator(
                 policy: _policy,
                 store: _storeFx.Store,
                 stateStore: _storeFx.StateStore,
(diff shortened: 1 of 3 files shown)


greymistcube:
Newly fixed test seems to only test FAV changed during fee processing. I think we should also test an `IAction` that uses gas **and** makes changes to FAV. 😶

Libplanet.Tests/Action/ActionEvaluatorTest.cs
```diff
@@ -968,25 +968,46 @@ public void EvaluatePolicyBlockAction()
...
+            var gas = Currency.Uncapped("GAS", 0, null);
+
+            var gasFillActions = privateKeys.Select(pk => new UseGasAction()
+            {
+                GasUsage = 0,
+                Memo = "Refill",
+                MintValue = gas * 100,
+                Receiver = pk.Address,
+            });
+
```
riemannulus, line 981: Maybe we can make a new test for this case, not a fixed previous test.

OnedgeLee:
> Newly fixed test seems to only test FAV changed during fee processing. I think we should also test an `IAction` that uses gas **and** makes changes to FAV. 😶

On `L1032`, checking `TotalUpdatedFungibleAssets` of `Gas` is equal to total count of transactions. I thought this test is for `TotalUpdatedFungibleAssets`, and it's enough.

For Gas processing, it seems other tests are covering.

riemannulus:
IMO this PR needs to log into changelog.

Libplanet.Tests/Action/ActionEvaluatorTest.cs
OnedgeLee: I thought coverage of previous test was too narrow(cannot cover tests with gas).
IMHO, this test have to cover cases with gas, since `TotalUpdatedFungibleAssets` is needed just for the purpose of gas-refilling, as I know. (thought previous test was useless)

riemannulus (approved):

Issue #3645, opened 2024-02-02
Refactor/remove `BaseIndex<K, V>` class.


greymistcube:
Not sure if we need this whole thing to be generic. It's only purpose is to be used as a parent class of `BlockSet`.

- Unless we'd want to reuse this to create other mappings, I'd say `IDictionary<BlockHash, Block>` would be more apt.
- Even if we do want to use it in different ways, such as to create mappings for `BlockHash` -> `IEnumerable<TxId>`, `BlockHash` -> `HashDigest<SHA256>`, etc. (although I can hardly imagine a scenario where this might be useful) I'd say the name is too vague as clearly `BaseIndex` is intended to support `IStore` indexing.
- Even then, we don't really use `BlockSet` for anything other than caching and `[]` indexed retrieval.
- The class itself is rather too vague as to what it's supposed to be useful for. I don't think having to manage an extra layer isn't all that beneficial.

PR #3686, opened 2024-02-23
Remove BaseIndex class.

s2quake:
https://github.com/planetarium/libplanet/issues/3645


CHANGES.md
```diff
@@ -15,6 +15,8 @@ To be released.
  -  Removed the '#nullable disable' from the Libplanet.Store project. [[#3644]]
  -  Removed the '#nullable disable' from the Libplanet.RocksDBStore project.
     [[#3651]]
+ -  Removed `BaseIndex` class and changed `BlockSet` base class
+    from `BaseIndex` to `IDictionary`.  [[#3686]]
```
greymistcube, line 19:
```suggestion
    from `BaseIndex<BlockHash, Block>` to `IDictionary<BlockHash, Block>`.  [[#3686]]
```

greymistcube:
Eh, I think implementing `IDictionary<K, V>` is still an overkill since `BlockSet` is supposed to be readonly really.
How about changing it to `IReadOnlydictionary<K, V>`?

s2quake:
> Eh, I think implementing `IDictionary<K, V>` is still an overkill since `BlockSet` is supposed to be readonly really. How about changing it to `IReadOnlydictionary<K, V>`?

The `_blocks` field in `BlockChain` has a write operation code, so I changed its type to `BlockSet`.
see [this](https://github.com/planetarium/libplanet/pull/3686/files#diff-ecd6275ea8686534dc160d341079910ede8da3d2ec020236582bdb0906cddc21R59)

greymistcube:
I guess I missed it somehow. Very well, I still think there are larger design issues pertaining to this problem, such as its scope and features, but I think this is enough for now.

greymistcube (approved):

PR #3688, opened 2024-03-05
✨  introduce: Libplanet.Store.Remote

riemannulus:
# Context
Since `Libplanet 4.0`, we decided to use the `Trie` base state machine. As a side-effect, `ActionEvaluator` has `StateStore`, and `BlockChain` asks if it wants some `State`.

It makes the `Libplanet` architecture clear, but some 2nd party services which dependent on `BlockChainStates`. Especially `RemoteBlockChainStates`. 

So, we make `Libplanet.Store.Remote`, to support remote state machines.

# Rationale
In `Libplanet.Store.Remote`, have one server and one client. it shares `.proto` schema for communicating as a `gRPC`.

- RemoteKeyValueStore: Client side of `RemoteKeyValueStore`.
- RemoteKeyValueService: Server side of `RemoteKeyValueStore`.

# Run
## Server
First, we need to make a server project.

```csharp
using Libplanet.Store.Remote.Server;
using Libplanet.Store.Trie;

var builder = WebApplication.CreateBuilder(args);

// Add services to the container.
builder.Services.AddGrpc();
builder.Services.AddSingleton<IKeyValueStore, MemoryKeyValueStore>();

// If use `RocksDBStore`, try this:
// builder.Services.AddSingleton<IKeyValueStore>(_ => new RocksDBKeyValueStore("path/to/rocksdb"));

var app = builder.Build();

// Configure the HTTP request pipeline.
app.MapGrpcService<RemoteKeyValueService>();
app.MapGet("/", () => "Communication with gRPC endpoints must be made through a gRPC client. To learn how to create a client, visit: https://go.microsoft.com/fwlink/?linkid=2086909");

app.Run();
```

Then you can test with any other gRPC client tool.
GRPC localhost:5239/libplanet.rpc.v1.KeyValueStore/Get is return some value

## Client
Just use any project if you want to use the remote store.

Async example:
```csharp
using Libplanet.Store.Remote.Client;
using Libplanet.Store.Trie;

var channel = GrpcChannel.ForAddress("https://localhost:5239");
var client = new RemoteKeyValueStoreClient(channel);

await client
        .SetAsync("key", new byte[] { 1, 2, 3, 4, 5 })
        .ConfigureAwait(false);

var value = await client
        .GetAsync("key")
        .ConfigureAwait(false);
```

Sync example:
```csharp
using Libplanet.Store.Remote.Client;
using Libplanet.Store.Trie;

var channel = GrpcChannel.ForAddress("https://localhost:5239");
var client = new RemoteKeyValueStoreClient(channel);

client.Set("key", new byte[] { 1, 2, 3, 4, 5 });

var value = client.Get("key");
```


Libplanet.Store.Remote/Server/RemoteKeyValueService.cs
s2quake, line 1: The _logger field is not being used. How about adding some simple code to it? It seems like it could be easily added using Copilot.

riemannulus: https://github.com/planetarium/libplanet/pull/3688/commits/a28b92a621c883d9a2aec1b5e2b9cb66d83fb215 Amended.

Libplanet.Store.Remote.Tests/Helpers/LoggerHelper.cs
```diff
@@ -0,0 +1,22 @@
+using Microsoft.Extensions.Logging;
```
greymistcube, line 1: Any reason why we are using `Microsoft.Extensions.Logging` instead of `Serilog`?

Libplanet.Store.Remote/Extensions/KeyValueStoreExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+            catch
+            {
+                value = Array.Empty<byte>();
+                result = false;
+            }
+
+            return result;
+        }
+    }
+}
```
greymistcube, line 24: This single use extension seems unnecessary. I think it should be part of `RemoteKeyValueService.Get()`.

Libplanet.Store.Remote.Tests/Helpers/LoggerHelper.cs
riemannulus: No. Since `Microsoft.Extensions.Logging` is provided by default without special Dependency injection, it was used for no particular reason during development. As you said, it would be better to change to Serilog.

Libplanet.Store.Remote/Server/RemoteKeyValueService.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public override Task<SetResponse> Set(SetRequest request, ServerCallContext context)
+        {
+            var key = new KeyBytes(request.Key.ToByteArray());
+            _baseStore.Set(key, request.Value.ToByteArray());
+            _logger.LogTrace("Set: {Key} <= {Value}", key, request.Value);
+            return Task.FromResult(new SetResponse());
+        }
+
+        public override Task<SetMapResponse> SetMap(SetMapRequest request, ServerCallContext context)
```
greymistcube, line 39: I was bit confused by the name. Why not `SetCollection` like `DeleteCollection`?

Libplanet.Store.Remote/Extensions/KeyValueStoreExtensions.cs
riemannulus: I don't wanna popping up data logic to the controller. I want to leave `RemoteKeyValueService.Get()` just a controller.

Libplanet.Store.Remote/Server/RemoteKeyValueService.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public override Task<ListKeysResponse> ListKeys(ListKeysRequest request, ServerCallContext context)
+        {
+            IEnumerable<KeyBytes> keys = _baseStore.ListKeys();
+            var response = new ListKeysResponse();
+            response.Keys.AddRange(keys.Select(KeyBytesExtensions.ToByteString));
+            _logger.LogTrace("ListKeys: {Count}", response.Keys.Count);
+            return Task.FromResult(response);
+        }
```
greymistcube, line 82: Not sure this should actually be implemented instead of just throwing a `NotSupportedException`. 🤔

Libplanet.Store.Remote/Server/RemoteKeyValueService.cs
riemannulus: Multiple `Set` gets `Dictionary`. On the other side, multiple `Delete` gets `List`.

riemannulus: https://github.com/planetarium/libplanet/pull/3688/files#diff-baf92b08d85437de7131f62dd9b70066064c72f21b241ca587248ef293c5833bR17 fyi.

CHANGES.md
```diff
@@ -35,6 +35,10 @@ To be released.
 ### Backward-incompatible storage format changes
 
 ### Added APIs
+ -  (Libplanet.Store.Remote) Introduce
+    `Libplanet.Store.Server.RemoteKeyValueService`  [[#3688]]
+ -  (Libplanet.Store.Remote) Introduce
+   `Libplanet.Store.Client.RemoteKeyValueStore`  [[#3688]]
```
greymistcube, line 41:
```suggestion

 -  (Libplanet.Store.Remote) Introduce
    `Libplanet.Store.Server.RemoteKeyValueService`  [[#3688]]
 -  (Libplanet.Store.Remote) Introduce
    `Libplanet.Store.Client.RemoteKeyValueStore`  [[#3688]]
```

Libplanet.Store.Remote/README.md
```diff
@@ -0,0 +1,73 @@
+Libplanet.Store.Remote
+=========
+A remote store service for Libplanet.
```
greymistcube, line 3:
```suggestion
Libplanet.Store.Remote

=========

A remote store service for Libplanet.
```

Also overall line spacing for the rest of the document for readability. 🙃

Libplanet.Store.Remote/Server/RemoteKeyValueService.cs
greymistcube: I understand. However, the term "collection" is underlying data type agnostic. I'd suggest either

```proto
message SetCollectionRequest {
    map<string, bytes> key = 1;
}

message DeleteCollectionRequest {
    repeated bytes key = 1;
}
```

or 

```proto
message SetMapRequest {
    map<string, bytes> map_field = 1;
}

message DeleteRepeatedRequest {
    repeated bytes repeated_field = 1;
}
```

for consistency. I prefer the former, but maybe I don't know enough about Protobuf's general conventions. 🙄

Libplanet.Store.Remote/Extensions/KeyValueStoreExtensions.cs
greymistcube: The way I see it, throwing an `Exception` is baked in to the model layer, as it is well documented in `IKeyValueStore` interface. If we look at it that way, as `Exception` could be considered as a valid "response" from the data model layer, I would say it is the controller's job to `catch` and convert that `Exception` to a serviceable response. It definitely isn't part of the current data model. 

One way to have so called "popping data logic" abstracted away would be to create a new `interface` or `class` that wraps the base `IKeyValueStore`, and pretend `RemoteKeyValueService` is to serve that abstracted model, but that would be an overkill and confusing. 🙄

I understand wanting to have the controller devoid of logic as much as possible, but I don't think having the transformation of the model behavior into a response just "floating around" as an extension method is necessarily a good idea.

Libplanet.Store.Remote/Libplanet.Store.Remote.csproj
```diff
@@ -0,0 +1,35 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net6.0</TargetFramework>
```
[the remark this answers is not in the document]
s2quake: `Grpc.AspNetCore` supports only net6.0 or greater, so `netstandard2.1` can not be used.
https://www.nuget.org/packages/Grpc.AspNetCore#supportedframeworks-body-tab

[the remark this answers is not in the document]
riemannulus: @s2quake right.

Libplanet.Store.Remote/README.md
riemannulus: d56b073137fdd560f2ea9dd1b8520cc02c89e429 amended

CHANGES.md
riemannulus: d56b073137fdd560f2ea9dd1b8520cc02c89e429 amended

Libplanet.Store.Remote/Server/RemoteKeyValueService.cs
riemannulus: https://cloud.google.com/apis/design/naming_convention?hl=ko#method_names

Ok, I found it. It said: `Use *VerbNone* naming convention`. So, I suggest...

```
message SetPairsRequest {
    map<string, bytes> map_field = 1;
}

message DeleteListRequest {
    repeated bytes repeated_field = 1;
}
```

How about this? @greymistcube

greymistcube: Looks good to me. 🙂

Libplanet.Store.Remote/Extensions/KeyValueStoreExtensions.cs
riemannulus: Check this method used in `RemoteKeyValueService`, not `RemoteKeyValueStore`. When throwing `Exception` on `RemoteKeyValueService.Get()` is a problem of the server response, not `IKeyValueStore`.

But yes, I understand what you said "floating around" means. I'll move this method into `RemoteKeyValueService`.

riemannulus: https://github.com/planetarium/libplanet/pull/3688/commits/9a4d4e08ee39a9a55f18346b34b1db5fa9ae3bd8 amended.

Libplanet.Store.Remote/Server/RemoteKeyValueService.cs
riemannulus: https://github.com/planetarium/libplanet/pull/3688/commits/60fb78868f5bf781e20ef631281fb048c2669955 amended. PTAL 🙏 @greymistcube

Libplanet.Store.Remote/Extensions/ByteStringExtensions.cs
s2quake, line 1: Some extension methods use a fixed @this name so that developers don't have to think about the name of the parameter. How about applying it to this code? I think that's good because it seems consistent.

Libplanet.Store.Remote/Libplanet.Store.Remote.csproj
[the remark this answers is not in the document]
s2quake: If this project only supports `net6.0`, it would be nice to add `net8.0` and use language version `12`. In the age of `net9.0`, isn't `net6.0` a relic? We need to go one step further.

Libplanet.Store.Remote.Tests/Helpers/LoggerHelper.cs
riemannulus: https://stackoverflow.com/questions/61544047/use-serilog-with-microsoft-extensions-logging-ilogger Ok, I found some information. use `Microsoft.Extensions.Logging` is ok even if the user wants to use `Serilog`.

Libplanet.Store.Remote/Extensions/ByteStringExtensions.cs
riemannulus: https://github.com/planetarium/libplanet/pull/3688/commits/da6ffe476c5294ce66b58f3608613c6aa317b227 Amended. PTAL 🙏

Libplanet.Store.Remote/Libplanet.Store.Remote.csproj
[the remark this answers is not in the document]
riemannulus: > it would be nice to add net8.0 and use language version 12.

The target of this service is a project which uses `Libplanet`. It means most projects use `net6.0` and language `8`. Therefore, we cannot bump the version immediately. @s2quake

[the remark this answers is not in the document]
riemannulus: Additionally, I tried bumping up to `net8.0` and lang version `12` to [`NineChronicles.Headless.`](https://github.com/planetarium/NineChronicles.Headless) But it causes a build error rn. I don't wanna bumping up for using this service. It should be another PR or task.

Libplanet.Store.Remote.Tests/Helpers/LoggerHelper.cs
greymistcube: Yes, I understand it's "doable", but the link above is about how to integrate `Serilog` to existing large codebase using `Microsoft.Extensions.Logging`. Our whole stack is based on `Serilog` and unless there is a specific reason to introduce using `Microsoft.Extensions.Logging`, this just increases the overall maintenance cost, which I don't see it being justified. 😐

riemannulus: AFAIK, `Microsoft.Extensions.Logging` is a dotnet-standard logging system. So in my opinion, the user can use only the serilog is not a proper way especially if it is a library.

What's your opinion about this?

greymistcube: > AFAIK, Microsoft.Extensions.Logging is a dotnet-standard logging system. So in my opinion, the user can use only the serilog is not a proper way especially if it is a library.

Yes, I understand your sentiment. If we are to choose, using `Microsoft.Extensions.Logging` could be an option, but I'd say this requires a larger consensus among the engineers working on this project. If we do decide to move away from `Serilog` and we set it as a goal, I would have less qualms about this change. If we are just introducing it, especially in a "subproject" of [libplanet](libplanet.io), this becomes an unmanaged technical debt.

riemannulus: Ok, I understand. I'll amend your request.

s2quake: For Asp users I think using Serilog can be inconvenient. I personally think it's better to create a Libplanet.ILogger and let the user decide which logger to use. 😀

riemannulus: https://github.com/planetarium/libplanet/pull/3688/commits/45d8987d803e10488d371e376804d698d6dbbdb6 amended. @greymistcube

greymistcube (approved):

PR #3696, opened 2024-03-15
Introduce evidence

s2quake:
### Features

- Definitions and inheritance evidence
- Collecting evidence during the consensus step
- Broadcasting evidence
- Evidence expiration
- Commit evidence
- Test code.

### DIFF
--- a/src/Libplanet.Action/ActionContext.cs
+++ b/src/Libplanet.Action/ActionContext.cs
@@ -4,6 +4,7 @@ using System.Collections.Immutable;
 using System.Threading;
 using Libplanet.Action.State;
 using Libplanet.Crypto;
+using Libplanet.Types.Evidence;
 using Libplanet.Types.Tx;
 
 namespace Libplanet.Action
@@ -25,7 +26,8 @@ namespace Libplanet.Action
             IWorld previousState,
             int randomSeed,
             long gasLimit,
-            IReadOnlyList<ITransaction>? txs = null)
+            IReadOnlyList<ITransaction>? txs = null,
+            IReadOnlyList<EvidenceBase>? evidence = null)
         {
             Signer = signer;
             TxId = txid;
@@ -36,6 +38,7 @@ namespace Libplanet.Action
             RandomSeed = randomSeed;
             _gasLimit = gasLimit;
             _txs = txs ?? ImmutableList<Transaction>.Empty;
+            Evidence = evidence ?? ImmutableList<EvidenceBase>.Empty;
 
             GetGasMeter.Value = new GasMeter(_gasLimit);
         }
@@ -68,6 +71,9 @@ namespace Libplanet.Action
         public IReadOnlyList<ITransaction> Txs => BlockAction
             ? _txs
             : ImmutableList<ITransaction>.Empty;
+
+        /// <inheritdoc cref="IActionContext.Evidence"/>
+        public IReadOnlyList<EvidenceBase> Evidence { get; }
 
         /// <inheritdoc cref="IActionContext.UseGas(long)"/>
         public void UseGas(long gas) => GetGasMeter.Value?.UseGas(gas);
--- a/src/Libplanet.Action/ActionEvaluator.cs
+++ b/src/Libplanet.Action/ActionEvaluator.cs
@@ -203,7 +203,8 @@ namespace Libplanet.Action
                     txs: block.Transactions,
                     previousState: prevState,
                     randomSeed: randomSeed,
-                    gasLimit: actionGasLimit);
+                    gasLimit: actionGasLimit,
+                    evidence: block.Evidence);
             }
 
             long gasLimit = tx?.GasLimit ?? long.MaxValue;
@@ -267,7 +268,8 @@ namespace Libplanet.Action
                     txs: inputContext.Txs,
                     previousState: newPrevState,
                     randomSeed: inputContext.RandomSeed,
-                    gasLimit: inputContext.GasLimit());
+                    gasLimit: inputContext.GasLimit(),
+                    evidence: inputContext.Evidence);
             }
 
             try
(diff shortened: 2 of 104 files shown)


src/Libplanet.Action/ActionEvaluator.cs
```diff
@@ -267,7 +268,8 @@ IActionContext CreateActionContext(IWorld newPrevState)
                     txs: inputContext.Txs,
                     previousState: newPrevState,
                     randomSeed: inputContext.RandomSeed,
-                    gasLimit: inputContext.GasLimit());
+                    gasLimit: inputContext.GasLimit(),
+                    evidences: inputContext.Evidences);
```
riemannulus, line 272: How do we collect evidence in action?

src/Libplanet.Action/ActionContext.cs
```diff
@@ -69,6 +72,9 @@ internal class ActionContext : IActionContext
             ? _txs
             : ImmutableList<ITransaction>.Empty;
 
+        /// <inheritdoc cref="IActionContext.Evidences"/>
+        public IReadOnlyList<Evidence> Evidences { get; }
```
riemannulus, line 76: Just `Evidence` stands for both plural and singular.

src/Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -1454,12 +1473,221 @@ public override void DeleteNextStateRootHash(BlockHash blockHash)
...
 
+        /// <inheritdoc/>
+        public override IEnumerable<EvidenceId> IteratePendingEvidenceIds()
+        {
+            foreach (Iterator it in IterateDb(_pendingEvidenceDb, PendingEvidenceKeyPrefix))
+            {
+                byte[] key = it.Key();
+                byte[] idBytes = key.Skip(PendingEvidenceKeyPrefix.Length).ToArray();
+                yield return new EvidenceId(idBytes);
+            }
```
riemannulus, line 1484: Can we use [prefix seek](https://github.com/facebook/rocksdb/wiki/Prefix-Seek) for performance?

src/Libplanet.RocksDBStore/RocksDBStore.cs
```diff
@@ -1454,12 +1473,221 @@ public override void DeleteNextStateRootHash(BlockHash blockHash)
...
+            catch (Exception e)
+            {
+                LogUnexpectedException(nameof(PutCommittedEvidence), e);
+                throw;
+            }
+            finally
+            {
+                _rwEvidenceLock.ExitWriteLock();
+            }
```
riemannulus, line 1642: Maybe we can store evidence separately with a block. like transactions.

src/Libplanet.Action/ActionContext.cs
s2quake: There are a lot of `evidences` that are used as identifiers (field, property). Should I rename them all to evidence?

riemannulus: IMO It is not a big deal when using `Find and replace`...

src/Libplanet.RocksDBStore/RocksDBStore.cs
s2quake: The store-related code was implemented by @OnedgeLee, so I have no understanding of this part. It will take some time to improve it.

src/Libplanet.Action/ActionEvaluator.cs
s2quake: Custom evidence can be added to the evidence pool of a `BlockChain` from the external layer of a library or application. However, the blockchain object cannot be accessed from within an action. Therefore, currently, there is no functionality to directly collect evidence within an action. If evidence can be collected based on failed transactions in a `BlockRender` and added to the `Blockchain`, it would be possible.

src/Libplanet.Action/ActionContext.cs
s2quake: By the way, the namespace is `Evidences`. How should this be handled?

src/Libplanet.Action/ActionEvaluator.cs
riemannulus: Then why need `Evidences` in `InputContext`?

src/Libplanet.Action/ActionContext.cs
riemannulus: Change needed I guess..?

s2quake: Then it will overlap with the `Evidence` class.

src/Libplanet.Action/ActionEvaluator.cs
s2quake: It is not needed now, but this property is required to execute the slash code in DPoS.

riemannulus: 👌

src/Libplanet.Action/ActionContext.cs
s2quake: - Changed namespace from `Evidences` to `Evidence`.
- Changed class name from `Evidence` to `EvidenceBase`.
- Changed key (used for serialisation, like bencode) from `evidences..` to `evidence..`.

riemannulus (approved):

src/Libplanet.RocksDBStore/RocksDBStore.cs
s2quake: This issue has been postponed.

src/Libplanet.RocksDBStore/RocksDBStore.cs
s2quake: This issue has been postponed.

PR #3697, opened 2024-03-19
Move FAV related operations from `IAccount` to `IWorld`

greymistcube:
Partially addresses #3694.

Things todo in a follow-up:
- Move `GetValidatorSet()` and `SetValidator()`
- Refactor `IWorldDelta`


Libplanet.Action/State/AccountDiff.cs
```diff
@@ -25,32 +24,22 @@ namespace Libplanet.Action.State
...
     ///     <item><description>
-    ///         Any <see langword="null"/> value, again, which is equivalent to non-existent value
-    ///         in the underlying storage, in the target for <see cref="FungibleAssetValue"/>
-    ///         and <see cref="ValidatorSet"/> is interpreted accordingly.  That is,
-    ///         0 amount of <see cref="FungibleAssetValue"/> and empty <see cref="ValidatorSet"/>
-    ///         are used.  This is in accordance with how <see cref="IAccountState.GetBalance"/>
-    ///         and <see cref="IAccountState.GetValidatorSet"/> would behave.
+    ///         This only shows the difference in "states" manipulated using
+    ///         <see cref="IAccount.SetState"/> and <see cref="IAccount.RemoveState"/>.  That is,
+    ///         this does care about differences in <see cref="FungibleAssetValue"/>s.
```
OnedgeLee, line 29: Are there any alternative way to compare differences in `FAV`s?

greymistcube: It is technically possible, but I don't really see the point. This isn't really being used anyway. Backward compatibility gets in the way in a big way since `ITrie` is completely data model agnostic. 🙄

OnedgeLee: When I think about it again, it feels closer to anxiety about what was gone. I think we can make it again if necessary later.

greymistcube: To be more specific, I'm working with the assumption that eventually we'd only use `KeyByte`s derived from `Address`es and possibly empty `KeyByte` (for metadata) as keys.

OnedgeLee (approved):

PR #3700, opened 2024-03-20
Introduce `ConsensusCryptoBackend`, `Proof`, `Lot`

OnedgeLee:
### To use VRF on consensus, three features are added.

1. ConsensusCryptoBackend : Implements ECVRF functionalities.
2. Proof : Wrapper class of VRF proof that contains not only verification of proof but also sortition logic.
3. Lot : Represents lots from each validators to participate in the sortition of proposer. Includes payload of proof, public key, and proof.

Scenario is as follows:

1. Each validator sends an `PreEvaluationBlock` with `Lot`. (`PreEvaluationBlock` contains `Proof` that is same to the `Lot.Proof`)
2. Each validator decides the proposer, and set proper `PreEvaluationBlock` as an proposed pre-evaluation block, and starts `Propose` step.
3. Proposer and validators evaluates `PreEvaluationBlock` simultaneously on the `Propose` step.


riemannulus (commented):
We need note to `CHANGES.md` about this.

Libplanet.Types/Consensus/Lot.cs
```diff
@@ -0,0 +1,113 @@
...
+        public PublicKey PublicKey { get; }
+
+        /// <inheritdoc cref="ILot.Proof"/>
+        public Proof Proof { get; }
+
+        [JsonIgnore]
+        public IValue Bencoded
+            => ((Dictionary)_metadata.Bencoded)
+                .Add(PublicKeyKey, PublicKey.Format(true))
+                .Add(ProofKey, Proof.ByteArray);
```
riemannulus, line 73: Seems need `Serialize`-`Deserialize` tests.

Libplanet.Types/Consensus/LotMetadata.cs
```diff
@@ -0,0 +1,118 @@
...
+            get
+            {
+                Dictionary bencoded = Dictionary.Empty
+                    .Add(HeightKey, Height)
+                    .Add(RoundKey, Round)
+                    .Add(LastProofKey, LastProof.ByteArray);
+
+                return bencoded;
+            }
+        }
```
riemannulus, line 86: Seems need `Serialize`-`Deserialize` tests.

OnedgeLee:
> We need note to `CHANGES.md` about this.

Noted!

PR #3712, opened 2024-03-27
Implementation of service provision to `IAction` via `ActionContext`.

s2quake:
`Action`, `BeginAction`, `EndAction` is implemented as the same `IAction`, 
but they have different purposes. Therefore, they have slightly different information and permissions.
Currently, `IActionContext` provides only the same information, so different information and permissions cannot be granted for each Action.
`IServiceProvider` is an interface that can solve this problem.
Each action can use `IServiceProvider.GetService` to get the information it needs.
And `ActionEvaluator` has the obligation to provide different information depending on the purpose of the Action.


riemannulus:
How do you think `Services` should be retrieved from Application?

riemannulus:
Note: I suggested https://github.com/riemannulus/ActionEvaluatorExp/blob/main/ActionEvaluatorExp.Modules/Bank/Actions/Transfer.cs before and I think your suggestion is also like this.

s2quake:
The code below is a simple example of retrieving services from application.
```csharp
sealed class SlashingAction : IAction
{
    IWorld IAction.Execute(IActionContext context)
    {
        var world = context.PreviousState;
        if (context.GetService(typeof(IEvidenceContext)) is IEvidenceContext evidenceContext)
        {
            // Console.WriteLine("SlashingAction executed!");
        }
        else
        {
            throw new NotSupportedException();
        }
        return world;
    }
}
```

This PR shows the fundamental way of providing information to `IAction` in libplanet.
I haven't thought about how to change the state or provide additional modules or services needed in the application.
Finally, Both are good ways, so it would be nice to think about it from various aspects.

PR #3713, opened 2024-03-27
✨ ♻️ Added `Txs` property to `IActionContext`

greymistcube:
In preparation to remove `TotalUpdatedFungibleAssets` from `IWorld`.
Having volatile non-state dependency in two different places, namely `IActionContext` and `IWorld`, is causing too much trouble. 🙄

Also since `ActionEvaluator` is doing more and more heavy lifting in terms of context management, I've widened the scope from which its evaluating methods can pull necessary context from (it was also to avoid adding an additional parameter to numerous methods). This might come in handy later when we introduce evidences. 😗

### DIFF
--- a/Libplanet.Action/ActionEvaluator.cs
+++ b/Libplanet.Action/ActionEvaluator.cs
@@ -202,6 +202,7 @@ namespace Libplanet.Action
                     miner: block.Miner,
                     blockIndex: block.Index,
                     blockProtocolVersion: block.ProtocolVersion,
+                    txs: block.Transactions,
                     previousState: prevState,
                     randomSeed: randomSeed,
                     gasLimit: actionGasLimit);
@@ -265,6 +266,7 @@ namespace Libplanet.Action
                     miner: inputContext.Miner,
                     blockIndex: inputContext.BlockIndex,
                     blockProtocolVersion: inputContext.BlockProtocolVersion,
+                    txs: inputContext.Txs,
                     previousState: newPrevState,
                     randomSeed: inputContext.RandomSeed,
                     gasLimit: inputContext.GasLimit());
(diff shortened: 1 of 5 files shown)


Libplanet.Action/ActionEvaluator.cs
```diff
@@ -555,8 +554,6 @@ internal IWorld PrepareInitialDelta(HashDigest<SHA256>? stateRootHash)
                     exception: evaluation.Exception);
                 committedEvaluations.Add(committedEvaluation);
 #pragma warning restore SA1118
-
-                trie = evaluation.OutputState.Trie;
```
riemannulus, line 559: is it intended?

riemannulus (commented):
We need to add a condition for adding a value to `Txs` to `Behavioral changes`.

Libplanet.Action/ActionEvaluator.cs
greymistcube: It seems like an artifact from the past? It isn't really being used.

greymistcube:
@riemannulus 
Can it be considered a behavioral change if it wasn't there in the first place? 🤔
The description of the property explains what it's supposed to do, but I don't think of it as being *changed*.

Libplanet.Action/ActionEvaluator.cs
riemannulus: Would it be a problem if you didn't put the previous `ITrie` into the next state?

riemannulus:
> The description of the property explains what it's supposed to do, but I don't think of it as being changed.

Fair enough.

Libplanet.Action/ActionEvaluator.cs
greymistcube: This simply converts `IActionEvaluation`s to `ICommittedActionEvaluation`s. Each `IActionEvaluation` already has input and output `ITrie`s. I can't remember exactly why reassigning next `ITrie` was used at the moment. 🙄

greymistcube: Note that `trie` variable was not being referenced in the first place. 🙃

riemannulus (approved):

PR #3729, opened 2024-04-05
⬆️ Bump Litedb to 5.0.15

greymistcube:
⬆️ Older versions have a known critical security vulnerability. See [Here](https://nvd.nist.gov/vuln/detail/CVE-2022-23535).

I haven't thoroughly gone through the changes, but the following options have been removed:
- `journal`: Seems like this option has been entirely removed by [LiteDB]. I presume this is controlled automatically depending on the type of disk used.
- `indexCahceSize`: This has been changed to a constant in [LiteDB] with no control possible on our end.
- `flush`: Seems like the option has been taken away in [LiteDB] and always flushes to disk when a disk is being used.

PS. Newer versions such as 5.0.17 and 5.0.19 doesn't work properly for some reason and I have failed to resolve the issue. 😥

[LiteDB]: https://github.com/mbdavid/LiteDB

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,7 +41,7 @@ with `allowNegativeBalance` as `true`.
 ### Dependencies
 
  -  Upgraded *LiteDB* from [4.1.4][LiteDB 4.1.4] to
-    [5.0.19][LiteDB 5.0.19].  [[#3729]]
+    [5.0.15][LiteDB 5.0.15].  [[#3729]]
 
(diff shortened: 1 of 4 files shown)


riemannulus:
Benchmark failed 😢

greymistcube:
@riemannulus 
Some performance change is to be expected since there was a security patch. I'd say patching the security hole takes precedence over performance. 🙄

PR #3730, opened 2024-04-05
Introduce `IAccount.SetValidatorSet()` method

limebell:
Actually, just opens private method to public.

### DIFF
--- a/Libplanet.Tests/Action/AccountTest.cs
+++ b/Libplanet.Tests/Action/AccountTest.cs
@@ -399,6 +399,42 @@ namespace Libplanet.Tests.Action
             Assert.Equal(0, delta.GetValidatorSet().TotalCount);
         }
 
+        [Fact]
+        public virtual void SetValidatorSet()
+        {
+            var initCount = _keys.Length;
+            var key3 = new PrivateKey().PublicKey;
+            var key4 = new PrivateKey().PublicKey;
+
+            IAccount delta = _initAccount;
+            // delta already has 3 validators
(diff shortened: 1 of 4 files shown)


OnedgeLee (approved):

Libplanet.Tests/Action/AccountTest.cs
s2quake, line 1: To be more certain, it would be nice to have a test code that verifies that the removed validator is not included, like the following

```csharp
            Assert.DoesNotContain(
                delta.GetValidatorSet().Validators,
                v => v.PublicKey.Equals(removedValidator.PublicKey));
```

limebell: > To be more certain, it would be nice to have a test code that verifies that the removed validator is not included, like the following
> 
> ```cs
>             Assert.DoesNotContain(
>                 delta.GetValidatorSet().Validators,
>                 v => v.PublicKey.Equals(removedValidator.PublicKey));
> ```

Ameded!

s2quake (approved):

Libplanet.Tests/Action/AccountTest.cs
```diff
@@ -399,6 +399,42 @@ public virtual void SetValidator()
             Assert.Equal(0, delta.GetValidatorSet().TotalCount);
         }
 
+        [Fact]
+        public virtual void UpdateValidator()
```
greymistcube, line 403:
```suggestion
        public virtual void SetValidatorSet()
```

greymistcube (approved):

PR #3745, opened 2024-04-16
✨ Introduce `ValidatorSetAccount`

greymistcube:
Similar to how `ITrie` migrated to `IWorld.Modern`, this introduces migration to a new backend model where validator sets are stored in a separate `IAccount` specifically for `ValidatorSet`.

### DIFF
--- a/Libplanet.Action/State/ValidatorSetAccount.cs
+++ b/Libplanet.Action/State/ValidatorSetAccount.cs
@@ -10,19 +10,19 @@ namespace Libplanet.Action.State
         public static readonly Address ValidatorSetAddress =
             new Address("1000000000000000000000000000000000000000");
 
-        public ValidatorSetAccount(ITrie trie, int worldTrieVersion)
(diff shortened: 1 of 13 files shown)


Libplanet.Action/State/ValidatorSetAccount.cs
```diff
@@ -0,0 +1,60 @@
+using Libplanet.Crypto;
+using Libplanet.Store.Trie;
+using Libplanet.Types.Blocks;
+using Libplanet.Types.Consensus;
+
+namespace Libplanet.Action.State
+{
+    public class ValidatorSetAccount
```
riemannulus, line 8: Could you add a more detailed comment?

greymistcube: Eh, sure. I just added a short description. Technically, this isn't supposed to be `public` considering how it is implemented. 🙄

OnedgeLee (approved):

PR #3764, opened 2024-04-22
Add BlockAction flag to ActionContext

limebell:
Previously `ActionContext.BlockAction` was determined by nullability of `TxId` field, became to take argument directly.

Fixed bug where TxActions' signer was block's proposer, not tx's.

Skips changelog because all API changes are internal.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,8 @@ Version DPoS
  -  (Libplanet.Action) `ActionEvaluator` constructor requires
     `PolicyActionsRegistry` parameter instead of the
     `policyBlockActionGetter` parameter.  [[#3701], [#3748]]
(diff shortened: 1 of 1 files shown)


s2quake (commented):
It is common practice in .NET to prefix a bool type identifier with 'Is'. For example, refer to the following link: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names.
The link does not specifically mention identifier names for bool types, but the example code in the link uses the `is` prefix. Your style preference may differ, but I would suggest using `isBlockAction`. `blockAction` seems like an instance of an action to me.

Libplanet.Action/ActionEvaluator.cs
s2quake, line 1: line 268, parameter `bool blockAction` needs to be renamed. 😀

s2quake (approved):

PR #3766, opened 2024-04-23
Add test for evaluation when policy actions fails

limebell:


Libplanet.Tests/Action/ActionEvaluatorTest.cs
```diff
@@ -294,6 +294,55 @@ public void EvaluateWithPolicyActions()
...
+
+            (_, Transaction[] txs) = MakeFixturesForAppendTests();
+            var block = chain.ProposeBlock(
+                GenesisProposer, txs.ToImmutableList(), CreateBlockCommit(chain.Tip));
+            var evaluations = actionEvaluator.Evaluate(
+                block, chain.Store.GetStateRootHash(chain.Tip.Hash)).ToArray();
+
+            // BeginBlockAction + (BeginTxAction + #Action + EndTxAction) * #Tx + EndBlockAction
+            Assert.Equal(
+                4 + txs.Length * 4 + txs.Aggregate(0, (sum, tx) => sum + tx.Actions.Count),
```
s2quake, line 339:
```suggestion
                4 + txs.Length * 4 + txs.SelectMany(item => item.Actions).Count(),
```
or

```suggestion
                4 + txs.Length * 4 + txs.Sum(item => item.Actions.Count),
```

I think it's better to use `Sum` instead of `Aggregate` because the test project targets the .NET 6.0 framework.

PR #3768, opened 2024-04-27
Sloth update for Libplanet.Net

OnedgeLee:
## Apply changes for version `Sloth`, on `Libplanet.Net` level

- Update `Swarm.AppendPreviousBlocks()`.
  - Affects preloading and block synchronization
- Update `Context.IsValid()`
  - Affects block appending via consensus

---

Found malfunction under time-consuming action evaluation, so converted as a draft temporarily

---

### Ramaining job
- Adjust timing (timeouts, block interval)
- Rebasing

### DIFF
--- a/Libplanet.Net/Consensus/ConsensusContext.cs
+++ b/Libplanet.Net/Consensus/ConsensusContext.cs
@@ -440,7 +440,9 @@ namespace Libplanet.Net.Consensus
                 height,
                 _privateKey,
                 _blockChain
-                    .GetNextWorldState(_blockChain[Height - 1].Hash)!
+                    .GetWorldState(
+                        _blockChain.GetNextStateRootHash(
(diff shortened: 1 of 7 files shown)


Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -434,14 +435,30 @@ private void OnTipChanged(object? sender, (Block OldTip, Block NewTip) e)
         private Context CreateContext(long height)
         {
             // blockchain may not contain block of Height - 1?
-            var context = new Context(
+            // blockchain may not contain block of Height - 1?
```
limebell, line 438: Maybe duplication?

OnedgeLee: Amended!

limebell (approved):

PR #3779, opened 2024-05-07
✨ Add `CurrencyAccount` to handle all `Currency` related logic

greymistcube:
The implementation is pretty convoluted at the moment. 🙄
There are numerous issues regarding `Currency`:

- It isn't at all clear who should be responsible of keeping a state "consistent" regarding a `Currency`.
  - "Consistency" is purely enforced by code, not stored data.
  - The whole `Currency` manipulation pipeline is running on top of the assumption that a state is "consistent"
  - There are two relations that can be checked purely from state + `Currency`:
    - Whether `TotalSupply` does not exceed `MaximumSupply`
    - Whether `TotalSupply` is being tracked and is **correct**.
- `TotalSupplyTrackable` loses its meaning starting with `BlockMetadata.CurrencyAccountProtocolVersion`.
- Read operation throwing an `Exception` for `TotalSupply`.
  - It is put in as a safety measure, but in reality, this does not achieve anything.
  - Handling of said `Exception` regarding `TotalSupply` must be handled from `IAction` side (or any outside service), and its interpretation requires additional knowledge about `Currency` and the behavior of read operation of `TotalSupply`.
  - Besides, it has never been used.
- Mint/burn authority validation has nothing to do with the underlying state.
  - One solution would be to enforce business logic through code analysis where a call to `MintAsset()` or `BurnAsset()` should be preceded by authority validation.
  - Falling short of that, the responsibility falls to the developers of `IAction`s.

This is partially due to information about `Currency` not being recorded in a state and migration to `CurrencyAccount` model
does not permit adding in this missing information.

I'll be making some `Currency` spec and API changes in separate PRs to mitigate some of these issues. 😶

### DIFF
--- a/Libplanet.Action/State/CurrencyAccount.cs
+++ b/Libplanet.Action/State/CurrencyAccount.cs
@@ -12,7 +12,7 @@ namespace Libplanet.Action.State
     /// A special "account" for managing <see cref="Currency"/> starting with
     /// <see cref="BlockMetadata.CurrencyAccountProtocolVersion"/>.
     /// </summary>
-    public class CurrencyAccount
+    public sealed class CurrencyAccount
     {
         /// <summary>
         /// The <see cref="Address"/> location within the account where
--- a/Libplanet.Action/State/CurrencyAccount.cs
+++ b/Libplanet.Action/State/CurrencyAccount.cs
@@ -1,0 +1,320 @@
+using System;
+using System.Numerics;
+using Bencodex.Types;
+using Libplanet.Crypto;
+using Libplanet.Store.Trie;
+using Libplanet.Types.Assets;
+using Libplanet.Types.Blocks;
+
+namespace Libplanet.Action.State
+{
+    /// <summary>
+    /// A special "account" for managing <see cref="Currency"/> starting with
+    /// <see cref="BlockMetadata.CurrencyAccountProtocolVersion"/>.
+    /// </summary>
+    public class CurrencyAccount
+    {
+        /// <summary>
+        /// The <see cref="Address"/> location within the account where
+        /// the total supply of the currency gets stored.
+        /// </summary>
+        public static readonly Address TotalSupplyAddress =
+            new Address("1000000000000000000000000000000000000000");
+
+        public CurrencyAccount(ITrie trie, int worldVersion, Currency currency)
+        {
+            Trie = trie;
+            WorldVersion = worldVersion;
+            Currency = currency;
+        }
+
+        public ITrie Trie { get; }
+
+        public int WorldVersion { get; }
+
+        public Currency Currency { get; }
+
+        public FungibleAssetValue GetBalance(Address address, Currency currency)
+        {
+            if (!Currency.Equals(currency))
+            {
+                throw new ArgumentException();
+            }
+
+            return FungibleAssetValue.FromRawValue(
+                Currency,
+                WorldVersion >= BlockMetadata.CurrencyAccountProtocolVersion
+                    ? GetRawBalanceV7(address)
+                    : GetRawBalanceV0(address));
+        }
+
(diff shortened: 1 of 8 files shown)


Libplanet.Action/State/CurrencyAccount.cs
s2quake, line 1: How do you think about the following pattern?

```csharp
abstract class CurrencyAccountVerifier
{
}
sealed class CurrencyAccountVerifierV0 : CurrencyAccountVerifier
{
}
sealed class CurrencyAccountVerifierV7 : CurrencyAccountVerifier
{
}
sealed class CurrencyAccount
{
    private readonly CurrencyAccountVerifier _verifier
        = WorldVersion >= BlockMetadata.CurrencyAccountProtocolVersion ? 
            new CurrencyAccountVerifierV7() : CurrencyAccountVerifierV0();
}
```

greymistcube: Can you elaborate on this? What is this for exactly? 🤔

OnedgeLee: As I understood, seems like creating two inner `CurrencyAccount` named `CurrencyAccountVerifierVx` and using them instead of private methods, though naming `Verifier` is confusing. Not sure I'm properly understood.

s2quake: It is to simplify the code of CurrencyAccount and separate the verification code into a separate class or file.
and to minimize the modification of CurrencyAccount code when a new version comes out and to easily add version verification code.

If the above pattern is used, the code would look like the following.

```csharp
public FungibleAssetValue GetBalance(Address address, Currency currency)
{
    CheckCurrency(currency);
    return FungibleAssetValue.FromRawValue(
        Currency,
        _verfier.GetRawBalanceV7(address));
}
```

s2quake: Sorry. It seems that I used the name `verifier` incorrectly. It was just to represent a class that contains functions used for each version.

greymistcube: We can maybe introduce `CurrencyAccountImplV0` and `CurrencyAccountImplV7`. I'm not so sure about using `abstract` and inheritance. As for having separate files and current code being more prone to mistakes, I agree. It's partially this way due to possible upcoming spec changes. 😐

Libplanet.Action/State/CurrencyAccount.cs
```diff
@@ -0,0 +1,400 @@
...
+using Libplanet.Types.Assets;
+using Libplanet.Types.Blocks;
+
+namespace Libplanet.Action.State
+{
+    /// <summary>
+    /// A special "account" for managing <see cref="Currency"/> starting with
+    /// <see cref="BlockMetadata.CurrencyAccountProtocolVersion"/>.
+    /// </summary>
+    public class CurrencyAccount
```
s2quake, line 15: I think there is no reason for this class to be inherited, if it is, this class should be sealed.

OnedgeLee (approved):
Do this PR contains state migration test? Or, existing test can cover currency migration?

Libplanet.Action/State/CurrencyAccount.cs
greymistcube: Applied 🙂

greymistcube:
@OnedgeLee 

> Do this PR contains state migration test? Or, existing test can cover currency migration?

Migration is mostly covered by `ActionEvaluatorTest.Migration.cs`. Although now that I look at it, I should add `Currency` manipulation to the action used in the test. 😗

greymistcube:
@OnedgeLee 

I've added an additional test regarding migration. See 4e5fc2fd6d08c8798f3e71dfdbe28b9eb8e93c29.

OnedgeLee (approved):

PR #3798, opened 2024-05-23
♻️ Moved balance and validator query from account to world

### LINKED ISSUE #3792: Fix `Currency` and `ValidatorSet` related GQL queries
`Currency` and `ValidatorSet` related queries under `WorldStateType` ➡️ `AccountType` are no longer applicable. Should be moved to be directly under `WorldStateType`.

greymistcube:
Closes #3792.

♻️ The old scheme is no longer applicable as currency and validator set handling is done on `IWorld` layer instead of `IAccount` layer. This changes the API. Let's just hope this doesn't break anything. 😣

Also, we need more tests to see if the queries work on both an old state and a new state. 😥

### DIFF
--- a/Libplanet.Explorer/GraphTypes/AccountStateType.cs
+++ b/Libplanet.Explorer/GraphTypes/AccountStateType.cs
@@ -3,7 +3,6 @@ using System.Security.Cryptography;
 using GraphQL;
 using GraphQL.Types;
 using Libplanet.Action.State;
-using Libplanet.Common;
 using Libplanet.Crypto;
 using Libplanet.Store.Trie;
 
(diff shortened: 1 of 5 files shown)


Libplanet.Explorer/GraphTypes/AccountStateType.cs
```diff
@@ -83,112 +82,6 @@ public AccountStateType()
                         .Select(address => context.Source.GetState(address))
                         .ToArray()
             );
-
-            Field<IValueType>(
```
riemannulus, line 87: Maybe we should check ok to remove those types. Previously, we had a problem as I remember...

greymistcube: Arg... you are right. I'll revert it back.
However, the solution to this predicament can't be "just keep adding stuff". 😕

greymistcube: Reverted.

PR #3808, opened 2024-06-05
Minor fixes for Sloth

OnedgeLee:
1. Regenerate NextStateRootHash on construct BlockChain
2. Fire-and-forget block appending
3. Rename protocol version for sloth

### DIFF
--- a/Libplanet.Types/Blocks/BlockMetadata.cs
+++ b/Libplanet.Types/Blocks/BlockMetadata.cs
@@ -82,7 +82,14 @@ namespace Libplanet.Types.Blocks
         /// </summary>
         public const int CurrencyAccountProtocolVersion = 7;
 
-        public const int StateRootHashPostponeProtocolVersion = 8;
+        /// <summary>
+        /// The starting protocol version where <see cref="Block.StateRootHash"/> is
+        /// postponed by one blocks.
+        /// Meaning of <see cref="Block.StateRootHash"/> has been changed from
+        /// the result state reference of <see cref="Block.Transactions"/> to state reference
+        /// before <see cref="Block.Transactions"/> were executed.
+        /// </summary>
+        public const int SlothProtocolVersion = 8;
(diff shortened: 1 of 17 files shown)


Libplanet.Types/Blocks/BlockMetadata.cs
```diff
@@ -82,7 +82,14 @@ public class BlockMetadata : IBlockMetadata
...
 
-        public const int StateRootHashPostponeProtocolVersion = 8;
+        /// <summary>
+        /// The starting protocol version where <see cref="Block.StateRootHash"/> is
+        /// postponed by one blocks.
+        /// Meaning of <see cref="Block.StateRootHash"/> has been changed from
+        /// the result state reference of <see cref="Block.Transactions"/> to state reference
+        /// before <see cref="Block.Transactions"/> were executed.
+        /// </summary>
+        public const int SlothProtocolVersion = 8;
```
greymistcube, line 92: This should probably be in the changelog.

Libplanet/Blockchain/BlockChain.cs
```diff
@@ -182,6 +182,21 @@ public partial class BlockChain : IBlockChainStates
                     stored: Genesis.Hash
                 );
             }
+
+            if (Tip.ProtocolVersion < BlockMetadata.SlothProtocolVersion)
+            {
+                return;
+            }
```
greymistcube, line 189: Can't we just put NSRH regardless of its protocol version? There are no safety measures anyway. We might as well make the definition of NSRH as "the executed result of a block". 😶

OnedgeLee: Since then, name `NSRH` seems to be weird a bit(For protocol version lower than sloth). I still think about that, but suspended those kind of thinking to after 5.0.0 release.

greymistcube (dismissed):

PR #3813, opened 2024-06-10
Make `IVoteMetadata.ValidatorPower` nullable

limebell:
To prevent `KeyNotFoundException` while deserializing old blocks, make field nullable.

### DIFF
--- a/Libplanet.Types/Consensus/IVoteMetadata.cs
+++ b/Libplanet.Types/Consensus/IVoteMetadata.cs
@@ -39,7 +39,7 @@ namespace Libplanet.Types.Consensus
         /// <summary>
         /// The voting power of the validator that voted.
         /// </summary>
-        BigInteger ValidatorPower { get; }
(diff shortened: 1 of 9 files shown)


Libplanet.Types/Consensus/IVoteMetadata.cs
```diff
@@ -39,7 +39,7 @@ public interface IVoteMetadata
         /// <summary>
         /// The voting power of the validator that voted.
         /// </summary>
-        BigInteger ValidatorPower { get; }
+        BigInteger? ValidatorPower { get; }
```
riemannulus, line 42: Don't need any changelog change about this?

limebell: Amended in https://github.com/planetarium/libplanet/pull/3813/commits/9e25244f1bd5eeef21a5e6479ef346540e5b9223

PR #3837, opened 2024-06-20
Remove Vote.Verify()

OnedgeLee:
### Context

1. `Vote` is verified on the constructor
2. `Vote` always have non-null `ValidatorPublicKey`.


OnedgeLee:
Seems to be fail on `Libplanet.Net.Tests.Consensus.ContextNonProposerTest.EnterPreVoteNilOneThird`

PR #3842, opened 2024-06-20
♻️ In memory "next state root hash"

greymistcube:
The general meaning of "next state root hash" has changed to be simpler:
- The next state root hash of a `Block` is meant as the state root hash resulting from an execution of the `Block` **regardless** of its protocol version number.
  - The next state root hash of a pre-Sloth `Block` is its state root hash, i.e. `Block.StateRootHash`.
  - The next state root hash of a post-Sloth `Block` is the state root hash **after** its evaluation.

On top of this, the next state root hash of any `Block` is considered ephemeral until it can be derived from the `BlockChain` itself:
- Note that any `Block` that is either pre-Sloth or that is included in `BlockChain` that is not the `Tip` can be fetched from the `BlockChain`.
  - This is notion is dependent upon the general assumption that the `BlockChain` itself **is always valid**.
- The next state root hash of the `Tip` **may** be missing **only if** the `Tip` is a post-Sloth `Block` and it is still under evaluation.

I have (temporarily) moved a possible lock inside `GetNextStateRootHash()` to a more appropriate place, i.e. `ConsensusContext`. This doesn't actually resolve the (issue)[#3833], but should be handled separately.

Internally, the next state root hash of the `BlockChain` (i.e., `BlockChain.GetNextStateRootHash()`) is kept up to date by managing it through `BlockChain()`, `Fork()`, and `Swap()`.


Libplanet.Net.Tests/Consensus/ContextTest.cs
```diff
@@ -321,7 +323,7 @@ public async Task CanPreCommitOnEndCommit()
                 1L,
                 TestUtils.PrivateKeys[0],
                 blockChain
-                    .GetNextWorldState(blockChain[0L].Hash)!
+                    .GetNextWorldState(0)
```
OnedgeLee, line 326: For consistency,
```suggestion
                    .GetNextWorldState(0L)
```

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -452,20 +452,28 @@ private Context CreateContext(long height)
         {
             // blockchain may not contain block of Height - 1?
             ValidatorSet validatorSet;
-            if (_blockChain.Tip.ProtocolVersion
-                < BlockMetadata.SlothProtocolVersion)
+            if (_blockChain.Tip.ProtocolVersion < BlockMetadata.SlothProtocolVersion)
             {
                 validatorSet = _blockChain
                     .GetWorldState(_blockChain[Height - 1].StateRootHash)
```
OnedgeLee, line 458: The operation would be same, but using parameter would be better for the meaning, IMO.
Or, we have to remove the parameter of this method I think.
```suggestion
                    .GetWorldState(_blockChain[height - 1].StateRootHash)
```

Libplanet.Net/Consensus/ConsensusContext.cs
```diff
@@ -452,20 +452,28 @@ private Context CreateContext(long height)
...
             else
             {
-                validatorSet = _blockChain
-                    .GetWorldState(
-                        _blockChain.GetNextStateRootHash(
-                            _blockChain[Height - 1].Hash, TimeSpan.Zero))
-                    .GetValidatorSet();
+                while (true)
+                {
+                    var nextStateRootHash = _blockChain.GetNextStateRootHash(Height - 1);
```
OnedgeLee, line 465: Same as above.
```suggestion
                    var nextStateRootHash = _blockChain.GetNextStateRootHash(height - 1);
```

Libplanet.Tests/BlockChainExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <returns>An <see cref="IWorldState"/> resulting from the execution of
+        /// a <see cref="Block"/> associated with given <paramref name="blockHash"/>.</returns>
+        public static IWorldState GetNextWorldState(
+            this BlockChain blockChain,
+            BlockHash blockHash) =>
+            blockChain.GetNextStateRootHash(blockHash) is HashDigest<SHA256> stateRootHash
+                ? blockChain.GetWorldState(stateRootHash)
+                : null;
+    }
+}
```
OnedgeLee, line 53: Isn't returning `IWorldState` on above methods nullable?

Libplanet.Tests/Action/WorldTest.cs
```diff
@@ -505,9 +505,7 @@ public virtual void TotalSupplyTracking()
         }
 
         protected static IWorldState GetLatestWorldState(BlockChain blockChain) =>
-            blockChain.Tip.ProtocolVersion < BlockMetadata.SlothProtocolVersion
-                ? blockChain.GetWorldState()
-                : blockChain.GetNextWorldState();
+            blockChain.GetNextWorldState();
```
OnedgeLee, line 508: Do result of this method always exists, though on asynchronous condition?
(I know we'll use synchronous method on tests though)

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2316,15 +2314,15 @@ private void ValidateNextBlockCommitOnValidatorSetChange()
 
             Assert.Equal(
                 blockChain
-                    .GetNextWorldState(blockChain[0].Hash)
+                    .GetNextWorldState(0)
```
OnedgeLee, line 2317: Trivial
```suggestion
                    .GetNextWorldState(0L)
```

Libplanet.Tests/Blockchain/BlockChainTest.cs
```diff
@@ -2316,15 +2314,15 @@ private void ValidateNextBlockCommitOnValidatorSetChange()
...
                     .GetValidatorSet(),
                 new ValidatorSet(
                     ValidatorPrivateKeys.Select(
                         pk => new Validator(pk.PublicKey, BigInteger.One)).ToList()));
 
             Assert.Equal(
                 blockChain
-                    .GetNextWorldState(blockChain[1].Hash)
+                    .GetNextWorldState(1)
```
OnedgeLee, line 2325: Trivial
```suggestion
                    .GetNextWorldState(1L)
```

Libplanet/Blockchain/BlockChain.ProposeBlock.cs
```diff
@@ -145,11 +145,10 @@ public partial class BlockChain
...
 
-            HashDigest<SHA256> stateRootHash =
-                _blocks[prevHash].ProtocolVersion <
-                BlockMetadata.SlothProtocolVersion
-                    ? _blocks[prevHash].StateRootHash
-                    : (HashDigest<SHA256>)GetNextStateRootHash(prevHash, TimeSpan.Zero);
+            HashDigest<SHA256> stateRootHash = GetNextStateRootHash(prevHash) ??
+                throw new InvalidOperationException(
+                    $"Cannot propose a block as the next state root hash " +
+                    $"for block {prevHash} is missing.");
```
OnedgeLee, line 151: Since we've already waited for NSRH from setting validator set on consensus, we can assume this won't throw on the consensus, right?

Libplanet/Blockchain/BlockChain.States.cs
```diff
@@ -27,33 +26,16 @@ public IWorldState GetWorldState(HashDigest<SHA256>? stateRootHash)
...
-        /// <exception cref="ArgumentException">Thrown when next state root hash exists,
-        /// but corresponding state root is not found in the <see cref="IStateStore"/>.
-        /// </exception>
-        /// <seealso cref="IWorldState"/>
-        public IWorldState? GetNextWorldState(BlockHash offset)
+        public IWorldState? GetNextWorldState()
         {
-            var nextSrh = Store.GetNextStateRootHash(offset);
-            if (nextSrh is { } srh)
+            if (GetNextStateRootHash() is { } srh)
```
OnedgeLee, line 31: trivial, but variable name nsrh would be better for here.
```suggestion
            if (GetNextStateRootHash() is { } nsrh)
```

Libplanet/Blockchain/BlockChain.States.cs
```diff
@@ -27,33 +26,16 @@ public IWorldState GetWorldState(HashDigest<SHA256>? stateRootHash)
...
             {
                 var trie = StateStore.GetStateRoot(srh);
```
OnedgeLee, line 33:
```suggestion
                var trie = StateStore.GetStateRoot(nsrh);
```

Libplanet/Blockchain/BlockChain.States.cs
```diff
@@ -27,33 +26,16 @@ public IWorldState GetWorldState(HashDigest<SHA256>? stateRootHash)
...
                 return trie.Recorded
-                    ? new WorldBaseState(StateStore.GetStateRoot(nextSrh), StateStore)
-                    : throw new ArgumentException(
-                        $"Could not find state root {srh} in {nameof(StateStore)}.",
-                        nameof(offset));
+                    ? new WorldBaseState(trie, StateStore)
+                    : throw new InvalidOperationException(
+                        $"Could not find state root {srh} in {nameof(StateStore)} for " +
```
OnedgeLee, line 37:
```suggestion
                        $"Could not find state root {nsrh} in {nameof(StateStore)} for " +
```

OnedgeLee (commented):
Looks much better 👍

Libplanet/Blockchain/BlockChain.ProposeBlock.cs
greymistcube: In essence, `ProposeBlock()` should only be called by consensus controller as a proposer. This should never be thrown if the PBFT algorithm is working as intended.

Libplanet.Net/Consensus/ConsensusContext.cs
greymistcube: Fixing this properly requires further rework and will be handled in a separate PR addressing #3833.

Libplanet.Tests/BlockChainExtensions.cs
greymistcube: It's due to this being in `Libplanet.Tests` which doesn't allow `nullable`. 😐

Libplanet.Tests/Action/WorldTest.cs
greymistcube: Actually, this helper method seems unnecessary.
As for whether it exists, it depends on the surrounding test code.
Since its used in a synchronous manner via `Append()` this should always exist.

Libplanet.Net/Consensus/ConsensusContext.cs
OnedgeLee: Ah, now I remember.
Missed it's also called on https://github.com/planetarium/libplanet/blob/cc268b022eb3569d2056d29174a51efaeb408d18/Libplanet.Net/Consensus/ConsensusContext.cs#L269
Seems like there are no validator replacing logic, and pre-generation of higher context would raises some bugs, right?

Libplanet.Net/Consensus/ConsensusContext.cs
OnedgeLee: 👌

OnedgeLee (dismissed):

greymistcube:
@OnedgeLee 
I've made the suggested changes except for the indexing issue in `ConsensusContext.NewHeight()` method, which will be handled in a separate PR.

OnedgeLee (approved):

PR #3845, opened 2024-06-24
Remove a possible `ConsensusContext` lock due to a future `ConsensusMsg`.

### LINKED ISSUE #3833: Fix ConsensusContext message handling lock

greymistcube:
Resolves #3833.

Changed to store `ConsensusMsg`s in a `HashSet<T>` instead of creating a new `Context`.
Also removed a requirement for needing `lastCommit` from `Context.Start()`.
As a result, a lock should no longer happen due to receiving a `ConsensusMsg` from other `Peer`s.

### DIFF
--- a/Libplanet.Net/Consensus/Context.Async.cs
+++ b/Libplanet.Net/Consensus/Context.Async.cs
@@ -11,15 +11,11 @@ namespace Libplanet.Net.Consensus
         /// <summary>
         /// Starts round #0 of consensus for <see cref="Height"/>.
         /// </summary>
-        /// <param name="lastCommit">A <see cref="Block.LastCommit"/> from previous block.
-        /// </param>
-        public void Start(BlockCommit? lastCommit = null)
+        public void Start()
         {
             _logger.Information(
-                "Starting context for height #{Height}, LastCommit: {LastCommit}",
-                Height,
-                lastCommit);
-            _lastCommit = lastCommit;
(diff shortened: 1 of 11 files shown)


Libplanet.Net/Consensus/Context.Async.cs
```diff
@@ -11,15 +11,11 @@ public partial class Context
         /// <summary>
         /// Starts round #0 of consensus for <see cref="Height"/>.
         /// </summary>
-        /// <param name="lastCommit">A <see cref="Block.LastCommit"/> from previous block.
-        /// </param>
-        public void Start(BlockCommit? lastCommit = null)
+        public void Start()
```
OnedgeLee, line 14: With this scheme, `Context.Start()` seems to be less meaningful than before, though still we have advantages separating constructor from this for tests.

OnedgeLee (approved):

Libplanet.Net/Consensus/Context.Async.cs
greymistcube: That's one of the reasons. Aside from testing, we might also want to control when a `Context` "starts" from a node. It might not be necessary, but as we will lose more control, it should then be clearly documented **when** a `Context` should be made. 🙄

OnedgeLee: Agree

PR #3851, opened 2024-06-26
♻️ Refactor `ConsensusContext`

greymistcube:
Overview:
- `ConsensusContext` now only holds a single `Context` internally. This is no longer `null`-able.
- `NewHeight()` is only responsible for switching the "view".
- New `Running` property determines whether a `Context` held by `ConsensusContext` will be running.


OnedgeLee (approved):
LGTM

PR #3880, opened 2024-07-15
Fix an issue where currency does not work correctly in dotnet6

s2quake:
In dotnet6, the following error occurs when building.

`CS0188, CS8885: The this object cannot be used before all of its fields are assigned to`

I tried to change the `Hash` field of `Currency` to a property, but it was not the correct way to modify it.

According to the error message, I just need to avoid calling this method, so I changed the method to static and used it.

The code is less readable, but I was able to solve the problem for now.

### DIFF
--- a/src/Libplanet.Types/Assets/Currency.cs
+++ b/src/Libplanet.Types/Assets/Currency.cs
@@ -80,12 +80,7 @@ namespace Libplanet.Types.Assets
         /// The deterministic hash derived from other fields.
         /// </summary>
         [JsonInclude]
-#if NETSTANDARD2_0_OR_GREATER
         public readonly HashDigest<SHA1> Hash;
-#else
-#pragma warning disable SA1201
-        public HashDigest<SHA1> Hash => GetHash();
-#endif
 
         /// <summary>
         /// Whether the total supply of this instance of <see cref="Currency"/> is trackable.
@@ -248,9 +243,7 @@ namespace Libplanet.Types.Assets
                 Minters = null;
             }
 
-#if NETSTANDARD2_0_OR_GREATER
-            Hash = GetHash();
-#endif // NETSTANDARD2_0_OR_GREATER
+            Hash = GetHash(Minters, Ticker, DecimalPlaces, _maximumSupply, TotalSupplyTrackable);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Libplanet.Types/Assets/Currency.cs
```diff
@@ -742,40 +728,56 @@ private static SHA1 GetSHA1()
...
         }
 
-        [Pure]
-        private IValue SerializeForHash()
+        private static IValue SerializeForHash(
+            IImmutableSet<Address>? minters,
+            string ticker,
+            byte decimalPlaces,
+            (BigInteger Major, BigInteger Minor)? maximumSupply,
+            bool totalSupplyTrackable)
```
riemannulus, line 736: Is it necessary?

s2quake: I think it's not, That code has been moved into `GetHash`.
amended https://github.com/planetarium/libplanet/pull/3880/commits/997922be4554673eba597395fe03f1efee5f4d6c

greymistcube: Wasn't there a reason why `GetHash()` and `SerializeForHash()` were separate in the first place? 🤔

greymistcube: Ah, nevermind, it was about `SerializeForHash()` and `Serialize()`. 🙄

greymistcube: It might be good to have a simple comment on whether we have a separate serialization scheme. 😐

s2quake: > It might be good to have a simple comment on whether we have a separate serialization scheme. 😐

https://github.com/planetarium/libplanet/pull/3880/commits/3808b22190d2bf242c44416b47315184856ba235#diff-7975b57b58acd250c0c5fb4f0001c98e298746481661b38ccb8e64c429dda7b5R731
How about this?

greymistcube: The comment doesn't clarify as to **why** the `Dictionary` construction is different from `Serialize()`, which I think would be the main point of confusion for those seeing the code without context. I'd say the history regarding `GetHash()` is barely relevant and doesn't need mentioning.

s2quake: > The comment doesn't clarify as to **why** the `Dictionary` construction is different from `Serialize()`, which I think would be the main point of confusion for those seeing the code without context. I'd say the history regarding `GetHash()` is barely relevant and doesn't need mentioning.

https://github.com/planetarium/libplanet/pull/3880/commits/9a8fd893bb89c127168ba77cfbb223123d0122c9#diff-7975b57b58acd250c0c5fb4f0001c98e298746481661b38ccb8e64c429dda7b5R731
amended

greymistcube (approved):

riemannulus (approved):

PR #3897, opened 2024-07-29
🔧  fix: use cache instead of get block from store

riemannulus:
RocksDBStore's `store.IterateIndexes` causes too much CPU usage. So we use `BlockChain`'s blocks cache temporarily.

### DIFF
--- a/tools/Libplanet.Explorer/Queries/ExplorerQuery.cs
+++ b/tools/Libplanet.Explorer/Queries/ExplorerQuery.cs
@@ -5,7 +5,6 @@ using GraphQL.Types;
 using Libplanet.Blockchain;
 using Libplanet.Crypto;
 using Libplanet.Explorer.GraphTypes;
-using Libplanet.Explorer.Indexing;
 using Libplanet.Explorer.Interfaces;
 using Libplanet.Store;
 using Libplanet.Types.Blocks;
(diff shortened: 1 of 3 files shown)


tools/Libplanet.Explorer/Queries/ExplorerQuery.cs
```diff
@@ -49,53 +49,13 @@ public ExplorerQuery(IBlockChainContext chainContext)
         {
             Block tip = Chain.Tip;
```
s2quake, line 50: This method does not use the parameters `excludeEmptyTxs` and `miner`.
If those parameters are useless, it might be a good idea to delete them.
Otherwise, I suggest the code below, which uses the parameters `excludeEmptyTxs` and `miner`. 
With a little help from the GitHub co-pilots. 😀
```csharp
var comparer = desc
    ? Comparer<long>.Create((x, y) => y.CompareTo(x))
    : Comparer<long>.Default;
var count = Chain.Tip.Index;
var min = Math.Max(0, offset);
var max = Math.Min(count, offset + (limit ?? 100));
return Range(min, max)
        .Where(item => excludeEmptyTxs != true || item.Transactions.Count > 0)
        .Where(item => miner is null || item.Miner == miner)
        .OrderBy(item => item.Index, comparer);

static IEnumerable<Block> Range(long min, long max)
{
    for (long i = min; i < max; i++)
    {
        yield return Chain[i];
    }
}
```

PR #3900, opened 2024-07-30
Bump dotnet to 8 in Libplanet SDK

s2quake:
SSIA

### DIFF
--- a/tools/Libplanet.Explorer.Cocona/Libplanet.Explorer.Cocona.csproj
+++ b/tools/Libplanet.Explorer.Cocona/Libplanet.Explorer.Cocona.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <NoWarn>$(NoWarn);CS1591;S1118;SA1118</NoWarn>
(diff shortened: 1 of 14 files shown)


tools/Libplanet.Explorer.Cocona/Libplanet.Explorer.Cocona.csproj
```diff
@@ -8,7 +8,7 @@
   <ItemGroup>
     <PackageReference Include="Cocona.Lite" Version="2.0.*" />
     <PackageReference Include="Serilog.Sinks.Console" Version="4.1.0" />
-    <PackageReference Include="System.Text.Json" Version="7.0.0" />
+    <PackageReference Include="System.Text.Json" Version="8.0.4" />
```
riemannulus, line 11: This change is not the sdk feature boundary I guess...

s2quake: But, in .NET 8, using version 7.0.0 causes the following error.
```plain
error NU1903: Warning As Error: Package 'System.Text.Json' 7.0.0 has a known high severity vulnerability,
```
The only way to solve this problem is to add `NU1903` into the NoWarn property.
So we may lose the chance to solve the vulnerability in the future when it raises.

PR #3868, opened 2024-07-03
wip: Introduce/gas tracer

riemannulus:
Since we track gas usage by transaction, `ActionContext.GasMeter` is not a good choice when tracking gas usage.

PR #3912, opened 2024-08-02
Preparation for PoS

limebell:
### Changes

 -  Added `MaxGasPrice` property to `IActionContext` interface and its implementations.  [[#3762]]
 -  (Libplanet.Explorer) Added `self` field to `NoteStateType`.  [[#3841]]
 -  (Libplanet.Action) Removed `IFeeCollector` interface
    and its implementations.  [[#3867]]
 -  (Libplanet.Action) Removed following methods from the
    `IActionContext` interface.  [[#3868]]
     -  Removed `IActionContext.UseGas(long)`.
     -  Removed `IActionContext.GasUsed()`.
     -  Removed `IActionContext.GasLimit()`.
 -  (Libplanet.Action) Added `GasTracer` static class.  [[#3868]]
 -  (Libplanet.Action) Added `LastCommit` property to `IActionContext`
    interface and its implementations.  [[#3912]]

### Bug Fixes

 - A vote without validator power(or null) cannot be added to `HeightVoteSet`.
 - Fixed an issue where the proposer could suggest a vote with null validator power.
 - Suppressed some warnings caused by blank lines.


src/Libplanet.Types/Consensus/ValidatorSet.cs
```diff
@@ -264,16 +263,84 @@ public Validator GetProposer(long height, int round)
...
-                           blockCommit.Votes.Select(vote => vote.ValidatorPublicKey).ToList()) &&
-                   blockCommit.Votes.All(
-                       v => v.ValidatorPower is null ||
-                            v.ValidatorPower == GetValidator(v.ValidatorPublicKey).Power);
+            ValidatorSet validatorSetFromCommit = new ValidatorSet(blockCommit.Votes.Select(
+                v => v.ValidatorPower is BigInteger power
+                    ? new Validator(v.ValidatorPublicKey, power)
+                    : throw new InvalidBlockCommitException(
+                        "All votes in the block commit after block protocol version 10 " +
+                        "must have power.")).ToList());
```
OnedgeLee, line 277: Do we need to construct `validatorSetFromCommit` variable?

src/Libplanet.Types/Consensus/ValidatorSet.cs
```diff
@@ -264,16 +263,84 @@ public Validator GetProposer(long height, int round)
...
+                    $"BlockCommit of BlockHash {blockCommit.BlockHash} " +
+                    $"has different validator power with chain state's validator set: \n" +
+                    $"in states | \n " +
+                    Validators.Aggregate(
+                        string.Empty, (s, v) => s + v.Power + ", \n") +
+                    $"in blockCommit | \n " +
+                    blockCommit.Votes.Aggregate(
+                        string.Empty, (s, v) => s + v.ValidatorPower + ", \n"));
+            }
+        }
```
OnedgeLee, line 307: Seems like merging this condition with above null condition will reduce 1 times of iteration.

src/Libplanet.Types/Consensus/ValidatorSet.cs
s2quake: > Do we need to construct `validatorSetFromCommit` variable?

I don't think this code is necessary, because there is power comparison between the vote and the validator below.
Therefore, this code should be deleted.

src/Libplanet.Types/Consensus/ValidatorSet.cs
s2quake: > Seems like merging this condition with above null condition will reduce 1 times of iteration.

It can be resolved by deleting the above code, rather than merging it with this code.

s2quake: Amended https://github.com/planetarium/libplanet/pull/3912/commits/5f5108e68967db43bea1070ac3a9e11aca5aacf4

CHANGES.md
```diff
@@ -24,6 +39,12 @@ To be released.
 
 ### CLI tools
 
+[#3762]: https://github.com/planetarium/libplanet/pull/3762
+[#3841]: https://github.com/planetarium/libplanet/pull/3841
+[#3867]: https://github.com/planetarium/libplanet/pull/3867
+[#3868]: https://github.com/planetarium/libplanet/pull/3868
+[#3912]: https://github.com/planetarium/libplanet/pull/3912
```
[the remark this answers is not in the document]
s2quake: See this https://github.com/planetarium/libplanet/pull/3912/commits/d65ff915dace0c5f695a77f04d2f79747d285971

riemannulus (approved):

OnedgeLee (approved):

PR #3913, opened 2024-08-02
⚡ ♻️ Refactor `BlockLocator` to include only one `BlockHash`

greymistcube:
This should help with overall performance when syncing `Block`s.
The update makes it incompatible with nodes using a previous version of `Libplanet` in terms of syncing `Block`s.


riemannulus (dismissed):
LGTM

PR #3916, opened 2024-08-08
🔧 Prepare 5.2.3

riemannulus:
SSIA


s2quake:
In the `Directory.Build.props` file in the root path, you need to change the `VersionPrefix` to `5.2.3`.

s2quake (approved):

PR #3917, opened 2024-08-09
Refactor/actionbase

greymistcube:


Libplanet.SDK.Action.Tests/SimpleRPG/Actions/AvatarAction.cs
```diff
@@ -0,0 +1,46 @@
...
+            string name = (Text)args;
+            Call<InfoAction, Info>("Create", new object?[] { name });
+            Call<InventoryAction, Inventory>("Create");
+        }
+
+        [Callable]
+        public Avatar GetAvatar(Address address)
+        {
+            Info info = Call<InfoAction, Info>(
+                "GetInfo",
```
s2quake, line 27: How about using `nameof(InfoAction.GetInfo)` instead of literal strings.

Libplanet.SDK.Action/Action/ActionBase.cs
```diff
@@ -11,14 +11,14 @@ namespace Libplanet.SDK.Action
...
+            .Add("args", _args ?? throw new NullReferenceException());
 
         public void LoadPlainValue(IValue plainValue)
         {
             Dictionary dict = (Dictionary)plainValue;
-            _name = (Text)dict["name"];
+            _name = (Text)dict["type_id"];
             _call = (Text)dict["call"];
             _args = dict["args"];
         }
```
s2quake: How about a pattern like below?

```csharp
        void IAction.LoadPlainValue(IValue plainValue)
        {
            if (_isLoaded)
            {
                throw new InvalidOperationException("Action is already loaded.");
            }

            Dictionary dict = (Dictionary)plainValue;
            _name = (Text)dict["type_id"];
            _exec = (Text)dict["exec"];
            _args = dict["args"];
            _isLoaded = true;
        }

```

It has two advantages: limited access and avoiding duplicate calls.

PR #3919, opened 2024-08-12
Add json schema builder for appsettings.json

s2quake:
### Feature

* A `OptionsSchemaBuilder` was added to create schemas to be used in `appsettings.json`.
* Added `OptionsAttribute` to options classes to add properties to the schema.
* Added schema file `appsettings-schema.json`.
* You can get the schema by connecting to http://localhost:5259/schema.

### DIFF
--- a/sdk/node/Libplanet.Node/Options/GenesisOptions.cs
+++ b/sdk/node/Libplanet.Node/Options/GenesisOptions.cs
@@ -1,3 +1,4 @@
+using System.ComponentModel;
 using Libplanet.Crypto;
 using Libplanet.Net;
 using Libplanet.Node.DataAnnotations;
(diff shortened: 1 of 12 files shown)


sdk/node/Libplanet.Node/Options/GenesisOptions.cs
```diff
@@ -17,10 +19,13 @@ public static readonly PrivateKey AppProtocolKey
         AppProtocolKey, 1);
 
     [PrivateKey]
+    [Description("The key of the genesis block.")]
     public string GenesisKey { get; set; } = string.Empty;
 
     [PublicKeyArray]
+    [Description("The hash of the genesis block.")]
```
riemannulus, line 26: This is not a `hash of the genesis block`.

s2quake: > This is not a `hash of the genesis block`.

amended https://github.com/planetarium/libplanet/pull/3919/commits/db6882cd8f42dec38b99a27e8505cd2d06ecd8bd

PR #3920, opened 2024-08-13
♻️ Refactor `GetDemandBlockHashes()`

greymistcube:
- Changed the return type to `List<T>` as the return type of `IAsyncEnumerable<T>` does not make much sense. Things were internally being unnecessarily cast to various collection types.
- Changed the behavior of `GetBlockDemandHashes()` to **not** aggregate `BlockHash`es from various peers. The rest of the block syncing stack pretty much assumes `BlockHash`es from `GetBlockDemandHashes()` is from a single continuous branch.

This is to mainly fix the behavior of `GetBlockDemandHashes()` in accordance to the whole block syncing process.
No changelog is provided as all the changes are internal.

### DIFF
--- a/src/Libplanet.Net/Swarm.BlockCandidate.cs
+++ b/src/Libplanet.Net/Swarm.BlockCandidate.cs
@@ -146,16 +146,15 @@ namespace Libplanet.Net
             bool render,
             IProgress<BlockSyncState> progress)
         {
-            BlockChain workspace = blockChain;
+            BlockChain workspace;
             List<Guid> scope = new List<Guid>();
             bool forked = false;
 
-            Block oldTip = workspace.Tip;
+            Block oldTip = blockChain.Tip;
             Block newTip = candidate.Blocks.Last().Item1;
             List<(Block, BlockCommit)> blocks = candidate.Blocks.ToList();
             Block branchpoint = FindBranchpoint(
                  oldTip,
-                 newTip,
                  blocks.Select(pair => pair.Item1).ToList());
 
             if (branchpoint.Equals(oldTip))
@@ -163,8 +162,9 @@ namespace Libplanet.Net
                 _logger.Debug(
                     "No need to fork. at {MethodName}()",
                     nameof(AppendPreviousBlocks));
+                workspace = blockChain;
             }
-            else if (!workspace.ContainsBlock(branchpoint.Hash))
+            else if (!blockChain.ContainsBlock(branchpoint.Hash))
             {
                 // FIXME: This behavior can unexpectedly terminate the swarm (and the game
                 // app) if it encounters a peer having a different blockchain, and therefore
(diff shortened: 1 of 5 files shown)


src/Libplanet.Net/Swarm.BlockCandidate.cs
```diff
@@ -146,25 +146,25 @@ private BlockChain AppendPreviousBlocks(
...
             IProgress<BlockSyncState> progress)
         {
-            BlockChain workspace = blockChain;
+            BlockChain workspace;
             List<Guid> scope = new List<Guid>();
             bool forked = false;
 
-            Block oldTip = workspace.Tip;
+            Block oldTip = blockChain.Tip;
             Block newTip = candidate.Blocks.Last().Item1;
```
OnedgeLee, line 154: Is this variable still needed?

greymistcube: Well, reorg hasn't been completely removed as you see in the code below. The goal is to remove it eventually. 😓

OnedgeLee: I could find reorg still exists, but seems like `FindBranchPoint()` does not receives `newTip` at all, and find it from `newBlocks` inside the method.

src/Libplanet.Net/Swarm.BlockCandidate.cs
```diff
@@ -179,15 +179,15 @@ private BlockChain AppendPreviousBlocks(
...
-                    workspace.Genesis.Hash);
+                    blockChain.Genesis.Hash);
             }
             else
             {
                 _logger.Debug(
                     "Trying to fork... at {MethodName}()",
                     nameof(AppendPreviousBlocks)
                 );
-                workspace = workspace.Fork(branchpoint.Hash);
```
riemannulus, line 190: Using BlockChain instead of workspace means that there is no point in forking here. To put it another way, in my opinion, the need for Fork is a situation where it is assumed that Reorg still exists. What do you think about removing Fork first?

riemannulus (commented):
Did you test on the 9c mainnet for check compatibillity?

src/Libplanet.Net/Swarm.BlockCandidate.cs
greymistcube: Ah, I misunderstood. Indeed, `newTip` can be removed. 😅

src/Libplanet.Net/Swarm.BlockCandidate.cs
greymistcube: The problem is that it is not guaranteed by the current API that forking would not occur. With PBFT, the whole `else` block would not be run in the first place, but it is left as it is to reduce the scope of this PR which mainly focuses on the internal logic of `GetDemandBlockHashes()`. Cleaning up `Fork`ing, other unnecessary conditional branching, reimplementation of `FindBranchPoint`, etc. to name the few will be done gradually in follow-up PRs. 😅

greymistcube:
@riemannulus 
Tested on 9c-mainnet without a problem. 🙂

riemannulus (approved):

PR #3931, opened 2024-08-23
♻️ Simplified the internal behavior of `GetDemandBlockHashes()`

greymistcube:
- Removed internal "stitching" of `BlockHash`es over multiple attempts.
- Removed `chunkSize` parameter. This was never clearly defined and was being used improperly in the first place.
- When attempting to get `BlockHash`es from a `Peer`, the `Peer` selection is done randomly (instead of trying to attempt from the "highest" `Peer` first).
- As a result, the overall behavior of preloading has changed.


riemannulus (approved):
LGTM

PR #3941, opened 2024-08-30
Add IStoreService to SDK

s2quake:
When creating a blockchain, you now need to create and inject a StoreService externally instead of creating a store internally.


s2quake:
This PR is included in https://github.com/planetarium/libplanet/pull/3946.

PR #3946, opened 2024-09-05
Add StoreService and Test codes for each service

s2quake:
### Changes

* Added `IStoreService`
  * `IStoreService` is created externally and injected into the `BlockChainService`.
* Fixed an issue with ValidatorService not being created.
* Added test codes for each service.


riemannulus (approved):
LGTM

PR #3948, opened 2024-09-05
🧹 Remove `Fork()` and `Swap()`

greymistcube:
We can finally assume forking and swapping of chains would not happen through normal operations. 🙃
As such, once 5.3.0 is released, we can perform an analysis on an `IStore` and migrate/prune the internal DB and simplify the `IStore` interface accordingly.


riemannulus:
Here, in [NineChronicles.Snapshot](https://github.com/planetarium/NineChronicles.Snapshot) use a fork and swap afaik. Could you check it?

greymistcube:
@riemannulus 

> Here, in [NineChronicles.Snapshot](https://github.com/planetarium/NineChronicles.Snapshot) use a fork and swap afaik. Could you check it?

I have checked it. It is all done externally without using `BlockChain.Fork()` or `BlockChain.Swap()`.
Although, I have been thinking maybe we should provide a tool to manipulate `IStore` and `IStateStore` more safely. 🙄

riemannulus:
> Although, I have been thinking maybe we should provide a tool to manipulate IStore and IStateStore more safely. 🙄

Agree.

riemannulus (approved):

PR #4004, opened 2024-11-29
Optimized `RocksDBStore.IterateIndexes()` for pruned chain storage.

greymistcube:


greymistcube:
Added 49ae37ff92d0ac852227410d9e2471dde551eb63 for further optimization.

The point is to terminate iteration early at the exact point it should terminate.
Although not a proper benchmark, a preliminary benchmark shows a vastly improved indexing speed.

```
For unpruned Odin chain:
Took 28 ms to access last 1000 hashes

For pruned Odin chain:
Took 2 ms to access last 1000 hashes
```

Just another reason why pruning should be done. 😗

greymistcube:
Unless we want to support partial chains, the removed test is invalid.

PR #4007, opened 2024-12-07
Delayed consensus step transition

limebell:
This PR provides an option to delay step transition, since libplanet allows for a single node to hold +2/3 vote power.


src/Libplanet.Net/Consensus/Context.cs
```diff
@@ -585,21 +590,30 @@ private Maj23 MakeMaj23(int round, BlockHash hash, VoteFlag flag)
...
 
         /// <summary>
-        /// Gets the proposed block and valid round of the given round.
+        /// Sets the proposal for the round.
         /// </summary>
-        /// <returns>Returns a tuple of proposer and valid round.  If proposal for the round
-        /// does not exist, returns <see langword="null"/> instead.
-        /// </returns>
-        private (Block, int)? GetProposal()
+        private void SetProposal(Proposal? proposal)
```
s2quake, line 595: How about improving the Proposal property instead of defining the SetProposal method as shown below?

```csharp

        public Proposal? Proposal
        {
            get => _proposal;
            set
            {
                if (value is not null)
                {
                    var marshaled = (Dictionary)_codec.Decode(value.MarshaledBlock);
                    _proposal = value;
                    _proposalBlock = BlockMarshaler.UnmarshalBlock(marshaled);
                }
                else
                {
                    _proposal = null;
                    _proposalBlock = null;
                }
            }
        }
```

s2quake (approved):

PR #4017, opened 2024-12-30
Add genesis config option for Node

limebell:
This provides genesis base state configuration option.

Use "Genesis:GenesisConfigurationPath" and pass JSON represented base genesis state.

The structure of the JSON file is:
```JSON
{
    "<AccountAddress>":
    {
        "<Address>":"<Serialized IValue as Hex string>"
    }
}
```

For more detail, please take a look at `sdk/node/Libplanet.Node.Tests/Services/BlockChainServiceTest.cs`

### DIFF
--- a/sdk/node/Libplanet.Node.Tests/Services/BlockChainServiceTest.cs
+++ b/sdk/node/Libplanet.Node.Tests/Services/BlockChainServiceTest.cs
@@ -1,7 +1,13 @@
+using Bencodex;
+using Bencodex.Types;
+using Libplanet.Common;
+using Libplanet.Crypto;
 using Libplanet.Node.Extensions;
 using Libplanet.Node.Services;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Libplanet.Node.Tests.Services;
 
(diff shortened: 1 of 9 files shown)


sdk/node/Libplanet.Node/Services/BlockChainService.cs
```diff
@@ -152,4 +174,83 @@ private static Block LoadGenesisBlockFromUrl(Uri genesisBlockUri)
...
+
+    private static byte[] LoadConfigurationFromUri(Uri configurationUri)
+    {
+        using var client = new HttpClient();
+        return configurationUri.IsFile
+            ? LoadConfigurationFromFilePath(configurationUri.AbsolutePath)
+            : client.GetByteArrayAsync(configurationUri).Result;
+    }
+
+    private static Block CreateGenesisBlockFromConfiguration(
```
s2quake, line 191: How about throwing an exception instead of handling the error or performing alternative actions if the data for creating the genesis block is invalid?

```csharp
        if (data == null || data.Count == 0)
        {
            throw new .....
        }
```
```csharp
                    catch (Exception e)
                    {
                        // skip state
                        throw e;
                    }
```

sdk/node/Libplanet.Node.Tests/Services/BlockChainServiceTest.cs
```diff
@@ -16,4 +22,70 @@ public void Create_Test()
...
+            Assert.Equal(
+                (Text)"B",
+                blockChain.GetNextWorldState()?.GetAccountState(accountB).GetState(addressA));
+            Assert.Equal(
+                (Integer)456,
+                blockChain.GetNextWorldState()?.GetAccountState(accountB).GetState(addressB));
+        }
+        finally
+        {
+            if (File.Exists(tempFilePath))
```
s2quake, line 85: The indent level can be reduced by one if the following code is used.
https://github.com/s2quake/libplanet-console/blob/main/src/common/LibplanetConsole.Common/IO/TempFile.cs
However, it's unfortunate that the code is not available in libplanet.node

sdk/node/Libplanet.Node/Options/GenesisOptions.cs
```diff
@@ -26,6 +27,12 @@ public sealed class GenesisOptions : OptionsBase<GenesisOptions>
...
         $"The path of the genesis block, which can be a file path or a URI." +
-        $"This property cannot be used with {nameof(GenesisKey)}.")]
+        $"This property cannot be used with {nameof(GenesisKey)} and " +
+        $"{nameof(GenesisConfigurationPath)}.")]
     public string GenesisBlockPath { get; set; } = string.Empty;
+
+    [Description(
+        $"The path of the genesis configuration, which can be a file path or a URI." +
+        $"This property cannot be used with {nameof(GenesisKey)} and {nameof(GenesisBlockPath)}.")]
+    public string GenesisConfigurationPath { get; set; } = string.Empty;
```
s2quake, line 37: If there are many ways to create Genesis blocks, it may be better to create a type variable rather than increase the properties.
If the properties increase, the complexity will increase because the exception handling code must be implemented.
For example,
```csharp
enum GenesisBlockType
{
    Binary,
    BinaryUrl,
    Json,
    JsonUrl
}
```
I don't know what the best way is, but I think it's worth considering.

tools/Libplanet.Explorer.Cocona/Libplanet.Explorer.Cocona.csproj
```diff
@@ -8,7 +8,7 @@
   <ItemGroup>
     <PackageReference Include="Cocona.Lite" Version="2.0.*" />
     <PackageReference Include="Serilog.Sinks.Console" Version="4.1.0" />
-    <PackageReference Include="System.Text.Json" Version="7.0.0" />
+    <PackageReference Include="System.Text.Json" Version="9.0.*" />
```
s2quake, line 11: Bumping to the latest version is always welcome. At least for me. 😀

sdk/node/Libplanet.Node.Tests/Services/BlockChainServiceTest.cs
```diff
@@ -16,4 +22,70 @@ public void Create_Test()
...
+                blockChain.GetNextWorldState()?.GetAccountState(accountA).GetState(addressA));
+            Assert.Equal(
+                (Integer)123,
+                blockChain.GetNextWorldState()?.GetAccountState(accountA).GetState(addressB));
+            Assert.Equal(
+                (Text)"B",
+                blockChain.GetNextWorldState()?.GetAccountState(accountB).GetState(addressA));
+            Assert.Equal(
+                (Integer)456,
+                blockChain.GetNextWorldState()?.GetAccountState(accountB).GetState(addressB));
```
s2quake, line 81: It can be expressed as a single line like this, but it's more about coding style, so you don't need to worry about it.
```suggestion
            var worldState = blockChain.GetNextWorldState();
            Assert.NotNull(worldState);

            Assert.Equal((Text)"A", worldState.GetAccountState(accountA).GetState(addressA));
            Assert.Equal((Integer)123, worldState.GetAccountState(accountA).GetState(addressB));
            Assert.Equal((Text)"B", worldState.GetAccountState(accountB).GetState(addressA));
            Assert.Equal((Integer)456, worldState.GetAccountState(accountB).GetState(addressB));
```

sdk/node/Libplanet.Node/Services/BlockChainService.cs
limebell: Throwing an error in the service, the exception bubble pops up to the program?

tools/Libplanet.Explorer.Cocona/Libplanet.Explorer.Cocona.csproj
limebell: Vulnerability in 7.0 causes failure in build 😢

sdk/node/Libplanet.Node/Options/GenesisOptions.cs
limebell: I guess this suggestion should be dealt in one separated PR

sdk/node/Libplanet.Node/Services/BlockChainService.cs
s2quake: Genesis is the beginning of blockchain, so shouldn't it be perfect? It seems to me that if the initialization is wrong, it's like not knowing which direction the process will go, so if it's wrong, the blockchain shouldn't start at all.

limebell: IMO this would be better to be in `planet` CLI :thinking:

